PR #189, opened 2017-12-22
Point to a git version of cism and get standalone build to work

billsacks:
Note that the cism version is based off of cism2_1_44, rather than the
previously-pointed-to cism2_1_40

### DIFF
--- a/.config_files.xml
+++ b/.config_files.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0"?>
+
+<?xml-stylesheet type="text/xsl" ?>
+
+<entry_id>
+
+  <!-- This is the same as the default entry in
+       cime/config/cesm/config_files.xml except for the value for clm:
+       In a standalone clm checkout, COMP_ROOT_DIR_LND is $SRCROOT
+       rather than $SRCROOT/components/clm.
+
+       However, because of the way overrides are handled, we need to
+       re-specify the full information here rather than just overriding
+       the value for clm.
+  -->
+  <entry id="COMP_ROOT_DIR_LND">
+    <type>char</type>
+    <default_value>unset</default_value>
+    <values>
+      <value component="clm"      >$SRCROOT</value>
+      <value component="dlnd"      >$CIMEROOT/src/components/data_comps/dlnd</value>
+      <value component="slnd"      >$CIMEROOT/src/components/stub_comps/slnd</value>
+      <value component="xlnd"      >$CIMEROOT/src/components/xcpl_comps/xlnd</value>
+    </values>
+    <group>case_comps</group>
+    <file>env_case.xml</file>
+    <desc>Root directory of the case land model component  </desc>
+    <schema>$CIMEROOT/config/xml_schemas/config_compsets.xsd</schema>
+  </entry>
+
+</entry_id>
--- a/CESM.cfg
+++ b/CESM.cfg
@@ -6,9 +6,10 @@ required = True
 
 [cism]
 local_path = components/cism
-protocol = svn
-repo_url = https://svn-ccsm-models.cgd.ucar.edu/glc
-tag = trunk_tags/cism2_1_40
+protocol = git
+repo_url = https://github.com/ESCOMP/cism-wrapper
+branch = add_manage_externals_file
+externals = CISM.cfg
 required = True
 
 [rtm]
--- a/CESM.cfg
+++ b/CESM.cfg
@@ -30,7 +30,7 @@ required = True
 local_path = cime
 protocol = git
 repo_url = https://github.com/CESM-Development/cime
-tag = billsacks/always_glcmec_n03
+tag = billsacks/always_glcmec_n04
 required = True
 
 [externals_description]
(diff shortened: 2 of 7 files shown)


CESM.cfg
```diff
@@ -6,9 +6,10 @@ required = True
 
 [cism]
 local_path = components/cism
-protocol = svn
-repo_url = https://svn-ccsm-models.cgd.ucar.edu/glc
-tag = trunk_tags/cism2_1_40
+protocol = git
+repo_url = https://github.com/ESCOMP/cism-wrapper
+branch = add_manage_externals_file
+externals = CISM.cfg
```
billsacks: @bandre-ucar do I need to explicitly point to an `externals` file like this, or can it be auto-detected if I name it appropriately?

.config_files.xml
```diff
@@ -0,0 +1,31 @@
...
+  <!-- This is the same as the default entry in
+       cime/config/cesm/config_files.xml except for the value for clm:
+       In a standalone clm checkout, COMP_ROOT_DIR_LND is $SRCROOT
+       rather than $SRCROOT/components/clm.
+
+       However, because of the way overrides are handled, we need to
+       re-specify the full information here rather than just overriding
+       the value for clm.
+  -->
+  <entry id="COMP_ROOT_DIR_LND">
```
billsacks: I was hoping that I could get away with just specifying the new `COMP_ROOT_DIR_LND` for clm, without duplicating all of the metadata and without duplicating the entries for `dlnd`, `slnd` or `xlnd`. Alas, this doesn't seem possible:

- When I didn't duplicate all of the metadata (group, file, etc.), the python complained about missing `file`

- When I didn't specify entries for `dlnd`, `slnd` and `xlnd`, I wasn't able to run a T compset (CISM forced by dlnd). I know it's kind of funny to run a compset without CLM from a standalone CLM checkout, but I sometimes do that for testing, and I can see some value in allowing it for scientists as well, so I don't want to lose that ability.

This isn't a big deal, but @bandre-ucar @jedwards4b - am I missing something that would allow me to make this override file briefer?

billsacks:
I have just pushed some changes that allow me to build a case from a standalone checkout. cime seemed to have everything that's needed (I just needed to cherry-pick some cime changes onto the cime branch that we're currently pointing to; later we'll do an update to cime master). So I just needed to make some changes to files in ctsm for this to work.

.config_files.xml
jedwards4b: You aren't missing anything, the override needs to be a complete entry.   I'm not aware of a way to make it shorter.

billsacks: Okay, thanks a lot for taking a look at this

billsacks:
Okay: With the changes on this branch, I'm able to get the aux_clm test suite to pass, and tests are bfb with the baselines that were generated from the latest CLM trunk tag in svn (clm4_5_18_r270)!

(Details to be precise: I did most testing with b07e503. Everything passed except for SSP, LII and LVG tests: the tests defined in CTSM's cime_config directory. I then reran the failing cheyenne tests from the latest commit on this branch: 0f6f402. These passed. I did not bother rerunning the two SSP tests on yellowstone, which are very similar to the now-passing cheyenne SSP test.)

A hard part of this was finding a good cime version to use. I'm currently pointing to a branch off of cime5.4.0-alpha.03. There were three cime PRs needed for the standalone checkout to work right; I have merged these onto my branch. These are all on `cime5.4.0-alpha.16`, but that tag has a broken version of cheyenne_gnu (which is fixed in a commit that is now on cime master but not yet in a tag: ESMCI/cime#2167, fixed in ESMCI/cime@d59c28516).

billsacks:
I have not tried running the build-namelist tests, tools tests or PTCLM tests. I expect some of those to fail due to hard-coded directory structures.

I am working on getting the unit tests to pass with this directory structure, and will push a commit that fixes that.

jedwards4b:
I would prefer the problem be characterised correctly as missing blas or lapack library dependency.  There is nothing wrong with the gnu compiler and the pr you refer to passed all cime testing before being merged to master.

billsacks:
Okay, unit tests work now, both from a standalone checkout and from within a cesm checkout (where we have components/clm/).

I consider this branch ready to merge now, once others are ready. We may still need to make changes to tools, tools testing and possibly build-namelist and ptclm testing, but I think it would be useful to get this branch merged so that we can start doing other work off of it – then we can fix those other pieces in a follow-up tag.

CESM.cfg
```diff
@@ -6,9 +6,10 @@ required = True
 
 [cism]
 local_path = components/cism
-protocol = svn
-repo_url = https://svn-ccsm-models.cgd.ucar.edu/glc
-tag = trunk_tags/cism2_1_40
+protocol = git
+repo_url = https://github.com/ESCOMP/cism-wrapper
+tag = cism2_1_45
+externals = CISM.cfg
```
billsacks: @bandre-ucar do I need to explicitly point to an externals file like this, or can it be auto-detected if I name it appropriately?

billsacks:
I should add: Thank you to @jedwards4b for implementing all of the cime changes that were needed for this standalone checkout to work smoothly!

CESM.cfg
bandre-ucar: It needs to be explicitly listed.

Issue #13, opened 2017-12-16
mkmapdata should use shared mapping tools


billsacks:
**Bill Sacks < sacks@ucar.edu > - 2013-05-07 15:48:42 -0600**
**Bugzilla Id:** [1683](http://bugz.cgd.ucar.edu/show_bug.cgi?id=1683)
**Bugzilla Depends:**  1938, 
**Bugzilla CC:**  anthony.p.craig@gmail.com,  erik@ucar.edu,  mlevy@ucar.edu,  rfisher@ucar.edu,  sacks@ucar.edu, 

Mike Levy suggested that CLM's mkmapdata tool should use the shared mapping tools, and in particular tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh -- rather than directly calling the ESMF tool. The main reason for this is that the shared tool can provide an insulating layer, requiring less maintenance of the CLM tool if the version or interface of the ESMF tool changes in the future.

We wanted to get this into the cesm1.2 release, but time limitations prevented doing so. 

Here are some notes on what needs to be done to accomplish this:

First, some notes from Mike Levy:

I've created

https://svn-ccsm-models.cgd.ucar.edu/tools/mapping/trunk_tags/mapping_130410

with all of the changes we talked about (and more). To generate the file

map_${grid}_${lmask}_to_${res}_nomask_aave_da_$CDATE.nc

You'll want to run

$CESMROOT/tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh -fsrc [file corresponding to ${grid}] -nsrc ${grid}_${lmask} -fdst [file corresponding to ${res}] -ndst ${res}_nomask -map aave --clm_name

You can add the following (see the README file in gen_ESMF_mapping_file for more options):
* If the source file is a regional map, use -tsrc regional
* If the destination file is a regional map, use -tdst regional
* If you want to run the serial ESMF implementation, use -serial
* If you need the 64bit offset, use -big
* If you want to pass any other flags to ESMF, use --pass2esmf $FLAGS

This should work on yellowstone, caldera, and geyser. It may even work on jaguar, but as I said I've never tried.

Now, my notes (some of this repeats what Mike said):

- need to pass -clm_name (this forces the tool to use the CLM naming convention, which differs somewhat from the naming convention used elsewhere for mapping files)

- need to pass src & dest names, NOT weight file name. Note that these names should include the mask so that the output name looks right

[...]

PR #190, opened 2017-12-28
Remove infiltration excess runoff for VIC

billsacks:
**NOTE: This PR was originally created in the temporary CTSM repository, Sep 13, 2017. I am moving over most of the comments from there: both my own and those of @martynpclark .**

**Here was the original comment:**

Martyn Clark reviewed the VIC implementation, and felt that the current
implementation of infiltration excess runoff is inconsistent with the
standard VIC implementation. It appears that what was being called VIC's
infiltration excess runoff was actually just an attempt to give a better
numerical approximation to the solution for saturated surface excess
runoff. So deleting this leaves only a first-order approximation to
VIC's saturated surface excess runoff.

Eventually we may want to put in place a more accurate solution for
VIC's saturated surface excess runoff. But Martyn's feeling is that this
can come in with other changes we want to make regarding numerical
solutions in CTSM.

### DIFF
--- a/src/biogeophys/InfiltrationExcessRunoffMod.F90
+++ b/src/biogeophys/InfiltrationExcessRunoffMod.F90
@@ -53,13 +53,22 @@ module InfiltrationExcessRunoffMod
      procedure, private :: InitCold
 
      procedure, private, nopass :: ComputeQinmaxHksat
-     procedure, private, nopass :: ComputeQinmaxVic
   end type infiltration_excess_runoff_type
 
   ! !PRIVATE DATA MEMBERS:
 
+  ! For methods that don't generate any infiltration excess runoff, we get this end result
+  ! by specifying a huge qinmax value - i.e., an effectively infinite max infiltration
+  ! rate.
+  !
+  ! 1e200 mm H2O /s seems large enough to be effectively infinite, while not being so
+  ! large as to cause floating point overflows elsewhere.
+  real(r8), parameter :: qinmax_unlimited = 1.e200_r8  ! mm H2O /s
+
+  ! The 'none' option avoids generating any infiltration excess runoff by setting qinmax
+  ! to a huge value
+  integer, parameter :: QINMAX_METHOD_NONE = 0
   integer, parameter :: QINMAX_METHOD_HKSAT = 1
-  integer, parameter :: QINMAX_METHOD_VIC  = 2
 
   character(len=*), parameter, private :: sourcefile = &
        __FILE__
@@ -155,7 +164,7 @@ contains
 
     ! TODO(wjs, 2017-08-14) We'll read qinmax_method from namelist.
     if (use_vichydro) then
-       this%qinmax_method = QINMAX_METHOD_VIC
+       this%qinmax_method = QINMAX_METHOD_NONE
     else
        this%qinmax_method = QINMAX_METHOD_HKSAT
     end if
@@ -203,15 +212,16 @@ contains
           )
 
     select case (this%qinmax_method)
+    case (QINMAX_METHOD_NONE)
+       ! NOTE(wjs, 2017-09-01) I'm putting this here for clarity, though it could be
+       ! moved to initialization for efficiency
+       do fc = 1, num_hydrologyc
+          c = filter_hydrologyc(fc)
+          qinmax_on_unsaturated_area(c) = qinmax_unlimited
+       end do
     case (QINMAX_METHOD_HKSAT)
        call this%ComputeQinmaxHksat(bounds, num_hydrologyc, filter_hydrologyc, &
             soilhydrology_inst, soilstate_inst, &
-            qinmax_on_unsaturated_area = qinmax_on_unsaturated_area(bounds%begc:bounds%endc))
-    case (QINMAX_METHOD_VIC)
-       call this%ComputeQinmaxVic(bounds, num_hydrologyc, filter_hydrologyc, &
-            soilhydrology_inst, &
-            fsat = fsat(bounds%begc:bounds%endc), &
-            qflx_in_soil = qflx_in_soil(bounds%begc:bounds%endc), &
             qinmax_on_unsaturated_area = qinmax_on_unsaturated_area(bounds%begc:bounds%endc))
     case default
        write(iulog,*) subname//' ERROR: Unrecognized qinmax_method: ', this%qinmax_method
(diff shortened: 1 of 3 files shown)


src/biogeophys/InfiltrationExcessRunoffMod.F90
```diff
@@ -53,13 +53,22 @@ module InfiltrationExcessRunoffMod
...
 
   ! !PRIVATE DATA MEMBERS:
 
+  ! For methods that don't generate any infiltration excess runoff, we get this end result
+  ! by specifying a huge qinmax value - i.e., an effectively infinite max infiltration
+  ! rate.
+  !
+  ! 1e200 mm H2O /s seems large enough to be effectively infinite, while not being so
+  ! large as to cause floating point overflows elsewhere.
+  real(r8), parameter :: qinmax_unlimited = 1.e200_r8  ! mm H2O /s
```
billsacks: **Comment from @billsacks Sep 13, 2017**

When we talked about this a few weeks ago, there were suggestions of using a global constant here.

Reasons I went with a locally-defined qinmax_huge rather than a globally-defined constant:

- Globally-defined is useful to eliminate the need for, "If I change this here, then I also need to change this over there". That doesn't seem to be the case here:
  - Nobody needs to check this
  - I couldn't see any benefits to having this shared between different parameterizations
- Problems with globally-defined
  - A different value might make more sense in other contexts
  - If another parameterization needs a different value and so changes the global constant, it would change this and other previous parameterizations that use it, which could inadvertently change behavior

billsacks: **Comment from @martynpclark Sep 13, 2017**

Agreed. This seems like a good approach.

billsacks:
**Comment from @martynpclark Sep 13, 2017**

@billsacks -- good description -- the VIC implementation was actually the analytical solution to saturation-excess runoff (which depends on saturated area), not infiltration-excess runoff (which depends on the flux at the upper boundary).

billsacks:
**Comment from @martynpclark Sep 13, 2017**

@billsacks -- we have already got to the point where we are making changes to the numerical solutions in CTSM. The VIC implementation provides an analytical solution, in the same way that we use analytical solutions for the ponded water store.

billsacks:
**Comment from @martynpclark Sep 13, 2017**

The code for the VIC analytical solution is by no means obvious. The derivation is in this pdf file:

[VIC_surfaceRunoff.pdf](https://github.com/ESCOMP/ctsm/files/1592028/VIC_surfaceRunoff.pdf)

src/biogeophys/InfiltrationExcessRunoffMod.F90
```diff
@@ -270,73 +280,4 @@ subroutine ComputeQinmaxHksat(bounds, num_hydrologyc, filter_hydrologyc, &
...
-       else if(max_infil <= 0._r8) then
-          rsurf_vic = qflx_in_soil(c)
-       else if((i_0 + qflx_in_soil(c)*dtime) > max_infil) then             !(Eq.(3a) Wood et al. 1992)
-          rsurf_vic = (qflx_in_soil(c)*dtime - top_max_moist(c) + top_moist(c))/dtime
-       else                                                                !(Eq.(3b) Wood et al. 1992)
-          basis = 1._r8 - (i_0 + qflx_in_soil(c)*dtime)/max_infil
-          rsurf_vic = (qflx_in_soil(c)*dtime - top_max_moist(c) + top_moist(c)    &
-               + top_max_moist(c) * basis**(1._r8 + b_infil(c)))/dtime
-       end if
-       rsurf_vic = min(qflx_in_soil(c), rsurf_vic)
```
billsacks: **Comment from @martynpclark Sep 13, 2017**

It is correct to delete this code here, and now that we have options for analytical numerical solutions, the code block below can be moved to the saturation excess runoff routine:

```fortran
       max_infil = (1._r8+b_infil(c)) * top_max_moist(c)
       i_0 = max_infil * (1._r8 - (1._r8 - fsat(c))**(1._r8/b_infil(c)))
       if(qflx_in_soil(c) <= 0._r8) then
          rsurf_vic = 0._r8
       else if(max_infil <= 0._r8) then
          rsurf_vic = qflx_in_soil(c)
       else if((i_0 + qflx_in_soil(c)*dtime) > max_infil) then             !(Eq.(3a) Wood et al. 1992)
          rsurf_vic = (qflx_in_soil(c)*dtime - top_max_moist(c) + top_moist(c))/dtime
       else                                                                !(Eq.(3b) Wood et al. 1992)
          basis = 1._r8 - (i_0 + qflx_in_soil(c)*dtime)/max_infil
          rsurf_vic = (qflx_in_soil(c)*dtime - top_max_moist(c) + top_moist(c)    &
               + top_max_moist(c) * basis**(1._r8 + b_infil(c)))/dtime
       end if
```
We can use the same logic as in the ponded water (options for explicit Euler [current code] and the analytical solution [this code]).

Again, the pdf document above provides derivations.

src/biogeophys/SaturatedExcessRunoffMod.F90
```diff
@@ -334,6 +334,10 @@ subroutine ComputeFsatVic(bounds, num_hydrologyc, filter_hydrologyc, &
     ! Citation: Wood et al. 1992, "A land-surface hydrology parameterization with subgrid
     ! variability for general circulation models", JGR 97(D3), 2717-2728.
     !
+    ! This implementation gives a first-order approximation to saturated excess runoff.
+    ! For now we're not including the more exact analytical solution, or even a better
+    ! numerical approximation.
+    !
```
billsacks: **Comment from @martynpclark Sep 13, 2017**

We can add the analytical solution here.

billsacks:
**Comment from @billsacks Sep 14, 2017**

@martynpclark I feel like there are still two somewhat unresolved issues, both really boiling down to how much complexity we want to support:

1. I recall that moving this code into the saturated surface excess runoff calculation would mean that different parameterizations for saturated surface excess runoff now potentially differ in more than just fsat. That's fine in principle, but my sense from the Aug 24 ctsm meeting was that nobody was very happy with what that would mean in terms of code complexity, and that we only wanted to go there if we really had to. It's unclear to me whether the vic option is important enough to justify this extra complexity itself, or if this is only worth doing if some other parameterizations also require this.

2. I feel like we haven't totally worked out how we want to select between different solution methods. In this case, I'd personally be more inclined to just _always_ include the more exact analytical solution, rather than providing an option to include it or exclude it, in order to avoid extra levels of complexity. Of course, I'm open to having this be an option, but I feel we should be careful to only include options like this where they're really warranted.

We can continue this discussion here if you want. Or we can defer this discussion to a later date if (as is my impression) this vic stuff isn't the highest priority.

billsacks:
**Comment from @martynpclark Sep 14, 2017**

Hi @billsacks -- I agree. Let's defer the decision until once we converge on the overall structure that we are happy with. Yes, I agree for the operator splitting implementation that we should (ideally) *always* use the more accurate method if it is available -- one question is if there is a desire to have what was present before, since different solutions can lead to different model behavior.

billsacks:
**Comment from @billsacks Sep 14, 2017**

@martynpclark 

> I agree. Let's defer the decision until once we converge on the overall structure that we are happy with.

So any feelings about what I should do in the meantime?:

1. Merge this PR as is (once I have tested it)

2. Hold off on merging this PR until we decide on the final solution

> one question is if there is a desire to have what was present before, since different solutions can lead to different model behavior.

But here, isn't the more accurate method what was there before? (Just now we'd be applying it to saturated surface excess runoff rather than infiltration excess runoff?)

billsacks:
**Comment from @martynpclark Sep 14, 2017**

Hi @billsacks

> So any feelings about what I should do in the meantime?

My preference is to merge this PR as is

> But here, isn't the more accurate method what was there before? (Just now we'd be applying it to saturated surface excess runoff rather than infiltration excess runoff?)

Exactly. My comments were intended to be more general. There may be cases where we add a more accurate solution (as in the case for ponded water), and we still may have the desire to retain what was present before (since different solutions can lead to different model behavior). In these cases we may not *always* want to use the more accurate solution, and we retain capabilities in order to compare against previous model versions.

billsacks:
**Comment from @billsacks Sep 14, 2017**

@martynpclark thanks for the clarifications - this all makes sense now. I'll run testing on this as soon as I'm done with #13

billsacks:
@martynpclark in case it wasn't clear: no action / replies are needed; I just wanted to move over our old conversation to save it in the new repository.

PR #192, opened 2017-12-28
Ponded water/alternative solutions - object-oriented

billsacks:
**NOTE: This PR was originally created in the temporary CTSM repository by @billsacks, Sep 8, 2017, where it was NCAR/clm-ctsm#14. I am moving over most of the comments from there: both my own and those of @martynpclark .**

**Original comment from @billsacks Sep 8, 2017 (with references to PRs and commits changed for this new repository)**

NOTE: This PR has not been tested (though it does compile). It is for discussion only.

This is a counterpart to #191 , but with an object-oriented implementation.

In order to facilitate comparison with @martynpclark 's implementation, I started with his changes, and then made changes on top of that. To see just what is different in the object-oriented version, see 98da3bfd . You can see from this that the two versions are very similar. The biggest addition in the OO version is the declaration of a type, which holds input arguments and a type-bound procedure declaration:

```fortran
  type, extends(flux_type) :: drainPond_type
     real(r8) :: smoothScale
     real(r8) :: drainMax
   contains
     procedure :: getFlux => drainPondFlux
  end type drainPond_type
```

I prefer this version over the internal subroutine for the reasons I laid out here: https://github.com/ESCOMP/ctsm/pull/191#issuecomment-354360438

But I'm open to discussion.

Compared with @bandre-ucar 's suggestion: I think the two would be pretty similar. I'm not sure I see any real downsides of this one... Ben's just wouldn't have drainPondFlux be a type-bound procedure. But making drainPondFlux a type-bound procedure opens some additional doors that I hope to show in some upcoming commits.

@mvertens I think you were also interested to see this.

### DIFF
--- a/src/biogeophys/SoilHydrologyMod.F90
+++ b/src/biogeophys/SoilHydrologyMod.F90
@@ -29,8 +29,8 @@ module SoilHydrologyMod
   use ColumnType        , only : col                
   use PatchType         , only : patch                
 
-  use computeFluxMod                            ! template for the flux module
-  use implicitEulerMod  , only : implicitEuler  ! implicit Euler solver for a single state
+  use computeFluxMod    , only : computeFlux_type ! template for the flux module
+  use implicitEulerMod  , only : implicitEuler    ! implicit Euler solver for a single state
 
   !
   ! !PUBLIC TYPES:
@@ -58,7 +58,7 @@ module SoilHydrologyMod
   private :: QflxH2osfcSurf      ! Compute qflx_h2osfc_surf
   private :: QflxH2osfcDrain     ! Compute qflx_h2osfc_drain
 
-  type, extends(flux_type) :: drainPond_type
+  type, extends(computeFlux_type) :: drainPond_type
      real(r8) :: smoothScale
      real(r8) :: drainMax
    contains
--- a/src/biogeophys/SoilHydrologyMod.F90
+++ b/src/biogeophys/SoilHydrologyMod.F90
@@ -28,6 +28,10 @@ module SoilHydrologyMod
   use LandunitType      , only : lun                
   use ColumnType        , only : col                
   use PatchType         , only : patch                
+
+  use computeFluxMod                            ! template for the flux module
+  use implicitEulerMod  , only : implicitEuler  ! implicit Euler solver for a single state
+
   !
   ! !PUBLIC TYPES:
   implicit none
@@ -581,8 +585,16 @@ contains
      logical           , intent(inout) :: truncate_h2osfc_to_zero( bounds%begc: ) ! whether h2osfc should be truncated to 0 to correct for roundoff errors, in order to maintain bit-for-bit the same answers as the old code
      !
      ! !LOCAL VARIABLES:
-     integer :: fc, c
-     real(r8) :: dtime         ! land model time step (sec)
+     procedure(fluxTemplate), pointer :: funcName ! function name
+     integer             :: fc, c               
+     real(r8)            :: dtime                 ! land model time step (sec)
+     real(r8)            :: h2osfc1               ! h2osfc at the end of the time step, given qflx_h2osfc_drain only
+     real(r8)            :: drainMax              ! maximum drainage rate of ponded water
+     real(r8)            :: const                 ! constant in analytical integral
+     real(r8)            :: uFunc                 ! analytical integral
+     real(r8), parameter :: smoothScale=0.05_r8   ! smoothing scale
+     integer(i4b)        :: err                   ! error code
+     character(len=128)  :: message               ! error message
 
      character(len=*), parameter :: subname = 'QflxH2osfcDrain'
      !-----------------------------------------------------------------------
@@ -603,9 +615,31 @@ contains
            truncate_h2osfc_to_zero(c) = .true.
         else
 
+           ! define maximum drainage
+           ! NOTE: check fraction
+           drainMax = frac_h2osfc(c)*qinmax(c)
 
+           ! switch between different numerical solutions
+           select case(ixSolution)
 
-           qflx_h2osfc_drain(c)=min(frac_h2osfc(c)*qinmax(c),h2osfc(c)/dtime)
+              ! constrained Explicit Euler solution with operator splitting (original)
+              case(ixExplicitEuler)
+                 qflx_h2osfc_drain(c)=min(drainMax,h2osfc(c)/dtime)
+
+              ! implicit Euler solution with operator splitting
+              case(ixImplicitEuler)
+                 funcName=>drainPond
+                 call implicitEuler(funcName, dtime, h2osfc(c), h2osfc1, err, message)
+                 if(err/=0) call endrun(subname // ':: '//trim(message))
+                 call drainPond(h2osfc1, qflx_h2osfc_drain(c))
+
+              ! analytical solution with operator splitting
+              case(ixAnalytical)
+                 const   = 1._r8 - 1._r8/(1._r8 - exp(-h2osfc(c)/smoothScale))
+                 uFunc   = -1._r8/(const*exp(drainMax*dtime/smoothScale) - 1._r8)
+                 h2osfc1 = -alog(1._r8 - uFunc)*smoothScale
+                 qflx_h2osfc_drain(c)= (h2osfc1 - h2osfc(c))/dtime
+
            if(h2osfcflag==0) then
               ! ensure no h2osfc
               qflx_h2osfc_drain(c)= max(0._r8,h2osfc(c)/dtime)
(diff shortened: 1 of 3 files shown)


billsacks:
**Comment from @martynpclark Sep 8, 2017**

@billsacks -- very nice.

src/utils/computeFluxMod.F90
```diff
@@ -5,24 +5,21 @@ module computeFluxMod
   use shr_kind_mod      , only : r8 => shr_kind_r8
   implicit none
 
+  type, abstract :: flux_type
```
billsacks: **Comment from @martynpclark Sep 8, 2017**

This is pedantic, I know, though can we call this `computeFlux_type` (or something similar)? The reason is that it is easy to confuse `flux_type` with `WaterFlux_type` and `EnergyFlux_type`, and folks could assume that the type only includes fluxes (whereas, in fact, it will include all variables required to compute fluxes).

billsacks: **Comment from @billsacks Sep 11, 2017**

Fixed - thanks. Not pedantic at all - this is a very good point.

billsacks:
**Comment from @billsacks Sep 11, 2017, edited to point to PRs and commits in this new repo**

One thing that was bothering me a bit with this prototype (and the same issue exists in the prototype in #191 ) is the extra complexity introduced into the science routines by having the alternative numerical solutions inline along with the science. With these long select case statements embedded in the science routines, it gets hard to see what's really going on at a higher level.

ea1549007310d67d4ae515aee4747ffc74623408 shows one possible partial solution to this problem. Downsides are: (1) There's a bit more boiler-plate needed to define the different subroutines; (2) the extra subroutine call could lead to a loss of efficiency, though I hope that the compiler is smart enough to inline this. (And (2) should be less of an issue – at least for global runs – once we rewrite this so the flux calculation routines operate on arrays rather than single points.) That said, I still like this solution because it makes it easier to see the high-level structure of the main routine without extra clutter. I'd be curious what others think about this.

billsacks:
**Comment from @billsacks Sep 12, 2017**

I was trying for some improvements in #193 , but I don't like how that turned out, so I'm closing that PR. (I just wanted to use the PR mechanism to record my ideas in case I want to come back to it later.)

billsacks:
**Comment from @billsacks Sep 13, 2017**

For this and similar situations: Probably it would make sense to move the `QflxH2osfcDrain` subroutine into the `drainPond_type` class. This would pull together all related routines into this class, and would allow following the convention (which we're trying to enforce in CLM) that instance variables in a class (even public ones) should only be set by methods of that class.

billsacks:
**Comment from @billsacks Sep 13, 2017, edited to refer to the PR in this new repo**

I was thinking about where the drainPond_inst instance should be stored, once we move to having this work on arrays rather than one element at a time. We want to avoid doing dynamic memory allocation in every time step.

So that leaves us with two options:

1. Allocate space for it and store it somewhere in initialization

2. Keep it declared as a local variable in the calling subroutine. Make it a parameterized derived type for the array lengths (i.e., parameterized on bounds%endc, etc.)

(1) is required if we want to save any history variables here, or if we want to do the namelist read in initialization in this class. This is more in line with what we do elsewhere in CLM.

(2) is more efficient in terms of memory (because the necessary memory doesn't need to stick around for the whole run) and _may_ be more efficient in time (performance of stack vs. heap memory). But this will only work right now (until we move to always using array lower bounds of 1) if we can specify array lower bounds in parameterized derived types (I'm not sure if this is allowed).

(Note that the notion of what the calling subroutine is would change if we followed the suggestion of https://github.com/ESCOMP/ctsm/pull/192#issuecomment-354388437 : In this case, the calling subroutine would become `UpdateH2osfc`, so the instance would become a local variable to that subroutine.)

billsacks:
**Comment from @billsacks Sep 13, 2017**

I was thinking more about https://github.com/ESCOMP/ctsm/pull/192#commitcomment-26526545 . I do like the function pointer idea, the start of whose implementation could look like this:

```diff
diff --git a/components/clm/src/biogeophys/SoilHydrologyMod.F90 b/components/clm/src/biogeophys/SoilHydrologyMod.F90
index 88b4572..ff9d9a4 100644
--- a/components/clm/src/biogeophys/SoilHydrologyMod.F90
+++ b/components/clm/src/biogeophys/SoilHydrologyMod.F90
@@ -65,11 +65,18 @@ module SoilHydrologyMod
      real(r8) :: smoothScale = 0.05_r8  ! smoothing scale
    contains
      procedure :: getFlux => drainPondFlux  ! required method to get the current flux estimate
-     procedure :: drainPondExplicitEuler    ! compute the drainPond flux using an explicit euler method
-     procedure :: drainPondImplicitEuler    ! compute the drainPond flux using an implicit euler method
-     procedure :: drainPondAnalytical       ! compute the drainPond flux using an analytical solution
+     procedure(computeDrainPond_interface), pointer :: computeDrainPond
   end type drainPond_type

+  abstract interface
+     subroutine computeDrainPond_interface(this, h2osfc, qflx_h2osfc_drain)
+       import :: drainPond_type
+       class(drainPond_type), intent(in) :: this
+       real(r8) , intent(in)  :: h2osfc            ! drainPond storage
+       real(r8) , intent(out) :: qflx_h2osfc_drain ! drainage flux
+     end subroutine computeDrainPond_interface
+  end interface
+
   !-----------------------------------------------------------------------
   real(r8), private :: baseflow_scalar = 1.e-2_r8
```

However, this only makes sense if we have a place to store the drainPond_inst instance persistently, so that it's set up in initialization (rather than having it be a local variable to the calling subroutine - in which case we'd need to recreate the function pointer every time step). We may want to solve that problem anyway, in order to (a) allocate necessary arrays, and (b) do a namelist read in that class. See also the related comment https://github.com/ESCOMP/ctsm/pull/192#issuecomment-354388562

src/biogeophys/SoilHydrologyMod.F90
```diff
@@ -55,9 +59,22 @@ module SoilHydrologyMod
...
+
   !-----------------------------------------------------------------------
   real(r8), private :: baseflow_scalar = 1.e-2_r8
 
+  ! !ALTERNATIVE NUMERICAL SOLUTIONS
+  integer, parameter :: ixExplicitEuler=1001  ! constrained Explicit Euler solution with operator splitting (original)
+  integer, parameter :: ixImplicitEuler=1002  ! implicit Euler solution with operator splitting
+  integer, parameter :: ixAnalytical=1003     ! analytical solution with operator splitting
+  integer  :: ixSolution=ixExplicitEuler
+
```
billsacks: **Comment from @martynpclark Sep 13, 2017**

We should talk about how this is implemented. In principle it is possible to have multiple different numerical solutions for every single flux. This would provide scope to mix-and-match different numerical solutions in different parts of the code (sort of the way that CLM is developed now, although CLM does not have multiple numerical options for each flux). This would be incredibly cumbersome and likely confusing. At the other end of the spectrum we use the same numerical options everywhere in the code. This could be too restrictive.

I initially included these alternative options as (1) an example of how we can increase the accuracy of the operator splitting constrained explicit Euler solution without any increase in cost, expecting that there may be cases where people desire to retain operator splitting; and (2) begin the process of separating the physics from the numerical solution.

In my opinion the alternative numerical options for a single flux (as implemented here) may not be a representative use case.

billsacks: **Comment from @billsacks Sep 14, 2017**

I'd also like to discuss this point further. And I'd be interested to see a more representative use case at some point.

One other concern I have is: If each flux has multiple possible parameterization options, each with multiple numerical solution options, the code is going to get very complex. (Much of the work I was doing earlier this week was to try to play around with ways to reduce that complexity, but there's no way to remove it entirely.)

Originally I had imagined that, for the most part, the choice of numerical solution method would just appear at high level in the code, not down at the level of individual fluxes. That would reduce this problem of a combinatorial explosion of complexity. I think that's what you're saying, too.

billsacks:
**Comment from @billsacks Sep 14, 2017, edited to point to commits in this new repo**

The general feeling from today's CTSM meeting was that people were happy with the main changes to make this object oriented - i.e., the change in 98da3bfd216872c6ce811bf97ff9493b68f0c964 . However, at least some people (e.g., @mvertens ) felt that the changes in ea1549007310d67d4ae515aee4747ffc74623408 had more negatives than positives.

(Note that the changes in the commits between 98da3bfd216872c6ce811bf97ff9493b68f0c964 and ea1549007310d67d4ae515aee4747ffc74623408 involved minor cleanup that @martynpclark and I discussed already. This means that people are happy to go with something like b872d2138d996dd2a6f87413240aceee5090da91, which is the last commit before ea1549007310d67d4ae515aee4747ffc74623408.)

billsacks:
**Comment from @billsacks Sep 14, 2017**

Okay, I have reverted the commit that people didn't like.

billsacks:
**Comment from @billsacks Sep 14, 2017**

@martynpclark I have updated this branch to include the latest changes on the main 'ctsm' branch, which include my reworked truncation code for h2osfc. So you can now base your next changes off of this branch. Feel free to either work directly on this branch or to create your own branch off of this one (at which point you could do a PR back into this branch to highlight your changes).

billsacks:
I'm closing this PR that was mainly opened for discussion purposes. I'm deleting the  pondedWater/alternativeSolutions_oo branch from the escomp/ctsm repository, but I'm pushing it to billsacks/ctsm (and it should also be available from the escomp repo for a while by checking out the last commit: 9133ffd8c9c9836e612194f251c8157ebbfe34b4).

PR #214, opened 2018-01-17
Use cimeroot rather than hard-coded relative path

billsacks:
This is needed to support a standalone checkout, where cime appears in a
different location relative to the bld directory

### DIFF
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -59,6 +59,7 @@ SYNOPSIS
 
      Create the namelist for CLM
 REQUIRED OPTIONS
+     -cimeroot "directory"    Path to cime directory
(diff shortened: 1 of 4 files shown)


billsacks:
@olyson and @swensosc - these changes should fix the problems you had running `preview_namelist` or `case.build` related to not being able to find perl5lib. I'm running some tests now, but so far everything looks good. So once @ekluzek reviews this and @bandre-ucar merges this, you can retry your testing.

billsacks:
Okay, both a clm50 and clm40 test built and ran successfully with a ctsm clone in a deeply-nested directory where it couldn't possibly find any cime directories using a relative path. @bandre-ucar this is ready to merge when you get a chance.

bandre-ucar:
I built and ran a clm5 test on the merged branch it is should be usable now.

PR #228, opened 2018-01-28
Get tools and their testing working as well as build-namelist testing

ekluzek:
These changes get the tools and their testing working as well as the build-namelist testing.  Currently it doesn't get PTCLM working though.


.gitignore
```diff
@@ -27,6 +27,24 @@ CMakeFiles/
 # don't ignore cism build utilities directory
 !components/cism/**/build
 
+# configure / build files
```
billsacks: I assume that all of these new entries are referring to files that appear in specific places in the directory tree? If so, the rules should be more specific, listing the exact paths where these files occur.

Also: I started doing this by having multiple `.gitignore` files in different places. However, I was just doing some googling, and it sounds like the most common convention is to put everything in this top-level .gitignore file. So I'd like to rework things to do that. So, for example, this entry in `src/unit_test_stubs/utils/.gitignore`: `restUtilMod_stub.F90` would go in the top-level as `/src/unit_test_stubs/utils/restUtilMod_stub.F90` (I think it's correct to have this leading slash here: I think this says to match exactly this path). I can do this myself in a following PR if you don't want to deal with it.

bld/queryDefaultNamelist.pl
```diff
@@ -34,7 +34,8 @@
 
 #-----------------------------------------------------------------------------------------------
 # Add $cfgdir to the list of paths that Perl searches for modules
-my @dirs = ( $cfgdir, "$cfgdir/../../../cime/utils/perl5lib" );
+my @dirs = ( $cfgdir, "$cfgdir/../../../cime/utils/perl5lib",
```
billsacks: Is this going to first look in `$cfgdir/../../../cime/utils/perl5lib`, then look in `$cfgdir/../cime/utils/perl5lib`? If so, I'd change the order of these. My rationale is: If you're within a standalone checkout, there's a chance that there would be an unrelated cime directory at `$cfgdir/../../../cime/utils/perl5lib`.

tools/mksurfdata_map/mksurfdata.pl
```diff
@@ -28,6 +28,7 @@
 #-----------------------------------------------------------------------------------------------
 # Add $scrdir to the list of paths that Perl searches for modules
 my @dirs = ( $scrdir, "$scrdir/../../../../cime/utils/perl5lib",
+             "$scrdir/../../cime/utils/perl5lib"
```
billsacks: As above, I'd change the order of these search paths.

PR #248, opened 2018-02-01
Fix documentation of where SMB is computed for vegetated columns

billsacks:
I changed this in the code a while ago, but never updated this
documentation


billsacks:
@ekluzek would it be easy for you to fold this into one of your upcoming tags? If not, I can get it in later. I don't want to hold you up at all, so only include this if doing so would be really quick and easy.

PR #250, opened 2018-02-02
update snow overburden compaction parameter

lvankampenhout:
Snow viscosity for dry compaction was changed around 2016 and is described in more detail by (Van Kampenhout et al., 2017). Basically we adopted the viscosity expression from Crocus (Vionnet et al, 2012) and changed one parameter (c_eta) to better fit firn core observations. The original value of the c_eta parameter in Vionnet2012 is 250 kg/m3. In our study we tested different values (like 450) before arriving at the final value of 358 kg/m3. In this commit we change the value of c_eta from 450 to 358 kg/m3.

### DIFF
--- a/src/biogeophys/SnowHydrologyMod.F90
+++ b/src/biogeophys/SnowHydrologyMod.F90
@@ -1905,7 +1905,7 @@ contains
     real(r8) :: f1, f2                          ! overburden compaction modifiers to viscosity
     real(r8) :: eta                             ! Viscosity
 
-    real(r8), parameter :: ceta = 450._r8       ! overburden compaction constant [kg/m3]
+    real(r8), parameter :: ceta = 358._r8       ! overburden compaction constant [kg/m3]
     real(r8), parameter :: aeta = 0.1_r8        ! overburden compaction constant [1/K]
     real(r8), parameter :: beta = 0.023_r8      ! overburden compaction constant [m3/kg]
     real(r8), parameter :: eta0 = 7.62237e6_r8  ! The Viscosity Coefficient Eta0 [kg-s/m2]


lvankampenhout:
FYI, this commit will be answer-changing but only to a very minor degree. Overburden compaction is a relatively weak force that only plays a dominant role in thick snow packs (> 1 m)

billsacks:
@lvankampenhout thanks for this. Have you run this by @dlawrenncar or others to determine the timing of bringing this to master? I'm particularly wondering if this is supposed to make it into CESM2, or if this is post-CESM2. If you haven't talked with @dlawrenncar about it yet, then I'm guessing this change will come in post-CESM2, since science is supposed to basically be frozen at this point.

Also, assuming this is for post-CESM2: We'll also want to consider pulling this out to a namelist parameter, so that its value can remain as is for CLM5.0 physics, but change for future physics versions.

dlawrenncar:
That's correct.  This should be a post-CESM2 change.

lvankampenhout:
OK, post-release is fine by me. In the Tech Note I wrote c_eta = 358 though (the new value) is that a problem? In that case we may want to change that to 450 to make it consistent to the code.

billsacks:
Yes, probably this should be changed in the tech note. I think it's not 
too late; can you make that change?

Thanks,
Bill

lvankampenhout:
OK I just made the change in the Tech note: https://github.com/ESCOMP/ctsm-docs/commit/073a899ece636222d52a0b04efa2965bf5e4e52d

billsacks:
@lvankampenhout sorry for the silence here; we haven't forgotten about this. We just discussed it at our ctsm software meeting today, and felt that we'll want to

- Introduce a new phys option (like clm4_5 and clm5_0, though we haven't figured out what to call the new one yet), triggered by compset

- Pull this parameter out to namelist

- Have the defaults for the parameter depend on the phys option (so clm5_0 would give the same science as before)

This might all take a while, so let us know if you need this sooner rather than later for any reason.

lvankampenhout:
Hi Bill,
No need to change this fast. For most runs I expect this will only make a marginal difference anyway (see my second post at the top). It's just a slight nuisance that right now the code is inconsistent with the paper that we published last year. 

Personally I feel that pulling this out to the namelist is somewhat overkill, but if you guys decided this is desirable, I won't object that.

adamrher:
Leo, we are configuring the refined arctic simulations and am wondering if you still feel the c_eta value should be changed to 358 kg/m3. We are trying to improve the surface mass balance over greenland so if the 358 value is more realistic then I would favor this change.

lvankampenhout:
Hi Adam, making this change is not going to do much, as it's only a slight deviation on a parameter of compaction. You can still try if it's not not a hassle. 

Much more important for melt are the parameters for snow initial grain size / refreezing grain size / grain growth rate / irreducible water content / bare ice albedo, all of which are only loosely constrained.

wwieder:
@oleson, can you add this to the hard coded parameter bfb tag you're making.
Then we'll need to update to the value for `ceta` suggested here

ekluzek:
This work is actually completed now with our updated paramsfile that sets ceta to 358 in ctsm5.1. This is the case in ctsm5.1.dev021. So the equivalent of this is done. So I'm closing this.

PR #305, opened 2018-02-23
Add tests for aux_cime_baselines edison tests for fates

ekluzek:
Add tests for aux_cime_baselines from the most highly tested cases.
And add some edison tests for fates test list, again for the most highly tested cases


ekluzek:
OK, I made some changes so SMS tests have preference and did three major different configurations 2000Clm45Sp, 1850CLM50BgcCrop and HistClm50BgcCrop with isotopes. I think those are three pretty major and quite divergent types of things to do.

@bandre-ucar you have a valid point about edison. We'll see if we actually do exercise testing there, we may not. Using a different test list when you don't test all the time, does make some sense. Since, Ryan and Charlie asked for it I'm going to leave it in. If it isn't used we should go ahead and remove it. So hopefully, I'll remember about that. But, yes we should evaluate if it does actually get used.

PR #307, opened 2018-02-27
Bug fixes for energy imbalance associated with surface water and lakes

olyson:
These are bug fixes for the land energy imbalance over land as determined by coupler diagnostics.  They include bug fixes for surface water phase change and lake/snow interactions developed by Sean Swenson and Keith Oleson.
The following modules have been modified:
src/biogeophys/EnergyFluxType.F90
src/biogeophys/SoilFluxesMod.F90
src/biogeophys/SoilTemperatureMod.F90
src/biogeophys/SurfaceAlbedoMod.F90
src/biogeophys/SurfaceRadiationMod.F90
Diagnostics compared to an offline control are here:
http://webext.cgd.ucar.edu/I1850/clm50_r272_1deg_GSWP3V1_iso_h2osfclakefix_1850/lnd/clm50_r272_1deg_GSWP3V1_iso_h2osfclakefix_1850.1443_1461-clm50_r272_1deg_GSWP3V1_iso_1850.1443_1461/setsIndex.html


billsacks:
@olyson - I know the distinction between assignee and reviewer is a bit unclear. Typically it works best to have a single assignee and multiple reviewers.

src/biogeophys/EnergyFluxType.F90
```diff
@@ -21,6 +21,9 @@ module EnergyFluxType
   type, public :: energyflux_type
 
      ! Fluxes
+!scs
```
billsacks: Need to remove comment lines like this, and other similar comments (scs comments and other commented-out lines).

@swensosc and @olyson - for the future, it makes code reviews easier if you can remove these commented-out things and commented initials before submitting the PR.

src/biogeophys/SurfaceRadiationMod.F90
```diff
@@ -664,7 +666,9 @@ subroutine SurfaceRadiation(bounds, num_nourbanp, filter_nourbanp, &
                 sabg_soil(p) = sabg(p)
              endif
              ! if no subgrid fluxes, make sure to set both components equal to weighted average
-             if (subgridflag == 0) then 
+!             if (subgridflag == 0) then
+!scs
+             if (subgridflag == 0 .or. lun%itype(l) == istdlak) then
```
billsacks: @swensosc There is one other occurrence of subgridflag, on line 763 of this module. Should that one also check if the landunit type is istdlak?

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -891,15 +916,15 @@ subroutine PhaseChangeH2osfc (bounds, num_nolakec, filter_nolakec, &
...
+!scs
+            hm(c) = frac_h2osfc(c)*(dhsdT(c)*tinc - tinc*c_h2osfc(c)/dtime)
+!scs
 
             ! mass of water converted from liquid to ice
             xm(c) = hm(c)*dtime/hfus  
             temp1 = h2osfc(c) + xm(c)
 
-            ! compute change in cv due to additional ice
-            dcv(c)=cpice*min(abs(xm(c)),h2osfc(c))
```
billsacks: Should also remove the declaration of this no-longer-used local variable, dcv

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -919,17 +944,51 @@ subroutine PhaseChangeH2osfc (bounds, num_nolakec, filter_nolakec, &
...
+!                     c2=frac_h2osfc(c)*(-cpliq*xm(c))
+                     c2=(-cpliq*xm(c))
+                  else
+                     c2=0.0_r8
+                  end if
+                  t_soisno(c,0) = (c1*t_soisno(c,0)+ c2*t_h2osfc(c)) &
+                       /(c1 + c2)             
+
+                  eflx_h2osfc_to_snow_col(c) =(t_h2osfc(c)-t_soisno(c,0))*c2/dtime
+               else
```
billsacks: It looks like this `else` block is exactly the same as the above `else if` block, except that here there is no `dhsdT` in the calculation of `c1`. To reduce this code duplication, can you do something like this?:

```Fortran
               if (snl(c) == 0) then
                  !initialize for next time step
                  t_soisno(c,0) = t_h2osfc(c)
                  eflx_h2osfc_to_snow_col(c) = 0.
               else
                  if (snl(c) == -1) then
                     somevar = dhsdT(c)
                  else
                     somevar = 0._r8
                  end if
                  c1=frac_sno(c)*(dtime/fact(c,0) - somevar*dtime)
                  ! Rest of the code here
```

swensosc:
or:

c1=frac_sno(c)*(dtime/fact(c,0)
if (snl(c) == -1) c1 = c1 - frac_sno(c)*dhsdT(c)*dtime

then no need for additional variable (?)

billsacks:
> or:
> 
> c1=frac_sno(c)*(dtime/fact(c,0)
> if (snl(c) == -1) c1 = c1 - frac_sno(c)*dhsdT(c)*dtime
> 
> then no need for additional variable (?)

Yes, that's fine, too

src/biogeophys/SurfaceRadiationMod.F90
olyson: I've pushed this change to the branch

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -919,17 +944,50 @@ subroutine PhaseChangeH2osfc (bounds, num_nolakec, filter_nolakec, &
...
+!                     c2=frac_h2osfc(c)*(-cpliq*xm(c))
+                     c2=(-cpliq*xm(c) - frac_h2osfc(c)*dhsdT(c)*dtime)
+                  else
+                     c2=0.0_r8
+                  end if
+                  t_soisno(c,0) = (c1*t_soisno(c,0)+ c2*t_h2osfc(c)) &
+                       /(c1 + c2)             
+
+                  eflx_h2osfc_to_snow_col(c) =(t_h2osfc(c)-t_soisno(c,0))*c2/dtime
+               else
```
ekluzek: Also note that there is a division by "fact" that isn't being checked for. Can fact be small or zero?

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1221,7 +1281,14 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
...
                      if (j == snl(c)+1) then ! top layer                   
-                        hm(c,j) = dhsdT(c)*tinc(c,j) - tinc(c,j)/fact(c,j)
+!                        hm(c,j) = dhsdT(c)*tinc(c,j) - tinc(c,j)/fact(c,j)
+!scs
+                        if(j > 0) then
+                           hm(c,j) = dhsdT(c)*tinc(c,j) - tinc(c,j)/fact(c,j)
+                        else
+                           hm(c,j) = frac_sno_eff(c)*(dhsdT(c)*tinc(c,j) - tinc(c,j)/fact(c,j))
+                        endif
+!scs
```
ekluzek: More divides by fact without checking.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -793,7 +802,14 @@ subroutine SoilThermProp (bounds,  num_nolakec, filter_nolakec, &
...
             if (snl(c)+1 < 1 .and. j >= snl(c)+1) then
-               cv(c,j) = cpliq*h2osoi_liq(c,j) + cpice*h2osoi_ice(c,j)
+!               cv(c,j) = cpliq*h2osoi_liq(c,j) + cpice*h2osoi_ice(c,j)
+!scs
+               if (frac_sno(c) > 0._r8) then
+                  cv(c,j) = max(1.0e-6_r8,(cpliq*h2osoi_liq(c,j) + cpice*h2osoi_ice(c,j))/frac_sno(c))
+               else
+                  cv(c,j) = 1.0e-6_r8
+               endif
+!scs
```
ekluzek: Probably should check for frac_sno > small value rather than greater than zero.

swensosc:
I think fact will always be nonzero.  It is basically the soil moisture,
multiplied by a couple of constants.

ekluzek:
The change I made to get past the divide by zero, does change answers. So I will need to rerun testing at minimum.

olyson:
The divide by zero fix looks ok to me, Erik.

src/biogeophys/SoilTemperatureMod.F90
olyson: You are probably right but I think Sean should weigh in on this.

swensosc: Yes, that's fine.

billsacks:
Thanks for doing this cleanup, @ekluzek . It looks like most of my comments are addressed now; the one remaining one is https://github.com/ESCOMP/ctsm/pull/307#discussion_r171079729 . However, given the time-criticality of this, I'll go ahead and mark this as accepted - and if it doesn't get resolved this time around, we can come back to it later.

ekluzek:
I'm having a bunch of problems fully completing testing on cheyenne. But, I'm only see expected fails, and testing is complete on hobart. So going to bring this to master and complete testing afterwards.

PR #327, opened 2018-03-16
fates-hydro boundary index fix

rgknox:
This PR fixes a bug in the indexing of the soil to root flux boundary condition from CLM to fates-hydro.  The input array to fates is indexed by soil layers, but the providing array was not the same dimension, thus bounds were needed instead of open indexing, ie "(:)".

Note: fates-hydro is only stub code.  This should have no impact on any code covered by active tests right now.

Collaborators: found and corrected originally by @xuchongang

Expected Changes:  No changes to user environment or workflow. This change also only effects stub code in FATES.  Fates-hydro is still being tested.  No answer changes expected.

Tests: TBD

### DIFF
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -2192,7 +2192,7 @@ contains
     
     do s = 1, this%fates(nc)%nsites
        c = this%f2hmap(nc)%fcolumn(s)
-       waterflux_inst%qflx_rootsoi_col(c,:) = this%fates(nc)%bc_out(s)%qflx_soil2root_sisl(:)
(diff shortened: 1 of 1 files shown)


rgknox:
This PR is preparations for full integration of fates-hydro:  https://github.com/NGEET/fates/pull/343

ekluzek:
@rgknox I took a look at this, and it looks like this is something that should just come in ASAP. Right? It isn't going to impact testing as it's not exercised, but it does correct a potential memory issue which we should just get corrected right away. Does that sound right?

rgknox:
Yeah, simple yet important fix once we start running fates-hydro.  It should not touch any current tests.

rgknox:
We will be integrating plant hydraulics into NGEET/FATES over the next couple of days.  Is there anything else required on my side for this PR?

rgknox:
This PR is superceded by #375 .

PR #328, opened 2018-03-24
Fixes for init_interp

### LINKED ISSUE #325: init_interp can change answers over the CISM domain
Our goal has been that, if you interpolate (via `use_init_interp`) a restart file from a given configuration onto an essentially identical configuration, answers should be bit-for-bit. That is, it shouldn't matter whether or not you ran the restart file through `init_interp`. I discovered that this is not always true for glacier points in configurations where CISM is included (as is the case for most of our configurations now), even if CISM is being run in `NOEVOLVE` mode.

The current operation of interpolation for glacier columns basically follows the design laid out a few years ago here: http://bugs.cgd.ucar.edu/show_bug.cgi?id=1447 , with the most relevant pieces copied below:

> We should extend interpinic so that, for glc_mec columns, it takes data from the input grid that had the closest topographic height. Our current thinking is to use topographic height just as a tie-breaker, as in the following algorithm:
>
> - determine the actual topographic height of this output column; call this T
>
> 
> - find the closest* column (in x-y space) whose landunit type is glc_mec. Note here that there is no restriction on the elevation class: we simply find the closest* glc_mec column in x-y space. 
> 
> - definition of closest*: This is the key: In practice, there will often be multiple columns that are equally close. Currently in the code, ties are broken by giving precedence to the first column found. But now, the actual topographic height of the columns will be the tie breaker: If two columns are equally close in x-y space, then the one whose topographic height is closer to that of the output column will be given precedence. 

and

> Note that my proposal has the important implication that an output column of elevation class N will NOT necessarily get its data from elevation class N in the input (even if the number of elevation classes is the same in both the input & output).
> 
> For example, cases where elevation class N in the output will get data from elevation class M in the input (M /= N) are:
> 
> (1) The closest grid cell in x-y space has data for elevation class M, but NOT for elevation class N. In this case we would take the data from elevation class M since it's closest horizontally, even though it's further vertically. (The alternative would be to keep searching further away for the appropriate elevation class.)
> 
> (2) The closest grid cell in x-y space has data both for elevation class N and elevation class M. However, we have a situation like this:
> 
> EC N: 200 - 400 m
> EC M: 400 - 700 m
> 
> In output, grid cell i: EC N has a mean elevation of 399 m
> 
> In input, closest grid cell to grid cell i: EC N has a mean elevation of 201 m, EC M has a mean elevation of 401 m.
> 
> Then, for better or for worse, EC N in the output would get its data from EC M in the input, because its mean elevation is much closer to that of EC M.
> 
> 
> --
> 
> But what makes my original proposal appealing is that it's simple to implement, and allows using the same general rule whether we're mapping from 10 EC -> 10 EC, from 5 EC -> 10 EC, from 10 EC -> 5 EC, from 10 EC -> 10 EC but with different definitions of the 10 elevation classes, etc. It would also handle the case where we have adaptive elevation classes - i.e., each grid cell has its own elevation classes that are defined based on the range of elevations in that grid cell, as Jeremy Fyke has suggested doing.

The problem is that, over the CISM domain, CTSM updates its glacier areas and elevations at the start of the first time step, rather than at initialization. (Ideally, we'd want this to happen in initialization, but it currently isn't possible to get this information through the coupler in initialization. Once we have a NUOPC-based coupler, I hope to be able to get this information from CISM to CTSM in initialization. That would simplify this issue and some others.) So, at the moment when init_interp is run, glacier column topographic heights come from the surface dataset. As described above, each destination point then finds the closest source point in both x-y space and vertical (topographic) space. However, the destination topographic heights have temporary values at this point. So this problem noted above:

> Note that my proposal has the important implication that an output column of elevation class N will NOT necessarily get its data from elevation class N in the input (even if the number of elevation classes is the same in both the input & output).

kicks in in some cases – even though, once the run gets underway (and so CTSM has received topographic height information from CISM), the topographic heights in the new run will exactly match those used in the run to generate the finidat file.

In an f09 case, it looks like this problem only affects a few grid cells in Greenland, and the scientific implications of this are pretty minor. However, for the sake of ensuring reproducibility and insensitivity to whether init_interp is run, I'd like to fix this problem.

### LINKED ISSUE #326: init_interp can change answers for absorbed radiation in non-existent snow layers
Our goal has been that, if you interpolate (via `use_init_interp`) a restart file from a given configuration onto an essentially identical configuration, answers should be bit-for-bit. That is, it shouldn't matter whether or not you ran the restart file through `init_interp`. However, I discovered a problem for the four variables `flx_absdv`, `flx_absdn`, `flx_absiv` and `flx_absin`, in snow layers: Sometimes these have non-zero values in the layer above the current-top snow layer, and it appears that these values in non-existent snow layers can affect answers. The current snow interpolation code (in `initInterpMultilevelSnow.F90`) only copies values for existing snow layers, so these absorbed radiation fluxes in non-existent layers do not get copied, resulting in changes in answers when using `use_init_interp`.

Note that these `flx_abs` terms are computed in subroutine SurfaceAlbedo (called from clm_driver, line 987, if doalb, over inactive and active points), and are used in subroutine SurfaceRadiation (called from clm_driver, line 470).

I'm not positive why these `flx_abs` values in non-existent snow layers matter, but my best guess is that it's related to the fact that SurfaceAlbedo is only called when doalb is true, not necessarily every time step. I'm imagining: Albedo calculations are done at the end of time step n. In time step n+1, a snow layer melts, but albedo calculations aren't redone. Then, in time step n+1, SurfaceRadiation is called. This seems to be supported by this comment in SurfaceRadiation:

```fortran
             ! Error handling: The situation below can occur when solar radiation is 
             ! NOT computed every timestep.
             ! When the number of snow layers has changed in between computations of the 
             ! absorbed solar energy in each layer, we must redistribute the absorbed energy
             ! to avoid physically unrealistic conditions. The assumptions made below are 
             ! somewhat arbitrary, but this situation does not arise very frequently. 
             ! This error handling is implemented to accomodate any value of the
             ! radiation frequency.
             ! change condition to match sabg_snow isntead of sabg
```

Another possible explanation is that the number of snow layers might change in CanopyHydrology, which is called just before SurfaceRadiation.

billsacks:
This PR fixes two bugs with init_interp that could cause undesirable answer changes (including changing answers when interpolating a file onto an identical configuration):

(1) Fixes #326 - Now we copy as many snow layers as possible, rather than just copying existing snow layers. This is needed in order to get bit-for-bit results in some radiation fluxes. This changes answers for many points, but in what I believe to be a very minor way (though I haven't done a super-careful investigation of the magnitude of the changes).

(2) Fixes #325 - Now, whenever possible, we find glcmec columns/patches of the same type, rather than ignoring column/patch type when looking for the closest glcmec point. This just changes answers for a few grid cells.

The new behavior is covered by unit tests. This PR also expands some unit testing infrastructure to facilitate the addition of some of these unit tests.

I have confirmed that this test passes: `LII_D_Ld9.f09_g16.I1850Clm50SpCru.hobart_intel.clm-default`, when I point to an initial conditions file that I generated from the dev 002 tag (because I wasn't sure if the current default initial conditions file was exactly compatible with the latest master). Before the changes in this PR, that test was failing.

I have not yet run the test suite on these changes.

### DIFF
--- a/src/biogeophys/test/TotalWaterAndHeat_test/test_total_water_and_heat.pf
+++ b/src/biogeophys/test/TotalWaterAndHeat_test/test_total_water_and_heat.pf
@@ -30,9 +30,7 @@ contains
   subroutine tearDown(this)
     class(TestTWAH), intent(inout) :: this
 
-    if (unittest_subgrid_needs_teardown) then
-       call unittest_subgrid_teardown()
-    end if
+    call unittest_subgrid_teardown()
   end subroutine tearDown
 
   ! ------------------------------------------------------------------------
--- a/src/init_interp/CMakeLists.txt
+++ b/src/init_interp/CMakeLists.txt
@@ -8,6 +8,7 @@ list(APPEND clm_sources
   initInterpMultilevelInterp.F90
   initInterpMultilevelSnow.F90
   initInterpMultilevelSplit.F90
+  initInterpUtils.F90
   )
 
 sourcelist_to_parent(clm_sources)
--- a/src/init_interp/initInterp.F90
+++ b/src/init_interp/initInterp.F90
@@ -12,6 +12,7 @@ module initInterpMod
   use initInterp2dvar, only: interp_2dvar_type
   use initInterpMultilevelBase, only : interp_multilevel_type
   use initInterpMultilevelContainer, only : interp_multilevel_container_type
+  use initInterpUtils, only: glc_elevclasses_are_same
   use shr_kind_mod   , only: r8 => shr_kind_r8, r4 => shr_kind_r4
   use shr_const_mod  , only: SHR_CONST_PI, SHR_CONST_REARTH
   use shr_sys_mod    , only: shr_sys_flush
@@ -158,7 +159,8 @@ contains
     integer            :: decomp_cascade_state_i, decomp_cascade_state_o 
     integer            :: npftsi, ncolsi, nlunsi, ngrcsi 
     integer            :: npftso, ncolso, nlunso, ngrcso 
-    integer , pointer  :: pftindx(:)     
+    logical            :: glc_elevclasses_same
+    integer , pointer  :: pftindx(:)
     integer , pointer  :: colindx(:)     
     integer , pointer  :: lunindx(:)     
     integer , pointer  :: grcindx(:) 
@@ -213,6 +215,8 @@ contains
     call check_dim_level(ncidi, ncido, dimname='levtot' , must_be_same=.false.)
     call check_dim_level(ncidi, ncido, dimname='levgrnd', must_be_same=.false.)
     call check_dim_level(ncidi, ncido, dimname='numrad' , must_be_same=.true.)
+
+    glc_elevclasses_same = glc_elevclasses_are_same(ncidi, ncido)
 
     ! --------------------------------------------
     ! Determine input file global attributes that are needed 
--- a/src/init_interp/initInterp.F90
+++ b/src/init_interp/initInterp.F90
@@ -22,6 +22,7 @@ module initInterpMod
   use abortutils     , only: endrun
   use spmdMod        , only: masterproc
   use restUtilMod    , only: iflag_interp, iflag_copy, iflag_skip
+  use glcBehaviorMod , only: glc_behavior_type
   use ncdio_utils    , only: find_var_on_file
   use ncdio_pio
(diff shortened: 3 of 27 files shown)


billsacks:
@ekluzek I'll leave it up to you as to what you want to review here. Many of the diffs are in unit test infrastructure.

PR #338, opened 2018-04-10
Radtemp

ekluzek:
Adds new diagnostic fields: TSKIN, TSL, LITFALL_N, GSSUNLN, GSHANLN from Ahmed Tawfik


olyson:
After examining the urban skin temperatures that are produced and looking at what Ahmed ended up doing for vegetated/soil surfaces, I think that the current approach I put in should be abandoned and I should simply assign the t_grnd_u to t_skin for urban surfaces, as Ahmed is doing for veg/soil.
I will do this on the branch at some point and commit, but just wanted to give a heads up since this is an active pull request.

ekluzek:
I have about 20 tests that fail in the current branch. If I remove @ckoven changes for litfall_n_patch to CNVegNitrogenFluxType.F90 it looks like things start working again. I need to make sure all of the failing tests work. I've done some looking at the changes and it's not obvious why they are failing. In looking at it in the ddt debugger, the arrays in the type itself aren't even allocated.

ckoven:
@ekluzek does this change in 2cdd0ce fix it?  I just looked for differences between how I'm using litfall_n_patch and some of the other variables and noticed one thing that I had seemed to have forgotten.

ekluzek:
@ckoven that is an important change that needs to be there. This tag has become time-critical, so I'm going to bring it in without littfall_n, but then we'll put it on the next tag to bring in. In the debugger it was looking like bad memory issues, so I don't know if it's that SetValue change that you added.

ekluzek:
But, @ckoven thanks so much for looking at it and getting back on a possible change so quickly. Appreciate it so much!

ckoven:
ok @ekluzek.  shall i submit this as a separate PR to master then?

ekluzek:
Yes, that would be awesome if you would do that @ckoven! I can also do it myself, but your creating the branch would save me a few steps, while I'm working on getting the main PR to master ASAP.

Thanks so much.

ekluzek:
OK, so tests are now working. There is one critically important test that is failing and that is the cmip6 output test, so I'm diving into what's wrong with that one.

ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-cmip6

There are also some tests that show answer changes and I'm verifying those answer changes are as expected. I'm pretty sure they are.

PR #350, opened 2018-04-24
Add science_support for more compsets

billsacks:
Based on direction from Dave Lawrence

@ekluzek @dlawrenncar can you review this to make sure I got this right? It's the same as what we came up with at yesterday's meeting, but it seems like it would be good to give it a final once-over.

### DIFF
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -154,6 +154,13 @@
   </compset>
 
   <compset>
+    <alias>IHistClm50Sp</alias>
+    <lname>HIST_DATM%GSWP3v1_CLM50%SP_SICE_SOCN_MOSART_CISM2%NOEVOLVE_SWAV</lname>
+    <science_support grid="f09_g17"/>
+    <science_support grid="f19_g17"/>
+  </compset>
+
+  <compset>
     <alias>IHistClm50Bgc</alias>
     <lname>HIST_DATM%GSWP3v1_CLM50%BGC_SICE_SOCN_MOSART_CISM2%NOEVOLVE_SWAV</lname>
   </compset>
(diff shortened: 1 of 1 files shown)


cime_config/config_compsets.xml
```diff
@@ -197,6 +205,8 @@
   <compset>
     <alias>I1850Clm45BgcCruGs</alias>
     <lname>1850_DATM%CRUv7_CLM45%BGC_SICE_SOCN_RTM_SGLC_SWAV</lname>
+    <science_support grid="f09_g17"/>
```
ekluzek: Is this the supported one, or is I1850Clm45BgcCru?

cime_config/config_compsets.xml
```diff
@@ -300,16 +316,22 @@
   <compset>
     <alias>I1850Clm40SpCruGs</alias>
     <lname>1850_DATM%CRUv7_CLM40%SP_SICE_SOCN_RTM_SGLC_SWAV</lname>
+    <science_support grid="f09_g17"/>
```
ekluzek: I think this should actually refer to the CN version rather than the SP version. The cases actually run were the CN version. We point to the CN startup files for SP, but the actual cases were CN on.

cime_config/config_compsets.xml
```diff
@@ -300,16 +316,22 @@
...
     <alias>I1850Clm40CnGswGs</alias>
     <lname>1850_DATM%GSWP3v1_CLM40%CN_SICE_SOCN_RTM_SGLC_SWAV</lname>
+    <science_support grid="f09_g17"/>
+    <science_support grid="f19_g17"/>
   </compset>
 
   <compset>
     <alias>IHistClm40SpCruGs</alias>
     <lname>HIST_DATM%CRUv7_CLM40%SP_SICE_SOCN_RTM_SGLC_SWAV</lname>
+    <science_support grid="f09_g17"/>
```
ekluzek: Same here. This should be for the CN version rather than SP.

dlawrenncar:
We did both CLM4SP and CLM4CN simulations so both should be scientifically
supported.

ekluzek:
OK, we don't have SP initial conditions. So I assume I should get some initial conditions from Keith from the cases he ran then right?

dlawrenncar:
Initial conditions from CN version are fine.

ekluzek:
They do work, and it takes less work to put them into place. But, using the version from a simulation helps document what the scientifically supported case is. I suppose if Keith's SP case likewise started from the CN case in the same manner, that would be fine. But, if SP was run separately, when they compare to Keith's case, they will get an apparent change of answers, that we'll have to explain to people. And likely have to explain to people over and over again.

olyson:
The CLM4 SP and CLM4 BGC initial conditions did end up being different because we reran the CLM4 BGC with the CMIP6 surface dataset and we did a short spinup at 1850 to get it into equilibrium with that dataset.

The CLM4 BGC initial file is now:

/glade/p/cgd/tss/people/oleson/CLM5_restarts/clm40_r243_1deg_GSWP3V2_CMIP6_1850.clm2.r.0361-01-01-00000.nc

The CLM4 SP initial file is the older BGC file:

/glade/p/cesmdata/cseg/inputdata/lnd/clm2/initdata_map/clmi.I1850Clm40CnGswGs.1441-01-01.0.9x1.25_gx1v6_simyr1850_c170823.nc

We didn't run a CLM4 SP CRUNCEP simulation.

ekluzek:
OK, we need to go over this in detail tomorrow so that we can bring something in for the next tag.

PR #352, opened 2018-04-26
New init_interp method

### LINKED ISSUE #345: Add a cmip6_evolving_icesheet usermods directory
@lvankampenhout requests the field `ICE_MODEL_FRACTION` as annual output in cmip6 runs with an evolving ice sheet. I'll do this by adding a new usermods directory, `cmip6_evolving_icesheet` (consistent with what we already have in CISM). For now this will:

- include the cmip6 usermods directory

- have a user_nl_clm file that adds ICE_MODEL_FRACTION to the annual output stream

### LINKED ISSUE #346: New mode of operation for init_interp: Copy subgrid areas, too
Currently, init_interp maintains the subgrid areas from the current configuration rather than copying them from the input file. This is the only way I can imagining it working in the general case, when you are interpolating from one resolution to another or from one configuration to a very different configuration (with a substantially different surface dataset, for example).

However, a common case for which we are now using init_interp involves interpolating from a given resolution and configuration to essentially the same resolution and configuration, but with different internal memory structures. In this case, we should be able to interpolate areas (and the related 'active' flags) from the input to the output.

Doing this would have a number of benefits:

1. I think this could help with the problem of interpolating FATES initial conditions files (see #76), at least for the specific (but common) case where the input and output are at the same resolution and we just need to handle changes in the memory structure. (@rgknox and @ekluzek you can let me know if you have thoughts on this)

2. The current operation leads to undesirable answer changes in a coupled context when running with CISM (even in NOEVOLVE mode): because we don't have glacier areas until the first time step, glacier areas over the CISM domain are wrong in initialization when running init_interp. This leads to wrong gridcell-average states/fluxes being sent to the coupler in initialization. While this doesn't impact an offline case, it does impact the evolution of a coupled case (probably not in a way that affects climate, but in a way that prevents bit-for-bit continuation). Putting in place this new mode of operation should let us get bit-for-bit results when running init_interp in a coupled context, e.g., when going from a spinup to a transient run (when init_interp is currently required) or when changing some aspect of the subgrid internal memory allocation in CTSM.

So I plan to add a namelist flag that can be set to use this alternative operation of init_interp. You could only use this when running at the same resolution with (essentially) the same surface dataset.

### LINKED ISSUE #347: In endrun: pass message to shr_sys_abort rather than writing it to iulog
I want to write some expected-exception tests for my init_interp changes. For these to work robustly, I want to check the error message returned against the expected error message (to make sure it's not aborting for some completely different reason). In doing so, I noticed that our endrun calls don't actually pass the abort message on to shr_sys_abort:

```fortran
    if (present (msg)) then
       write(iulog,*)'ENDRUN:', msg
    else
       write(iulog,*)'ENDRUN: called without a message string'
    end if

    call shr_sys_abort()
```

@ekluzek - Do you know if that's intentional?

I'm thinking of changing this to pass msg on to shr_sys_abort rather than writing it to iulog, but don't want to mess with this if the current behavior is intentional.

Actually, the full implementation of what I'm thinking is: do what I'm saying above for the existing msg argument. But also add a new optional argument to endrun, additional_msg, which is written to iulog but not passed along to shr_sys_abort. This is to support expected-exception unit testing, particularly thinking about the common practice of writing out the line number: this is too volatile to test against in unit tests. So a call could look like this:

```fortran
call endrun(msg='Informative message', additional_msg=errmsg(__FILE__, __LINE__))
```

@ekluzek please let me know if you have any thoughts on this.

billsacks:
The main change in this PR is to introduce a new method for init_interp. This is needed in order for subgrid areas to be the same in initialization in runs with and without init_interp.

The choice of init_interp method is controlled by a new namelist variable:

```xml
<entry id="init_interp_method" type="char*64" category="datasets"
       group="clm_initinterp_inparm"
       valid_values="general,use_finidat_areas" >
Method to use for init_interp. Only applies when use_init_interp = .true.

'general': The general-purpose method that can be used when changing
grids, configurations, etc. This starts off with subgrid areas taken
from the surface dataset.

'use_finidat_areas': This starts off with subgrid areas taken from the
input finidat file. This is important to achieve bit-for-bit results in
a coupled case (where areas in initialization impact initial fields sent
to the atmosphere). For this method to work, the input finidat file
needs to be at the same resolution as the current configuration. So this
is a less general form of init_interp. However, it can be used when
transitioning from a spinup run to a transient run, or in other cases
where the only difference is in internal memory allocation. In order to
catch possible problems, this uses a different algorithm for finding the
input point for each output point, which ensures that each active output
point is associated with exactly one input point with the same latitude,
longitude and type. This method requires (a) the same grid for input and
output, within roundoff; (b) any non-zero-weight point in the input must
have memory allocated for it in this grid cell in the output (this will
be satisfied if the point is non-zero-weight on the surface dataset or
if it's a point for which we allocate memory even for zero-weight
points); (c) any active point in the output (based on the surface
dataset and rules for determining active points) must have a matching
point in this grid cell in the input.

</entry>
```

Fixes #346 

This PR also has some other changes: 

- Fixes #347 (endrun message behavior: needed for some of the new unit tests) (see https://github.com/ESCOMP/ctsm/issues/347#issuecomment-384267630 for detailed documentation of new behavior)

- Fixes #345 (unrelated, but folded in here for convenience of testing)


billsacks:
@ekluzek Given limited time to review this, the things I'd most like your feedback on are:

- Are you okay with the init_interp_method option names and clarity of documentation for this new namelist variable?

- I wanted to make `init_interp_method` only settable when `use_init_interp` is true, with these diffs:

   ```diff
   diff --git a/bld/CLMBuildNamelist.pm b/bld/CLMBuildNamelist.pm
   index 568e663b..3595ec74 100755
   --- a/bld/CLMBuildNamelist.pm
   +++ b/bld/CLMBuildNamelist.pm
   @@ -3734,7 +3734,14 @@ sub setup_logic_initinterp {
       my ($opts, $nl_flags, $definition, $defaults, $nl, $physv) = @_;
       if ($physv->as_long() >= $physv->as_long("clm4_5")) {
   -      add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'init_interp_method');
   +      my $var = 'init_interp_method';
   +      if ( &value_is_true($nl->get_value("use_init_interp"))) {
   +         add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var);
   +      } else {
   +         if (defined($nl->get_value($var))) {
   +            $log->fatal_error("$var can only be set if use_init_interp is true");
   +         }
   +      }
       }
    }
   ```

   However, that wouldn't work for the LII test that tests the new method via a new testmods directory, because the testmod adds this namelist variable to both runs of the LII test. In principle I could handle this by introducing a new test type that's similar to LII but adds use_finidat_areas to just the run with use_init_interp = true. This felt like overkill to me, but I could do that if you feel it's important to prevent the user from trying to specify this variable without specifying use_init_interp = .true.

- The new LII test uses the out-of-the-box finidat file. This means that this particular finidat file needs to be kept up-to-date with the current configuration... though we can always move this to user_nl_clm for the test at a later date if that turns out to be a problem. I thought that starting out by trying to use the out-of-the-box finidat file for this test might be less of a maintenance headache, but am happy to just put this in the user_nl_clm file from the get-go if you prefer.

billsacks:
Per discussion with @ekluzek I'll take a stab at introducing a new test type so I can put the error check back into CLMBuildNamelist. I'll call this LII_FINIDAT_AREAS.

PR #368, opened 2018-05-03
Allow FATES to run with inventory initialization files

ckoven:
The current code in CTSM doesn't let FATES run when an inventory initialization is specified. This was first pointed out by @serbinsh in https://github.com/NGEET/fates/issues/338.  This PR includes the fix to allow it to run.  It may not be the most elegant solution to the problem though -- I just commented out the lines that were preventing the code from running.


ckoven:
@rgknox suggested I change this PR to pull into the fates-next branch rather than master. I'll try that now.

ckoven:
that seems to have worked.

rgknox:
While not having an understanding of exactly why that line had to be removed, I can say that I tested the change and it allowed a setup that uses that namelist argument to complete that wouldn't before.

It looked like the offending line was testing to see if the file was present, but was not finding it?  Maybe the file presence evaluation was not working correctly?

bld/CLMBuildNamelist.pm
```diff
@@ -3742,8 +3742,8 @@ sub setup_logic_fates {
               $var = "fates_inventory_ctrl_filename";
               if ( ! defined($nl->get_value($var))  ) {
                  $log->fatal_error("$var is required when use_fates_inventory_init is set" );
-              } elsif ( ! -f "$nl->get_value($var)" ) {
-                 $log->fatal_error("$var does NOT point to a valid filename" );
+              # } elsif ( ! -f "$nl->get_value($var)" ) {
+              #    $log->fatal_error("$var does NOT point to a valid filename" );
```
ekluzek: My bet is that extra quotes in the string are making the check for file existence to fail. So I think I can fix this so it will work.

rgknox:
I believe this PR was intended to fix that error you refer to @serbinsh.  
@ckoven and I have been merging these commits into any runs that we need census initialization.

ckoven:
this keeps coming up for me, so I tried to fix it more properly.  @ekluzek please advise if this is ok to merge now.

ekluzek:
OK, the fixed version looks fine.

PR #375, opened 2018-05-12
variable soil depth compatibility with FATES

rgknox:
### Description:

THIS IS A BACKWARDS INCOMPATIBLE CHANGE WILL INCREMENT FATES API TAG to 4.0.0

https://github.com/NGEET/fates/pull/381

CLM5 introduced variable soil depths at the column level.  Thus, we need to pass these depths into fates at the site level, and modify usage of arrays dictated by this depth accordingly.  This also affected the decomposition depth.

These changes focus on the clm side of the interface, specifically how we pass in soil depth at the column level.

### Collaborators:

@ckoven @dlawrenncar 

### Expectation of Answer Changes:

There may be some round-off differences just because code was moved around, but differences should be small.

### Checklist:

- [ ] My change requires a change to the documentation.
- [x] I have updated the in-code documentation AND wiki accordingly.
- [x] FATES PASS/FAIL regression tests were run
- [x] If answers were expected to change, evaluation was performed and provided

### Test Results:

Test Output:

All PASS


ekluzek:
This needs the update to the FATES version in the externals. This can point to NGEET/fates.

rgknox:
Tested as merged with #390.   All pass.  Is b4b with base, when temporarily reverting a small fix to normalized rooting depths.

Pending any FATES side review/changes, will update the tag in externals.

rgknox:
@ekluzek , are #375 and #390 satisfactory from your end?  
Once you approve, I will merge on NGEET/fates and update the externals config to the new tag.

Issue #380, opened 2018-05-15
Some low resolution gnu tests were accidentally removed...


ekluzek:
Some low resolution gnu tests were accidentally removed.

```
diff --git a/cime_config/testdefs/testlist_clm.xml b/cime_config/testdefs/testlist_clm.xml
index 8c8de09..de438bf 100644
--- a/cime_config/testdefs/testlist_clm.xml
+++ b/cime_config/testdefs/testlist_clm.xml
@@ -2,7 +2,7 @@
 <testlist version="2.0">
   <test name="ERI_D_Ld9" grid="f10_f10_musgs" compset="I1850Clm50Bgc" testmods="clm/default1850">
     <machines>
-      <machine name="cheyenne" compiler="intel" category="aux_clm"/>
+      <machine name="cheyenne" compiler="gnu" category="aux_clm"/>
     </machines>
     <options>
       <option name="wallclock">00:20:00</option>
@@ -26,6 +26,7 @@
   </test>
   <test name="ERI_D_Ld9" grid="f10_f10_musgs" compset="I2000Clm50BgcCruGs" testmods="clm/default">
     <machines>
+      <machine name="cheyenne" compiler="gnu" category="aux_clm"/>
       <machine name="cheyenne" compiler="intel" category="aux_clm"/>
     </machines>
     <options>
@@ -142,6 +143,7 @@
   </test>
   <test name="ERI_Ld9" grid="f10_f10_musgs" compset="I2000Clm50BgcCruGs" testmods="clm/default">
     <machines>
+      <machine name="cheyenne" compiler="gnu" category="aux_clm"/>
       <machine name="cheyenne" compiler="intel" category="aux_clm"/>
     </machines>
     <options>
@@ -174,6 +176,7 @@
   </test>
   <test name="ERP_D" grid="f10_f10_musgs" compset="IHistClm50Bgc" testmods="clm/decStart">
     <machines>
+      <machine name="cheyenne" compiler="gnu" category="aux_clm"/>
       <machine name="cheyenne" compiler="intel" category="aux_clm"/>
     </machines>
     <options>
@@ -248,7 +251,7 @@
   </test>
   <test name="ERP_D_Ld5" grid="f10_f10_musgs" compset="I2000Clm50BgcCruGs" testmods="clm/ciso_flexCN_FUN">
     <machines>
-      <machine name="cheyenne" compiler="intel" category="aux_clm"/>
[...]

Issue #374, opened 2018-05-10
PE layout change affects answers for some configurations to a greater extent than expected...


ekluzek:
In updating the cime version I also updated the PE layout for f45, f09, and f19. This changed answers to a minor extent (just in Volr from ROF component or uvel, vvel from cism and only order E-19) for some Clm40 tests as well as the following:

ERI_D_Ld9.f09_g16.I1850Clm45BgcCruGs.cheyenne_intel.clm-default1850
ERI_Ld9.f09_g17.I2000Clm50BgcCruGs.cheyenne_intel.clm-cn_conly.
ERI_Ld9.f45_g37.I2000Clm50BgcCruGs.cheyenne_intel.clm-nofire
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm50Bgc.cheyenne_gnu.clm-monthly.
SMS_Ld1.f09_g17.I2000Clm50BgcCruGs.cheyenne_intel.clm-af_bias_v7

This was expected.

However, answers also changed for several tests to a much greater degree and for a wide variety of variables. The tests that changed are:

ERI_Ld9.f09_g17.I1850Clm50Bgc.cheyenne_intel.clm-drydepnomegan1850
ERI_N2_Ld9.f19_g17.I2000Clm50BgcCrop.cheyenne_intel.clm-default
ERS_D.f19_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-reseedresetsnow
ERS_D_Ld3.f09_g17_gl4.I1850Clm50BgcCrop.cheyenne_intel.clm-clm50KitchenSink
ERS_D_Ld3.f19_g17_gl4.I1850Clm50BgcCrop.cheyenne_intel.clm-clm50dynroots1850
ERS_D_Ld5.f09_g17.I2000Clm45Fates.cheyenne_intel.clm-FatesColdDef
ERS_D_Ld5.f09_g17.I2000Clm50Fates.cheyenne_intel.clm-FatesColdDef
ERS_Ld3.f09_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-rad_hrly_light_res_half
LII_D_Ld3.f19_g17_gl4.I2000Clm50BgcCrop.cheyenne_intel.clm-glcMEC_spunup_1way
SMS.f19_g17.I2000Clm50Cn.cheyenne_intel.clm-default
SMS_D_Ld5.f45_f45_mg37.I2000Clm45Fates.cheyenne_intel.clm-FatesColdDef
[...]

Issue #363, opened 2018-05-02
Lower number of nodes used on cheyenne for three key configurations


ekluzek:
We want to lower the number of nodes on cheyenne as follows:

f45 -- 4 nodes
f09 -- 50 nodes
f19 -- 40 nodes (this last one is a guess)

ekluzek:
This will be done in clm5.0.dev011

PR #381, opened 2018-05-15
1850 fully coupled ndep update, update cism, update PE layouts, turn BFBFLAG on for testing

ekluzek:
Update cism to latest release branch: release-v2.1.00
Update PE layouts for f45, f09, f19 #363 #374 #380
Turn BFBFLAG=TRUE for most CTSM testing


Externals.cfg
```diff
@@ -8,7 +8,7 @@ required = True
 local_path = components/cism
 protocol = git
 repo_url = https://github.com/ESCOMP/cism-wrapper
-tag = cism2_1_54
+tag = release-v2.1.00
```
billsacks: There is now a 2.1.01 tag, which changes answers relative to 2.1.00. If you have already done testing, don't worry about changing this, but I mention it in case you need to rerun testing anyway.

PR #385, opened 2018-05-16
Update max day length (max_dayl) each time step

### LINKED ISSUE #260: max daylength doesn't change over time for varying orbital parameters
As documented in #19 , @ekluzek recently changed the max daylength function so that it is set properly based on the starting orbital parameters for a run. However, this still doesn't change over time, if running with time-varying orbital parameters.

### LINKED ISSUE #379: Problems with Hist decStart restart tests due to variable orbital year
In clm5.0.dev010 I have problems with the following test:

ERP_D_P48x1.f10_f10_musgs.IHistClm50Bgc.hobart_nag.clm-decStart

I ran it three times. The first and last it failed the exact restart test, for a bunch of fields. So it is a problem. The second time it failed on the build of cism, which might have just been a random system glitch.

Since, I don't see other problems I'm going ahead with dev010, but we need to continue to investigate this.

billsacks:
This is needed for max day length to remain correct with time-varying
orbital parameters, which in turn is needed to achieve exact restart
in a run that crosses the year boundary with time-varying orbital
parameters.

Fixes #379
Fixes #260

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -30,7 +30,7 @@ required = True
 local_path = cime
 protocol = git
 repo_url = https://github.com/ESMCI/cime
-tag = cime5.6.6
+hash = a67dceae95567cc7529c4cfed153017f4b65b210
 required = True
 
 [externals_description]
--- a/cime_config/testdefs/ExpectedTestFails.xml
+++ b/cime_config/testdefs/ExpectedTestFails.xml
@@ -2,27 +2,11 @@
 
 <expectedFails>
   <category name="aux_clm">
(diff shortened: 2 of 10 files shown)


billsacks:
@ekluzek - in case you'd like to give this a quick review

billsacks:
I did some preliminary testing and this looked good (changed answers in a HIST test that crossed the year boundary, even without the cime change; did not change answers in a couple of single-year HIST tests). I have just kicked off the full test suite. If all goes well, I should be able to make this tag later today.

PR #392, opened 2018-05-23
Remove old code for snw_rds restart that looks no longer needed

billsacks:
It looks like this was there for backwards compatibility, but I can't
imagine it being needed any more. The removed code block was messing
with our ability to cleanly separate WaterStateType for water isotopes.

### DIFF
--- a/src/biogeophys/WaterStateType.F90
+++ b/src/biogeophys/WaterStateType.F90
@@ -13,6 +13,7 @@ module WaterstateType
   use clm_varctl     , only : use_vancouver, use_mexicocity, use_cn, iulog, use_luna
   use clm_varpar     , only : nlevgrnd, nlevurb, nlevsno   
   use clm_varcon     , only : spval
+  use abortutils     , only : endrun
   use LandunitType   , only : lun                
   use ColumnType     , only : col                
   !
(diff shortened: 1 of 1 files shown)


billsacks:
@ekluzek - this code block was tripping @mathewvrothstein up in his work to separate WaterStateType. I'm kicking off testing overnight to make sure it's safe to remove this according to the test suite, but please let me know if you can see any reason why it wouldn't be safe to remove this.

Assuming all tests pass, can I go ahead and bring this to master (making a little tag with just this change)? If that's not convenient, I can just give this change to @mathewvrothstein to merge to his branch.

ekluzek:
Since, this isn't required for the release, I'd like to have it stay away from master for a little bit longer. Sounds like that would be OK as well.

billsacks:
Thanks, @ekluzek - I'll have @mathewvrothstein just merge these changes to his branch, then. I'll go ahead and close this PR.

billsacks:
The full test suite passed and was bit-for-bit with clm5.0.dev012, so I think this is a safe change to make.

billsacks:
Actually, I'm going to reopen this and let it auto-close when it comes to master, because I think that will give better traceability.

billsacks:
I have closed this because I reimplemented these changes off of the latest version of master in a separate branch that will come to master soon.

PR #400, opened 2018-06-01
Update CanopyStateType.F90

serbinsh:
Update CanopyStateType.F90 to include missing LAI/SAI units missing from TLAI, TSAI, LAISUN, LAISHA.  Seems like these should have units of m2/m2 unless I am missing something


ekluzek:
@dlawrenncar @danicalombardozzi @ckoven @rosiealice can one or all of you approve this change? This is just changing the units of the "*AI" fields like LAI,SAI, etc. from "none" to "m2/m2". Does that sound like the right thing to do?

rosiealice:
Yes, this is fine. Well spotted @serbinsh :)

PR #405, opened 2018-06-06
Protected fates calculation of bare-soil area to not go below 0

rgknox:
This PR ensures that when FATES passes back area weights to CTSM, the bare-soil patch is not given a negative area.  The precisions on our patch area calculations can stray up to 1e-12, which is small but had made it possible that an nearly-zero negative bare-soil patch area was possible.  This fix simply bounds it at 0.


rgknox:
Ran the fates regression suite, compared with fates-next-api, all b4b, all pass.

/glade2/scratch2/rgknox/clmed-tests/fates.cheyenne.intel.Cfdce2b2-F91785e3-master-v3

PR #407, opened 2018-06-07
Initial version of templates for contributing, PR's and issues

ekluzek:
Templates for Pull Requests and issues, to make sure more information is given on each of those. Also an initial cut at a contributing document to outline expectations of how code changes come into CTSM. Please review and give feedback. But, since this is non-code, we can make changes at any time, so I'll probably bring it in before I get feedback from everyone.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,12 +1,27 @@
+# Type of Issue:
+Some example types are:
+* bug in science or answers
+* bug in software
+* science enhancement (changes answers and requires science development)
+* software enhancement
+* FATES change
+* Discussion
+
+
 ## Summary of Issue:
 
-## Expected behavior and actual behavior:
+## Details of Issue:
+(Be sure to give enough details that the issue can understood sufficiently to replicate
+ or to make the required changes)
 
-## Steps to reproduce the problem:
+# For bugs be sure to fill out the following (otherwise remove):
+-----------------------------------------------------------------
+# Expected behavior and actual behavior:
 
-## Screen output or log file showing the error message and context:
+# Steps to reproduce the problem:
 
-## What is the CTSM version you are using (output of git describe)?:
+# What is the CTSM version you are using (output of git describe)?:
 
-## What is the machine you are running on?:
+# What is the machine you are running on?:
 
+-----------------------------------------------------------------
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,27 +1,22 @@
-# Type of Issue:
+### General Type of Issue:
 Some example types are:
-* bug in science or answers
-* bug in software
-* science enhancement (changes answers and requires science development)
-* software enhancement
-* FATES change
-* Discussion
+* bug
+* enhancement
+* question
 
+### Details of Issue:
+(Be sure to give enough details that the issue can be understood sufficiently 
+to replicate or to make the required changes)
 
-## Summary of Issue:
-
-## Details of Issue:
-(Be sure to give enough details that the issue can understood sufficiently to replicate
- or to make the required changes)
-
-# For bugs be sure to fill out the following (otherwise remove):
+### For bugs be sure to fill out the following:
+(Otherwise remove, and remove any prompts that aren't useful)
 -----------------------------------------------------------------
-# Expected behavior and actual behavior:
+### Important Details of your setup/configuration (compset, settings, etc.):
 
-# Steps to reproduce the problem:
+### Important output or errors that show the problem:
 
-# What is the CTSM version you are using (output of git describe)?:
+#### What is the CTSM version you are using (output of git describe)?:
 
-# What is the machine you are running on?:
+#### What is the machine you are running on:?
 
 -----------------------------------------------------------------
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,15 +1,15 @@
 ### General Type of Issue:
 Some example types are:
-* bug
-* enhancement
-* question
+*bug
+*enhancement
+*question
 
 ### Details of Issue:
 (Be sure to give enough details that the issue can be understood sufficiently 
 to replicate or to make the required changes)
 
 ### For bugs be sure to fill out the following:
-(Otherwise remove, and remove any prompts that aren't useful)
+`(Otherwise remove, and remove any prompts that aren't useful)`
 -----------------------------------------------------------------
 ### Important Details of your setup/configuration (compset, settings, etc.):
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,19 +1,18 @@
 ### General Type of Issue:
 Some example types are:
-*bug
-*enhancement
-*question
+- bug
+- enhancement
+- question
 
 ### Details of Issue:
 (Be sure to give enough details that the issue can be understood sufficiently 
 to replicate or to make the required changes)
 
-### For bugs be sure to fill out the following:
-`(Otherwise remove, and remove any prompts that aren't useful)`
+### For bugs fill out the following: (Otherwise remove, and remove any prompts that aren't useful)
 -----------------------------------------------------------------
-### Important Details of your setup/configuration (compset, settings, etc.):
+#### Important Details of your setup/configuration (compset, settings, etc.):
 
-### Important output or errors that show the problem:
+#### Important output or errors that show the problem:
 
 #### What is the CTSM version you are using (output of git describe)?:
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,12 @@
+## Summary of Issue:
+
+## Expected behavior and actual behavior:
+
+## Steps to reproduce the problem:
+
+## Screen output or log file showing the error message and context:
+
+## What is the CTSM version you are using (output of git describe)?:
+
+## What is the machine you are running on?:
+
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+## Description of changes:
+
+## CTSM Issues Fixed (include github issue #):
+(https://github.com/ESCOMP/ctsm/issues)
+
+## Are answers expected to change?:
+
+## Any User Interface Changes?:
+
+## Externals updated (if any):
+
+## Testing performed:
+
+## Did you follow the conventions in the CTSM Developers Guide?:
+(https://wiki.ucar.edu/display/ccsm/CLM+Coding+Conventions)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,15 +1,14 @@
-## Description of changes:
+# Description of changes:
 
-## CTSM Issues Fixed (include github issue #):
-(https://github.com/ESCOMP/ctsm/issues)
+# CTSM Issues Fixed (include github issue #):
 
-## Are answers expected to change?:
+# Are answers expected to change? (and if so in what way):
 
-## Any User Interface Changes?:
+# Any User Interface Changes (namelist or namelist defaults changes)?:
 
-## Externals updated (if any):
+# Testing performed:
+(List what testing you did to show your changes worked as expected)
+(This can be manual testing or running of the different test suites)
 
-## Testing performed:
-
-## Did you follow the conventions in the CTSM Developers Guide?:
-(https://wiki.ucar.edu/display/ccsm/CLM+Coding+Conventions)
+# NOTE: Be sure to check your Coding style against the standard:
+https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,14 +1,14 @@
-# Description of changes:
+### Description of changes:
 
-# CTSM Issues Fixed (include github issue #):
+### CTSM Issues Fixed (include github issue #):
 
-# Are answers expected to change? (and if so in what way):
+### Are answers expected to change? (and if so in what way):
 
-# Any User Interface Changes (namelist or namelist defaults changes)?:
+### Any User Interface Changes (namelist or namelist defaults changes)?:
 
-# Testing performed:
+### Testing performed:
 (List what testing you did to show your changes worked as expected)
 (This can be manual testing or running of the different test suites)
 
-# NOTE: Be sure to check your Coding style against the standard:
+### NOTE: Be sure to check your Coding style against the standard:
 https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,34 +10,35 @@ The ctsm-dev group is here:
 
 https://groups.google.com/a/ucar.edu/forum/#!forum/ctsm-dev
 
-Use the help from the WIKI below to setup a fork and personal branch in github to put your developments
-on and keep up to date with the main development CTSM. Once the changes are sufficiently advanced you
-can form a Pull Request on github under
+Use the help from the wiki below to setup a fork and personal branch in GitHub to put your developments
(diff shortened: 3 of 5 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,12 @@
+## Summary of Issue:
```
dlawrenncar: Seems like this only applies to issues that are bugs.  We have a range of ways that we utilize issues and this template would not be appropriate for all of them.  So, maybe that can be clarified somehow, that this is how users should report a bug?

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
+### How to contribute:
+
+It's recommended that you first contact ctsm-software@ucar.edu to discuss the changes or additions that you propose.
+Work with them to think about the design and implemention of the changes as well as the planning
+and scheduling for the change to come into CTSM.
+
+It's also recommened that you join the ctsm-dev@ucar.edu google groups in order
```
dlawrenncar: I removed ctsm-core because that list of meant mainly for NCAR staff and close collaborators.

.github/ISSUE_TEMPLATE.md
billsacks: I agree with @dlawrenncar . But I'd go further than this: I have found issue templates (such as the one in place for manage_externals) to be more annoying than helpful, both when submitting an issue and when reading other people's issues. The problem is that many of these sections end up not applying to any given issue, so it just adds extra work for both writing an issue and reading an issue (because you have to read through the sections filled out with N/A or whatever).

If we do anything here, I feel this should not be a template like this (where people feel they need to fill something in for every section), but rather a short paragraph saying something like, "A good bug report contains a brief summary of the issue, followed by more detailed information on how to reproduce the problem... [etc.]"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
+## Description of changes:
```
billsacks: I don't like having such a big header level for each section, either in the PR or the issue template. I find it makes the headers overwhelm the actual information. You could keep these at a lower header level, or (my preference) remove the header `##` signs entirely.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+## CTSM Issues Fixed (include github issue #):
+(https://github.com/ESCOMP/ctsm/issues)
```
billsacks: I don't think the link to issues is necessary here. If someone has fixed an issue, they presumably know where to find issues.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Are answers expected to change?:
+
+## Any User Interface Changes?:
```
billsacks: What do you mean by this? Does this mean "Any new xml or namelist variables"? If so, we should say that explicitly rather than making people guess what we mean by user interface changes.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Externals updated (if any):
+
+## Testing performed:
+
+## Did you follow the conventions in the CTSM Developers Guide?:
```
billsacks: I think this last question is unnecessary. We should point to the developer's guide in the contributing document, but forcing everyone to type "Yes" here for every PR just seems kind of bothersome.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+Work with them to think about the design and implemention of the changes as well as the planning
+and scheduling for the change to come into CTSM.
+
+It's also recommened that you join the ctsm-dev@ucar.edu google groups in order 
+to be informed about the science that is happening in CTSM as well as the latest develepments and tags.
+The ctsm-dev group is here:
+
+https://groups.google.com/a/ucar.edu/forum/#!forum/ctsm-dev
+
+Use the help from the WIKI below to setup a fork and personal branch in github to put your developments
```
billsacks: I don't think WIKI needs to be capitalized (all-lowercase wiki seems to be the general convention). However, github should be "GitHub".

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+To show your proposed changes and start getting feedback on your code changes. This also allows your
+changes to be planned for and slated for a time to come into CTSM master. In most cases you won't merge
+the Pull Request yourself, but a software engineer responsible for CTSM will do additional testing and
+bring the changes to CTSM master.
+
+## CTSM Developers Guide:
+
+ https://wiki.ucar.edu/display/ccsm/Community+Land+Model+Developers+Guide
+
+## Getting Started:
```
billsacks: Is something supposed to be in this section?

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+https://github.com/ESCOMP/ctsm/pulls
+
+To show your proposed changes and start getting feedback on your code changes. This also allows your
+changes to be planned for and slated for a time to come into CTSM master. In most cases you won't merge
+the Pull Request yourself, but a software engineer responsible for CTSM will do additional testing and
+bring the changes to CTSM master.
+
+## CTSM Developers Guide:
+
+ https://wiki.ucar.edu/display/ccsm/Community+Land+Model+Developers+Guide
```
billsacks: We should move this to the github wiki at some point soon.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+and scheduling for the change to come into CTSM.
+
+It's also recommened that you join the ctsm-dev@ucar.edu google groups in order 
+to be informed about the science that is happening in CTSM as well as the latest develepments and tags.
+The ctsm-dev group is here:
+
+https://groups.google.com/a/ucar.edu/forum/#!forum/ctsm-dev
+
+Use the help from the WIKI below to setup a fork and personal branch in github to put your developments
+on and keep up to date with the main development CTSM. Once the changes are sufficiently advanced you
```
billsacks: "main development CTSM" doesn't sound right to me. How about "master branch of CTSM" or "main development line of CTSM"?

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+can form a Pull Request on github under
```
billsacks: Again, github should be GitHub

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+https://github.com/ESCOMP/ctsm/pulls
```
billsacks: Is this really the way to make a PR? When I went there, it didn't show me my branches on my personal fork. I think you need to create a PR from your fork.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+To show your proposed changes and start getting feedback on your code changes. This also allows your
```
billsacks: Move this first sentence fragment "To show..." to the above text (before the link), because it feels awkward here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+## CTSM Developers Guide:
+
+ https://wiki.ucar.edu/display/ccsm/Community+Land+Model+Developers+Guide
+
+## Getting Started:
+
+## CTSM WIKI:
+
+https://github.com/ESCOMP/ctsm/wiki
+https://github.com/ESCOMP/ctsm/wiki/Getting-started-with-CTSM-in-git
```
billsacks: I'm not sure you need the two git-focused links called out explicitly, since they're already mentioned prominently in the main wiki home page.

CONTRIBUTING.md
ekluzek: You can do either one. On your fork you can make a PR from your list of branches. On ESCOMP when you first make the PR you hit the "compare across forks" link at the top of the page. Then you tell it your fork and it'll have your list of branches after you do that.

.github/ISSUE_TEMPLATE.md
ekluzek: @dlawrenncar good point on the different types. @billsacks I do find templates to be useful. Templates help me to remember important things to include, or prompt users on those things. As long as the user knows it's OK to delete prompts as appropriate I think that's OK. So, I'm working on a compromise. It allows you to specify the general type of issue, and for bugs has a few things to fill out, with directions to remove as appropriate. Since, standard users don't have access to adding labels, having the list of general types will be useful for us to categorize after the issue is created.

ekluzek:
@billsacks @dlawrenncar I've updated, can you take a look, especially at the Issue template again?

CONTRIBUTING.md
billsacks: Ah, thanks for clarifying.

CONTRIBUTING.md
```diff
@@ -0,0 +1,44 @@
...
+the pull request).
+
+https://github.com/ESCOMP/ctsm/pulls
+
+This allows you to show your proposed changes and start getting feedback on them (even if they aren't finished). 
+This also allows your changes to be planned for and slated for a time to come into CTSM master. In most 
+cases you won't merge the Pull Request yourself, but a software engineer responsible for CTSM will do 
+additional testing and bring the changes to CTSM master.
+
+# CTSM Developers Guide:
```
billsacks: (I thought I made this comment in my first review, but I don't see it anywhere, so I guess I never submitted it.)

I don't think we need a heading for each link. Could this instead look like this?:

```markdown
# Helpful resources

- CTSM Developers Guide: URL

- CTSM Wiki: URL

- etc.
```

But maybe I'll be more okay with headers once they're fixed to be smaller (see the above comment about header levels).

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Are answers expected to change? (and if so in what way):
+
+# Any User Interface Changes (namelist or namelist defaults changes)?:
+
+# Testing performed:
+(List what testing you did to show your changes worked as expected)
+(This can be manual testing or running of the different test suites)
+
+# NOTE: Be sure to check your Coding style against the standard:
```
billsacks: I like this better (having a note but not expecting them to write "yes") - thanks.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
+# Type of Issue:
```
billsacks: I see your rationale for adding this, but I'm not sure it's worth asking users to enter the type of issue: I don't feel it's fair to make standard users understand all of the different issue types, and if they don't understand them all (and/or we don't list all of the possible values), then there's a high likelihood that they'll enter the wrong type. i.e., it seems like this puts extra burden on the issue submitter without providing us with a reliable benefit.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+Some example types are:
+* bug in science or answers
+* bug in software
+* science enhancement (changes answers and requires science development)
+* software enhancement
+* FATES change
+* Discussion
+
+
+## Summary of Issue:
```
billsacks: Often the summary of the issue appears in the subject / title line, so doesn't need to be called out separately in the body.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Details of Issue:
+(Be sure to give enough details that the issue can understood sufficiently to replicate
+ or to make the required changes)
+
+# For bugs be sure to fill out the following (otherwise remove):
+-----------------------------------------------------------------
+# Expected behavior and actual behavior:
```
billsacks: With manage_externals, I've found this section on expected/actual behavior to be more bothersome than helpful: In some cases it doesn't apply, and in others it duplicates what was already given in the details of issue section.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+* science enhancement (changes answers and requires science development)
+* software enhancement
+* FATES change
+* Discussion
+
+
+## Summary of Issue:
+
+## Details of Issue:
+(Be sure to give enough details that the issue can understood sufficiently to replicate
```
billsacks: "can understood" should be "can be understood"

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+ or to make the required changes)
+
+# For bugs be sure to fill out the following (otherwise remove):
+-----------------------------------------------------------------
+# Expected behavior and actual behavior:
+
+# Steps to reproduce the problem:
```
billsacks: Steps to reproduce the problem can be useful in some cases. However, I'd prefer if this and the machine (if relevant... usually it isn't) were folded into the "details" section. You already pretty much ask for this in the parenthetical in that section... we should just make it clear that they should give enough information to reproduce the issue there, and not ask for it again separately.

ekluzek:
OK. I think we need to resolve this in person. I'm going to take one more shot at getting something more acceptable. And then I'll merge it to master, and we can meet in person and work out something for the issue template. The main disagreement is just around it, so there's no reason to not start moving forward on this. Since, this is just documentation, this is something that can be easily changed at anytime.

ekluzek:
OK, I made more changes. We'll meet to go over this more. We should probably go over these templates as a group as well.

billsacks:
@jhamman - Martyn suggested that you might have some thoughts on the CONTRIBUTING document. Even though this has been merged, feel free to comment here or open a new PR with any thoughts you have for improving that document.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+Some example types are:
+- bug
+- enhancement
+- question
+
+### Details of Issue:
+(Be sure to give enough details that the issue can be understood sufficiently 
+to replicate or to make the required changes)
+
+### For bugs fill out the following: (Otherwise remove, and remove any prompts that aren't useful)
```
jhamman: typo "...remove, and remove..."

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+### General Type of Issue:
+Some example types are:
+- bug
+- enhancement
+- question
```
jhamman: You may consider asking people to prefix their title with one of the following codes:

- ENH: Enhancement, new functionality
- BUG: Bug fix
- DOC: Additions/updates to documentation
- TST: Additions/updates to tests
- BLD: Updates to the build process/scripts
- PERF: Performance improvement
- CLN: Code cleanup

These sort of issue tips can also be hidden in a html comment, e.g. 

```
<!---
your comment goes here
and here
-->
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+### Description of changes:
+
+### CTSM Issues Fixed (include github issue #):
```
jhamman: On other projects, we've used something like the following template: 

```
 - [ ] Closes #xxxx (remove if there is no corresponding issue, which should only be the case for minor changes)
 - [ ] Tests added (for all bug fixes or enhancements)
 - [ ] Tests passed (for all non-documentation changes)
 - [ ] Fully documented, ...)
```

This seems to help new/old developers actually provide this information.

CONTRIBUTING.md
```diff
@@ -0,0 +1,44 @@
+## How to contribute:
+
+It's recommended that you first contact ctsm-software@ucar.edu to discuss the changes or additions that you propose.
```
jhamman: I would suggest encouraging developers to open github issues instead of sending an email. Discussing new feature additions in the open will help provide transparency in the development process.

.github/ISSUE_TEMPLATE.md
billsacks: Thanks for this suggestion. I should have mentioned that we're going to try having separate issue templates for a few different types of issues (using a relatively new GitHub feature). That partially accomplishes what you're suggesting here. For now, I'd like to avoid making users classify their issues more specifically than that, but we might revisit that decision in the future.

I like the suggestion of separating out test-focused issues, and I have created a new label for that purpose.

CONTRIBUTING.md
billsacks: Good point.

billsacks:
@jhamman Thanks for your feedback here. I've incorporated some suggestions into #422. We might make some more changes along the lines of what you suggest down the line.

PR #416, opened 2018-06-11
Start a document of design notes

billsacks:
### Description of changes:

Start a document of design notes.

@mathewvrothstein asked about this particular design question, so I thought it was worth noting for others as well.

Eventually we may want to have multiple files here. But I'm not sure
what organization we'll want in the end, so for now I'm starting with a
single file.


billsacks:
@ekluzek and others: I'm not sure if this is the best place for design notes like this. I'd like them to live in the main repository alongside the code, so it's easier to keep them in sync with the code, but I'm open to suggestions about where exactly they should live.

ekluzek:
Hmmm. I like the idea of having design documents. And it's good to have design documents with the code. I suppose this directory location under the "doc" directory makes sense as well. The concern I have is how to organize this, so it doesn't become just a mess of different little notes on a variety of things. It would be good to start with at least some level of organization.

billsacks:
I agree this should be organized somehow, but I couldn't think of how to organize it. I thought it would be a lot easier to come up with an organization once we have a few different notes to organize, rather than trying to impose something up-front before we have a sense of what will be here.

That said, if you have any ideas, I'm very open to them.

billsacks:
@ekluzek gave his permission to merge once I made the renames in ee6f586.

PR #422, opened 2018-06-14
Work on templates

billsacks:
### Description of changes

- Multiple issue templates (as discussed at yesterday's CTSM meeting)

- Minor tweaks to PR template and contributing guidelines based on feedback from yesterday's CTSM meeting and @jhamman

### Specific notes

Testing performed: none


billsacks:
Some things that were suggested at yesterday's meeting that I have NOT done:

- PR template: @rosiealice suggested adding a link to documentation on testing, but I'm not sure if we have anything good & complete yet

- Support request template: @dlawrenncar suggested adding a link to the user's guide, but I don't think this is complete yet

billsacks:
I named the issue templates starting with 01, etc., so that we can control the ordering (the ordering is alphabetical)

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,21 +0,0 @@
...
------------------------------------------------------------------
-#### Important Details of your setup/configuration (compset, settings, etc.):
-
-#### Important output or errors that show the problem:
-
-#### What is the CTSM version you are using (output of git describe)?:
-
-#### What is the machine you are running on:?
-
------------------------------------------------------------------
```
jhamman: My understanding is that you will still need a default issue template? Should that be a copy of `04_other_...`

CONTRIBUTING.md
```diff
@@ -1,11 +1,12 @@
...
+We recommend that you first open an issue (https://github.com/escomp/ctsm/issues/) to
+discuss the changes or additions that you propose. Through the issue discussion, you can
+work with other CTSM developers to think about the design and implementation of the changes
+as well as the planning and scheduling for the change to come into CTSM.
 
-It's also recommened that you join the ctsm-dev@ucar.edu google groups in order 
-to be informed about the science that is happening in CTSM as well as the latest develepments and tags.
+We also recommend that you join the ctsm-dev@ucar.edu google groups in order 
+to be informed about the science that is happening in CTSM as well as the latest developments and tags.
 The ctsm-dev group is here:
```
jhamman: I think you'll want to provide links to the new issue interface here in the contributing guide. This shows up as a banner when opening a new issue. If I understand correctly, the workflow would be:

- come to github
- open a new issue
- see the contributing banner (click)
- read contributing banner and see links for multiple issue types (click)
- fill out the issue template

I guess, an alternative workflow would be to put those links in the default issue template.

BTW, all I know on this subject came from here: https://blog.github.com/2018-01-25-multiple-issue-and-pull-request-templates/

billsacks: Sorry, I'm not sure what change you're suggesting we make here....

.github/ISSUE_TEMPLATE.md
billsacks: In our case, if the issue doesn't fall under one of the given categories, then we don't have any particular issue template – a blank slate is sufficient.

If you'd like to see how this plays out in practice, I've set up a demo repo where you can open an issue:

https://github.com/billsacks/ctsm-github-multiple-issue-templates/issues

jhamman: oooo, this demo is great. I think you can ignore all my comments on this PR now.

billsacks:
@ekluzek thanks for your changes. I just made a couple of commits where I reformatted things a bit, but kept your added content.

If we're going to suggest the option of contacting ctsm-software, the permissions need to be opened up on that list to allow anyone in the world to post to it (if I'm understanding the permissions correctly). Is that something we want to do? I'm a little worried about doing this in terms of spam... but I guess we can give it a shot and then pull back if it's a problem.

billsacks:
Okay, I have opened up posting permissions on ctsm-software.

PR #425, opened 2018-06-22
Add missing --append flag to decStart xmlchange command

### LINKED ISSUE #424: decStart testmod is missing an `--append` in an xmlchange command
1b6b6e26a (in clm5.0.dev013 and release-clm5.0.03) introduced an xmlchange of `CLM_BLDNML_OPTS` in the decStart testmod. However, this xmlchange command was missing an `--append` flag. This means that tests using this testmod (or ciso_decStart, which includes decStart) end up testing the wrong thing, because any existing options in `CLM_BLDNML_OPTS` get overwritten.

I'll put in a fix for this.

billsacks:
Add missing --append flag to decStart xmlchange command

### Specific notes

Fixes #424 

Answers will change for some tests: decStart and ciso_decStart

Testing performed:
Tested creation of `ERP_D.f10_f10_musgs.IHistClm50Bgc.roo2_gnu.clm-decStart`

### DIFF
--- a/cime_config/testdefs/testmods_dirs/clm/decStart/shell_commands
+++ b/cime_config/testdefs/testmods_dirs/clm/decStart/shell_commands
@@ -1,2 +1,2 @@
 ./xmlchange RUN_STARTDATE=2001-12-30
-./xmlchange CLM_BLDNML_OPTS=-ignore_warnings
+./xmlchange CLM_BLDNML_OPTS=-ignore_warnings --append


billsacks:
@ekluzek note that this is a PR into the release branch (I have already brought this into master).

ekluzek:
I suppose we could put these on the branch right away. But, I think we might as well wait a bit until we have something we want to make a tag with. I might forget about it otherwise.

PR #437, opened 2018-07-06
ChangeLog: ask about significant changes to various configurations

billsacks:
It no longer will work to denote these changes via a tag numbering
convention, so Dave, Erik and I decided that we should call this out in
the ChangeLog.

### DIFF
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -24,11 +24,14 @@ Known bugs found since the previous tag (include github issue ID): [If none, rem
 Significant changes to scientifically-supported configurations
 --------------------------------------------------------------
 
-Does this tag change answers significantly for any of the following configurations?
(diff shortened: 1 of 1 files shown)


billsacks:
@dlawrenncar @ekluzek I remembered that we said we'd add this once we started with the new tag numbering. I am not tied to the specific location or wording of this new text in the ChangeLog, so feel free to edit as you see fit.

billsacks:
Okay, looks good - thanks.

billsacks:
@ekluzek I changed the format for marking which configurations change answers. I think this will make it easier for readers to see which configurations are changed if they are just quickly skimming through the ChangeLog. See what you think.

ekluzek:
Yes, that's simpler and makes the default to be nothing significant happened, which is the usual case.

PR #449, opened 2018-07-26
Bring in snowmip diagnostic fields

ekluzek:
### Description of changes

Adds in new fields needed for SNOWMIP

### Specific notes

Contributors other than yourself, if any: flanner, swensosc, justinperket

Are answers expected to change (and if so in what way)? No (only new diagnostic fields added)

Add use_SSRE namelist item

Testing performed, if any:

Ran aux_clm test on cheyenne and hobart and tests work as expected and are bit-for-bit
(however, new fields weren't on by default)


ekluzek:
OK, looking at this I realize more work needs to be done on this. There's some name comments that need to come out. It also currently doesn't test the new fields, so I've done some testing with use_ssre turned on and that's looking fine. So I also need to add a test with the new fields. And it looks like there maybe some duplication that could be eliminated that I'm going to look into.

ekluzek:
The following tests all work fine with use_ssre=T. So that gives me some confidence that the new option is working.

SMS_Lm1_D.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-snowlayers_3_monthly
SMS_Ld5_D.f09_g16.I1850Clm50BgcCrop.cheyenne_intel.clm-cmip6
SMS_Lm37.f10_f10_musgs.I1850Clm50SpG.cheyenne_intel.clm-glcMEC_long
SMS_D_Lm6_P144x1.f45_f45_mg37.I2000Clm50Fates.cheyenne_intel.clm-FatesColdDef
SMS_D_Ly2.1x1_brazil.IHistClm50BgcQianGs.cheyenne_intel.clm-ciso_bombspike1963
ERP_D_P36x2_Ld30.f10_f10_musgs.I2000Clm50BgcCruGs.cheyenne_intel.clm-default
SMS_D_Ly2.1x1_numaIA.IHistClm50BgcCropGs.cheyenne_intel.clm-ciso_bombspike1963

ekluzek:
I can see that the duplication has already caused problems, because bug 2431 was fixed in the old, but not new version of TwoStream in SurfaceAlbedoMod.

ekluzek:
I removed the duplication and showed that results are still the same. This is much more maintainable now.

src/biogeophys/test/Irrigation_test/test_irrigation.pf
```diff
@@ -645,26 +630,26 @@ contains
...
-       this%time_prev = this%time_prev + dtime
+       call advance_timestep()
     end do
     ! The following assertion is mainly here to make sure the test is working as intended
     @assertEqual(0._r8, this%waterflux%qflx_irrig_patch(bounds%begp))
 
     ! Now reset time, change soil moisture, and make sure that irrigation happens as expected
-    this%time_prev = time_prev_orig
+    call unittest_timemgr_teardown()
+    call unittest_timemgr_setup(dtime=dtime)
```
billsacks: I don't think you need to teardown and re-setup the time manager here: you should be able to simply reset the time. (If necessary, you could use the same time but the following day.)

src/utils/test/clm_time_manager_test/test_clm_time_manager.pf
```diff
@@ -220,4 +221,156 @@ contains
...
+    ! Check for local noon at Greenich
+    londeg = 0.0_r8
+    secs = 3600*12
+    call set_date(yr=2, mon=3, day=1, tod=secs)
+    expected = secs
+    @assertEqual( expected, get_local_time( londeg ) )
+    londeg = 360.0_r8
+    @assertEqual( expected, get_local_time( londeg ) )
+
+    ! Check for local noon on otherside of the world
```
billsacks: For this and some of the other tests you added: I find it significantly easier to understand what tests are doing if they follow the generally-recommended unit testing organization of arrange-act-assert (or equivalently: setup-exercise-verify). That is, each test should test one thing. It gets hard to follow tests when they call the system under test multiple times in a single test.

src/utils/clm_time_manager.F90
```diff
@@ -1484,6 +1488,89 @@ end subroutine get_rest_date
...
+    integer  :: yr, mon, day    ! year, month, day, unused
+    integer  :: secs            ! seconds into the day
+    real(r8) :: lon             ! positive longitude
+    !---------------------------------------------------------------------------------
+    SHR_ASSERT( londeg >= -180.0_r8, "londeg must be greater than -180" )
+    SHR_ASSERT( londeg <= 360.0_r8,  "londeg must be less than 360" )
+    call  get_curr_date(yr, mon, day, secs, offset )
+    lon = londeg
+    if ( lon < 0.0_r8 ) lon = lon + 360.0_r8
+    get_local_time  = secs + nint((lon/degpsec)/real(dtime,r8))*dtime
```
billsacks: The division by dtime here yields incorrect results. See https://github.com/ESCOMP/ctsm/issues/491#issuecomment-422640470 for details.

ekluzek:
@billsacks thanks for the review! I'll go back through and make changes and/or comments as appropriate. 

Note, that I went through several iterations of the compare_local_time unit test and part of how I knew it was working, was changing the delta degree. The last thing I did was to change the time it ran over from just a day, to a full year. I never got it to fail running with gnu on my laptop, but I didn't run it anywhere else. I'll comment more later.

src/utils/clm_time_manager.F90
ekluzek: @billsacks and I went over this with @olyson and we figured out why this was done this way. This is just to make sure the code in SurfaceRadiationMod.F90 would match the closest time-step to local noon. This can be handled appropriately with the is_near_local_noon function. So we can remove the dtime logic that's here. The bottom line is that we want to use the method used in Irrigation for going forward. I still plan to have a bit-for-bit tag, that has two different methods and then we'll remove one with a tag that has answer changes.

src/utils/test/clm_time_manager_test/test_clm_time_manager.pf
ekluzek: @billsacks and I talked about this. The general advice is to have one test per thing tested to make it easiest to read. It also helps in the theoretical sense so that broken tests will all be shown rather than just the first assert that fails within a test. Bill didn't think this had to change for this PR, but we do want to try to follow best practices for unit testing in general.

src/biogeophys/test/Irrigation_test/test_irrigation.pf
ekluzek: Yep, you are right.

PR #462, opened 2018-08-02
Add namelist item to calculate all heat stress indices only if requested; to speed up model

olyson:
### Description of changes
Add a namelist item (calc_human_stress_indices_all) to calculate the full set of heat stress indices only if the namelist item is true.  This item is in addition to the calc_human_stress_indices namelist item that already existed in the code.
The motivation for this is to speed up the model.  While most heat index history fields are turned off by default, the computationally expensive wet bulb temperature, which is required by several heat indices, was still being calculated.
Now, with the following settings (default):
calc_human_stress_indices = .true.
calc_human_stress_indices_all = .false.
the wet bulb temperature calculation is turned off and the only heat indices that are output are SWBGT, HUMIDEX, WBT, and HIA.
Setting calc_human_stress_indices_all=.true. adds in the the wet bulb temperature calculation and additional heat indices.
In theory, a user could naively set:
calc_human_stress_indices = .false.
calc_human_stress_indices_all = .true.
but I've added in a log error in CLMBuildNamelist.pm to check for this.

### Specific notes

No

None

No

The namelist item calc_human_stress_indices_all was added (default is false).

Testing performed, if any:
I tested the different combinations of calc_human_stress_indices and calc_human_stress_indices_all in a I2000 case for two years.
Per Erik, I also ran the aux_clm test suite on cheyenne with the intel compiler (158 tests) and all passed (no pending) with three failures that are listed as EXPECTED.


olyson:
Sounds like a very good idea, I'll work on that, thanks.

olyson:
I've pushed a commit to use logical variables within the code.

ekluzek:
OK, I added in indentation on some of the new if statements and pushed it to the branch. Checked it by running the case: SMS_Lm1_D.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-snowlayers_3_monthly and ensured answers were same as base version.

PR #468, opened 2018-08-04
Minor bug fixes, cleanup, documentation and enhancements

### LINKED ISSUE #24: ncd_io_1d_log_glob is broken
**Bill Sacks < sacks@ucar.edu > - 2014-01-23 11:14:21 -0700**
**Bugzilla Id:** [1906](http://bugz.cgd.ucar.edu/show_bug.cgi?id=1906)
**Bugzilla CC:**  muszala@ucar.edu,  mvertens@ucar.edu,  rfisher@ucar.edu, 

When trying to read in logical values with ncd_io in clm4_5_59, I get this message:

(shr_sys_abort) ERROR: ncd_io_1d_log_glob ERROR: read in bad integer value(s) for logical data

It looks like the relevant variable is never set: starting line 1216 in ncdio_pio.F90:

          allocate(idata1d(size(data))) 
          data = (idata == 1)
          if ( any(idata1d /= 0 .and. idata1d /= 1) )then
             call shr_sys_abort( subname// &
                  ' ERROR: read in bad integer value(s) for logical data' )
          end if

This worked in clm4_5_52. My guess is that it got broken in clm4_5_53, which created the .in version of ncdio_pio.F90.

### LINKED ISSUE #120: Incorrect comments in Biogeophysics1Mod.F90
**Liang Chen < lchen15 > - 2015-08-18 13:11:28 -0600**
**Bugzilla Id:** [2202](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2202)
**Bugzilla CC:**  rfisher, 

In Biogeophysics1Mod.F90, the variables hr is not relative humidity, but the alpha soil (Equation 5.73 in CLM45_Tech_Note). 

    real(r8) :: hr      !relative humidity
    real(r8) :: hr_road_perv  !relative humidity for urban pervious road

### LINKED ISSUE #217: Change some cheyenne_gnu tests to cheyenne_intel
Following from the yellowstone to cheyenne testlist migration:

gnu can't catch all of the issues that intel, pgi or nag can in debug mode: see https://github.com/ESMCI/cime/issues/1974 . Therefore, it would be best if we avoided having our only debug tests of certain options be with gnu. So I'd like to change the compiler used for these tests: Currently the only `_D` tests of these compsets or testmods are with gnu:

`ERP_P180x2_D_Ld5.f19_g17_gl4.I1850Clm50BgcCropG.cheyenne_gnu.clm-glcMEC_increase`: change to `cheyenne_intel`

`ERP_P180x2_D_Ld5.f19_g17.I2000Clm50BgcDvCrop.cheyenne_gnu.clm-crop`: change to `cheyenne_intel`

`SMS_D_Ld3.f10_f10_musgs.I1850Clm50Bgc.cheyenne_gnu.clm-interp_f19_crop`: change to `hobart_nag`

`SMS_D_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm45BgcCropQianGs.cheyenne_gnu.clm-cropMonthOutput`: change to `cheyenne_intel`

At least for now, I'm planning to keep these two with gnu, even though they're the only `_D` tests with interp_f19_noncrop, since they were in gnu to begin with on yellowstone:

```
SMS_D_Ld3.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_gnu.clm-interp_f19_noncrop
PET_P36x2_D.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-interp_f19_noncrop
```

@ekluzek I'm not sure whether we should also put these on a different compiler; do you have thoughts on whether they should be moved to intel or nag?:

1. `ERP_D_Ld5.f10_f10_musgs.I2000Clm50BgcCruGs.cheyenne_gnu.clm-ciso_flexCN_FUN` - this is the only `_D` test with this testmod. But we have `_D` tests of other ciso testmods with intel. Now that flexCN and FUN are defaults (right?) I'm not sure if this is really a unique configuration.

2. `ERS_D_Ld6.f10_f10_musgs.I1850Clm45BgcCrop.cheyenne_gnu.clm-clm50CMIP6frc` - this testmod only appears in this test. This test pre-dates the yellowstone to cheyenne testlist migration.

I'll wait to do this until after the git transition, to keep the git transition tag to the minimum of what's needed.

### LINKED ISSUE #245: Put all .gitignore entries in top-level file
I started having multiple `.gitignore` files in different places. However, I was just doing some googling, and it sounds like the most common convention is to put everything in the top-level .gitignore file. So I'd like to rework things to do that. So, for example, this entry in `src/unit_test_stubs/utils/.gitignore`: `restUtilMod_stub.F90` would go in the top-level as `/src/unit_test_stubs/utils/restUtilMod_stub.F90` (I think it's correct to have this leading slash here: I think this says to match exactly this path).

### LINKED ISSUE #272: code should error on missing mxsoil_color when SOIL_COLOR is used
This bug only manifests under unique circumstances, namely when using a custom surface dataset and the  variable `mxsoil_color` has accidentally been forgotten (this happened to me due to a bug or "feature" in CDO's merge).

When `mxsoil_color` is not present, the init code in `SurfaceAlbedoMod.F90` simply assumes 8 color classes. However my input dataset assumes 20 classes, which are reflected in the values of variable SOIL_COLOR. This by itself is an erroneous model state, but it also led to the production of weird errors in Snicar en in the production of NaN's. Root cause: code in subroutine SoilAlbedo accesses memory that has not been allocated.

This mismatch state can be prevented simply by requiring `mxsoil_color` and erroring out when it does not exist, and `SOIL_COLOR` does. A more complicated solution would be to somehow guess the value of `mxsoil_color` from the input , but this requires more work.

### LINKED ISSUE #283: Add more helpful message about need to do init_interp with wrong number of vertical layers
**Original comment from 2015-11-17. It's possible that we already resolved this.**

I believe that you do NOT currently get any useful error messages about needing to set use_init_interp = .true. if you try to run with an initial conditions file with the wrong number of vertical layers: you just get a cryptic death in this case.

### LINKED ISSUE #367: For cmip6 runs: Turn on cpl hist output needed to drive a TG compset
Land Ice Working Group members request that, for cmip6 runs, we output the coupler history fields needed to drive a later TG compset (CISM-only forced by dlnd). The easiest way to do this is to add a `user_nl_cpl` file in CTSM's `cime_config/usermods_dirs/cmip6_output` that contains:

```
histaux_l2x1yrg = .true.
```

This needs to wait until we point to a cime version that includes https://github.com/ESMCI/cime/pull/2530

### LINKED ISSUE #412: Fix documentation of init_interp_method
The documentation for `init_interp_method` says that the method `use_finidat_areas` can be used when transitioning from a spinup to a transient run. But I realized that this isn't true in general, because spinup runs typically have irrigation off and transient runs have irrigation on.

I'll fix this documentation.

### LINKED ISSUE #419: Do not allow SOYFIXN diagnostic field with FUN
-bug
Nitrogen fixation in soy should be handled by FUN in CLM5, thus CNSoyFix is not called in CNDriverMod.F90. The variable SOYFIXN should be off by default if FUN is active.  Otherwise the output is zero for default CLM5 runs, which may confuse users.  
When fun is on, users will need to turn of PFT level output of NFIX to get fixation rates from soy.

This can be done by changing line 954 of CNVegNitrogenFluxType.F90 as follows
    if (use_crop .and. .not. use_fun) then

### LINKED ISSUE #467: Increase wallclock limit for test
@fischer-ncar requests:

> Whenever you get a chance, can you increase the wallclock limit for.
>
> ERP_P36x2_Lm36.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-irrigOn_reduceOutput

billsacks:
### Description of changes

A collection of minor bug fixes, code cleanup, documentation and enhancements, all bit-for-bit.

This includes most of what came into ctsm1.0.dev006, with the exception of:
- Some commits that only apply to dev005, which isn't coming to this release branch (5fbbab83, 5207582a)
- Some tweaks to the test list that I made after merging dev005 into the branch destined for master (5c226bdf, 48ac8d75)
- The ChangeLog update for dev006 (89983415)

@ekluzek this PR follows the plan we discussed yesterday. However, you would then not have the test list tweaks or the ChangeLog update; I'm not sure if that matters to you.

An alternative would be to do what you've been doing up until now: merging the latest master into the release branch after first selectively reverting commits that you don't want on the release branch. In this case, you would revert the entire dev005 change, and also revert commits 5fbbab83 and 5207582a. The advantage of that would be consistency. It might be worth thinking about how you want these two to look:

```
git log --no-merges escomp/release-clm5.0..master
git log --no-merges master..escomp/release-clm5.0
```

To see how they looked until recently:

```
git log --no-merges escomp/release-clm5.0..ctsm1.0.dev004
git log --no-merges ctsm1.0.dev004..escomp/release-clm5.0
```

Currently the first is empty and the second shows a revert and a few commits specific to the release branch. If you want to stay consistent with that, then you could stick with what you've been doing up until now.

### Specific notes

- Fixes #24 (ncd_io_1d_log_glob is broken)
- Fixes #120 (Incorrect comments in Biogeophysics1Mod.F90
- Fixes #217 (Change some cheyenne_gnu tests to cheyenne_intel)
- Fixes #245 (Put all .gitignore entries in top-level file)
- Fixes #272 (Code should error on missing mxsoil_color when SOIL_COLOR
  is used)
- Fixes #283 (Add more helpful message about need to do init_interp with
  wrong number of vertical layers)
- Fixes #367 (For cmip6 runs: Turn on cpl hist output needed to drive a
  TG compset)
- Fixes #412 (Fix documentation of init_interp_method)
- Fixes #419 (Do not allow SOYFIXN diagnostic field with FUN)
- Fixes #467 (Increase wallclock limit for test)

Testing performed, if any: Ran full test suite on the version merged up to dev005


ekluzek:
I'm going to bring this change in on the release branch, without extensive testing because it already had extensive testing on master. I won't tag it though. But, will run extensive testing after the next PR.

PR #470, opened 2018-08-05
Avoid glacier dynamic landunit adjustments in first time step

### LINKED ISSUE #340: Avoid generating dynamic landunit adjustment fluxes for glacier changes in the first timestep
I have struggled with how to handle changes in glacier area in the first timestep of a run. Specifically, I've struggled with the case: If you're doing a hybrid run or a startup run that points to a spunup initial conditions file, and the CTSM initial file disagrees with the CISM initial file. In this case, CTSM will see a change in glacier area in the first time step.

Currently, I've been treating this the same as if a change in glacier area arises during the run loop, with the consequent dynamic landunit adjustment fluxes. I can still envision scenarios where this may be what's wanted. However, the recent experience with an inadvertent glacier area change resulting in huge dynamic landunit adjustment fluxes has led me to feel that, in general, we should let CTSM update to the initial CISM glacier areas in the first time step without generating any adjustment fluxes. That is, it should effectively sync up with CISM areas in initialization, even if the CISM areas differ from what's on the CTSM restart file.

Example of when you might want the current behavior (dynamic landunit adjustment fluxes generated in the first time step): If you run a case with an evolving ice sheet, then run CISM offline for some time to evolve the ice sheet further, then plug in the updated CISM initial conditions file into a hybrid case and expect CLM to generate adjustment fluxes as if you had just continued the first run with CISM coupled in the whole time.

Example of when this causes problems, and the new behavior will be better: If you run a case without an evolving ice sheet, then (unknown to you) CISM changes in a way that affects its initial glacier cover, then you do a hybrid run off of the original case with the new CISM code: With the current code, you'd get big adjustment fluxes in the first time step, but this generally isn't what you want.

I'm becoming more concerned about supporting the second example than the first example.

Incidentally: It turns out that the big adjustment fluxes due to water and energy conservation currently don't affect the coupled system. This is because they are generated in time step -1 of the model (a pre-time step that is done before the real run loop gets underway), and apparently fluxes at that time don't affect the rest of the system. But this is still relevant for its impact on state variables that are adjusted more rigorously with dynamic landunits - currently, carbon and nitrogen states, but eventually some water and energy states as well. I feel that the new, proposed solution will more often do the right thing in this respect - keeping glacier state variables at the values from spunup glacier columns, rather than possibly adjusting them to be a blend of glacier and natural veg state variables in the first time step, if CTSM's glacier areas disagree with CISM glacier areas.

A final reason why I like the new, proposed solution is that I think it is consistent with what we'd get if we could update CTSM to match CISM's glacier areas in initialization, rather than needing to wait until the run loop. That's what I'd really like to do, but the current mct-based driver doesn't allow that. I hope we'll be able to do that when we switch to a NUOPC-based driver/mediator. (Although if we really wanted to maintain the current behavior, we could probably come up with a way to do so even if we're generally getting CISM's glacier area in initialization.)

Note that, in all cases, changes to CISM's glacier areas in the first time step of a branch or restart run *will* result in adjustment fluxes (the same as if you had run straight through without stopping).

@whlipscomb @lofverstrom @lvankampenhout  do any of you have any thoughts on this? I realize these points are subtle, and I'm not sure if I'm explaining this very well. Don't feel the need to spend a lot of time thinking about this, but I wanted to give you the opportunity to weigh in if you want to.

billsacks:
### Description of changes

Always avoid generating dynamic landunit adjustments for glacier area changes in the first timestep of a startup or hybrid run - not just for cold start or interpolated start. See discussion in #340 for rationale.

### Specific notes

- Fixes #340 

 In general, can have answer changes in startup / hybrid tests that do not use init_interp, and use CISM (even NOEVOLVE), due to:
- different PE layout in new case vs. the one that generated the finidat (roundoff-level diffs)
- different glacier area on finidat file vs. what's in CISM (greater than roundoff-level diffs)

Testing performed, if any: Ran full aux_clm test suite


ekluzek:
The one comment I have are about variables that predate this change. is_cold_start and is_interpolated_start. I'm wondering if instead of having several logical variables that have to be coordinated, if it should be a single variable that has a few valid options. Something like: start_type = 'COLD', 'INTERPOLATED', 'RESTART'.

But, that's a comment that essentially predates this PR, hence we should think about it as a clean-up step rather than a change here.

PR #472, opened 2018-08-06
Always avoid generating dynamic landunit adjustments for glacier area

ekluzek:
changes in the first timestep of a startup or hybrid run - not just for
cold start or interpolated start.

This is the change in ctsm1.0.dev007.

### Specific notes

CTSM Issues Fixed (include github issue #): #340

In general, can have answer changes in startup / hybrid tests that do not use init_interp, and use CISM (even NOEVOLVE), due to:

different PE layout in new case vs. the one that generated the finidat (roundoff-level diffs)
different glacier area on finidat file vs. what's in CISM (greater than roundoff-level diffs)

Any User Interface Changes (namelist or namelist defaults changes)? None

Testing performed, if any: None so far Testing was done on master


billsacks:
We need to get confirmation from @dlawrenncar and possibly others before bringing this to the release branch. The main thing I'm unsure of is what simulations/configurations we are / are not allowed to change answers for on the release branch.

ekluzek:
@billsacks, we are changing answers due to changes in forcing with simulations right now. We are also changing answers due to updates in initial conditions as well. This change only effects answers at startup, so is smaller than either of those things. As a matter of fact with my testing many cases appear identical, I'm not even sure if there are tests that DO change answers.

My understanding of Gokhan's statement "I am comfortable with this path. That is i) having virtual columns on with the land initial conditions coming from a 1-year additional land simulation with identical layout as in the coupled simulation for the land model and ii) when this option is off, the land initial conditions would still produce bfb results....." is that this PR is REQUIRED to satisfy his requirements, since the simulations that will be done will be done with release branch tags. Am I missing something?

dlawrenncar:
My understanding is the same as Erik's, but I could be misunderstanding
something as it seems that there are a lot of subtleties here.

ekluzek:
Only two tests show change in answers

ERI_N2_Ld9.f19_g17.I2000Clm50BgcCrop.cheyenne_intel.clm-default
SMS_Lm13.f19_g17.I2000Clm50BgcCrop.cheyenne_intel.clm-cropMonthOutput

There's also two single-point tests (that are in the share queue and CISL says the share queue has problems) that I'm having trouble with.

Issue #461, opened 2018-08-01
year_last for CTSM streams should be increased to 2015


ekluzek:
### Brief summary of bug

For transient simulations most of the stream files that CTSM uses only go to 2000, 2005, or 2010. Since, simulations are now going to 2015 and most of the files do as well, we need to increase this.

### General bug information

**CTSM version you are using:** ctsm1.0dev004

This should be changed both in namelist_defaults file and in the 20thC_transient use case.

### Important details of your setup / configuration so we can reproduce the bug

IHist cases for Clm45 or Clm50 would show this issue.

### Important output or errors that show the problem

Mike Mills sent us this snippet of the default namelist for a transient case.

```
&ndepdyn_nml
 model_year_align_ndep = 1850
 ndepmapalgo = 'bilinear'
 stream_fldfilename_ndep = '/glade/p/cesmdata/cseg/inputdata/lnd/clm2/ndepdata/fndep_clm_hist_simyr1849-2006_1.9x2.5_c100428.nc'
 stream_year_first_ndep = 1850
 stream_year_last_ndep = 2005
/
&popd_streams
 model_year_align_popdens = 1850
 popdensmapalgo = 'bilinear'
 stream_fldfilename_popdens = '/glade/p/cesmdata/cseg/inputdata/lnd/clm2/firedata/clmforc.Li_2017_HYDEv3.2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2016_c180202.nc'
 stream_year_first_popdens = 1850
 stream_year_last_popdens = 2010
/
&urbantv_streams
 model_year_align_urbantv = 1850
 stream_fldfilename_urbantv = '/glade/p/cesmdata/cseg/inputdata/lnd/clm2/urbandata/CLM50_tbuildmax_Oleson_2016_0.9x1.25_simyr1849-2106_c160923.nc'
 stream_year_first_urbantv = 1850
 stream_year_last_urbantv = 2005
 urbantvmapalgo = 'nn'
/
&light_streams
 lightngmapalgo = 'bilinear'
 stream_fldfilename_lightng = '/glade/p/cesmdata/cseg/inputdata/atm/datm7/NASA_LIS/clmforc.Li_2012_climo1995-2011.T62.lnfm_Total_c140423.nc'
 stream_year_first_lightng = 0001
 stream_year_last_lightng = 0001
/
```

Note, we might as well make the last year the last year data is available on the particular file, rather than having it truncate earlier than the end of the data on the file.

[...]

PR #477, opened 2018-08-09
Move changes from release-clm5.0.05 onto master

ekluzek:
### Description of changes

Update to use of new 1850 Nitrogen-deposition file as well as correcting year_last for some of the CLM streams. These are the changes that went into release-clm5.0.05 being moved to master.

### Specific notes

CTSM Issues Fixed (include github issue #): #461

  For 1850 cases with BGC on
  And for post 2005 for IHist cases.

Testing performed, if any: testing was done on the release branch
Will be repeated here before this comes to master.


ekluzek:
Testing now looks as expected, with just expected fails failing. Differences are shown only for 1850 cases.

PR #483, opened 2018-08-24
Collapse2gencrop branch: Allow the model to not need to read 16-pft datasets and rather read 78-pft datasets

slevisconsulting:
### Description of changes
Erik converted his collapse2gencrop branch from svn to git. Sam cloned Erik's branch and updated to ctsm1.0.dev008. Resolved conflicts and ran this test successfully:
qcmd -- ./create_test ERP_D_P36x2_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default --compare ctsm1.0.dev008 --walltime 0:30 --project UIDA0003

Details documented here:
https://docs.google.com/document/d/1ey5DDYWail8T0Jy7X8K3Mo-cOaDtKTSoXQjVMrZADGI/edit?usp=sharing

### Specific notes

No, except in special case of collapsing pfts and code for that case is in development.

Testing performed, if any:
qcmd -- ./create_test ERP_D_P36x2_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default --compare ctsm1.0.dev008 --walltime 0:30 --project UIDA0003


ekluzek:
Next steps that I see would be:

- Make sure all standard testing works (cheyenne intel/gnu, hobart intel/gnu/nag)
- Make sure unit testing is working (also review the new tests I added and make sure it's complete, and if not add more unit tests for the new capability)
- Keep the branch updated to the latest ctsm tags on master
- Add some specific tests for the new capability to run SP mode from 79-pft surface and landuse.timeseries datasets.
- Run the above tests by hand and also add to the test suite
- Review everything make sure it all makes sense
- Look at opportunities to improve/cleanup the code that's changed
- Ask @billsacks and I to review it once it's both working and you think it's complete

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_78_to_16/user_nl_clm
```diff
@@ -0,0 +1,3 @@
+for_testing_allow_non_annual_changes = .true.
+fsurdat = '/glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/surfdata_10x15_78pfts_CMIP6_simyr1850_c170824.nc'
```
billsacks: Please add a comment about the origin of these files. e.g., are these the files that are used out-of-the-box in some other configuration? If so, please state something like, "These are the files that are used out-of-the-box for compset X, resolution Y". This is important to keep this test maintainable as we introduce new surface datasets: we need to know how to update this testmod.

It would also be helpful if the README file for this test pointed out that it only works at f10 resolution.

ekluzek: You should also change the name of the test to have "f10" in the name, so it's obvious that it can only work at f10 resolution. That's something we've done in the past when we have a resolution dependent test.

src/main/clm_varpar.F90
```diff
@@ -104,16 +105,19 @@ subroutine clm_varpar_init()
...
     !------------------------------------------------------------------------------
 
     ! Crop settings and consistency checks
 
-    if (use_crop) then
-       numpft      = mxpft   ! actual # of patches (without bare)
-       numcft      =  64     ! actual # of crops
+    ! slevis: I tried ".or. get_do_transient_crops()" & got unhelpful bld error
+    !         so added this fsurdat check instead
+    if (use_crop .or. index(fsurdat,'78pfts') > 0) then
```
billsacks: I am very uncomfortable with this change. We should never have logic based on the file name.

ekluzek: Yes, absolutely. It should query the metadata for anything it needs to know.

billsacks:
@slevisconsulting Thanks a lot for adding some new tests covering this behavior! However, what was your reason for making the two new tests based on the glcMEC_decrease testmod? Is that a fundamentally important part of this test? I'm thinking that these tests are orthogonal to glcMEC stuff, and so shouldn't be based off of that testmod.

If you _do_ want these to be based off of the glcMEC_decrease testmod, then you should do this by having the `include_user_mods` file point to the `glcMEC_decrease` directory, and then only having the differences from that testmod in your new testmods directories. We try hard to avoid copying and pasting one testmod directory to another, by leveraging the include_user_mods mechanism.

I also noticed that you have some unused files in one of your new testmods directories. Can you please remove these? In git, you should try to avoid committing large files to the repo if they aren't really needed, because once they're in a commit, they're in history forever, bloating the repository. So my preference would be if you rewrote the last commit to remove these big new files (using git commit --amend followed by a force push... I can help you with that if you don't know how to do that).

As for the change of maxpatch_pft vs. numpft+1: I think @ekluzek can address this better than I can. However, if we do stick with numpft+1 in the places where you have changed it, then I'd request that a new variable be added in clm_varpar that is currently set to numpft+1 and is named with something meaningful, to make this code more clear and to make it easier to change this in the future (if numpft+1 is no longer the correct expression).

Finally, a note about commit messages: your commit messages will show up nicer if you wrap your lines to 72 characters. It's especially helpful if keep your first line to 72 characters or less. Sometimes it takes some work to summarize what you did succinctly in 72 characters or less, but this keeps things more readable, especially on github.

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_78_to_16/user_nl_clm
slevisconsulting: Bill: I added comments in the README following these suggestions, though I'm not sure whether they are you had in mind. Pls confirm after next commit.

Erik: This is the test name that I added in testlist_clm.xml:
<test name="ERS_D_Ld10" grid="f10_f10_musgs" compset="IHistClm50SpG" testmods="clm/collapse_pfts_78_to_16">
so the f10 shows up in the grid entry. Are you recommending that it should appear elsewhere, as well?

src/main/clm_varpar.F90
slevisconsulting: I looked for an alternative:

As far as I can tell, this piece of code executes before the model reads any surface or landuse data and before the model defines get_do_transient_crops. As a result, I believe that I do not have access to numcft or ncid or get_do_transient_crops.

This does not justify keeping my hack; however, I think our goal is to converge to one set of input data so that the if-then-else block will go away altogether. We will be left only with these two lines (not surrounded by ifs):
`numpft        = mxpft   ! actual # of patches (without bare)
 actual_numcft =  64     ! actual # of crops`

Let me know if you disagree and think that I should invest the time making this if-then-else more robust.

slevisconsulting:
Bill,  I would not mind your help with the 
`git commit --amend ... `
`force push ... `
sequence, keeping in mind that I already merged to the latest master since that commit and resolved corresponding conflicts and committed again. So I hope we can still make the correction.

I used the glcMEC_decrease test as a template to set up my new tests only because these were the first clm tests that I ever created. I will follow your advice and make the new tests distinct from glcMEC_decrease.

I agree about assigning numpft+1 to a variable. I will wait for confirmation from Erik before I proceed with that.

I think the 72-char issue arises because I often do
`git commit -m 'text text text on and on'`
and I think I can solve this by just typing `git commit` and using the editor.

slevisconsulting:
Bill, I ended up with an error that seems like a can of worms to me, so I'm aborting the amend. This is what I tried:
`git rebase --interactive 'b97f0fd'
`In the editor I changed "pick" to "edit" for the "1b1e7b5" commit and saved and quit out of the editor.
`git rm -r modified_paramfile_not_needed
`git commit --all --amend --no-edit
`git rebase --continue

error: could not apply dd9cd6d... added TracerConsistencyCheck; breakout of isotope lnd2atm/atm2lnd vars
When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
Could not apply dd9cd6d4b16a176fd03bb9e397838cf0855f6f39... added TracerConsistencyCheck; breakout of isotope lnd2atm/atm2lnd vars

`git rebase --abort

billsacks:
@slevisconsulting I'm sorry to have sent you down the rabbit hole of `git rebase --interactive`. I realized that I did not give proper consideration to your caveat:

> keeping in mind that I already merged to the latest master since that commit and resolved corresponding conflicts and committed again

which is what made this case difficult: I forgot that rebasing is difficult when you've already done a merge from master into your branch. I guess I'm still learning what's reasonably doable and what's a real pain in git, and I'm sorry to have dragged you along on that....

That said: Partly out of wanting to hone my git skills, and partly out of stubbornness, I have managed to accomplish the necessary rebase. The two keys were: (1) adding the `--rebase-merges` option to the `git rebase -i` command; and (2) even with that option, git forces you to redo the conflict resolution that you did when merging master into your branch (fortunately it was pretty easy in this case).

So if you're willing to go along with me just a bit more here, can you do the following so we can get a cleaner history? (I'm assuming here that you're in a git clone with collapse2gencrop checked out.)

(1) `git remote add billsacks https://github.com/billsacks/ctsm.git`

(2) `git fetch billsacks`

(3) `git diff billsacks/collapse2gencrop`

This should show no diffs; if there are any diffs, stop. (It probably means that you have committed locally something beyond what's pushed to your fork. But looking at your sandbox on cheyenne right now, you should be good here.)

(4) `git status`

The status should show that you have a clean sandbox before doing the following – i.e., no uncommitted changes. (Untracked files are okay.) (Again, looking at your sandbox on cheyenne right now, you should be good here.)

(5) `git reset --hard billsacks/collapse2gencrop`

This is a potentially destructive command, but it's safe here as long as (3) shows no diffs and (4) shows no uncommitted changes.

(6) `git push slevisconsulting +collapse2gencrop`

The `+` does a force push, which is necessary given the modified history.

billsacks:
Oh, and as to this:

> By the way 2: Here in github, I saw a trash icon and wondered whether I could simply click on that to delete the files. I tried this and it may have worked. However, my `git push` gets rejected now. I need to `git pull` first.

That is effectively the same as doing a `git rm` of the files followed by a `git commit` on the command line: it doesn't actually remove the files from history.

slevisconsulting:
Bill, I executed the list of commands, so we should be good.

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_16_to_16/user_nl_clm
```diff
@@ -0,0 +1 @@
+for_testing_allow_non_annual_changes = .true.
```
billsacks: I wouldn't think this user_nl_clm is needed for this test. Can this test simply use the 'default' testmod rather than having its own testmod?

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_78_to_16/user_nl_clm
```diff
@@ -0,0 +1,3 @@
+for_testing_allow_non_annual_changes = .true.
```
billsacks: I don't think this 'allow_non_annual_changes' is needed for this test, and should be removed.

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_16_to_16/user_nl_clm
slevisconsulting: Removing.
(This was my original thought process: I did not know whether it helped with "testing" in general and assumed that it would not likely hurt anything to leave there.)

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_78_to_16/user_nl_clm
slevisconsulting: Removing.

billsacks:
Thanks a lot, @slevisconsulting !

Just a minor note about commit messages: GitHub and many other tools are set up assuming that you follow the convention: First line no more than 72 characters, then a blank line, then additional lines, also limited to 72 characters (though it's less important that the additional lines follow this convention). If you can try to make the initial line of your commit message short like this, it's easier for others to read a summary of activity.

slevisconsulting:
Sorry about that. I interpreted your previous caution about that (7 days ago) to mean 72 chars/line even if my first sentence does not complete in the first line. I will try to get it right next time :-)

slevisconsulting:
I appreciate your patience, and I also suspect that it's hard to train everyone! But worth it.

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_78_to_16/user_nl_clm
billsacks: What Erik meant, I think, is that the testmod should have f10 in the name.

src/main/clm_varpar.F90
billsacks: @ekluzek 's suggestion: have a new routine like surfrdGetMaxVeg that gets the necessary dimension size. Call that before setting the various values here.

cime_config/testdefs/testmods_dirs/clm/collapse_pfts_78_to_16/user_nl_clm
slevisconsulting: Done in this commit:
0ed02f6

src/main/clm_varpar.F90
slevisconsulting: Notes from our meeting on 10/1:

in clm_initializeMod.F90
after call control_init
enter call surfrd_get_maxveg  # similar to other subroutines in surfrd...
before call clm_varpar_init

slevisconsulting: This now done.

slevisconsulting:
As a result of numerous tests in the test-suites FAILing, I have come across a problem with the baseline whereby %crop remain fixed at 1850 values in old-style transient runs. I will open a new Issue for this and will elaborate there.

slevisconsulting:
Cheyenne test-suites PASS except for some TPUTCOMP & MEMCOMP.

Hobart test-suites [in progress]

src/dyn_subgrid/dyncropFileMod.F90
```diff
@@ -165,8 +166,8 @@ subroutine dyncrop_interp(bounds,crop_inst)
     allocate(fertcft_cur(bounds%begg:bounds%endg, cft_lb:cft_ub))
     call fertcft%get_current_data(fertcft_cur)
 
-    if (use_crop) then
-       call collapse_crop_types(wtcft_cur, fertcft_cur, bounds%begg, bounds%endg, verbose = .false.)
+    if (cft_size > 2) then
```
billsacks: Can you explain the need for this change, and the rationale for this particular condition (cft_size > 2)? A comment in the code would help here.

slevisconsulting: With the new "collapse pft" capability, 
`if (use_crop)`
is incorrect here because we intend to read 78-pft data and collapse to 16 pfts when use_crop = .false.

In surfrdMod.F90 we have the following options (including one that calls collapse_crop_types):
`    if ( actual_numcft > 0 ) then
       ! Full crop on file and in CLM
       if ( create_crop_landunit .and. actual_numcft > 2 .and. use_crop)then
          call surfrd_cftformat( ncid, begg, endg, wt_cft, fert_cft, cft_size, natpft_size )  ! Format where CFT's is read in a seperate landunit
       ! Generic crop in file and generic crop in model
       else if ( create_crop_landunit .and. actual_numcft == 2 .and. cft_size == 2 )then
          call surfrd_cftformat( ncid, begg, endg, wt_cft, fert_cft, cft_size, natpft_size )  ! Format where CFT's is read in a seperate landunit
       ! Full crop in file, but model is only with generic crop
       else if ( create_crop_landunit .and. actual_numcft > 2 .and. .not. use_crop)then
          cftsize = actual_numcft
          allocate(array2DCFT (begg:endg,cft_lb:cftsize-1+cft_lb))
          allocate(array2DFERT(begg:endg,cft_lb:cftsize-1+cft_lb))
          call surfrd_cftformat( ncid, begg, endg, array2DCFT, array2DFERT, cftsize, natpft_size )
          call collapse_crop_types(array2DCFT(begg:endg, :), array2DFERT(begg:endg, :), cftsize, begg, endg, &
                                   verbose=.true., sumto=100._r8)
          wt_cft  (begg:,cft_lb:) = array2DCFT (begg:,cft_lb:cft_ub)
          fert_cft(begg:,cft_lb:) = array2DFERT(begg:,cft_lb:cft_ub)
          deallocate(array2DCFT)
          deallocate(array2DFERT)
       else if ( .not. create_crop_landunit )then`

For consistency I will change dyncropFileMod.F90 to the same:
`if (.not. use_crop .and. create_crop_landunit .and. cft_size > 2) then`

I will add a comment in dyncropFileMod.F90.

Long term I see the `cft_size > 2` part becoming obsolete because we will stop using all but the 78-pft datasets.

billsacks:
Thanks, @slevisconsulting !

src/biogeochem/CNPrecisionControlMod.F90
```diff
@@ -24,7 +24,7 @@ module CNPrecisionControlMod
   real(r8), public :: ccrit    =  1.e-8_r8              ! critical carbon state value for truncation (gC/m2)
   real(r8), public :: cnegcrit = -6.e+1_r8              ! critical negative carbon state value for abort (gC/m2)
   real(r8), public :: ncrit    =  1.e-8_r8              ! critical nitrogen state value for truncation (gN/m2)
-  real(r8), public :: nnegcrit = -6.e+0_r8              ! critical negative nitrogen state value for abort (gN/m2)
```
ekluzek: Is this an intentional change?

src/dyn_subgrid/dyncropFileMod.F90
```diff
@@ -165,8 +166,12 @@ subroutine dyncrop_interp(bounds,crop_inst)
...
     call fertcft%get_current_data(fertcft_cur)
 
-    if (use_crop) then
-       call collapse_crop_types(wtcft_cur, fertcft_cur, bounds%begg, bounds%endg, verbose = .false.)
+    ! Call collapse_crop_types when we read 78-pft datasets (ie cft_size > 2),
+    ! while running with use_crop = .false.
+    ! This call also requires create_crop_landunit = .true.
+    ! Same if-statement and call appear in surfrdMod.F90.
+    if (.not. use_crop .and. create_crop_landunit .and. cft_size > 2) then
+       call collapse_crop_types(wtcft_cur, fertcft_cur, cft_size, bounds%begg, bounds%endg, verbose = .false.)
```
ekluzek: This doesn't look like the right logic to me. This might be complex enough that it should have a unit test to make sure it's right. I think you want this to be done when crop is on (so that you collapse some of the crop types onto each other), but also for the case when crop is off and cft_size > 2.

src/dyn_subgrid/dyncropFileMod.F90
ekluzek: @slevisconsulting just one note on your very last statement "cft_size > 2 part becoming obsolete because we will stop using all but the 78-pft datasets" because of the NWP use-case needs, I'm not sure if this is really the case. For climate we will certainly want to transition to 78-pfts only, but I'm not so sure for NWP cases. I think we will need to support more options for them. Just something we have to start thinking about...

src/biogeochem/CNPrecisionControlMod.F90
slevisconsulting: I did not touch this file, so it came in with the merge.

ekluzek: OK, I thought that it could've been my fault. And it is.

So I'm going to check with @olyson and we'll decide if we should keep this or leave this change out.

src/dyn_subgrid/dyncropFileMod.F90
slevisconsulting: You're probably right, and I think this means that I had it right the first time when I just checked for cft_size > 2...

I will look into adding a unit test and will ask for help if I get lost.

billsacks: I can't envision what @ekluzek is thinking here in terms of a unit test. I think this is going to be very hard to unit test as is (i.e., with the logic embedded in this routine as it is now), and I probably wouldn't go there - at least, not as the first unit tests you write for ctsm. @ekluzek were you thinking that the conditional could be extracted to a little standalone logical function like 'cft_needs_collapse'? I was tempted to suggest something like that, but that mainly seems to make sense if the logic is shared between surfrdMod and this file, which based on your comments, sounds like it may not be.

So if the logic ends up differing between the two, I'll just reiterate my original comment that I'd like to see some comments explaining why the logic is what it is (and why it differs from the logic with a similar purpose in surfrdMod, if relevant).

slevisconsulting: The test suite flags the last change as a problem, producing FAILures such as:

I will return to just checking for cft_size > 2 and will modify the corresponding comment.

Knowing that the test-suite FAILs, is it still necessary to also make a unit test? If so, what do you have in mind?

slevisconsulting:
The last commit, returned the code to what I had tested earlier today and everything PASSed, except for TPUTCOMP and MEMCOMP FAILures.

src/biogeochem/CNPrecisionControlMod.F90
ekluzek: OK, heard from @olyson and we should back this change out. Since, it's my fault I can do that. It's won't affect testing, except it will change namelists, so technically it would be good to resend the tests so that the namelists get done again.

ekluzek:
@olyson I looked through my email about changing nnegcrit. I finally found out that @danicalombardozzi changed it this way to get the TRENDY simulations working. So I think on second thought we should let this go through as it is.

src/biogeochem/CNPrecisionControlMod.F90
ekluzek: OK, reversing what I just said, this was important for the TRENDY simulations, so we should leave this in.

src/dyn_subgrid/dyncropFileMod.F90
ekluzek: I guess what I was envisioning is making it a function call that you then test. And my main point is that complex logic always tends to be something that's easily incorrect. So doing testing on complex logic is a good thing. But, really what you want to make sure of is that it's being called when it should. One check for that, for example, is we want to make sure that crop simulations don't have a change in answers. So the system tests can verify that. Another approach to this is to call it more often than you should as long as it doesn't do the wrong thing. I think it's possible that collapse_crop_types call would in some cases do something important, but in others essentially amount to doing nothing, so it's OK to call it too much -- just not too little.

src/dyn_subgrid/dyncropFileMod.F90
billsacks: @slevisconsulting thank you for adding this comment - this is very helpful. I compared this with the comment at the top of `collapse_crop_types` and the code in that subroutine, and I think the comments and/or logic could still use some reconciliation: 

The comment at the top of `collapse_crop_types` says "Should only be called if using prognostic crops - otherwise, wt_cft is meaningless". The code inside seems to assume that the datasets have 78 pfts. I'm thinking that the comment at the top of that routine should be changed to, "Should only be called if (1) using create_crop_landunit (otherwise, wt_cft is meaningless), and (2) using datasets with the full set of npcropmax crops."

Then the logic that you have here (cft_size > 2) is effectively checking condition (2): I guess we're assuming that datasets always have cft_size <= 2 or cft_size = npcropmax, and never in between. You're not explicitly checking condition (1) (create_crop_landunit) here. I think that's safe at the moment, because we pretty much always use create_crop_landunit, at least by default (is this something a user can change? I'm not sure); the exception is when we're using FATES, but FATES doesn't work with transient right now, so that's currently safe. I'm a little worried about the fragility of this, though: if someone tries to do a transient run with create_crop_landunit false, what would currently happen? Should this be checked somewhere? At the very least, we should probably add to the comment here documenting that assumption (i.e., documenting that, if we get to this point, then we are assuming that create_crop_landunit is true).

src/dyn_subgrid/dyncropFileMod.F90
billsacks: @ekluzek I was starting to think along similar lines as you (I think) – specifically, thinking that we should just call collapse_crop_types all the time, and just let it be a no-op if you're already using a dataset CFTs that don't need collapsing (e.g., a 2-cft dataset). However, if I'm reading `collapse_crop_types` correctly, it looks hard-coded to assume that the inputs have CFTs that go up to npcropmax (hard-coded at 78). My intuitive feeling is that what should really be done here is to generalize `collapse_crop_types` so we can just call it always and have it Just Work. This routine is a perfect candidate for unit testing (maybe that's already done), so we can throw a variety of scenarios at it to make sure they're handled properly. But until that's done, I think we need a conditional around the call to collapse_crop_types.

slevisconsulting:
Makes sense to me that we should call collapse_crop_types more rather than less and - in fact - every time. Seems as though I should go ahead and do this now, though I'm not clear whether @billsacks you're suggesting wrapping up this PR and opening a separate issue that generalizes the call collapse_crop_types for all cases.

billsacks:
@slevisconsulting I'm open to either approach - either wrapping up this PR then generalizing collapse_crop_landunits as a follow-up, or doing it all together. I'm happy to defer to your and @ekluzek 's thoughts on what's best here.

ekluzek:
@slevisconsulting @billsacks and I just discussed this. And we agree with your finishing the improvement we've asked you to do here, before bringing to master. We also agreed that having collapse_crop_types called all the time, but added some unit tests to make sure it works under all conditions to be the right thing to do. So this all sounds good. Thanks for continuing to work on this to get it right.

slevisconsulting:
My cheyenne tests point to a baseline directory that I generated prior to the official dev016:

nohup qcmd -l walltime=06:00:00 -- ./create_test --xml-category aux_clm --xml-machine cheyenne --xml-compiler gnu -r /glade/scratch/slevis/collapse_pfts_cheyenne_gnu --project P93300641 --compare /glade/scratch/slevis/cesm_baselines/ctsm1.0.dev016 &   **PASS**

nohup qcmd -l walltime=12:00:00 -- ./create_test --xml-category aux_clm --xml-machine cheyenne --xml-compiler intel -r /glade/scratch/slevis/collapse_pfts_cheyenne_intel --project P93300641 --compare /glade/scratch/slevis/cesm_baselines/ctsm1.0.dev016 &   **PASS except**

nohup qcmd -- ./create_test <FAILing_test-names_here> --project P93300641 --compare /glade/scratch/slevis/cesm_baselines/ctsm1.0.dev016 --walltime 01:00 &

On second & third attempts, all three FAIL with the same indecipherable error:
ERP_D_Ld9.f09_g16.I1850Clm50SpCru.cheyenne_intel.clm-default.C.20181103_184724_byqx99   **[2nd and 3rd attempts]**
ERP_D_Ld5.f19_g17.I2000Clm50BgcCruGs.cheyenne_intel.clm-default   **[2nd attempt]**
ERP_D_Ld5.f19_g17_gl4.I1850Clm50BgcCrop.cheyenne_intel.clm-glcMEC_changeFlags   **[2nd attempt]**

These return an error that I do not know how to resolve, yet:
MPT: Launch error on r2i1n33.ib0.cheyenne.ucar.edu
MPT ERROR: could not run executable. If this is a non-MPT application,
you may need to set MPI_SHEPHERD=true.

        (HPE MPT 2.16  05/03/17 01:08:45)
/glade/u/apps/ch/opt/mpt/2.16/bin/mpiexec_mpt: line 399:  7815 Killed                  $mpicmdline_prefix -f $paramfile

src/dyn_subgrid/dyncropFileMod.F90
```diff
@@ -166,15 +166,13 @@ subroutine dyncrop_interp(bounds,crop_inst)
     allocate(fertcft_cur(bounds%begg:bounds%endg, cft_lb:cft_ub))
     call fertcft%get_current_data(fertcft_cur)
 
-    ! Call collapse_crop_types when we read 78-pft datasets (ie cft_size > 2).
-    ! For use_crop = .false. we collapse from 78 to 16 pfts.
-    ! For use_crop = .true. we collapse to the known list of crops and should
-    ! end up with more than 16 pfts.
-    ! The call collapse_crop_types appears in subroutine surfrd_veg_all twice:
-    ! once for use_crop = .false. and once later for use_crop = .true.
-    if (cft_size > 2) then
```
slevisconsulting: Removed if statement.

src/main/surfrdMod.F90
```diff
@@ -780,20 +780,14 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
 
     ! Check the file format for CFT's and handle accordingly
     if ( actual_numcft > 0 ) then
```
slevisconsulting: Collapsed if-elseif-elseif into a single if

src/main/surfrdMod.F90
```diff
@@ -780,20 +780,14 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
...
-       ! Full crop in file, but model is only with generic crop
-       else if ( create_crop_landunit .and. actual_numcft > 2 .and. .not. use_crop)then
+       if ( create_crop_landunit )then
+          ! Cases: Full crop in file and in model
+          !        Generic crop in file and in model
+          !        Full crop in file, generic crop in model
           cftsize = actual_numcft
           allocate(array2DCFT (begg:endg,cft_lb:cftsize-1+cft_lb))
           allocate(array2DFERT(begg:endg,cft_lb:cftsize-1+cft_lb))
           call surfrd_cftformat( ncid, begg, endg, array2DCFT, array2DFERT, cftsize, natpft_size )
```
slevisconsulting: Removed call collapse_crop_types from here, because the same call a few lines down now occurs for the combination of cases, leading to the same result.

src/main/surfrdMod.F90
```diff
@@ -832,18 +826,24 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
                ' must also have a separate crop landunit, and vice versa)'//&
                errMsg(sourcefile, __LINE__))
     end if
```
slevisconsulting: Rearranged a little here:
1) Moved wt_cft a few lines down without an if statement.
2) Moved both call check_sums a few lines down, one without an if and the other with if (.not. use_fates)
3) Kept call collapse_crop_types in the new if (.not. use_fates) 
which covers many more cases than the old if (use_crop)

olyson:
I'm getting some of these MPT launch errors today as well.
I think they had traced this previously to some bad switches of some type.
There is a function you can use to report these errors (if you think it is likely to be a machine error):

A new lightweight, easy-to-use command is available for reporting Cheyenne job failures. The reportfailure command is intended for reporting failures that users suspect were caused by system issues such as node problems.

To report a suspected failure, from your Cheyenne command prompt enter reportfailure followed by a one-line description that includes the job ID. Here are two examples:

reportfailure  Job 1234567  MPT Warning, timeout with communication to r3i7n21

reportfailure  MPT: shepherd terminated: r2i2n21.ib0.cheyenne.ucar.edu   JobID: 2705231

The command is not intended to be a substitute for opening ExtraView tickets and users will not receive a response. Rather, use of reportfailure will help CISL more quickly identify and respond to potential system issues. If you have questions about this, contact CISL at 303-497-2400 or cislhelp@ucar.edu.

src/main/surfrdUtilsMod.F90
```diff
@@ -149,12 +149,10 @@ subroutine collapse_crop_types(wt_cft, fert_cft, cftsize, begg, endg, verbose, s
     ! !DESCRIPTION:
     ! Collapse unused crop types into types used in this run.
     !
-    ! Should only be called if using prognostic crops - otherwise, wt_cft is meaningless
-    !
     ! !USES:
     use clm_varctl , only : irrigate, use_crop
```
slevisconsulting: Had to replace npcropmax (which = 78) with maxveg to make this work for all cases.

src/main/surfrdMod.F90
```diff
@@ -780,20 +780,14 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
...
+       if ( create_crop_landunit )then
+          ! Cases: Full crop in file and in model
+          !        Generic crop in file and in model
+          !        Full crop in file, generic crop in model
           cftsize = actual_numcft
           allocate(array2DCFT (begg:endg,cft_lb:cftsize-1+cft_lb))
           allocate(array2DFERT(begg:endg,cft_lb:cftsize-1+cft_lb))
           call surfrd_cftformat( ncid, begg, endg, array2DCFT, array2DFERT, cftsize, natpft_size )
-          call collapse_crop_types(array2DCFT(begg:endg, :), array2DFERT(begg:endg, :), cftsize, begg, endg, &
-                                   verbose=.true., sumto=100._r8)
```
slevisconsulting: Removed call collapse_crop_types from here, because the same call a few lines down now occurs for the combination of cases, leading to the same result.

src/main/surfrdMod.F90
```diff
@@ -780,20 +780,14 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
...
     ! Check the file format for CFT's and handle accordingly
     if ( actual_numcft > 0 ) then
-       ! Full crop on file and in CLM
-       if ( create_crop_landunit .and. actual_numcft > 2 .and. use_crop)then
-          call surfrd_cftformat( ncid, begg, endg, wt_cft, fert_cft, cft_size, natpft_size )  ! Format where CFT's is read in a seperate landunit
-       ! Generic crop in file and generic crop in model
-       else if ( create_crop_landunit .and. actual_numcft == 2 .and. cft_size == 2 )then
-          call surfrd_cftformat( ncid, begg, endg, wt_cft, fert_cft, cft_size, natpft_size )  ! Format where CFT's is read in a seperate landunit
-       ! Full crop in file, but model is only with generic crop
-       else if ( create_crop_landunit .and. actual_numcft > 2 .and. .not. use_crop)then
```
slevisconsulting: Collapsed if-elseif-elseif into a single if
because all three just needed to call surfrd_cftformat

slevisconsulting:
Thanks @olyson. I saw this posted in the Daily Bulletin. Can we do anything, other than report it, to make progress with our work?

olyson:
I just generally try running it again, a couple of times if necessary, to see if it will run on a different node

slevisconsulting:
Taking a different approach for now, because even my 3rd attempt with one of the three FAILing tests returned the same error...

For each of the FAILs I have found a similar test that PASSed:

1a) ERP_D_Ld9.f09_g16.I1850Clm50SpCru.cheyenne_intel.clm-default   FAIL
1b) ERI_D_Ld9.f09_g16.I1850Clm50Sp.cheyenne_intel.clm-default   PASSes

2a) ERP_D_Ld5.f19_g17.I2000Clm50BgcCruGs.cheyenne_intel.clm-default   FAIL
2b) SAME TEST WITH clm-fire_emis PASSes

3a) ERP_D_Ld5.f19_g17_gl4.I1850Clm50BgcCrop.cheyenne_intel.clm-glcMEC_changeFlags FAIL 
3b) SAME TEST WITH CropG PASSes

@ekluzek and @billsacks, is this reasonable or do I have to just keep resubmitting the tests until they run?

slevisconsulting:
@ekluzek and @billsacks, I pushed my first draft of a unit test for collapse_crop_types. 

I repeat here the notes from the commit (with slight corrections) because they include questions:

I used subroutine test_convert_cft_to_pft() as template for my work.
I do not expect the new unit test to work, yet.
I am committing & pushing so as to begin asking questions:

1) HOW DO I KNOW WHAT VALUES
subroutine collapse_crop_types
GETS FOR irrigate, maxveg, AND use_crop WHEN CALLED FROM THE UNIT TEST?

2) DO I NEED TO ADD THE NEW UNIT TEST TO A MASTER LIST WHEN I'M READY
TO RUN IT?

ekluzek:
Hey @slevisconsulting. Sorry for not seeing and responding sooner.

irrigate, maxveg and use_crop are all control variables that you can set just before the call to collapse_crop_types. So you'll know exactly what they are set to. You want to control all of the inputs to collapse_crop_types whether it's input arguments or module variables from somewhere else. So just explicitly use and set them in the unit test for it.

And no, if you add a new subroutine unit test with the @Test before it, PF unit will automatically add it to the list that it runs over. The only thing that's more tricky is adding a completely new unit tester with a top level driver for it.

slevisconsulting:
Thanks, @ekluzek. Let's keep the scheduled call on Monday's calendar for now, but we may not need to talk because your answers may be all I needed to make new headway.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -165,7 +165,6 @@ contains
      ! Now check that are correct
      call check_sums_equal_1( wt_cft/100.0_r8, begg, "test_check_sums_add_to_1", &
                               "should not trigger an error")
-     write(*,*) 'EDW!!', wt_cft, fert_cft
```
slevisconsulting: was just a diagnostic write statement...

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -112,6 +116,120 @@ contains
...
+     cft_lb = natpft_ub + 1
+     cft_ub = cft_lb + cftsize - 1
+
+     ! Collapse crop types
+     call collapse_crop_types( wt_cft, fert_cft, cftsize, begg, endg, verbose = .true., sumto = 100._r8 )
+
+     ! Now check that are correct
+     call check_sums_equal_1( wt_cft/100.0_r8, begg, "test_check_sums_add_to_1", &
+                              "should not trigger an error")
+     write(*,*) 'EDW!!', wt_cft, fert_cft
```
slevisconsulting: forgot a diagnostic write statement here; removed in the next commit/push...

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -112,6 +116,64 @@ contains
...
```
slevisconsulting: forgot diagnostic write statement; will remove in next commit/push...

src/main/surfrdUtilsMod.F90
```diff
@@ -181,10 +181,8 @@ subroutine collapse_crop_types(wt_cft, fert_cft, cftsize, begg, endg, verbose, s
     character(len=*), parameter :: subname = 'collapse_crop_types'
     !-----------------------------------------------------------------------
 
-    if (cftsize <= 0) then  ! Currently this applies only if use_fates
-       call endrun(msg = subname//' can only be called if cftsize > 0' // &
-            errMsg(sourcefile, __LINE__))
-    end if
+    if (cftsize > 0) then  ! The opposite applies only if use_fates
```
slevisconsulting: So if use_fates then do nothing instead of aborting.

src/main/surfrdMod.F90
```diff
@@ -835,7 +835,7 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
     ! Check sum of vegetation adds to 1
     call check_sums_equal_1(wt_nat_patch, begg, 'wt_nat_patch', subname)
     ! if ( use_fates ) wt_cft = 0 because called convert_cft_to_pft, else...
-    if ( .not. use_fates ) then
+    if ( .not. use_fates ) then  ! this if-statement may be redundant here
```
slevisconsulting: Testing removal of this if statement before committing.

src/main/surfrdMod.F90
```diff
@@ -835,17 +835,17 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns, actual_numcft)
...
-       ! The call collapse_crop_types also appears in subroutine dyncrop_interp
-       call collapse_crop_types(wt_cft(begg:endg, :), fert_cft(begg:endg, :), cft_size, begg, endg, verbose=.true.)
     end if
+    ! Call collapse_crop_types:
+    ! For use_crop = .false. collapsing 78->16 pfts or 16->16 or some new
+    !    configuration
+    ! For use_crop = .true. most likely collapsing 78 to the list of crops for
+    !    which the CLM includes parameterizations
+    ! The call collapse_crop_types also appears in subroutine dyncrop_interp
+    call collapse_crop_types(wt_cft(begg:endg, :), fert_cft(begg:endg, :), cft_size, begg, endg, verbose=.true.)
```
slevisconsulting: Moved call collapse_crop_types outside the if statement so now it's called for ALL cases.

src/main/surfrdUtilsMod.F90
```diff
@@ -183,78 +183,78 @@ subroutine collapse_crop_types(wt_cft, fert_cft, cftsize, begg, endg, verbose, s
 
     if (cftsize > 0) then  ! The opposite applies only if use_fates
```
slevisconsulting: All changes in this file are from indenting the lines within the outer if statement.

slevisconsulting:
Unit tests: PASS
Cheyenne gnu: PASS
Cheyenne intel: in progress

src/main/surfrdMod.F90
slevisconsulting: Gnu tests PASS
Intel tests PASS except some TPUTCOMP & MEMCOMP

slevisconsulting:
Running test suites after the merge.

Assuming I do not have to resolve failures, we're close to finishing, so @ekluzek please review my commits since you last looked.

slevisconsulting:
Cheyenne gnu PASS
Cheyenne intel PASS

doc/ChangeLog
```diff
@@ -1,4 +1,98 @@
...
+maxpatch_pft made obsolete
+
+Substantial timing or memory changes:
+Cheyenne gnu 7 TPUTCOMP FAILs
+Cheyenne intel 4 TPUTCOMP & 8 MEMCOMP FAILs
+Hobart gnu 1 TPUTCOMP FAIL
+Hobart nag 4 TPUTCOMP FAILs
+
+Notes of particular relevance for developers: (including Code reviews and testing)
+---------------------------------------------
```
ekluzek: Add a note that the name of the parameter: maxpatch_pft was changed to maxsoil_patchs and numpft was changed to maxveg and numpft_ed was changed to numveg_ed

doc/ChangeLog
```diff
@@ -1,4 +1,98 @@
...
+
+[ ] clm4_0
+
+Notes of particular relevance for users
+---------------------------------------
+
+Changes made to namelist defaults (e.g., changed parameter values):
+maxpatch_pft made obsolete
+
+Substantial timing or memory changes:
```
ekluzek: Instead of reporting on the TPUTCOMP and MEMCOMP fails, report here if we truly believe the code is slower or uses more memory than the baseline code. Do we really think that's true?

doc/ChangeLog
```diff
@@ -1,4 +1,98 @@
...
+[ ] clm5_0
+
+[ ] clm4_5
+
+[ ] clm4_0
+
+Notes of particular relevance for users
+---------------------------------------
+
+Changes made to namelist defaults (e.g., changed parameter values):
```
ekluzek: Note that nnegcrit was increased because of failure we saw in the TRENDY simulations. This was done in the source, rather than the namelist, but does have an effect on user behavior.

doc/ChangeLog
```diff
@@ -1,4 +1,98 @@
...
+maxpatch_pft made obsolete
+
+Substantial timing or memory changes:
+Cheyenne gnu 7 TPUTCOMP FAILs
+Cheyenne intel 4 TPUTCOMP & 8 MEMCOMP FAILs
+Hobart gnu 1 TPUTCOMP FAIL
+Hobart nag 4 TPUTCOMP FAILs
+
+Notes of particular relevance for developers: (including Code reviews and testing)
```
ekluzek: Also add a note that we should remove maxpatch_pft from the code and namelist as a cleanup step. Currently it's still there, but should be removed.

doc/ChangeLog
ekluzek: And note that several modules were made default private, and the list of variables in them that are public are explicitly listed now. It would also be good to note which modules that was done for.

src/main/surfrdMod.F90
```diff
@@ -753,20 +826,26 @@ subroutine surfrd_veg_all(begg, endg, ncid, ns)
...
-    ! Convert from percent to fraction, check sums of nat vegetation add to 1
-    if ( cft_size > 0 )then
-       wt_cft(begg:endg,:) = wt_cft(begg:endg,:) / 100._r8
-       call check_sums_equal_1(wt_cft, begg, 'wt_cft', subname)
-    end if
+    ! Convert from percent to fraction
     wt_lunit(begg:endg,istsoil) = wt_lunit(begg:endg,istsoil) / 100._r8
     wt_lunit(begg:endg,istcrop) = wt_lunit(begg:endg,istcrop) / 100._r8
     wt_nat_patch(begg:endg,:)   = wt_nat_patch(begg:endg,:) / 100._r8
-    call check_sums_equal_1(wt_nat_patch, begg, 'wt_nat_patch', subname)
```
ekluzek: Why was this changed from a subroutine call to inline code?

src/main/surfrdMod.F90
ekluzek: Sounds good.

src/main/surfrdMod.F90
ekluzek: Good

src/main/surfrdUtilsMod.F90
ekluzek: OK

src/main/surfrdMod.F90
ekluzek: OK

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -46,7 +46,6 @@ contains
      call collapse_crop_types( wt_cft, fert_cft, cftsize, begg, endg, verbose = .true.)
 
      ! Now check that are correct
-     write(*,*) 'EDW!', wt_cft, fert_cft
```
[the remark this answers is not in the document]
ekluzek: Good.

src/main/surfrdUtilsMod.F90
ekluzek: Good

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -13,15 +13,402 @@ module test_surfrdUtils
 
 contains
   
+  @Test
+  subroutine test_collapse_crop_types_none()
```
ekluzek: I'm glad you have all these different tests here. They seem to be comprehensive and extensive. There are two things I wonder about. The first thing I wonder about is that it looks like there is quite a bit of setup that's similar if not identical between them all. So you might be able to simplify it by having a shared setup subroutine. The other thing to think about is that the checks at the end are all hardcoded to match the setup at the beginning, meaning that it's not totally obvious that it matches. A goal of unit-testing is to try to make it obvious that it's doing the right thing. If it could be made more obvious that would be helpful. One way to do that would be to have the "expect" array settings based on the input settings. So then it's clear what part is supposed to be the same, and what part is different because it's should be.

ekluzek:
@slevisconsulting I added a bunch of comments and questions. Since, you've done a lot of testing we don't necessarily need to do all of them. But, some could be done without redoing all of the testing. I also had some suggestions for the ChangeLog, and of course that can be done without redoing testing.

I also think you should change the name of this PR to just describe the change that's coming in, rather than a statement about it's history (which is already documented in the PR).

I also resolved a bunch of the comments, just so we don't think they are issues that need to be looked at.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
ekluzek: I see this one was resolved.

doc/ChangeLog
slevisconsulting: DONE.

doc/ChangeLog
slevisconsulting: DONE.

doc/ChangeLog
slevisconsulting: DONE.

src/main/surfrdMod.F90
slevisconsulting: It's still a subroutine call, but I moved it a few lines down.

ekluzek: Ahhh, good!

doc/ChangeLog
slevisconsulting: @ekluzek, instead of giving a "handwavy" response, I should look at timing files, right? I'm not experienced looking at those, but I will give it a shot.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
slevisconsulting: This comment will take a little thought to address, so I will handle it after I commit the simpler revisions.

slevisconsulting:
Thank you, @ekluzek! I will commit/push the easy stuff first. Then I will work on the last couple of comments still remaining.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
slevisconsulting: Now added "expect" array settings as recommended.

slevisconsulting: And next added subroutine setup_test_collapse_crop_types as recommended. (Unit tests continue to PASS.)

src/dyn_subgrid/dyncropFileMod.F90
ekluzek: @slevisconsulting it looks like this issue still needs to be addressed. I think we decided that we should remove the "if ( cft_size > 2 )" around the call to "collapse_crop_types" right? So it looks like that should be done. That possibly means redoing (or rerunning) the non crop tests.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
ekluzek: Excellent. This looks much better and so much easier to read! The setup subroutine moves off the busy work, so you just see the core of what's happening.

I have one more suggestion if you are willing to hang on for just a little more...

You set the input arrays element by element and then do the same for the expected arrays. I think it would be easier to see if you did this using array syntax, and also set the expected arrays based on the input arrays (since typically there is some simple change that is going on). For example, in the 20_to_18 test you have...

``` fortran
     ! Input values
     wt_cft(begg:,1) = 5.0_r8
     wt_cft(begg:,2) = 10.0_r8
     wt_cft(begg:,3) = 20.0_r8
     wt_cft(begg:,4) = 30.0_r8
     wt_cft(begg:,5) = 15.0_r8
     wt_cft(begg:,6) = 20.0_r8
     call check_sums_equal_1( (wt_cft/100.0_r8), begg, "test_check_sums_add_to_1", &
                              "should not trigger an error")
     fert_cft(begg:,1) = 5.0_r8
     fert_cft(begg:,2) = 10.0_r8
     fert_cft(begg:,3) = 20.0_r8
     fert_cft(begg:,4) = 30.0_r8
     fert_cft(begg:,5) = 15.0_r8
     fert_cft(begg:,6) = 20.0_r8
      ! Output values expected
     wt_cft_expected(begg:,1) = 5.0_r8
     wt_cft_expected(begg:,2) = 10.0_r8
     wt_cft_expected(begg:,3) = 0.0_r8
     wt_cft_expected(begg:,4) = 0.0_r8
     wt_cft_expected(begg:,5) = 35.0_r8
     wt_cft_expected(begg:,6) = 50.0_r8
```

Here's what I suggest you change it to...

``` fortran
     ! Input values
     wt_cft(begg:,:) = (/ 5.0_r8, 10.0_r8, 20.0_r8, 30.0_r8, 15.0_r8, 20.0_r8 /)
     call check_sums_equal_1( (wt_cft/100.0_r8), begg, "test_check_sums_add_to_1", &
                              "should not trigger an error")
     fert_cft(begg:,:) = wt_cft(begg:,:)

      ! Output values expected, same as before with 3 and 4 zeroed, but added into 5 and 6
     wt_cft_expected(begg:,:) = wt_cft(begg:,:)
     wt_cft_expected(begg:,3:4) = 0.0_r8
     wt_cft_expected(begg:,5) = wt_cft(begg:3) + wt_cft(begg:5)
     wt_cft_expected(begg:,6) = wt_cft(begg:4) + wt_cft(begg:6)
```

When I read the unit test, to see what's going on I have to scan through all the values, just to realize they are duplicated from what was before. Since, most of them are duplicated it helps to just see which ones change. Technically, that would also allow you to change the values in the original setting of wt_cft and the unit test would still work. But, the main thing I'm thinking about is readability.

A goal of a unit test is to make it as readable as possible, so you don't have to think about it much to parse what it's doing. If you can just scan it and see that it's correct that's ideal, but often hard to achieve. But, working toward that is desirable. Looking at the above the second one is much shorter and I would say easier to read and see what it's doing. Hopefully, that makes sense.

ekluzek:
Sam I committed a few changes to the ChangeLog, just seemed easier than trying to explain it. The bottom line is that it's not useful to have detailed notes on TPUTCOMP and MEMCOMP fails, since they aren't reliable or consistent for all tests. So a simpler answer is better here.

slevisconsulting:
Ok, thank you for doing that, Erik.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
slevisconsulting: Will do. I will commit and push by the end of today.

slevisconsulting: Done.

src/dyn_subgrid/dyncropFileMod.F90
ekluzek: Ahh, OK. The diff at the top of this conversation is labeled as "Outdated". I thought I looked at the latest code and saw the same thing, but obviously not. So awesome I'm glad this is done. Thanks for clarifying.

ekluzek:
OK, this looks good to go. @slevisconsulting I'm not sure you can do the next steps, because of permission. Do you see the "merge pull request" button above? If not I think everything is done, so I can do the next two steps which are to merge to master, and then create an annotated tag and push that to escomp. I'll also update the time-stamp in the Change files just before doing the merge and tag.

billsacks:
@ekluzek - @slevisconsulting doesn't have write permission, so you should do this final step (see https://github.com/orgs/ESCOMP/teams/ctsm-write/members)

Issue #357, opened 2018-04-30
Put in place water isotope consistency checks


billsacks:
An early step to support implementation of water isotopes should be putting in place consistency checks that confirm that isotopic quantities are being updated properly as they pass through the system. I believe that Ben Andre put something like this in place on the old water isotope branch.

billsacks:
@mathewvrothstein I'm tentatively assigning this to you, thinking you might have a better sense of what needs to be done here, given your involvement with Ben earlier in this project. But I'm happy to help with it if you know as little as I do about it :-)

davidnoone:
Hi Bill, 

1) The scattering of consistency checks will be super valuable for code development (and maintenance), but would not need to be in for production runs. In essence, you're checking that the tracer mass balance meets the same mass balance as the normal/bulk water. Doing this check after initialization and after each hydrology routine allows you to isolate any possible issue.

However, once the code development debugging is finished, you can be more sparing about where these are placed. One call at the end of the initialization would be fine. I'd keep one call after the canopy, soil, urban and lake routines too as a check. Finally one at the end of the step.

On entry to CLM from the driver, I suggest checking the inputs (precip and vapor), and on exit check the fluxes passed back to the coupler (qflx, runoff to RTM, etc). This way we can be sure that CTSM is communicating consistently. 



2)  "Is it important that we maintain the checks of all tracers if wiso_clm_unit_fractionation is true?"
Yes, as a unit check. The key goal here is to ensure the tracer scheme conserves mass when the full isotopic physics is used. During development this is used to test the mass balance is working with the isotope code, and for longer maintenance it allows future developers to check that they have accounted for the tracers in any new code changes.  With fractional set to 1, there should be no difference between the tracer water and the bulk water (to the desired numerical precision). 

Your point that they would all be the same is a good one, and correct. And this is the criteria for knowing the check has passed. There would only be a difference if there had been a code error.

I expect that wiso_clm_unit_fractionation only triggers changes in the tracer registration at the point where one configures what type of water tracer is in each array position. With this unit test flagged, the fractionation factors (and similar physical properties) would be set to 1. That should be it.

billsacks:
@davidnoone thanks a lot for this. Your reply on (1) makes sense. My understanding from your reply is that having multiple checks throughout the driver loop does not catch any more problems, but helps people pinpoint where, exactly a problem occurred. Long-term, we'll try to strike a balance between having this extra pinpointing without cluttering the code too much.

[...]

PR #497, opened 2018-09-05
Add water tracer consistency checks, and other water tracer work

### LINKED ISSUE #479: More water states and fluxes that need to be replicated for tracers
Looking more closely at the older water isotope branch (branch wiso-clm5-r249 on https://github.com/ncar/water-isotopes), I realized that we're still missing a water state and a number of fluxes that need to be replicated for tracers:

- wa_col, currently defined in SoilHydrologyType.F90. This should be a straightforward matter of moving this to WaterStateType.

- Many water fluxes currently defined in atm2lndType and lnd2atmType. I'm not sure what to do here: The easiest solution would probably be to move all water fluxes out of atm2lndType and lnd2atmType and into WaterFluxType. But that means that these atm2lnd and lnd2atm variables would no longer all be together; I've particularly liked having the atm2lnd variables together because there's some complexity with respect to downscaling, and it's nice to have that all in one place. Still, my current feeling is that it's probably best to move these to WaterFluxType, but I'd like to give this more thought and here some thoughts of others. @mathewvrothstein and @mvertens - do you have any thoughts on this?

### LINKED ISSUE #492: Make a system test that enables the tracer consistency checks
We want an out-of-the-box system test (via a testmod, and included in the aux_clm test suite) that turns on the new tracer consistency checks (which are not yet on master).

To do this, we'll need a namelist item that controls whether these checks are done. If true, the appropriate tracer(s) would be added that are used in these consistency checks.

Note that, for now, this test needs to use cold start, because we have not yet put in place the code to initialize water tracers given a restart file that did not have water tracers on it.

billsacks:
### Description of changes

1. Add water tracer consistency checks (@mathewvrothstein )

2. Breakout of atm2lnd and lnd2atm water variables, needed for water tracers (@mathewvrothstein )

3. Add some namelist control over the addition of water tracers (@billsacks)

4. Add a system test that exercises the water tracer consistency checks (@billsacks)

### Specific notes

Contributors other than yourself, if any: @mathewvrothstein 

- Partially addresses #357 
- Fixes #479 
- Fixes #492 

- Adds two new namelist variables: enable_water_tracer_consistency_checks and enable_water_isotopes

Testing performed, if any:
- Ran much of the aux_clm test suite; will run the rest soon


billsacks:
@ekluzek - @mathewvrothstein and I have worked together on this, so I consider this to be sufficiently reviewed, but I'm opening this PR in case you or others want to look at it.

@mathewvrothstein feel free to look at my two most recent commits and comment on them if you'd like.

billsacks:
@davidnoone - I thought you'd be interested to know: With some infrastructure I have just added as part of this PR, we automatically ensure that the tracer consistency checks are run on every variable that is added to any of the water types that have instances for each tracer variable. This was driven by my worry that people would add new variables but not realize they needed to call the tracer consistency check on those new variables, and this omission might be hard to notice in a code review.

If you're curious: The infrastructure that enables this is in WaterTracerContainerType.F90; WaterTracerUtils.F90: AllocateVar1d and AllocateVar2d; and WaterType.F90: TracerConsistencyCheck. You can see how it's used in, for example, WaterStateType.F90: InitAllocate. As long as new variables are added with this AllocateVar infrastructure rather than a simple allocate call (which should be relatively easy to spot in a code review, since the new lines would stand out relative to surrounding code), they will automatically have the tracer consistency check run on them.

src/biogeophys/test/WaterTracerUtils_test/test_compare_bulk_to_tracer.pf
```diff
@@ -0,0 +1,127 @@
+module test_compare_bulk_to_tracer
```
billsacks: @mathewvrothstein - I added these unit tests largely for the sake of ensuring that nobody breaks the CompareBulkToTracer routine when it is modified in the future. You may want to take a quick look at these unit tests as a fairly simple example of pfunit-based unit tests, and also because you'll likely need to modify them a bit when you add the ratio argument.

PR #523, opened 2018-09-27
Groundwater irrigation

### LINKED ISSUE #81: Restart files are different for CLM when run over different number of tasks
**Erik Kluzek < erik > - 2011-03-31 10:24:06 -0600**
**Bugzilla Id:** [1310](http://bugz.cgd.ucar.edu/show_bug.cgi?id=1310)
**Bugzilla CC:**  dlawren,  jedwards,  jshollen,  mvertens,  oleson,  rfisher,  sacks, 

This is from Bill Sacks...

With Mariana's help, I believe I have uncovered a minor bug in the urban model; this came up while testing the new CLM multi-instance code that I have been working on. This appears as a difference in the lnd restart files depending on the number of processors. It's possible that this is unimportant, but I thought I'd let you know anyway.

In particular, the two variables albgrd and albgri differ in some urban landunits in the CLM restart files. I have confirmed this with the latest clm tag (clm4_0_26), doing a 5-day run with resolution f19_g16, and comparing results using 64 vs. 16 tasks for the land model. You can see the output of cprnc in /ptmp/sacks/archive/clm4_0_26.init.quarterPEs/rest/0001-01-06-00000/cprnc.out

I believe that what is going on is the following: 

(1) In UrbanMod.F90: UrbanAlbedo: A count is made of the number of urban landunits with coszen > 0 (num_solar); note that this count is just of the number of landunits that this processor is responsible for; thus, this is where the # PE-dependence comes in, I think.

(2) Later in that subroutine, a bunch of calculations are done if num_solar > 0 -- i.e., if this PE is responsible for at least one urban landunit with coszen > 0. Note that many of these calculations are done for all landunits, even ones for which coszen = 0. This introduces the possibility for different results depending on the decomposition.

(3) The particular difference that I am seeing is in albgrd & albgri. These are initialized to 0 at the start of the subroutine, and so remain 0 on any PE for which num_solar = 0. However, for PEs with num_solar > 0, landunits that have coszen = 0 end up getting albgrd = albgri = 1. This is because the calculation of albgrd & albgri depends on the values of the sref_* variables, which are initialized to 1 (and stay at 1 for any landunit for which coszen = 0).

swensosc:
### Description of changes
Add irrigation from groundwater sources (unconfined [from the soil column] and confined [from wa] aquifers).  Add irrigation application method (sprinkler [above canopy] and drip [below canopy]).  Both of these changes are activated via namelist.  Add option to set saturated fraction to zero for crop columns.
### Specific notes
Irrigation method also requires input data on surface data file, otherwise will default to previous behavior (drip).  

Fixes #81 

maybe for changes related to irrigation method due to new variables used to 
apply irrigation in CanopyHydrologyMod

two new namelist variables: use_groundwater_irrigation and crop_fsat_equals_zero
the latter name might not be great and could be changed

Testing performed, if any:

performed 1 year simulation with new options activated


src/biogeophys/IrrigationMod.F90
```diff
@@ -762,27 +819,69 @@ subroutine ApplyIrrigation(this, bounds, waterfluxbulk_inst)
...
+         qflx_gw_uncon_irrig_col   => waterfluxbulk_inst%qflx_gw_uncon_irrig_col   , & ! Output: [real(r8) (:)] col unconfined groundwater irrigation flux (mm H2O/s)
+         qflx_gw_con_irrig_patch   => waterfluxbulk_inst%qflx_gw_con_irrig_patch   , & ! Output: [real(r8) (:)] patch confined groundwater irrigation flux (mm H2O/s)
+         qflx_gw_con_irrig_col     => waterfluxbulk_inst%qflx_gw_con_irrig_col     , & ! Output: [real(r8) (:)] col confined groundwater irrigation flux (mm H2O/s)
+         qflx_irrig_drip_patch     => waterfluxbulk_inst%qflx_irrig_drip_patch     , & ! Output: [real(r8) (:)] patch drip irrigation flux (mm H2O/s)
+         qflx_irrig_drip_col       => waterfluxbulk_inst%qflx_irrig_drip_col       , & ! Output: [real(r8) (:)] col drip irrigation flux (mm H2O/s)
+         qflx_irrig_sprinkler_patch=> waterfluxbulk_inst%qflx_irrig_sprinkler_patch, & ! Output: [real(r8) (:)] patch sprinkler irrigation flux (mm H2O/s)
+         qflx_irrig_sprinkler_col  => waterfluxbulk_inst%qflx_irrig_sprinkler_col    & ! Output: [real(r8) (:)] col sprinkler irrigation flux (mm H2O/s)
          )
 
     do p = bounds%begp, bounds%endp
```
billsacks: Now that this loop is doing more work, I'd like to see it changed to use a filter for performance reasons. It seems like soilp would be an appropriate filter here (though it's possible that there's a more appropriate one). Similarly, the p2c calls could be changed to use their filter version (in which case you should use a column filter that is analogous to the patch filter used in this loop - e.g., soilc together with soilp).

It's possible that, if you make this change, you might also need to initialize some / all of these fluxes to 0 in initCold.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -78,15 +78,6 @@ subroutine BeginWaterBalance(bounds, &
          begwb        =>    waterbalancebulk_inst%begwb_col        & ! Output: [real(r8) (:)   ]  water mass begining of the time step    
          )
 
-   do fc = 1, num_nolakec
-       c = filter_nolakec(fc)
-       if (col%hydrologically_active(c)) then
-          if(zwt(c) <= zi(c,nlevsoi)) then
-             wa(c) = aquifer_water_baseline
```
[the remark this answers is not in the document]
billsacks: Also, can you comment on whether the references to `aquifer_water_baseline` in TotalWaterAndHeatMod.F90 should be changed?

billsacks:
@swensosc Regarding our in-person conversation about why the volr-based limitation is in calcIrrigationNeeded yet I'm suggesting moving the groundwater-related logic into ApplyIrrigation: I found this comment in the ChangeLog entry for `clm4_5_13_r208`:

> In addition, this reworks the river volume-based limitation. One change to the volr-based limitation is that we now apply the limit to the deficit computed in calcIrrigationNeeded - rather than applying the limit in applyIrrigation. The point of this is to avoid problems that arise due to evolving VOLR - e.g., if we had calculated irrigation demand such that VOLR was just barely big enough, then we apply irrigation for some time steps, then we get an updated VOLR which is lower due to these irrigation withdrawals - the old implementation would impose an unnecessary limit on the irrigation to be applied for the rest of this day. It's hard to see a clean way around this problem, given that VOLR is updated in some but not all CLM time steps. So we're going to impose the limit on irrigation deficit, rather than on the time step-by-time-step flux. The downside here is that it's somewhat more likely that irrigation would try to draw volr negative, if volr has been evolving for other reasons.

The key difference is that, for volr, we don't have an easy way to determine if this is a time step when we have an updated volr value, or if we still have volr from a few time steps ago (since ROF isn't coupled in every time step). So in that case, having the possibility that we'd exceed available volr seemed like the lesser of evils.

In the case at hand, however, we can rely on having an up-to-date view of available groundwater, so the argument against having the volr limitation in ApplyIrrigation doesn't apply, and that seems like a more robust place to do this limitation. That said, I also see the argument that it would be nice to have both volr-based and groundwater-based irrigation in the same place, if we're willing to live with one of them being in a less-than-ideal place.

I'm happy to discuss this more if you want.

src/main/surfrdMod.F90
```diff
@@ -688,6 +701,16 @@ subroutine surfrd_pftformat( begg, endg, ncid )
...
 
+    call ncd_io(ncid=ncid, varname='irrigation_method', flag='read', data=irrig_method, &
+            dim1name=grlnd, readvar=readvar)
+    if (readvar) then
+       call endrun( msg= ' ERROR: unexpectedly found irrigation_method on dataset when cft_size=0'// &
+               ' (if the surface dataset has a separate crop landunit, then the code'// &
+               ' must also have a separate crop landunit, and vice versa)'//&
+               errMsg(sourcefile, __LINE__))
+    end if
+    irrig_method = 0
```
billsacks: One more place where you should use irrig_method_unset

src/biogeophys/CanopyHydrologyMod.F90
```diff
@@ -308,12 +310,14 @@ subroutine CanopyHydrology(bounds, &
...
-                if (frac_veg_nosno(p) == 1 .and. (forc_rain(c) + forc_snow(c)) > 0._r8) then
-
+                   ! total liquid water inputs above canopy
+                   qflx_liq_above_canopy(p) = forc_rain(c)+ qflx_irrig_sprinkler(p)
+                   
                    ! determine fraction of input precipitation that is snow and rain
-                   fracsnow(p) = forc_snow(c)/(forc_snow(c) + forc_rain(c))
-                   fracrain(p) = forc_rain(c)/(forc_snow(c) + forc_rain(c))
+                   fracsnow(p) = forc_snow(c)/(forc_snow(c) + qflx_liq_above_canopy(p))
+                   fracrain(p) = forc_rain(c)/(forc_snow(c) + qflx_liq_above_canopy(p))
```
billsacks: Is it correct to have forc_rain in the numerator here, or should the numerator be qflx_liq_above_canopy? As written, fracsnow + fracrain may be less than 1. I'm not sure what's right here, but just wanted to double-check.

src/biogeophys/CanopyHydrologyMod.F90
```diff
@@ -339,21 +343,20 @@ subroutine CanopyHydrology(bounds, &
...
                    end if
-
                    ! Direct throughfall
-                   qflx_through_rain(p) = forc_rain(c) * (1._r8-fpi)
+                   qflx_through_rain(p) = qflx_liq_above_canopy(p) * (1._r8-fpi)
 
                    if (snowveg_on .or. snowveg_onrad) then
                       ! Intercepted precipitation [mm/s]
-                      qflx_prec_intr(p) = forc_snow(c)*fpisnow + forc_rain(c)*fpi
+                      qflx_prec_intr(p) = forc_snow(c)*fpisnow + (forc_rain(c) + qflx_irrig_sprinkler(p))*fpi
```
billsacks: The term in parentheses could be replaced with qflx_liq_above_canopy.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -272,7 +283,10 @@ subroutine BalanceCheck( bounds, &
                   - (forc_rain_col(c)        &
                   + forc_snow_col(c)         &
                   + qflx_floodc(c)           &
-                  + qflx_irrig(c)            &
+                  + qflx_irrig_drip(c)       &
```
billsacks: One more comment: Looking back through the changes in BalanceCheckMod, I realized that these changes are a little incongruous with the surrounding code: For the most part, the only water fluxes we consider here are ones that come from outside CTSM (rain, snow, flood) or leave CTSM (runoff, evaporation, etc.). But some of these irrigation fluxes can just be due to rearrangements of water within the system (extracting from groundwater and adding to the top of the soil, for example).

I think you should get rid of all of these qflx_irrig terms and instead just add qflx_sfc_irrig. I think this should amount to the same thing, since I think qflx_sfc_irrig = (qflx_irrig_drip + qflx_irrig_sprinkler - qflx_gw_uncon_irrig - qflx_gw_con_irrig). However, this is more congruous with the surrounding code. And, more importantly, this will do a better job at catching balance errors. (For example, I think that, as currently written, you could put something like 'qflx_irrig_drip = qflx_irrig_drip + 1' at the end of ApplyIrrigation, and this balance check as written would still pass, when it shouldn't.)

billsacks:
@swensosc I have merged this up to the latest master, and tested the build (but not a run).

src/biogeophys/SoilHydrologyMod.F90
```diff
@@ -2490,6 +2495,176 @@ subroutine RenewCondensation(bounds, num_hydrologyc, filter_hydrologyc, &
...
    end subroutine RenewCondensation
+!#8
+   !-----------------------------------------------------------------------
+   subroutine CalcAvailableUnconfinedAquifer(bounds, num_hydrologyc, filter_hydrologyc, &
+        soilhydrology_inst, soilstate_inst, waterdiagnosticbulk_inst) 
+     !
+     ! !DESCRIPTION:
+     ! Calculate water in unconfined aquifer (i.e. soil column) that
+     ! is available for groundwater recharge.
+     !
```
billsacks: Thanks a lot @swensosc - this looks good.

Can you please add a comment saying which options this is / is not compatible with? I know there are checks of this elsewhere in the code, but I think it would help to have this noted in the top-level documentation of this function, too.

Specifically, I'm thinking (but would like your confirmation):

- The removal of irrigation from the confined aquifer assumes use_aquifer_layer is false.

- The removal of irrigation from the unconfined aquifer follows the same pattern used in LateralFlowPowerLaw. Does this implicitly assume that we're using a particular formulation for the aquifer layer and/or other parameterizations? If so, it would be good to note that here.

(Maintaining some semblance of sanity with a multi-parameterization model is hard!)

src/main/clm_driver.F90
```diff
@@ -423,11 +425,25 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
...
        ! input is main channel storage
-       call irrigation_inst%ApplyIrrigation(bounds_clump, water_inst%waterfluxbulk_inst)
+       call irrigation_inst%ApplyIrrigation(bounds_clump, filter(nc)%num_soilc, &
+            filter(nc)%soilc, filter(nc)%num_soilp, filter(nc)%soilp, &
+            water_inst%waterfluxbulk_inst, &
+            available_gw_uncon = water_inst%waterdiagnosticbulk_inst%available_gw_uncon_col(bounds_clump%begc:bounds_clump%endc))
        call t_stopf('drvinit')
 
+       ! Remove groundwater irrigation
+       call WithdrawGroundwaterIrrigation(bounds_clump, filter(nc)%num_hydrologyc, &
```
billsacks: Were you going to wrap this in a conditional on whether we're using groundwater irrigation? I don't care strongly if that's done, though I did like your suggestion of doing it. If so, would it be correct to also wrap CalcAvailableUnconfinedAquifer in a similar conditional?

Since the irrigation params are private, we'd need to put in a simple logical function in IrrigationMod that returns the value of this%params%use_groundwater_irrigation:

```Fortran
  function UseGroundwaterIrrigation(this)
    !
    ! !DESCRIPTION:
    ! Returns true if we're using groundwater irrigation in this urn
    !
    ! !ARGUMENTS:
    logical :: UseGroundwaterIrrigation  ! function result
    class(irrigation_type), intent(in) :: this

    UseGroundwaterIrrigation = this%params%use_groundwater_irrigation

  end function UseGroundwaterIrrigation
```

billsacks: @swensosc wrapped the call to WithdrawGroundwaterIrrigation in a conditional. He said he could go either way on the call to CalcAvailableUnconfinedAquifer... I might add a conditional around that in a final cleanup.

bld/CLMBuildNamelist.pm
```diff
@@ -2779,7 +2780,7 @@ sub setup_logic_irrigation_parameters {
      my $var;
      foreach $var ("irrig_min_lai", "irrig_start_time", "irrig_length",
                    "irrig_target_smp", "irrig_depth", "irrig_threshold_fraction",
-                   "limit_irrigation_if_rof_enabled") {
+                   "limit_irrigation_if_rof_enabled","use_groundwater_irrigation") {
         add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var);
      }
```
ekluzek: You should add a check that use_groundwater_irrigation and lower_boundary_condition==3|4 aren't both set. The FORTRAN code checks this so it should be checked sooner in build-namelist as well. It's good to have it check for a valid namelist as soon as possible.

setup_logic_hydrology_params is already called before setup_logic_irrigation_parameters.

You just add something like this:

``` perl
my $lower = $nl->get_value( 'lower_boundary_condition'  );
if ( ($lower == 3 || $lower == 4)) & (&value_is_true($nl->get_value( 'use_groundwater_irrigation'  )) ) {
 $log->fatal_error("use_groundwater_irrigation can only be used when lower_boundary_condition is NOT 3 or 4");
}
```

src/main/surfrdMod.F90
```diff
@@ -613,6 +614,18 @@ subroutine surfrd_cftformat( ncid, begg, endg, wt_cft, cftsize, natpft_size )
        fert_cft = 0.0_r8
     end if
 
+    if ( cft_size > 0 )then
+       call ncd_io(ncid=ncid, varname='irrigation_method', flag='read', data=irrig_method, &
```
ekluzek: I actually think that irrigation_method shouldn't be on the surface dataset, but a separate stream that can be handled independently, like the pop-density streams and ndep. If you do it that way, you don't have to add more complexity to mksurfdata_map, and you have better control over how it's interpolated both spatially and in time. And you keep it as a separate file that doesn't require you to make all new surface datasets, which is a task we try NOT to do. I assume this ultimately will be something that changes both in space and time correct? Reading it as a stream would both be better and easier now as well as in the future when you want it to vary in time.

src/main/histFileMod.F90
```diff
@@ -3058,10 +3059,8 @@ subroutine hfields_1dinfo(t, mode)
           call ncd_defvar(varname='land1d_jxy', xtype=ncd_int, dim1name=namel, &
                long_name='2d latitude index of corresponding landunit', ncid=ncid)
 
-          ! --- EBK Do NOT write out indices that are incorrect 4/1/2011 --- Bug 1310
-          !call ncd_defvar(varname='land1d_gi', xtype=ncd_int, dim1name='landunit', &
-          !     long_name='1d grid index of corresponding landunit', ncid=ncid)
-          ! ----------------------------------------------------------------
+          call ncd_defvar(varname='land1d_gi', xtype=ncd_int, dim1name='landunit', &
```
ekluzek: This is missing the bit of code needed from the bottom of #523. There's a do loop that loops over called to GetGlobalIndex that needs to be called. The GetGlobalIndex was added as something that can be used, but the calls to it seem to be missing.

src/main/surfrdMod.F90
billsacks: I'd ask that we defer this to a later conversation. We need to get these changes onto master in order to move forward with the isotope implementation, and I don't want to delay them for this to be rewritten to use streams (particularly because, @ekluzek I think you're the only one who is familiar enough with streams to be able to do this pretty quickly). For now we know that more work is needed to get this spatially-varying bit to work out-of-the-box (see #565 ): what's here now is just enough for further testing / development.

ekluzek: I think it's OK to defer this for later. This bit right now doesn't actually add this to mksurfdata nor create new surface datasets. But, rather than do that I think we should move it to a streams file. If I'm the only familiar with it, more people should get spunup on it. People do the solutions that they are most familiar with, even if there are better options available.

billsacks: @ekluzek I'd suggest that, at a Thursday CTSM-Software meeting, you facilitate a discussion on streams vs. surface dataset / landuse.timeseries for things like this. It feels like we need a (written-down) policy about where we want things to go moving forward which accounts for various factors at play - and I'd like to hear from you as well as scientists about the arguments for and against the different options.

bld/CLMBuildNamelist.pm
billsacks: Okay, I added this and tested by adding use_groundwater_irrigation to user_nl_clm for both a clm50 case (works) and clm45 case (dies with the error).

src/main/histFileMod.F90
ekluzek: Yes, I marked the wrong place. @swensosc and I talked about this a bit as well. He had it correct in at least one version he had, but it was missing from the final branch. The code he showed me had it correct so he's going to add it to the right place.

ekluzek: This looks correct now.

billsacks:
Thanks, @ekluzek . Can you please comment on whether this should mark #81 as being fixed, or if it just addresses part of that issue?

billsacks:
@swensosc - I'm getting differences in `cols1d_gi`, `pfts1d_gi` and `pfts1d_li` when I change processor / thread count. At a glance, it looks like this may be due to the fact that you use `clmlevel=namel` for the GetGlobalIndex call for `cols1d_gi` (should that be `clmlevel=nameg`?), and `clmlevel=namec` for the calls for `pfts1d_gi` and `pfts1d_li` (should those be `clmlevel=nameg` and `clmlevel=namel`?).

I haven't taken the time to understand this fully; can you confirm that I should make the changes in parentheses above or some different changes to fix this?

swensosc:
I think you're correct.  I made the changes and one other place where
'namel' should have been used.

billsacks:
There are baseline differences in select fields in a number of tests. Here is an email I just sent to @swensosc about this:

I have been looking more into the diffs in WA that we were discussing yesterday. Here are a few findings:

(1) All of the differences are indeed in tests with cold start initial conditions, except for one test that shows diffs that is a CLM50 test that points to CLM45 initial conditions.

(2) Spot-checking one of these tests: On master, WA was exactly 5000 everywhere at the end of the run. On the branch, WA was roughly 4000 everywhere (in some grid cells it is 4000, in others it is about 3999.9990). So this is a big change! I think this arises because wa_col is initialized to 4000; on master, it was then changed to 5000 in the first time step, but this no longer happens on the branch. Note that the aquifer_water_baseline parameter is 5000, and there are some notes in TotalWaterAndHeatMod about assuming that wa = aquifer_water_baseline in CLM50... so if we are changing wa to typically be ~ 4000, then we should give some thought to this TotalWaterAndHeat code.

(3) While I'm not 100% confident of this, I've managed to convince myself that all of the differences we're seeing can be traced back to this change in wa from 5000 to 4000: this leads to roundoff-level differences in both the pre and post-landuse state, and since the dynbal adjustment fluxes are the small differences between large values, the flux differences end up looking greater than roundoff (see also https://github.com/ESCOMP/ctsm/issues/570).

Yesterday, we talked about my doing some extra baseline comparisons where I back out Jinyun's changes in SoilWaterMovement in both the baselines and the branch. But I don't think this is worth doing until (2) is resolved, because I'm pretty sure we're still going to see differences from baseline as long as wa has been changed by this much.

We can talk more when you're back the week after Thanksgiving. One thought you could consider at that point – if this is going to take more time than you are able to devote to it before AGU – is that I could move ahead with the groundwater_irrigation branch but with the BalanceCheckMod change backed out, and we could open an issue to fix this in a follow-on tag.

PR #536, opened 2018-10-09
Allow holes in the number of history tapes and reorganize cmip6 usermods

billsacks:
### Description of changes

This PR will eventually contain three things:

1. Allowing holes in the number of history tapes

2. Reorganizing the cmip6 usermods directory

3. Some miscellaneous minor fixes

For now, it just includes (1) - and that isn't tested yet.

Here are the detailed comments for (1):

> Holes are cases where, for example, we have h0, h1 and h3 tapes, but no
> h2 tape (because there are no fields on the h2 tape).
> 
> The key part of this implementation is cycling some loops if a given
> history tape is not in use. There are a few other loops for which it
> probably would have been fine to do this cycling, but it wasn't
> necessary - e.g., doing initialization of some per-tape variables. In
> most of those cases, I didn't bother doing this cycling because it
> seemed like it could keep the code more robust to still initialize these
> per-tape variables for the not-in-use tapes.
> 
> Note that I removed this error check:
> 
>     ! Check that the number of history files declared does not exceed
>     ! the maximum allowed.
> 
>     if (ntapes > max_tapes) then
>        write(iulog,*) trim(subname),' ERROR: Too many history files declared, max_tapes=',max_tapes
>        call endrun(msg=errMsg(sourcefile, __LINE__))
>     end if
> 
> because, with the way ntapes is determined, it's impossible for ntapes
> to exceed max_tapes - i.e., this error check wasn't actually doing
> anything.
> 
> Also note that I put in some error-checking code to ensure that, in a
> restart run, if a history file is in use in the current run, then it was
> in use on the restart file, and vice versa.
> 

### Specific notes

Contributors other than yourself, if any: For now, just myself

CTSM Issues Fixed (include github issue #): None yet

- Allows holes in history tapes


billsacks:
@ekluzek as I note above, this PR is still a work in progress, but I wanted to get it out there so you can have a chance to look at it. I'm doing some manual testing of this new "hole" functionality now. Once I am convinced that is working right, I'll do the reorganization of the cmip6 usermods directory, then fold in some other fixes we want in this tag.

billsacks:
@ekluzek note that I have added a real fix for #24 , because it turns out I needed a fix for that for this new functionality to work right.

billsacks:
For the record: I have tested these changes for their ability to correctly write and read both 0-d and 1-d logicals.

billsacks:
@ekluzek I'd be happy to get your review of https://github.com/ESCOMP/ctsm/pull/536/commits/55f343e285739bdeabe7923b680704c07d9ccf6a if you have time. More generally: This branch is nearing completion, so let me know if you'd like to review any of it before I kick off final testing.

billsacks:
@ekluzek hold off on making any line comments on this until I have rebased onto master, which I'll do later this morning.

billsacks:
@ekluzek I have rebased onto master and force pushed. This is ready for your review: I just need to add a CMIP6WACCMDECK modifier and add tests (as well as two compsets for testing, as we discussed).

billsacks:
@ekluzek gave verbal approval of these changes, so I'm going ahead and merging this despite the fact that it hasn't officially been accepted.

PR #537, opened 2018-10-15
Fates parteh api

rgknox:
### Description of changes

FATES has undergone a significant refactor of plant allocation, growth and turnover (aka PARTEH). To facilitate these refactors, a namelist variable has been added which swaps the different modes or hypotheses that are executed.  Some history variables are also no longer in the output files and had to be removed from tests.

Synced with: https://github.com/NGEET/fates/pull/421

### Specific notes

The FATES developer team.

None

These changes must be syncronized with a new FATES release.  The FATES release should not be B4B, no non-fates code or tests should be effected.

yes!   Added  fates_parteh_mode =  INT

Testing performed, if any:

fates test suite, cheyenne, intel, all PASS (not b4b)

30 year, 14 PFT, f45 grid, smoke-test (PASS)

multi-decade comparison of biomass, leaf area and carbon fluxes against master at a tropical site.


src/main/clm_varctl.F90
```diff
@@ -196,6 +196,9 @@ module clm_varctl
   logical, public :: use_fates = .false.            ! true => use fates
 
   ! These are INTERNAL to the FATES module
+  integer, public            :: fates_parteh_mode = -9                 ! 1 => carbon only
```
ckoven: @rgknox what does the -9 mean here?

rgknox: it is a bogus initialization value, that will force the model to fail if not set correctly.

ckoven: ok, let's add a comment to that effect in case its not obvious.

```suggestion
  integer, public            :: fates_parteh_mode = -9                 ! 1 => carbon only; -9 => junk value that will force fail unless reset later
```

rgknox:
FATES side changes have been integrated.  All tests pass.
1 team-member reviewed.  Merging into fates_next_api

PR #539, opened 2018-10-19
Support a %CMIP6DECK compset modifier

billsacks:
### Description of changes

This was suggested by @rljacob and supported by @mvertens @jedwards4b
and @bertinia as a mechanism to provide CMIP6-specific compsets.

This allows us to have a compset like this:

```xml
  <compset>
    <alias>B1850CMIP6</alias>
    <lname>1850_CAM60_CLM50%BGC-CROP%CMIP6DECK_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD</lname>
    <science_support grid="f09_g17_gl4"/>
    <science_support grid="f09_g17"/>
  </compset>
```

I have tentatively chosen the modifier %CMIP6DECK rather than just
%CMIP6, mainly because I imagine we might want some other %CMIP6XXX
modifiers later, and if we had a simple %CMIP6 modifier, it would be
harder to write a correct regular expression that matches %CMIP6XXX
without accidentally matching %CMIP6. For CTSM, my impression is that we
want the same mods for all DECK runs. Other components might want
something different, and they don't necessarily have to use %CMIP6DECK:
they can come up with their own set of modifiers.

Finally, a specific note about the regex match: Elsewhere we use regular
expressions like `"_CLM50%[^_]*BGC"`, but here I have put the "%"
immediately before the modifier, like `"_CLM50[^_]*%CMIP6DECK"`: this
seems slightly more robust.

### Specific notes

Allows compsets with %CMIP6DECK modifier

Testing performed, if any:
From a cesm sandbox, created a case using the compset given above; confirmed that user_nl_cpl and user_nl_clm were filled out correctly.

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -230,7 +230,7 @@
     <valid_values></valid_values>
     <default_value></default_value>
     <values match="last">
-      <value compset="_CLM50%[^_]*CMIP6DECK">$SRCROOT/components/clm/cime_config/usermods_dirs/cmip6</value>
+      <value compset="_CLM50%[^_]*CMIP6DECK[%_]">$SRCROOT/components/clm/cime_config/usermods_dirs/cmip6</value>
     </values>
     <group>run_component_clm</group>
     <file>env_case.xml</file>
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -13,10 +13,10 @@
   -->
 
   <!-- Descriptions of all the different valid configurations for different model versions -->
-  <description modifier_mode="+">
+  <description modifier_mode="1">
     <desc lnd="CLM40[%SP][%CN][%CNDV][%CN-CROP][%CNDV-CROP]"                                                        >clm4.0:</desc>
-    <desc lnd="CLM45[%SP][%SP-VIC][%CN][%CNDV][%CN-CROP][%CNDV-CROP][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%CMIP6DECK]" >clm4.5:</desc>
-    <desc lnd="CLM50[%SP][%SP-VIC][%CN][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%CMIP6DECK]"                              >clm5.0:</desc>
+    <desc lnd="CLM45[%SP][%SP-VIC][%CN][%CNDV][%CN-CROP][%CNDV-CROP][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP]" >clm4.5:</desc>
+    <desc lnd="CLM50[%SP][%SP-VIC][%CN][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%BGC-CROP-CMIP6DECK]"         >clm5.0:</desc>
     <desc option="SP"              >Satellite phenology:</desc>
     <desc option="CN"              >CN: Carbon Nitrogen model</desc>
     <desc option="CNDV"            >CNDV: CN with Dynamic Vegetation</desc>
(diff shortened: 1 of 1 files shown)


dlawrenncar:
I'm fine with CMIP6DECK, but immediately this raises minor issues that we just aren't going to be able to easily get around.  I don't think we need to have isotopes on for the 1%CO2 or 4XCO2 or AMIP runs in the DECK.  Obviously, we can run with isotopes on anyway (at a small computational cost and a potentially more significant, though hopefully still relatively small data storage cost).  Or, whoever actually sets up and completes the runs can manually turn off isotopes.

billsacks:
In discussion with @dlawrenncar we'll just stick with CMIP6DECK. The isotopes aren't needed for everything, but Dave's feeling is that we should just keep things simple with a single modifier and usermods directory for all the cmip6 experiments.

billsacks:
@ekluzek please let me know if this new version looks good to you. Please use GitHub's review functionality to note whether you approve or request changes (that makes it easier for me to see the status of PRs at a glance).

billsacks:
@ekluzek @dlawrenncar I realized that at least some of the WACCM cases have not used virtual glaciers over Antarctica... I honestly can't remember if that was intentional or unintentional, but in order to maintain compatibility with the REFCASES for those compsets, we need to keep virtual Antarctica glaciers off in the associated usermods. So I'll need a separate cmip6 usermods directory that includes all cmip6 settings except for glacier_region_behavior, and a different compset modifier that chooses that directory.

**My question for you is: Do you have any feelings about how the usermods directories and compset modifiers should be named?** I'm not sure at this point whether virtual Antarctica glaciers will be off for all WACCM runs or just some of them; if all, I might go with something like:

Compset modifiers:
- %CMIP6DECK (or CMIP6CAMDECK?)
- %CMIP6WACCMDECK

Usermods directories, matching them:
- cmip6_deck (or cmip6_cam_deck)
- cmip6_waccm_deck

But I'm very open to suggestions.

billsacks:
Follow-up from my last comment: I have confirmed that we'll be consistent for all WACCM runs: we will NOT set glacier_region_behavior for any WACCM runs. So it could be appropriate to distinguish between CAM and WACCM in the name.

rljacob:
Hang on.  I was not suggesting an additional "%" modifer.  Does CIME even know how to process that?  What I was suggesting was more like:
`<lname>1850_CAM60_CLM50%BGC-CROPCMIP6DECK_CICE_POP2%ECOABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD</lname>`

Also the modifier completely replaces the need for a CLM_USER_MODS block.   Just define more modifiers for all the setting combinations you need.

billsacks:
@rljacob For CTSM, we've ended up NOT doing an extra "%" modifier (so sticking with something close to your suggestion), but for CAM and CICE we will have the extra "%" modifier, in line with how they handle having multiple options. Yes, CIME knows how to parse multiple "%" modifiers, and the advantage of doing it that way is that it avoids the combinatoric problem. (But for CTSM, we're only supporting CMIP6DECK along with BGC-CROP, so we decided that something close to your suggestion makes more sense.)

I don't understand your suggestion about not needing a CLM_USER_MODS block: As far as I know, the only effect a compset modifier can have is to set an xml variable in config_component.xml (or config_compsets.xml), and in this case, the effect we want is to pull in some user mods. Maybe you can explain your thinking a little more, if you see something that I'm missing?

rljacob:
What kind of user mods?  Source code mods?  namelist mods?  In either case if they are supposed to be a permanently available feature of the code, then the mods should be committed and the namelist variables should be made settable by xml.

billsacks:
@rljacob I think it would be best to continue this discussion live, with screen sharing, to demo what we currently have in place. Maybe we can do this during tomorrow's cime telecon?

billsacks:
For the record: After demoing this to @rljacob , he was okay with the way it's being done. However, he feels – and I'm coming to agree – that, long-term, it would be better to do various settings (at least those that are not simply related to output fields) via xml variables (i.e., a compset modifier sets an xml variable which influences the namelist defaults) rather than via usermods.

PR #527, opened 2018-10-02
Add carbon_isotope user-mod directory to turn on c13 and c14

ekluzek:
User-mod directory for carbon_isotopes and include it for the cmip6 user-mod

### Description of changes

Add carbon_isotopes user-mod directory to turn on c13 and c14 carbon isotopes.

### Specific notes

Contributors other than yourself, if any: none

CTSM Issues Fixed (include github issue #): none

Are answers expected to change (and if so in what way)?

Yes, just for the cmip6 user-mod test cases

Any User Interface Changes (namelist or namelist defaults changes)? Changes cmip6

Testing performed, if any: none yet
(aux_clm on cheyenne for gnu/pgi and hobart for gnu/pgi/nag is the standard for tags on master)

**NOTE: Be sure to check your Coding style against the standard:**
https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines

PR #551, opened 2018-10-29
Add history fields for vertically-resolved sums of SOIL1C, SOIL2C, SO…

olyson:
### Description of changes

Add history fields for vertically-resolved sums of SOIL1C, SOIL2C, and SOIL3C for C12, C13, C14.

### Specific notes
New fields are SOILC_vr, C13_SOILC_vr, and C14_SOILC_vr.
The C12 and C14 versions are active by default and C13 is inactive.

Contributors other than yourself, if any:  None

CTSM Issues Fixed (include github issue #):  This is associated with pull request #527 which is now closed.

Are answers expected to change (and if so in what way)?  No

For CMIP6, we would want to exclude the following from the primary history tape (C13_SOILC_vr is off by default):
hist_fexcl1 = 'SOILC_vr','C14_SOILC_vr'
and add the following to the h3 stream (yearly average, gridcell level):
hist_fincl4 = 'SOILC_vr','C14_SOILC_vr'

Testing performed, if any:
Since we had core-hours to spend by Oct 31, I ran testing on cheyenne with intel compiler.  All passed except for 2 expected failures.


billsacks:
@dlawrenncar I wanted to confirm: In an earlier thread you said this should come into CESM 2.1.1. However, that would mean that the out-of-the-box set of diagnostic fields will differ for CMIP (and other) runs done from 2.1.1 vs. 2.1.0. Do you feel that's acceptable, or would it be better to just put this on master but not try to use this for any of the cmip runs given that we've already done some without these diagnostics?

dlawrenncar:
It would be good to get this into CESM2.1.1.

PR #559, opened 2018-11-06
Updated fates external tag

rgknox:
### Description of changes

This is just an update of the tag for fates in externals.


ekluzek:
OK, I'm just accepting and assuming this is good to go.

PR #561, opened 2018-11-07
Add list of checks that should be done when making new surface datasets

billsacks:
### Specific notes


ekluzek:
We talked about this at our last CLM-CMT (Dec/17/2018) meeting and agreed to the list. We also realized that had we done this every time we would HAVE found the wetlands Antartica problem. So that seems like a good reason to do it.

In creating landuse.timeseries files for SSP's I've found some new things that I'll add in.

ekluzek:

Issue #545, opened 2018-10-25
Antarctica ice shelves are being treated as wetlands rather than glaciers


billsacks:
### Brief summary of bug

In recent surface datasets, including the ones being used for CMIP6 runs, Antarctica's ice shelves are being treated as wetlands rather than glaciers.

### General bug information

**CTSM version you are using:** As far as I can tell, affects all versions for the last year or two.

**Does this bug cause significantly incorrect results in the model's science?** No

Although the landunit type is incorrect over the ice shelves, at least in the one run I checked (b.e21.BHIST.f09_g17.CMIP6-historical.001), the affected regions are still covered in deep snow year-round, so the impact on land-atmosphere fluxes should be minimal.

**Configurations affected:** All (I think)

#### General description

Jan Lenaerts noticed that there are no glaciers over Antarctica's ice shelves in the CMIP6 runs (for example, 80 S, 180 E/W (Ross Ice Shelf)). This is very apparent if you look at the 4th element (glc_mec) of PCT_LANDUNIT in a history file from one of these runs.

I think I have traced the problem back to a change in the LANDMASK field on the new vegetation raw data files that we've been using for the last year or two: This field is 0 over the ice shelves in /glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/pftcftdynharv.0.25x0.25.LUH2.histsimyr1850-2015.c170629/mksrf_landuse_histclm50_LUH2_1850.c170629.nc, whereas it is 1 over ice shelves in the files we used before a couple of years ago (/glade/p/cesmdata/cseg/inputdata/lnd/clm2/rawdata/pftlandusedyn.0.5x0.5.simyr1850-2005.c090630/mksrf_landuse_rc1850_c090630.nc).

#### Probable cause

For better or worse (probably worse), mksurfdata_map uses the landmask on the PFT rawdata file to determine where to set landcover to 100% wetland:

https://github.com/ESCOMP/ctsm/blob/a34404419aaa81f5021602963222c50daa10c0f6/tools/mksurfdata_map/src/mksurfdat.F90#L707-L724

My understanding of the code is that the LANDMASK field isn't used directly in the mksurfdata_map code, but (if I remember correctly) I think it's used indirectly: I think the LANDMASK field on that file is used to set the mask field on the mapping files, and the mask field on the mapping files determines the mask in mksurfdata_map.

#### What this does and doesn't impact in CMIP6 runs

[...]

PR #564, opened 2018-11-08
Update mksurfdata_map to include glaciers outside of pft landmask

lawrencepj1:
… as well as lakes and wetlands outside of the landmask

with new code:

```
<        ! Assume wetland, glacier and/or lake when dataset landmask implies ocean

<           if (pctgla(n) < 1.e-6_r8) then
<               pctwet(n)    = 100._r8 - pctlak(n)
<               pctgla(n)    = 0._r8
<           else
<               pctwet(n)    = 100._r8 - pctgla(n) - pctlak(n)
<           end if
```

from

```
>        ! Assume wetland and/or lake when dataset landmask implies ocean

>           pctwet(n)        = 100._r8 - pctlak(n)
>           pctgla(n)        = 0._r8
```

### Specific notes

Addresses #545 (fixing it will require new surface datasets generated with this code)

Testing performed, if any:


billsacks:
Update: Based on today's co-chairs discussion, we actually want this on both master and the release branch. We want both the updated mksurfdata_map and all new surface datasets to come out-of-the-box starting in CESM2.1.1. To be clear: Starting with CESM2.1.1, all CTSM surface datasets should have this bug fix, and so should have glaciers rather than wetlands over the Antarctic ice shelves. This will mean that CESM2.1.1 changes answers relative to CESM2.1.0, and that is deemed acceptable. (We should just carefully check the new surface datasets against the old to ensure that the only thing that changes is this one aspect.)

In discussion over lunch, @ekluzek and I agreed that he would make this change in time for the CESM2.1.1 release. The earlier the better, probably, in terms of supporting upcoming cmip6 runs out-of-the-box, but there isn't a critical need for it before that release.

ekluzek:

PR #567, opened 2018-11-09
Add a bunch of ignore patterns for vim

billsacks:
From
https://github.com/github/gitignore/blob/master/Global/Vim.gitignore

Motivated by the fact that a .swp file appeared in a PR. We probably
don't need all of these, but it looks like they should be safe to
include (i.e., I can't imagine a real file having a name like one of
these patterns, given that most of them begin with '.' or '_').

### Specific notes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,23 @@ manage_externals.log
 *.swp
 *~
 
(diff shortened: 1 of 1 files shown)


ekluzek:
Like other's I might put this on the release branch first.

ekluzek:

PR #572, opened 2018-11-21
Water tracer updates for initial things in driver loop

### LINKED ISSUE #487: Set tracer versions of atm2lnd water variables
As @mathewvrothstein recognized: One of the first steps in getting water tracers consistent with bulk will be setting tracer versions of atm2lnd water variables. For tracers that actually come from other components (like particular water isotopes), I imagine that we'll set the tracer quantities alongside the bulk in lnd_import. For now, though, we just want tracers set up artificially, as fixed constants times the incoming bulk quantity.

There are three pieces to this:

(1) The lower-level function that does this updating.

This is kind of like the function in CNCIsoFluxMod, but we'll also need a special form that, rather than accepting two states, just accepts a fixed ratio.

(2) When is this called?

This ties in with a broader question of where in the driver loop we'll do the updating of the water tracer fluxes. Ideally we'd do this only in a few places - similar to the few places where this is done in CNCIsoFluxMod. At least in the near-term, though, we may have to do this in more places, due to the fact that the biogeophysics has state updates at a lot of points.

That said: it's not too important to resolve that question right now. For now, we can have a separate call to calculate tracer fluxes for the atm2lnd fluxes, near the start of the driver loop. Later, we may be able to merge this with some other update, if the atm2lnd tracer fluxes aren't actually needed until later in the driver loop: We can update a given tracer flux any time up until the time it's actually needed, as long as this update happens before its "source" state has changed (see also the discussion in #490 ).

We can probably put this call after the call to `downscale_forcings`:  I think that no water-related atm2lnd stuff is used before that point in the driver loop, so it's safe to wait until then to calculate the tracer versions of atm2lnd variables. However, note that this means that tracers and bulk will be inconsistent for these variables until this point. I think that's okay; it just means that you can't arbitrarily put the consistency checks at any point in the driver loop. i.e., we can put a tracer consistency check call after this call to update atm2lnd tracers post-downscale_forcings, but not before then. In the very near-term, we may want to split this into two pieces: (i) update tracers for the non-downscaled quantities very first thing in the driver loop; call consistency checks (should pass); (ii) after downscale_forcings, update tracers for the downscaled quantities; again call consistency checks (should still pass). Pretty soon, though (perhaps even before merging this to master), I think I'd like to merge these two pieces, so that all of the atm2lnd tracer updates happen in (ii), and we do the first tracer consistency check at or after (ii). The main reason to separate these initially is to support incrementally getting tracer consistency in place - particularly for the pieces of code between the start of the driver loop and the call to downscale_forcings.

(3) How is this called?

I'm imagining a routine in WaterType like set_atm2lnd_tracers

This will only set the tracers that aren't set directly in lnd_import. I expect that tracers that actually come from the atm / rof will be set in lnd_import, so this routine will just be responsible for computing atm2lnd values for tracers for which we don't have explicit input values, and have to assume some value based on the bulk quantity.

Then this will have code like this:

```Fortran
  ! ------------------------------------------------------------------------
  ! WaterType.F90: set_atm2lnd_tracers()
  ! ------------------------------------------------------------------------

  do i = 1, this%num_tracers
     if (this%tracer_info(i)%has_fixed_ratio()) then
        call this%wateratm2lnd_tracer_inst(i)%set_vals_from_bulk(this%wateratm2lndbulk_inst)
     end if
  end do

  ! ------------------------------------------------------------------------
  ! Wateratm2lndType.F90: set_vals_from_bulk(bulk_inst)
  ! ------------------------------------------------------------------------

  ! forc_q isn't actually a flux, but we're abusing the flux routine for this purpose
  ! because the calculation we want is the same as for a flux
  call water_tracer_flux_calc_fixed_ratio( &
       tracer_flux = this%forc_q_not_downscaled_grc(begg:endg), &
       bulk_flux = bulk_inst%forc_q_not_downscaled_grc(begg:endg), &
       ratio = this%info%get_fixed_ratio())

  call water_tracer_flux_calc_fixed_ratio( &
       tracer_flux = this%forc_rain_not_downscaled_grc(begg:endg), &
       bulk_flux = bulk_inst%forc_rain_not_downscaled_grc(begg:endg), &
       ratio = this%info%get_fixed_ratio())

  call water_tracer_flux_calc_fixed_ratio( &
       tracer_flux = this%forc_snow_not_downscaled_grc(begg:endg), &
       bulk_flux = bulk_inst%forc_snow_not_downscaled_grc(begg:endg), &
       ratio = this%info%get_fixed_ratio())

  do c = begc, endc
     if (col%active(c)) then
        g = col%gridcell(c)
        forc_q_not_downscaled_tracer_col(c) = this%forc_q_not_downscaled_grc(g)
        precip_not_downscaled_tracer_col(c) = &
             this%forc_rain_not_downscaled_grc(g) + &
             this%forc_snow_not_downscaled_grc(g)

        ! the following lines will be done redundantly for every tracer; I'm not sure if
        ! it's worth trying to optimize this:
        forc_q_not_downscaled_bulk_col(c) = bulk_inst%forc_q_not_downscaled_grc(g)
        precip_not_downscaled_bulk_col(c) = &
             bulk_inst%forc_rain_not_downscaled_grc(g) + &
             bulk_inst%forc_snow_not_downscaled_grc(g)
     end if
  end do

  ! forc_q isn't actually a flux, but we're abusing the flux routine for this purpose
  ! because the calculation we want is the same as for a flux
  call water_tracer_flux_calc( &
       tracer_flux = this%forc_q_downscaled_col(begc:endc), &
       bulk_flux = bulk_inst%forc_q_downscaled_col(begc:endc), &
       tracer_source = forc_q_not_downscaled_tracer_col(begc:endc), &
       bulk_source = forc_q_not_downscaled_bulk_col(begc:endc))

  call water_tracer_flux_calc( &
       tracer_flux = this%forc_rain_downscaled_col(begc:endc), &
       bulk_flux = bulk_inst%forc_rain_downscaled_col(begc:endc), &
       tracer_source = precip_not_downscaled_tracer_col(begc:endc), &
       bulk_source = precip_not_downscaled_bulk_col(begc:endc))

  call water_tracer_flux_calc( &
       tracer_flux = this%forc_snow_downscaled_col(begc:endc), &
       bulk_flux = bulk_inst%forc_snow_downscaled_col(begc:endc), &
       tracer_source = precip_not_downscaled_tracer_col(begc:endc), &
       bulk_source = precip_not_downscaled_bulk_col(begc:endc))

  ! ------------------------------------------------------------------------
  ! water_tracer_flux_calc_fixed_ratio(tracer_flux, bulk_flux, ratio)
  !
  ! We may also need a version of this with a filter, as is done in CIsoFluxCalc (or at
  ! least using an 'active' array). But the gridcell-level quantities don't have a
  ! filter, and all gridcells are active implicitly.
  ! ------------------------------------------------------------------------

  my_lbound = lbound(tracer_flux)
  my_ubound = ubound(tracer_flux)
  do i = my_lbound, my_ubound
     tracer_flux(i) = bulk_flux(i) * ratio
  end do

  ! ------------------------------------------------------------------------
  ! water_tracer_flux_calc(tracer_flux, bulk_flux, tracer_source, bulk_source)
  ! Calculates a tracer flux based on a bulk flux, for a flux without fractionation
  !
  ! We may also need a version of this with a filter, as is done in CIsoFluxCalc. Or at
  ! least use an 'active' array.
  ! ------------------------------------------------------------------------

  my_lbound = lbound(tracer_flux)
  my_ubound = ubound(tracer_flux)
  do i = my_lbound, my_ubound
     if (bulk_source(i) == 0._r8) then
        if (tracer_source(i) == 0._r8) then
           tracer_flux(i) = 0._r8
        else
           ! tracer_source is non-zero despite bulk_source being 0. Carbon isotope code
           ! sets the flux to 0 in this case. This seems like an error, though: How can we
           ! have tracer state when we have 0 bulk state? It seems like this either would
           ! have arisen from a coding error or a numerical issue; to deal with the
           ! latter, we might need some code (elsewhere) that sets the tracer state to 0
           ! if the bulk state is 0 and the tracer state is non-zero. In any case, I feel
           ! better about aborting in this case.
           call endrun()  ! fill this in with a message
        end if
     else
        tracer_flux(i) = bulk_flux(i) * (tracer_source(i)/bulk_source(i))
     end if
  end do
```

(Note that #486 raises some similar questions regarding code structure, such as whether we have a wrapper routine in WaterType.)

### LINKED ISSUE #488: Do any updates of water tracers needed in dynSubgrid_driver
One of the first driver calls is dynSubgrid_driver. To get tracer consistency checks to pass early in the driver loop, we'll need to update any tracer variables related to variables updated in dynSubgrid_driver.

### LINKED ISSUE #489: Do any updates of water tracers needed in BeginWaterBalance
One of the first driver calls is BeginWaterBalance. To get tracer consistency checks to pass early in the driver loop, we'll need to update any tracer variables related to variables updated in BeginWaterBalance.

billsacks:
### Description of changes

Update water tracers for initial stuff done in driver loop. This includes atm2lnd forcings (non-downscaled and downscaled), balance check initialization, and dyn subgrid updates.

This is a large PR that includes a lot of uninteresting stuff along with just a bit that's interesting. Broadly, speaking, the changes here are:

(1) Reworked WaterType to make it easier / more robust for other code to loop over tracers or bulk+tracers

(2) The most interesting changes are probably the implementation of the pseudo-code given in #487 (in Wateratm2lndType.F90 and WaterTracerUtils.F90)

(3) In various other places, do some infrastructurey stuff (initializing water balance, doing dyn subgrid stuff) for tracers as well as bulk

(4) Supporting unit tests and unit test infrastructure

### Specific notes

Portions of the design (and possibly code) have been reviewed by @mathewvrothstein , @davidnoone and @mvertens 

Resolves #487 
Resolves #488 
Resolves #489 

Are answers expected to change (and if so in what way)? Roundoff-level changes in sensible heat flux from precip conversion. (And significant changes in configurations with water tracers, but that configuration isn't actually working yet.)

Any User Interface Changes (namelist or namelist defaults changes)? none

Testing performed, if any: So far just:
- SMS_D_Ln1.f10_f10_musgs.I2000Clm50BgcCropGs.hobart_nag.clm-tracer_consistency
- ERP_D_P36x2_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default with baseline comparisons


billsacks:
@ekluzek @mathewvrothstein @davidnoone - I'm planning to merge this to master early next week. Mat and David have already reviewed the pseudo-code of the most interesting parts of this PR (point (2) in my initial comment),  and (if I remember correctly) I think I showed some of the actual code to Mat and Mariana a while ago (specifically, point (1) in my initial comment).

I'd like to get this in sooner rather than later, so that it can be merged with some of Mat's work and we can move ahead with other pieces of the isotope implementation.

That said, anyone should let me know by next Monday if you'd like to review this more carefully before it goes in.

billsacks:
I ran the full test suite. All tests pass and are bit-for-bit except for expected roundoff-level changes in FSH_PRECIP_CONVERSION, FSH_TO_COUPLER and l2x_Fall_sen (along with the expected changes in SMS_D_Ln1.f10_f10_musgs.I2000Clm50BgcCropGs.hobart_nag.clm-tracer_consistency).

@ekluzek we can talk on Monday about the timing of getting this to master.

Issue #201, opened 2017-12-29
Perform gridcell-level water balance checks bracketing the entire run loop


billsacks:
_From @billsacks on March 17, 2017 20:38_

Currently, the CLM driver loop looks like this, in terms of pieces relevant to this issue:

- Compute pre-dynlu gridcell-level water mass

- Adjust column areas via dynamic landunits

- Compute post-dynlu gridcell-level water mass

- Compute adjustment fluxes based on difference between pre- and post-dynlu gridcell-level water mass

- Compute pre-physics column-level water mass

- Do physics

- Compute post-physics column-level water mass

- Do column-level balance checks using the pre- and post-physics column-level water mass and the relevant fluxes

As long as you trust:

1. that the calculation of the dynlu adjustment fluxes is done correctly, and 

2. that the column-level fluxes used in the column-level balance checks correctly mirror the relevant gridcell-level fluxes

then you can convince yourself that this sequence ensures gridcell-level conservation.

However, it would be more robust and more reassuring if we performed gridcell-level water conservation checks, bracketing essentially the entire driver run loop (including dynamic landunits adjustments as well as the physics). **Ideally, the fluxes considered in these gridcell-level checks would be the actual fluxes received by and sent from CLM (gridcell-level precipitation, runoff, etc.).** This last point is particularly important for runoff, which is now split into a few different terms, and it seems easy for some term to accidentally not be included in any of the runoff terms sent to the coupler - so we should do these gridcell-level balance checks using the runoff terms that are actually sent to the coupler.

The sequence would then look like:

- Compute beginning gridcell-level water mass

- Adjust column areas via dynamic landunits

- Compute post-dynlu gridcell-level water mass

- Compute adjustment fluxes based on difference between pre- and post-dynlu gridcell-level water mass

- Compute pre-physics column-level water mass

- Do physics

- Compute post-physics column-level water mass; also average this to the gridcell level

- Do column-level balance checks using the pre- and post-physics column-level water mass and the relevant fluxes

- Do gridcell-level balance checks using the beginning gridcell-level water mass and ending gridcell-level water mass. For this check, ideally use the gridcell-level fluxes that are passed to and from CLM

Note that this plan still maintains the column-level balance checks: These are useful for seeing which column(s) have balance errors. But it's possible that we could avoid doing these in general, and just drill down to the column-level if a gridcell-level balance check fails.

_Copied from original issue: NCAR/CLM#3_

billsacks:
[...]

PR #586, opened 2018-12-05
Option for rain-to-snow to immediately run off in some regions

billsacks:
### Description of changes

Up until now: When repartition_rain_snow is .true. (which is the default
for CLM5), rain that falls when the near-surface temperature is cold is
converted to snow. This repartitioning was put in place for two
reasons: (1) Downscaling to elevation classes: changing the balance
between rain and snow for different elevation classes; (2) Correcting
problems in CAM. However, members of the Land Ice Working Group would
like to change this behavior so that, when CAM produces cold-temperature
rain, this rain immediately runs off rather than being converted to
snow. The purpose of this is to reduce the too-high SMB over portions of
Greenland in CESM2 coupled runs (which results in part from CAM's
generation of liquid precipitation despite very cold temperatures).

This new behavior is implemented in a glacier region-specific manner,
based on a new namelist flag, glacier_region_rain_to_snow_behavior. It
is not at all ideal to make this aspect of the physics differ by region,
but this has been requested by members of the Land Ice Working Group in
order to address biases over Greenland while having minimal impact on
the climate (so that the climate can stay very similar to that of the
official CMIP6 runs). Note that, unlike other glacier region-specific
behaviors, this one applies to all landunits, not just glaciers. This
also seems a bit non-ideal, but we want the physics to be the same for
all landunit types in a given region, and we also want this behavior to
apply to vegetated columns because they are used for glacial
inception (and we want this alternate behavior to apply to glacial
inception, too, in order to decrease some instances of inception).

The justification for this new physics is: In the case of (1) above: If
CAM is generating rain at a given elevation / temperature, that doesn't
necessarily imply that an equal water equivalent of snow would be
generated at a higher elevation / lower temperature: indeed, in reality,
there might not be any precipitation falling at that higher elevation /
lower temperature. In the case of (2) above: There seem to be problems
with CAM's microphysics that cause it to produce too much rain when
temperatures are very cold; it seems (at least to some people) equally
justifiable to throw this cold rain away (by sending it to the ocean as
runoff) as it is to convert this cold rain to snow.

Note: I don't think any changes are needed in
BalanceCheck (unfortunately), since BalanceCheck currently uses the
post-downscaling precipitation fluxes, and the pre-lnd2atm runoff
fluxes (i.e., the new runoff flux isn't included in the terms in
BalanceCheck, and it doesn't need to be because BalanceCheck uses the
post-downscaling precipitation fluxes). (See also
https://github.com/ESCOMP/ctsm/issues/201#issuecomment-444264954 .)

### Specific notes

Contributors other than yourself, if any: @whlipscomb contributed conceptual thoughts

- New namelist option: glacier_region_rain_to_snow_behavior

Testing performed, if any:
- So far just ran unit tests
- Will run full test suite, and will add a new system test covering this behavior


ekluzek:
@billsacks I looked at it briefly just now. In general it makes sense to me, and I didn't see any glaring problems. The one thing that I did see that I was going to look into further is if you do a bunch of character string comparisons inside of nested do loops.

I do think it would be worthwhile for me to look at it more. But, I don't know that should stand in the way of your continuing testing? But, if I could make sure I look at it by tomorrow say?

billsacks:
Okay, I'll go ahead and start testing. Regarding this:

> The one thing that I did see that I was going to look into further is if you do a bunch of character string comparisons inside of nested do loops.

A few comments:

1. I may be forgetting something, but I think these character string comparisons are just inside loops over min_glacier_region_id to max_glacier_region_id (in the translate_*) routines. Here I translate the string-based namelist input into integers for the sake of the rest of the code. (In fact, part of the complexity of this module is due to the pains I took to turn things from strings to integers as soon as possible.)

2. This is just at initialization

3. I don't want to do any general refactoring of glcBehaviorMod on this branch, given that it will only be on the release branch and not master, so any refactoring would not make it to master. I'm open to ideas on refactoring that module if you see something worthwhile, but I think that should be done on master, not here.

ekluzek:
OK, that was the thing I wanted to check that the character string comparisons were just at initialization. Since, you've already done that I'm not concerned about that. That's the kind of thing that I think we should do, use character strings in the namelist for readability, but use fixed integers in the run-time code for fast comparisons. It just takes some work to do that...

billsacks:
@ekluzek Testing is all good; I'll wait for your go-ahead later today before merging.

lvankampenhout:
Nice work Bill, thanks for the very clear description at the top.

adamrher:
@billsacks if I'm reading the namelist_defaults correctly in this PR, the cold rain straight to runoff option is not turned on by default, anywhere. It's only turned on for a clm test.

billsacks:
> if I'm reading the namelist_defaults correctly in this PR, the cold rain straight to runoff option is not turned on by default, anywhere. It's only turned on over Greenland for a clm test.

Yes, that's correct. This PR provided the option but didn't turn it on by default anywhere. I believe some runs by members of the LIWG (including runs for ISMIP6?) turned this on.

PR #600, opened 2018-12-28
Set tracer version of Irrigation fluxes

### LINKED ISSUE #521: Set tracer versions of fluxes set by ApplyIrrigation
ApplyIrrigation is called before CanopyHydrology, so we should get the tracer versions of variables in that routine set before moving on to CanopyHydrology.

### LINKED ISSUE #593: Generalize groundwater irrigation availability to handle multiple patches per column
Groundwater irrigation (introduced by @swensosc in ctsm1.0.dev020) will compute the wrong water availability if there are multiple irrigated patches in a single column. Currently this is never the case, but the code should be flexible enough to handle this case.

The problem is in this block of code:

https://github.com/ESCOMP/ctsm/blob/7755baf3f729ff9adb0e39088c02113bbc9ebb58/src/biogeophys/IrrigationMod.F90#L917-L923

Note that this compares patch-level demand / fluxes with column-level availability. I think this can lead to the irrigation flux being overly constrained (e.g., if there are two patches on a column, with one being irrigated and the other not: I think the irrigated patch is currently constrained as if it took up the entire column).

Off-hand, I'm not sure how to solve this problem, but I don't think we want code like this that implicitly assumes one patch per column.

### LINKED ISSUE #595: Inconsistency between CalcAvailableUnconfinedAquifer and WithdrawGroundwaterIrrigation
### Brief summary of bug

The code for calculating per-layer water availability is inconsistent between CalcAvailableUnconfinedAquifer and WithdrawGroundwaterIrrigation.

### General bug information

**CTSM version you are using:** ctsm1.0.dev020

**Configurations affected:** Only configurations with groundwater irrigation (which is newly available in ctsm1.0.dev020)

### Details of bug

In CalcAvailableUnconfinedAquifer:

```Fortran
             if (j==jwt(c)+1) then
                available_water_layer=max(0._r8,(s_y*(zi(c,j) - zwt(c))*1.e3))
             else
                available_water_layer=max(0._r8,(s_y*(zi(c,j) - zi(c,j-1))*1.e3))
             endif
```

In WithdrawGroundwaterIrrigation:

```Fortran
                irrig_layer=min(irrig_total,(s_y*(zi(c,j) - zwt(c))*1.e3))
```

It seems incorrect that irrig_layer always has zi - zwt, instead of sometimes having zi(c,j) - zi(c,j-1) as in CalcAvailableUnconfinedAquifer.

I think this problem is related to the more general bug in the calculation of rsub_top_layer that @swensosc initially fixed in #523 but I asked him to back out of that PR (deferring its fix to its own PR). Based on the fix he proposed there, it seems that the code in CalcAvailableUnconfinedAquifer is correct, and the code in WithdrawGroundwaterIrrigation is incorrect.

billsacks:
### Description of changes

Set tracer version of irrigation fluxes.

This required a substantial rewrite of ApplyIrrigation (now renamed to
CalcIrrigationFluxes). The problem was that the code was written to
compute the total irrigation withdrawal from groundwater, then use this
to compute the application fluxes (drip/sprinkler), then later divide
this total withdrawal by layer. But for tracer fluxes to be computed
correctly, we needed to reorder this, so that the per-layer withdrawals
are determined before determining the application fluxes. This is
because the application flux needs to know the tracer concentrations of
the source, which requires knowing which layers the source is drawing
from. This was made more challenging because of the mix of patch-level
and column-level variables at play: irrigation demand is patch-level,
groundwater availability and extraction is column-level, but application
is back to patch-level.

In a somewhat-related change, I also reworked the passing of information
between soil hydrology (groundwater availability) and irrigation:
Previously, there was some near-duplicate code in
CalcAvailableUnconfinedAquifer (which was called prior to
ApplyIrrigation) and WithdrawGroundwaterIrrigation (which was called
after ApplyIrrigation). I have reworked the code to remove this
duplication, calling the new CalcIrrigWithdrawals in the midst of
CalcIrrigationFluxes. In doing so, I reconciled an accidental
discrepancy between the two original routines (see #595).

### Specific notes

Contributors other than yourself, if any: none, but @swensosc reviewed
the initial parts of the rework of ApplyIrrigation and the passing of
information between soil hydrology and irrigation

- Resolves #521 (Set tracer versions of fluxes set by ApplyIrrigation)
- Resolves #593 (Generalize groundwater irrigation availability to
  handle multiple patches per column)
- Resolves #595 (Inconsistency between CalcAvailableUnconfinedAquifer
  and WithdrawGroundwaterIrrigation)

- Small answer changes when groundwater irrigation is enabled. In
  principle, could change answers by greater than roundoff due to fix of
  #595, but I didn't see any answer changes in my 7-month test due to
  that change. Other than that, just roundoff-level changes.
- Answer changes when tracers are enabled.
- Roundoff-level changes in the diagnostic field, QIRRIG_FROM_SURFACE,
  for many tests

Testing performed, if any:
I have run the unit tests and selected system tests so far. I am running
the full aux_clm test suite now.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -108,6 +108,44 @@ CTSM testing:
 
     ok means tests pass, expected baseline failures as noted below
 
+  Additional manual testing:
+
+  - Out of the box, the one-timestep tracer consistency test
+    (SMS_D_Ln1.f10_f10_musgs.I2000Clm50BgcCropGs.hobart_nag.clm-tracer_consistency)
+    passed when the new check was inserted after the irrigation call,
+    even without the changes in this tag. Presumably this is because no
+    irrigation was being done in the first time step of this test. To
+    confirm that the changes here were both necessary and effective, I
+    hacked the code to force irrigation of all types (surface and
+    groundwater) on the first time step (see
+    https://github.com/billsacks/ctsm/commit/d28a03145). I confirmed
+    that this test, with those code hacks, failed before the changes in
+    this tag, and passed with these changes in place.
+
+  - I used a multi-step process to confirm that these changes are only
+    roundoff-level different for groundwater irrigation (because
+    feedbacks in the system result in
+    SMS_D_Ld5.f10_f10_musgs.I2000Clm50BgcCrop.hobart_nag.clm-irrig_alternate
+    looking greater-than-roundoff-level different from the baseline):
+
+    (1) Confirmed that e507fe603 is only roundoff-level different from
+        baseline using cprnc output from this test.
+
+    (2) Confirmed that a4a1a626c is only roundoff-level different from
+        e507fe603. This was the tricky part. I did this by introducing
+        some temporary code that (a) computed some fluxes in both the
+        old and new ways, (b) compared the old and new methods for each
+        point and time step, confirming that they were no more than
+        roundoff-level different, then (c) set the fluxes to the old
+        method. I confirmed that this was bit-for-bit with e507fe603,
+        and the checks for greater-than-roundoff-level differences
+        between the old and new methods were never triggered. (See
+        https://github.com/billsacks/ctsm/commit/7d23e9e and the earlier
+        commits on that branch.)
+
+    (3) Confirmed that remaining changes an the branch are bit-for-bit
+        with a4a1a626c.
+
 CTSM tag used for the baseline comparisons: ctsm1.0.dev021
 
 
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,162 @@
 ===============================================================
+Tag name:  ctsm1.0.dev022
+Originator(s):  sacks (Bill Sacks)
+Date: Tue Jan  8 11:01:38 MST 2019
+One-line Summary: Set tracer version of irrigation fluxes
+
+Purpose of changes
+------------------
+
+Set tracer version of irrigation fluxes.
+
+This required a substantial rewrite of ApplyIrrigation (now renamed to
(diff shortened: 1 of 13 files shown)


billsacks:
@swensosc @davidnoone @mathewvrothstein Would any of you like to review these changes before I bring them to master? I'd like to get this branch onto master by, say, the end of next week, but I can hold off until then to give you time to look it over if you'd like. Please let me know if you'd like to review it, and if you need more time than a week (particularly given the holidays). (But I'm also happy to make follow-up changes to this after it's on master if you need significantly more time.)

It will probably be hard to review this by looking at the diffs. Instead, I recommend looking at the new version of IrrigationMod.F90: look at the CalcIrrigationFluxes method and the four methods it calls. This will show you the bulk of the new code that hasn't already been reviewed by @swensosc .

@davidnoone I'm calling you out here because this is the first instance of calculating tracer fluxes for a parameterization that feels more sciencey than infrastructurey. I'd welcome any feedback you have. That said, I'm hoping that some other examples will turn out to be simpler than this one: This one was tricky because (a) there were multiple sources of irrigation withdrawals, including multiple soil layers, which then needed to be summed to create an overall irrigation application flux, and (b) there was a complex interplay between column-level and patch-level fluxes. Because of these factors, I put the calculation of tracer fluxes in the middle of the routine. In general, I hope to be able to wait to calculate tracer fluxes until after a particular science routine is complete.

billsacks:
Update: I have run the full test suite and all tests pass other than expected baseline failures, as noted above.

PR #602, opened 2019-01-03
Hook up expected fails functionality in run_sys_tests

billsacks:
### Description of changes

The custom cs.status files produced by run_sys_tests now points to CTSM's expected fails file, so that the output of these cs.status scripts will be annotated with expected failures.

This requires the changes in https://github.com/ESMCI/cime/pull/2962, and should not be brought to master until that cime PR is on master and we can point to a cime master tag (I am labeling this PR "work in progress" for this reason). See the comments in that PR for example output.

Note that this required reworking the format of the expected fails xml file to be more easily machine-parseable. I welcome any feedback on the new xml schema, either here or in the cime PR.

### Specific notes

Testing performed, if any:
- Tests in the python directory
- Manual testing of new cs.status file and examination of output from this command (i.e., confirming that expected failures are annotated correctly)


cime_config/testdefs/ExpectedTestFails.xml
```diff
@@ -1,24 +1,123 @@
...
+      <status>FAIL</status>
+      <issue>ESMCI/cime#2917</issue>
+    </phase>
+    <phase name="BASELINE">
+      <status>FAIL</status>
+      <issue>ESMCI/cime#2917</issue>
+    </phase>
+  </test>
+
+  <!-- fates test suite failures -->
```
billsacks: Note that different test suites are now delineated by comments, not by xml sections. This is because, in the machine parsing of this file, we don't have an easy way to determine which category is appropriate, and in any case, it's not really important: if a test from one category is supposed to fail, then it will fail regardless of how we're running the test – including if we're running individual tests, not part of any test suite at all.

ekluzek: I'd kind of like to add an optional field that gives the test name. I don't think I even care if it isn't read in now, but I think it helps to organize this for humans.

`<category>fates</category>`

I propose you just enter it for the non aux_clm tests. But, it wouldn't be bad to put for all of them. The comments technically also take care of this, but I like making it explicit in the XML.

So I'd say this is a preference. I think it will help in the future. But, it's not a must have.

billsacks: Okay, take a look at the commit I just pushed and let me know if that's what you were thinking. Are you happy with this other than that, or do you still want to look some more?

ekluzek: This is great. Thanks for doing that, as well as adding the header at the top.

billsacks:
@ekluzek thanks. Can you also please mark the cime pr (https://github.com/ESMCI/cime/pull/2962) as approved (or remove yourself as a reviewer)?

Regarding the cime hash: Yes. We'll wait to bring this ctsm pr to master until we can point to a tagged version of cime. Probably it makes sense to fold this pr into our next cime-update.

PR #609, opened 2019-01-09
Remove CLM4.0

### LINKED ISSUE #455: Remove CLM4.0
We'd like to remove support for CLM4.0 soon, largely in order to clean up build-namelist.

We are awaiting final approval from Stan Soloman, Dan Marsh and Hanli Liu that they won't need CLM4 for WACCM-X for any releases after CESM2.1.

billsacks:
### Description of changes

Remove CLM4.0. This consists of removing code and other files that are
specific to CLM4.0, and doing some resulting cleanup of code that was
shared between CLM4.0 and later versions (primarily in build-namelist).

### Specific notes

- Resolves #455 (Remove CLM4.0)

- No longer possible to run CLM40 cases

Testing performed, if any:
- Ran build-namelist tests
- Full system tests and tools tests in progress


billsacks:
@ekluzek a few notes:

This currently points to a hash on my cime branch. Before merging this to master, I'll change this to point to a cime tag (either by pointing to a cime tag on master, if possible, or pointing to a branch tag).

There are a lot of tedious, straightforward deletions and changes mixed in with a few more interesting changes. If you want to review this, it might be best if you review individual commits. I recommend reviewing:

- 7e77ad220 (Remove references to clm4 from various build scripts)

- 32067e287 (Remove CLM_CONFIG_OPTS env_build variable)

- f9e8a7117 (Remove CLM4.0 support from CLMBuildNamelist)
  - review this one without showing whitespace differences

- 675754963 (Remove numeric representation of physv)

- 789491dd0 (Remove clm4 build-namelist tests)

- 7117d6c67 (Replace call to configure with a function that writes config_cache.xml)

- 4e4c6d0e0 (Remove some settings in listDefaultNamelist that are no longer needed)
  - **This one could use particular review: I've never run this before so I don't know how to confirm that I'm not breaking anything (though I did do a bit of testing: ran `./listDefaultNamelist.pl -res 10x15` before and after 4e4c6d0e0)**

billsacks:
Here are results of various testing, once I cherry-picked cime commit ESMCI/cime@148bc2158 onto both the cime branch used in this PR and the cime version used in the baseline code:

### build-namelist tests

Pass (comparison against ctsm1.0.dev021)

### PTCLM (tools/PTCLM/test)

```
Total number of tests             = 52
Number of tests that PASS         = 51
Number of tests that Fail         = 1
Number of compare tests that PASS = 37
Number of compare tests that Fail = 4
Number of tests without compare   = 11
```

The one failing test is:

```
FAIL fail-no_towerdata.-d+/glade/p/cesmdata/cseg/inputdata++-s+RF-Bra+--debug+--pftgrid+--soilgrid.
```

I get the same failures when running these tests on ctsm1.0.dev022.

### System tests (aux_clm)

hobart: pass and bit-for-bit

cheyenne: pass and bit-for-bit except this one unexpected failure: `FAIL DAE_N2_D_Lh12.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-DA_multidrv SUBMIT`. This is a system issue related to the inputdata ftp server that Jim Edwards and Mark Moore are looking into.

### tools tests

ongoing; I'll update this when they're complete

billsacks:
I am now pointing to a cime tag (albeit a branch tag rather than a tag on master), so I feel this is ready once @ekluzek has a chance to look at it.

ekluzek:
I've started going through this. One thing I saw to ask about is the removal from passing $physv to subroutines in CLMBuildnamelist.pm. The reason that's done is that in practice these subroutines just need it to distinguish between clm4.0 and NOT clm4.0. However, in the future, $physv will be used to distinguish between more things NWP and HYD modes for example. And we know clm5.5 is coming. so I suggest you continue to pass them down even if it isn't needed now, because it could in the future.

billsacks:
> I've started going through this. One thing I saw to ask about is the removal from passing $physv to subroutines in CLMBuildnamelist.pm. The reason that's done is that in practice these subroutines just need it to distinguish between clm4.0 and NOT clm4.0. However, in the future, $physv will be used to distinguish between more things NWP and HYD modes for example. And we know clm5.5 is coming. so I suggest you continue to pass them down even if it isn't needed now, because it could in the future.

I disagree for two reasons:

(1) I am generally opposed to complicating an interface because "we might need it later". See https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it , and Martin Fowler's discussion of the "Remove Parameter" refactoring in his "Refactoring" book:

> Programmers often add parameters but are reluctant to remove them. After all, a spurious parameter doesn't cause any problems, and you might ned it again later.
>
> This is the demon _Obfuscatis_ speaking; purge him from your soul! A parameter indicates information that is neded; different values make a difference. Your caller has to worry about what values to pass. By not removing the parameter you are making further work for everyone who uses the method. That's not a good trade-off, especially because removing parameters is an easy refactoring.

It will be just as easy (or easier) to add this back in later, when it's actually needed.

(2) Even if this is needed, the string version is already accessible via other means.

For example:

```perl
  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults,
              $nl, 'rootstem_acc', 'phys'=>$nl_flags->{'phys'});
```

(The code is inconsistent about whether it gets the physics version from physv or from nl_flags for add_default calls.)

bld/CLMBuildNamelist.pm
```diff
@@ -2856,39 +2627,28 @@ sub setup_logic_hydrology_switches {
   if ( $h2osfcflag == 1 && $subgrid != 1 ) {
     $log->fatal_error("if h2osfcflag is ON, subgridflag can NOT be off!");
   }
-  # These should NOT be set for CLM5.0 and beyond
```
ekluzek: It's not clear to me if this should be removed. I thought of these as variables we'd be able to remove at some point. But, they might be useful for NWP. So perhaps this should just be removed. The if could be changed, to just allow it for clm4_5 (since the numeric representation was removed).

billsacks: I felt it would be a maintenance headache to hard-code specific physics versions these options are and are not allowed for. As you say, they might be useful for NWP, so we'd have to have an ever-growing list of specific versions these are and are not allowed for. We can possibly revisit this when we have a more robust way to get the general physics option ("clm", "nwp", etc.) and the version of that option. However, my opinion on this is that it's unnecessary work on our part, and it's good enough to document these in namelist_definition as being deprecated.

Issue #624, opened 2019-01-22
QSNOEVAP history output incorrect


olyson:
### Brief summary of bug

QSNOEVAP history output is incorrect.

### General bug information

**CTSM version you are using:** release-clm5.0.15

**Does this bug cause significantly incorrect results in the model's science?** Yes, if QSNOEVAP is used for analysis

**Configurations affected:** All

### Details of bug

In WaterfluxType.F90, QSNOEVAP is defined as:

    this%qflx_ev_snow_patch(begp:endp) = spval
    call hist_addfld1d (fname='QSNOEVAP', units='mm/s',  &
         avgflag='A', long_name='evaporation from snow', &
         ptr_patch=this%qflx_tran_veg_patch, set_lake=0._r8, c2l_scale_type='urbanf')

but last line should be:

         ptr_patch=this%qflx_ev_snow_patch, set_lake=0._r8, c2l_scale_type='urbanf')

ekluzek:
@olyson @dlawrenncar should this fix go into release-clm5.0 ASAP then?

ekluzek:
I have a branch with the fix for this. I've also done some looking for this same type of problem elsewhere where ptr_patch points to the same array. There are some legit examples of this, where there is an ice mask for it. And there are some where a different name is allowed for the same variable.

ekluzek:
So fixed on release branch will be fixed on master when those changes come in.

PR #625, opened 2019-01-23
Qsnoevapfix

ekluzek:
### Description of changes

QSNOEVAP for history was pointing to the wrong internal variable. This corrects it to point to the correct one.

### Specific notes

CTSM Issues Fixed (include github issue #): #624

Are answers expected to change (and if so in what way)? Yes (just for QSNOEVAP)

Testing performed, if any:
 Ran SMS_Lm1_D.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-output_crop_highfreq
and made sure only the QSNOEVAP variable changed on history


billsacks:
I think you set this up to merge into master when you meant for the base to be the release branch. (You can change this via the 'edit' button near the top of the page.)

ekluzek:
@dlawrenncar verified that this doesn't have to be imminent, but can be combined with other answer changing things. So I'm going a head with that plan.

Issue #619, opened 2019-01-17
Begin adding a "contrib" tools directory for important useful scripts/tools with looser standards


ekluzek:
We've had a few discussions about this now. There's a set of a few tools that are on cheyenne under:

/gpfs/fs1/p/cgd/tss/CLM50_data_tools

We want to bring them into github, currently into the standard CTSM repository. But, there will be looser restrictions on the standards for these tools. They won't be supported and only provided on a "user-beware" basis. They may not be tested or even have tests for them. And as such they may not work with the current version of CTSM. Pull Requests (PR's) for changes in this area will likely be approved without review or testing. NetCDF files for these tools should also be provided outside of CTSM (in inputdata). Specialized development of the tools won't be done in the main repository. Users can make particular changes on their own. Changes that may be important for the majority of the community could be done and added as well as bug fixes. The main developers for this area will be: @olyson and @swensosc.

PR #626, opened 2019-01-24
Initial add of a contributor tool directory for various scripts

ekluzek:
### Description of changes

Initial add of a contributor tool directory for various scripts to be used fo pre and post processing, as well as case management of CTSM.

### Specific notes

Contributors other than yourself, if any: @swensosc @olyson @wwieder 

CTSM Issues Fixed (include github issue #): #619


olyson:
There is an updated version of the SpinupStability script here:

/gpfs/u/home/oleson/lnd_diag/run/SpinupStability_v8.ncl

An updated version of the run_historical script here:

/gpfs/u/home/oleson/run_hist_1850_files/BGC/run_clm_historical.v7.csh

and an updated version of the run_clmtowers script here:

/gpfs/fs1/work/oleson/release-clm5.0.12/tools/PTCLM/run_clmtowers.csh

ekluzek:
Some of the scripts have usernames hardwired in, and personal paths as well. The tower script points to data that @olyson has so in cases like that we'll need to leave that in. But, in other cases it should create data and expect data according to the user's name rather than hardcoded to a specific individual.

Also we want it setup to run the Tutorial cases out of the box, without having to make changes to it. We still want to educate users on how you make those changes, but don't want them to HAVE to make changes to get something to work.

They might also be hardwired to only work on cheyenne. That's probably OK for now. But, that could be an issue for general use.

tools/contrib/singlept
```diff
@@ -0,0 +1,329 @@
...
+#  Import libraries
+import sys
+import os
+from getpass import getuser
+import string
+import subprocess
+import numpy as np
+import xarray as xr
+
+def mprint(mstr):
```
billsacks: Instead of this, use

```
from __future__ import print_function
```

Then just use

```
print(...)
```

ekluzek:
OK, I made some changes to get it working for me. And added some more printing about what it's doing. I'm going to approve this as it is for now. But, we will likely continue working on it.

PR #632, opened 2019-02-04
update default fates parameter file

rgknox:
### Description of changes

This changeset updates the default fates parameter file for testing.  Changes since last change are most notably, the addition of the leaf age dimension.  This is a minor API compatibility change.

### Specific notes

See corresponding NGEET/fates PR: https://github.com/NGEET/fates/pull/462

@xuchongang

FATES simulations are expected to have small differences because of subtle changes in the default parameter configuration. FATES is also expected to have roundoff/machine-precision differences when parameters are kept the same because of changes in order of operations.

Testing performed, if any:

TBD

### DIFF
--- a/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -355,7 +355,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- FATES default parameter file                                       -->
 <!-- ================================================================== -->
 
-<fates_paramfile>lnd/clm2/paramdata/fates_params_default_2trop.c190114.nc</fates_paramfile>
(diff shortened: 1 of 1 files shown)


rgknox:
Tests pass,
/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.leafage-2trop-merged-v2.0-C366dc26-F1191c88

@ekluzek , I have not been updating the tag in the externals config file for non-major API updates (ie default parameter file updates and not changes in the interface), but I can if requested.

ekluzek:
@rgknox yeah non API changes don't have to be put in place here. An individual user could do that on their own. The API changing updates are the ones that really need to come in here. I think the reason to include a non-API change though is it's something that's important for users to start using that change. The best way to indicate that is to do it as an update. So you should make your best judgement in terms of "should everyone be using this new version of FATES at this point". You are the best judge for that. So it's up to you.

PR #635, opened 2019-02-13
remove flushing flags for fates history variables

rgknox:
### Description of changes

Fates will internally set the parts of this history output arrays that are not under its control to either zero or ignore flags.   The zeroing and flagging in the interface call was redundant, and was actually making it impossible to output certain variables (like flags) that are only relevant on fates columns.

As a long-term goal, it would be potentially better for FATES to not handle all of the flushing and zero'ing of its buffers, but given the system we currently have, its best not to also do this via the interface.

### Specific notes

This change should not effect non-fates runs in any way.  It also should not effect fates runs, as I believe we were flushing everything to zero up until now, which was consistent with the flushing flag set in the interface call.

Testing performed, if any:

TBD

### DIFF
--- a/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -355,7 +355,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- FATES default parameter file                                       -->
 <!-- ================================================================== -->
 
-<fates_paramfile>lnd/clm2/paramdata/fates_params_default_2trop.c190205.nc</fates_paramfile>
+<fates_paramfile>/glade/u/home/rgknox/ctsm/src/fates/parameter_files/fates_params_default_14pft.c190221.nc</fates_paramfile>
+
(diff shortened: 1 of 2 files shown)


rgknox:
running this through the test mill now.
To answer your question @ekluzek , I think that depends on the variables that are represented in our tests.  These changes should not change the model state, so the coupler output should be bit-for-bit.  But there may be some variables that we on the fates side are flagging to be set as ignore, and those values will now be correctly omitting this forced zero on non-fates columns.

rgknox:
Tests are b4b with baselines:

/gpfs/fs1/scratch/rgknox/clmed-tests/fates-sci.1.26.3_api.7.2.0-no-zeros-C7ce9619-Fd2f0907.fates.cheyenne.intel

However, I'm noticing that some of the tests have increased memory usage by >10 percent.  So I am guessing that by passing those flags, array sizes are reduced in the history IO machinery.

Lets hold on this, and I will check in with the FATES team.

rgknox:
This PR is now redundant with #766, closing

PR #641, opened 2019-02-23
Ability to remove individual landunits for faster runs

### LINKED ISSUE #581: Allow zeroing out special landunits if their area is below some threshold
Connected with #457 - particularly for the NWP configuration, we want to allow using higher thresholds for whether special landunits are present. For example, maybe lake is only present if it takes up 10% or more of the grid cell? We may want these thresholds to be selectable via the mksurfdata_map namelist (currently I think they're hardcoded in mksurfdata_map), or – maybe better – we may want this selectable at runtime, as we're planning for #457 .

slevisconsulting:
### Description of changes
Introduce user-defined namelist parameters:
toosmall_* (soil, crop, glacier, lake, wetland, and urban) 
Use as thresholds above which landunit areas are kept and below which landunit areas are removed

### Specific notes

@billsacks @ekluzek 

Fixes #581

Only when user sets above-listed parameters to values > 0

New namelist parameters listed above

Testing performed, if any:
This far I have performed only this new test:
nohup qcmd -- ./create_test ERS_D_Ld10.f10_f10_musgs.IHistClm50BgcCropGs.cheyenne_intel.clm-rm_individual_lunits_decStart &


slevisconsulting:
I will create unit tests next.

slevisconsulting:
This branch is also open for review now.

I'm running the cheyenne test suite on this branch and on the n_dom_landunits branch.

slevisconsulting:
Ready for code review and for final steps in preparation for the merge. Pls let me know if this one can get in line as dev030 or if it needs to wait.

slevisconsulting:
Running std testing on cheyenne and hobart.

slevisconsulting:
Cheyenne OK.

Hobart still shows PEND on many tests after I submitted this morning. I do not know how to check whether they are in a queue somewhere or whether I need to restart them. At this point I will restart, but pls let me know how to check. Thank you!

slevisconsulting:
Aha, I tried qstat and it worked on hobart. I didn't realize :-)

bld/CLMBuildNamelist.pm
```diff
@@ -2316,12 +2316,25 @@ sub setup_logic_do_transient_pfts {
...
+   my $toosmall_urb_md = $nl->get_value( 'toosmall_urb_md' );
+
    if (string_is_undef_or_empty($nl->get_value('flanduse_timeseries'))) {
       $cannot_be_true = "$var can only be set to true when running a transient case (flanduse_timeseries non-blank)";
    } elsif (&value_is_true($nl->get_value('use_cndv'))) {
       $cannot_be_true = "$var cannot be combined with use_cndv";
    } elsif (&value_is_true($nl->get_value('use_fates'))) {
       $cannot_be_true = "$var cannot be combined with use_fates";
+   } elsif ($n_dom_pfts > 0 || $n_dom_landunits > 0 || $toosmall_soil > 0 || $toosmall_crop > 0 || $toosmall_glacier > 0 || $toosmall_lake > 0 || $toosmall_wetland > 0 || $toosmall_urb_tbd > 0 || $toosmall_urb_hd > 0 || $toosmall_urb_md > 0) {
+      $cannot_be_true = "$var cannot be combined with any of the of the following > 0: n_dom_pfts > 0, n_dom_landunit > 0, toosmall_soi > 0, toosmall_crop > 0, toosmall_glacier > 0, toosmall_lake > 0, toosmall_wetland > 0, toosmall_urb_tbd > 0, toosmall_urb_hd > 0, toosmall_urb_md > 0";
```
billsacks: Here and below: this should not be done with the cannot_be_true logic: This logic ends up saying: if the user sets n_dom_pfts > 0, then that will silently turn off transient pfts. That is not what we want here. Instead, this error check should be done more similarly to how other error checking is done in build-namelist, not via this cannot_be_true mechanism.

src/main/glc2lndMod.F90
```diff
@@ -293,6 +299,23 @@ subroutine set_glc2lnd_fields(this, bounds, glc_present, x2l, &
        end if
     end if
 
+    if (glc_do_dynglacier) then
```
billsacks: You put this in a routine that is called every time in the run loop. Can you move this to initialization so that it is only done once?

slevisconsulting:
@billsacks do you recommend that I rerun standard testing? If another tag will come in before mine, then I'll rerun standard testing no matter what.

bld/CLMBuildNamelist.pm
```diff
@@ -2324,6 +2335,12 @@ sub setup_logic_do_transient_pfts {
       $cannot_be_true = "$var cannot be combined with use_fates";
    }
 
+   if (&value_is_true($nl->get_value($var))) {
```
billsacks: Here and the similar new code below: I think this is closer to what's wanted, but I think you do this check too early: at a glance, I think you need this *after* the add_default call a few lines down. Probably the best place for it is at the end of the function.

slevisconsulting: Yes, I see why now. If the functions will switch the do_transient_* options to .false. then the new error checks can pass without aborting and causing the run to fail for no good reason.

Thanks @billsacks

slevisconsulting: And, yes, I agree about waiting to rerun testing until at least a complete review comes through.

bld/CLMBuildNamelist.pm
```diff
@@ -2343,6 +2354,12 @@ sub setup_logic_do_transient_pfts {
    }
 }
 
+   if (&value_is_true($nl->get_value($var))) {
```
billsacks: It looks like now this block is too low - after the end of the function. It should be before the ending curly brace that is currently on line 2355.

slevisconsulting: Oops, thanks for catching that! It's prob. clear from this back and forth that I'm not experienced with this scripting language (what language is it btw?). Though my detail oriented self should have noticed that ending curly brace...

slevisconsulting:
Started standard testing.

slevisconsulting:
Standard testing on
cheyenne: OK
hobart: OK
i.e. only expected failures FAIL

slevisconsulting:
@ekluzek I saw that there was a new tag made today. I will wait to repeat standard testing after your review. Does that make sense?

bld/CLMBuildNamelist.pm
```diff
@@ -2316,6 +2316,17 @@ sub setup_logic_do_transient_pfts {
...
    my $cannot_be_true = "";
 
+   my $n_dom_pfts = $nl->get_value( 'n_dom_pfts' );
+   my $n_dom_landunits = $nl->get_value( 'n_dom_landunits' );
+   my $toosmall_soil = $nl->get_value( 'toosmall_soil' );
+   my $toosmall_crop = $nl->get_value( 'toosmall_crop' );
+   my $toosmall_glacier = $nl->get_value( 'toosmall_glacier' );
+   my $toosmall_lake = $nl->get_value( 'toosmall_lake' );
+   my $toosmall_wetland = $nl->get_value( 'toosmall_wetland' );
+   my $toosmall_urb_tbd = $nl->get_value( 'toosmall_urb_tbd' );
```
ekluzek: @slevisconsulting a question that comes to my mind for urban is if people would really want to break it out into the three different landunit types. This may be the right way to do it for flexibility, and just the natural way that the code works. But, I'd like to check with people about what might be desired. @barlage and @negin513 can you maybe comment here? I wonder if the desired behavior might more be toward turning urban off completely, or off if the total urban is more than some total amount? Would there be a use for tweaking the three different types differently? I'm not sure, so it's worth talking to some people who would be using this feature.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -1085,6 +1085,62 @@ Number of dominant landunits, so this determines the number of active landunits.
 <default>Default: 0</default>
 </entry>
 
+<entry id="toosmall_soil" type="integer" category="physics"
+       group="clm_inparm"
+       valid_values="0,1,2,3,4,5,6,7,8,9,10" value="0">
+Threshold above which the model keeps the soil landunit. Selecting the value 0 means DO NOTHING. Valid values reach 100 but practical values are not expected above 10 at this time.
```
ekluzek: What is the reason for limiting from 0 to 10? I suppose it makes sense to limit to integers, rather than real, but that's artificial as well.

bld/CLMBuildNamelist.pm
```diff
@@ -2397,6 +2425,12 @@ sub setup_logic_do_transient_crops {
       $log->fatal_error($cannot_be_true);
    }
 
+   if (&value_is_true($nl->get_value($var))) {
+      if ($n_dom_pfts > 0 || $n_dom_landunits > 0 || $toosmall_soil > 0 || $toosmall_crop > 0 || $toosmall_glacier > 0 || $toosmall_lake > 0 || $toosmall_wetland > 0 || $toosmall_urb_tbd > 0 || $toosmall_urb_hd > 0 || $toosmall_urb_md > 0) {
```
ekluzek: What is trying to limited here? I'm pretty sure having n_dom_pfts and n_dom_landunits together is fine. And I think any of the toosmall's with n_dom_pfts should be fine (other than maybe toosmall_soil). I see the point of checking the toosmalls with n_dom_landunits. I'm not sure if there are problems with the other conditions though. So we should clarify what we are trying to check here.

ekluzek:
@slevisconsulting I pushed a change to the build-namelist tester to check that the toosmalls weren't being used with n_dom_landunit. It's currently doing the wrong thing there, so it needs some work. But, you can run the tester to show that things are OK. Other's have complained that it's hard to work with, so I can help you with it. I think just using it shouldn't be a problem though.

src/main/controlMod.F90
```diff
@@ -259,6 +259,17 @@ subroutine control_init( )
     namelist /clm_inparm/ n_dom_pfts
     namelist /clm_inparm/ n_dom_landunits
 
+    ! Thresholds above which the model keeps the soil, crop, glacier, lake,
+    ! wetland, and urban landunits
+    namelist /clm_inparm/ toosmall_soil
```
ekluzek: This is a nitpick, but you don't have to have each of these on their own line. They could be comma delimited on a line. For grepability it might be good to not NOT & to extend. IF you prefer this syntax, that's fine, I mostly want to check so you know it's not required.

bld/namelist_files/namelist_definition_ctsm.xml
ekluzek: Somewhere in here and in the code, I think it would be good to say that this is a percentage of the landunits in the gridcell. It gives the range as 0 to 100, which implies it's a percentage, but it doesn't explicitly say that. It would be helpful to have the documention talk about it being a percentage.

src/main/surfrdUtilsMod.F90
```diff
@@ -145,6 +146,86 @@ subroutine convert_cft_to_pft( begg, endg, cftsize, wt_cft )
...
+    real(r8) :: residual(max_lunit)  ! Array of wt_lunit residuals (fraction)
+    !-----------------------------------------------------------------------
+
+    SHR_ASSERT_ALL((ubound(wt_lunit) == (/endg, max_lunit/)), errMsg(sourcefile, __LINE__))
+
+    toosmall_any = toosmall_soil + toosmall_crop + toosmall_glacier + &
+                   toosmall_lake + toosmall_wetland + toosmall_urb_tbd + &
+                   toosmall_urb_hd + toosmall_urb_md
+
+    if (toosmall_any > 0) then
```
ekluzek: There should also be a check that toosmall_any isn't > 100 right? The namelist puts a limit on that, but that can change, so the code should check the actual issue in case we change the namelist.

bld/CLMBuildNamelist.pm
ekluzek: @slevisconsulting do you have a way to debug the build-namelist changes? You can run build-namelist standalone, so that can help to work on it. Here's something that I do. Setup a case, and then run "preview_namelist -debug" and go back through and get the call to CLM build-namelist from it. You can then mouse it and run it standalone, and change the arguments to do the tests you want to do. The other thing to do is to run it with "perl -d" so it runs in the perl debugger, so you can "s" (step into) "n" (next over) or "c #" (continue to a given line number) to see what's happening.

bld/CLMBuildNamelist.pm
barlage: @ekluzek it may be that I just don't understand the nomenclature, but I'm not quite sure what you are asking.

ekluzek: @barlage Urban is divided into three different types: tall building district, high density, and medium density. What I'm thinking is that you probably aren't going to want to treat the three types differently. And more than that, since you don't care about the different types of urban, you probably really want to control it based on the total amount of urban in the grid-cell -- regardless of type. Having it broken up into the three types doesn't allow you to have control over total urban in the grid cell. So I'm thinking that having the namelist control item be toosmall_urban (rather than toosmall_urb_(tbd,hd,md)) would be more useful for users. And toosmall_urban is probably more obvious as to it's meaning.

src/main/surfrdUtilsMod.F90
ekluzek: OK, I realized this was wrong. There isn't an upper limit on toosmall_any. Each of the components can only be up to 100, But, one grid cell could be 100 urban and another 100 soil and that's fine.

bld/CLMBuildNamelist.pm
barlage: @ekluzek OK, I get it. I do like the added flexibility to say e.g., if the total urban (tbd+hd+md) is below some threshold, don't use it; if it is above, do something else (like use dominant of tbd, hd, md). I can tell you that currently in WRF, in "default" mode, only the dominant of all types (including the tbd, hd, md equivalent) is used to define the grid (to complicate things, if it is one of these "urban" types, the non-urban fraction will be a vegetation class...). This dominant approach can underestimate urban cover and was one of the main motivations for adding a "mosaic" option to the Noah LSM in WRF. When using this option, the n most dominant vegetation classes are used, still regardless of similar typing (like total urban, total forests, etc.).

src/main/surfrdUtilsMod.F90
```diff
@@ -145,6 +146,86 @@ subroutine convert_cft_to_pft( begg, endg, cftsize, wt_cft )
...
+       do g = begg, endg
+          residual = 0._r8  ! initialize
+          do m = 1, max_lunit
+             ! Remove landunits that are too small
+             if (wt_lunit(g,m) > 0._r8 .and. wt_lunit(g,m) <= toosmall(m)) then
+                residual(m) = wt_lunit(g,m)
+                wt_lunit(g,m) = 0._r8
+             end if
+          end do
+          ! If all landunits got removed, go back and keep the largest landunit
```
ekluzek: OK, this might need to be documented better. I think it makes sense, you have to fill the grid cell with at least one landunit, so it should be the biggest one. The user might want to be aware that could be the case though. I'm also wondering if this should have a warning if it's triggered, but I'm not sure.

Along the same lines I was wondering if you should be able to set one of the toosmalls to something over 100%, which would essentially trigger "I don't want to have ANY of the given landunit anywhere on my grid". If there's a desire to treat it that way, you would want different behavior in this case.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -43,6 +43,114 @@ contains
...
+
+     ! Set relevant pftcon values to defaults; override where necessary
+     call pftcon%InitForTesting()
+     allocate( wt_lunit(begg:endg,istsoil:max_lunit) )
+     allocate( wt_expected(begg:endg,istsoil:max_lunit) )
+     allocate( wt_in_out(begg:endg,istsoil:max_lunit) )
+
+     ! INPUT VALUES
+     wt_lunit(begg:,:) = 0._r8  ! initialize
+     wt_lunit(begg:,istsoil) = (/ input_value(1), input_value(8), input_value(4)/)
```
ekluzek: It's hard to make sense of the array indices here. I don't know why index 1, 8, and 4 are chosen here.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -43,6 +43,114 @@ contains
...
+     wt_lunit(begg:,isturb_hd) = (/ input_value(7),  input_value(2), input_value(2)/)
+     wt_lunit(begg:,isturb_md) = (/ input_value(8),  input_value(1), input_value(3)/)
+     wt_lunit(:,:) = wt_lunit(:,:) / 100._r8
+     call check_sums_equal_1( wt_lunit, begg, "test_check_sums_add_to_1", &
+                              "should not trigger an error")
+
+     do test = 1, tests
+
+        ! OUTPUT VALUES EXPECTED
+        if (test == 1) then
```
ekluzek: Little more documentation on what the different tests are would be good here (and possibly making the hardcoded test numbers into parameters with a meaning could be helpful).

bld/CLMBuildNamelist.pm
slevisconsulting: @barlage let me say it a bit differently:
- The way I wrote the code, you can set 3 "toosmall_urb" namelist parameters to decide whether one urban type is more important to you than another. So you may keep the tbd (tall building districts) and remove the hd and md, or vice versa, or keep no urban, or keep all urban, or any combination you choose.
- The suggested alternative simplifies this to one "toosmall_urban" namelist parameter, but you end up with all three urban landunits or lose all three.

I am happy to go with the alternative if you prefer it. I just wanted to make sure.

bld/namelist_files/namelist_definition_ctsm.xml
slevisconsulting: I added the word "Percentage" in the front of the comments for each of these namelist parameters. I added similar reference to the corresponding comment in surfrdMod.F90.

My comments here state:
"Valid values reach 100 but practical values are not expected above 10 at this time."
However, would it work if I changed what I have to valid_values="0 to 100"? I don't know how valid_values is used and whether that would still work @ekluzek 

Regarding reals vs. integers, do you prefer that I change these to reals?

bld/CLMBuildNamelist.pm
slevisconsulting: Added comments:
`if do_transient_pfts is .true. and any of these (n_dom_* or toosmall_*) are > 0, then give fatal error`
`if do_transient_crops is .true. and any of these (n_dom_* or toosmall_*) are > 0, then give fatal error`

bld/unit_testers/build-namelist_test.pl
```diff
@@ -138,7 +138,7 @@ sub make_config_cache {
 #
 # Figure out number of tests that will run
 #
-my $ntests = 914;
+my $ntests = 924;
```
slevisconsulting: @ekluzek I will try it; however, I disagree that the "toosmalls" may not be used with n_dom_pfts and n_dom_landunits. Here's an example: a user may wish to run with 2 dominant pfts and 2 dominant landunits and may still want to specify that a particular landunit shall be used if greater than 10%. As written, the code can do that.

> @slevisconsulting I pushed a change to the build-namelist tester to check that the toosmalls weren't being used with n_dom_landunit. It's currently doing the wrong thing there, so it needs some work. But, you can run the tester to show that things are OK. Other's have complained that it's hard to work with, so I can help you with it. I think just using it shouldn't be a problem though.

src/main/controlMod.F90
slevisconsulting: No strong preference, so I extended.

src/main/surfrdUtilsMod.F90
slevisconsulting: @ekluzek I like the idea of a warning, though a user could end up with very many warnings if this happened in a large number of grid cells. I will include this warning:
`if ( masterproc ) then`
`write(iulog,*) 'WARNING: The values of namelist parameters toosmall_* have resulted in the removal of all landunits in grid cell g = ', g, '; however, the model requires at least one landunit per grid cell, so the model has put back in this grid cell the largest landunit.'`
`end if`

Currently 100% means NONE of this landunit will appear UNLESS the model needs it in an empty grid cell. You're suggesting that we add a > 100% option to ensure that a landunit never appears. I feel skeptical about pursuing this for a couple of reasons:
1) A grid cell in some dataset may really have just one landunit and it could be the same landunit that the user chose to eliminate.
2) The toosmall_* parameters make sense as 0-100%, while >100% makes them potentially unintuitive, confusing, and prone to error.
3) If a user really wants a particular landunit entirely eliminated (say wetland), they can set toosmall_wetland = 100, the other toosmall_* = 0, and n_dom_landunits = some small number. This way wetland will only appear if all else is 0.

What do you think?

slevisconsulting: ...removing
`if ( masterproc )`
from the warning above...

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
slevisconsulting: Added comment in the code to explain.

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
slevisconsulting: Done.

bld/namelist_files/namelist_definition_ctsm.xml
slevisconsulting: Something that we could start doing is including the units in all parameter and variable names. For example these would become toosmall_*_percent

What do you think?

bld/unit_testers/build-namelist_test.pl
slevisconsulting: @ekluzek I will wait to hear from you on this before I `git pull` your changes.

bld/CLMBuildNamelist.pm
slevisconsulting: Thank you, @ekluzek , this helps.

bld/unit_testers/build-namelist_test.pl
slevisconsulting: @ekluzek does it work if I `git commit` and `git push` my revisions so that you may take a look, and then later I `git pull` your changes? Or will that order mess things up?

ekluzek: The commits and push that I did just changed bld/unit_testers/build-namelist_test.pl, so it won't interfere with anything you are doing outside of that. You will need to pull those changes before you push anything else though.

slevisconsulting: Ran `./build-namelist_test.pl`
and got
`Successfully ran all testing for build-namelist`
so I think we're good, other than still examining whether the `toosmall_*` parameters should coexist with the `n_dom_*` parameters. I vote YES, as explained earlier in this thread.

bld/CLMBuildNamelist.pm
ekluzek: I see how n_dom_landunits, could interfere with the too-smalls, but I don't see how n_dom_pfts could interfere. Is there really a problem with setting both n_dom_pfts and the too-smalls?

bld/CLMBuildNamelist.pm
slevisconsulting: Just to be clear, I will not change the implementation until I get the go-ahead that people prefer the alternative.

bld/CLMBuildNamelist.pm
slevisconsulting: Pls note that these comments say nothing of the n_doms interfering with the toosmalls. Instead they address the incompatibility with transient: The n_doms and the toosmalls are incompatible with **transient**.

@ekluzek pls explain what you mean by "n_dom_landunits could interfere with the toosmalls."

This was my response to the same concern elsewhere in the PR:
"I disagree that the toosmalls may not be used with n_dom_pfts and n_dom_landunits. Here's an example: A user may wish to run with 2 dominant pfts and 2 dominant landunits and may still want to specify that a particular landunit shall be used if greater than 10%. As written, the code can do that."

slevisconsulting: @ekluzek let me know if it would be helpful to go over some of this in person by phone. If so, feel free to book a time on my google calendar.

bld/namelist_files/namelist_definition_ctsm.xml
slevisconsulting: Now I know that setting `valid_values="0 to 100"` would not work, so I will write out every integer from 0 to 100.

ekluzek: @slevisconsulting we talked about this in our CLM software meeting, and actually we don't think you should limit it to integers for the namelist input. We think you should allow for double precision floating point input between 0.0 and 100.0.

We aren't sure that it's that useful to input say 9.5 rather than 9, but didn't see why to limit to integers.

slevisconsulting: Ok, I will just replace "integer" with "real" from what I see elsewhere in the file and will remove `valid_values`.

bld/CLMBuildNamelist.pm
slevisconsulting: @ekluzek , @negin513 , @barlage , and I decided that I will modify to the following:

Order of new subroutine calls at the end of subroutine surfrd_get_data:
1) Newest call will happen first: collapse_urban = .true. or .false.  If true, this will put the three urban landunit areas into the dominant urban landunit.
2) Recently added call using n_dom_landunits will not change
3) Recently added call using too_small_* will be modified so that a single too_small_urban value (namelist controlled) will apply to each urban landunit that's present.

bld/namelist_files/namelist_definition_ctsm.xml
slevisconsulting: ...and will modify elsewhere in the code accordingly.

slevisconsulting: Committed change of the toosmalls from integer to real.

bld/CLMBuildNamelist.pm
slevisconsulting: From above list...
Step 3: complete in b3dcad9
Step 2: no change
Step 1: I will work on next

slevisconsulting: Step 1 complete in 99645bb

slevisconsulting:
@ekluzek , this is ready for a new review.

slevisconsulting:
In #663 Bill suggested making the namelist defaults appear in the namelist even when the user didn't set them explicitly. I will go ahead and do that in this PR, as well.

billsacks:
> In #663 Bill suggested making the namelist defaults appear in the namelist even when the user didn't set them explicitly. I will go ahead and do that in this PR, as well.

Thanks, @slevisconsulting 

@ekluzek I think that's what we have agreed on in general, right?

ekluzek:
I'm fine with the standard of having defaults in the namelist. We are inconsistent with that in CTSM right now. So things are done both ways. I do see the point with being able to figure out defaults from the namelist, rather than search the source. But, right now you do have to search both.

In CTSM we should confirm that this is a goal we are trying to reach if that is indeed the case.

billsacks:
I thought I remembered this being discussed and decided upon at some point, but can't find a definitive written record of that. What I can find:

From a June 8, 2015 discussion at a CLM-CMT meeting:

> Should we give namelist items reasonable hard-coded defaults, or initialize them to garbage values and then ensure that they are set explicitly in the namelist?
>
> People agree that, at least for "important" namelist items, it makes sense to move towards initializing them to garbage values in the code, and ensuring that they are on the namelist.

From a November 14, 2016 CSEG meeting (referring to CESM as a whole):

> Mariana has worked hard to make sure more fields appear in the namelist – explicitly listing fields.

And I seem to remember Cecile in particular requesting that namelists be more explicit because it really helps her (but I may be remembering wrong).

slevisconsulting:
Test ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_individual_lunits  PASSes

dlawrenncar:
I agree that it would be preferable for everything to appear in namelist.

On Mar 21, 2019, at 1:08 PM, Bill Sacks <notifications@github.com> wrote:

I thought I remembered this being discussed and decided upon at some point,
but can't find a definitive written record of that. What I can find:

From a June 8, 2015 discussion at a CLM-CMT meeting:

Should we give namelist items reasonable hard-coded defaults, or initialize
them to garbage values and then ensure that they are set explicitly in the
namelist?

People agree that, at least for "important" namelist items, it makes sense
to move towards initializing them to garbage values in the code, and
ensuring that they are on the namelist.

From a November 14, 2016 CSEG meeting (referring to CESM as a whole):

Mariana has worked hard to make sure more fields appear in the namelist –
explicitly listing fields.

And I seem to remember Cecile in particular requesting that namelists be
more explicit because it really helps her (but I may be remembering wrong).

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ESCOMP/ctsm/pull/641#issuecomment-475364596>, or mute
the thread

ekluzek:
@dlawrenncar just to absolutely clarify. When we add new namelist items should we always make sure we add the default value to the namelist? And when we change things in the namelist, to try to head that direction as well? I suppose there still might be really odd optional things that we wouldn't want to do this for, but if we make it the standard, you'd then have to think about why it shouldn't be in the namelist. But, that should be the rare exception, if this is what we want to make the standard.

And is there agreement among CLM and CTSM scientists that this is the direction we should head? Might be something to make sure we have agreement on. Or at least agreement that we should move that direction.

Again, I think this makes sense. But, we need to clarify it as a goal and make sure everyone is aware of it if we want it to happen.

dlawrenncar:
Let’s discuss at forthcoming meeting. I don’t see many downsides at this
point, other than a potentially very large namelist, but maybe there is
something I’m not aware of.

On Mar 21, 2019, at 10:29 PM, Erik Kluzek <notifications@github.com> wrote:

@dlawrenncar <https://github.com/dlawrenncar> just to absolutely clarify.
When we add new namelist items should we always make sure we add the
default value to the namelist? And when we change things in the namelist,
to try to head that direction as well? I suppose there still might be
really odd optional things that we wouldn't want to do this for, but if we
make it the standard, you'd then have to think about why it shouldn't be in
the namelist. But, that should be the rare exception, if this is what we
want to make the standard.

And is there agreement among CLM and CTSM scientists that this is the
direction we should head? Might be something to make sure we have agreement
on. Or at least agreement that we should move that direction.

Again, I think this makes sense. But, we need to clarify it as a goal and
make sure everyone is aware of it if we want it to happen.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/ESCOMP/ctsm/pull/641#issuecomment-475427274>, or mute
the thread

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
```diff
@@ -43,6 +43,195 @@ contains
...
+     deallocate( wt_expected )
+     deallocate( wt_in_out )
+     deallocate( wt_lunit )
+
+     call pftcon%clean()
+
+  end subroutine test_collapse_individual_lunits
+
+  @Test
+  subroutine test_collapse_urban()
```
ekluzek: Add another test subroutine, that tests using the n_dom_* and the too_small_* all together at once.

src/main/surfrdUtilsMod.F90
```diff
@@ -219,7 +302,7 @@ subroutine collapse_to_dominant(weight, lower_bound, upper_bound, begg, endg, n_
        end do
 
        ! Error checks
-       call check_sums_equal_1(weight, begg, 'weight', subname)
+       if (wt_sum == 1._r8) call check_sums_equal_1(weight, begg, 'weight', subname)
```
ekluzek: Why is this only being done when wt_sum==1?

src/main/surfrdUtilsMod.F90
```diff
@@ -145,6 +146,83 @@ subroutine convert_cft_to_pft( begg, endg, cftsize, wt_cft )
...
+             if (wt_lunit(g,m) > 0._r8 .and. wt_lunit(g,m) <= toosmall(m)) then
+                residual(m) = wt_lunit(g,m)
+                wt_lunit(g,m) = 0._r8
+             end if
+          end do
+          ! If all landunits got removed, go back and keep the largest landunit
+          if (sum(wt_lunit(g,:)) == 0._r8) then
+             max_landunit = maxloc(residual, 1)
+             wt_lunit(g,max_landunit) = residual(max_landunit)
+             write(iulog,*) 'WARNING: The values of namelist parameters toosmall_* have resulted in the removal of all landunits in grid cell g = ', g, '. The model requires at least one landunit per grid cell, so the model has put back in this grid cell the largest landunit.'
```
ekluzek: I think we should check with Mike here on if this should be a warning or an error. @barlage if the use of too_small_* and n_dom_* results in the removal of all land-units -- should the model print a warning and add back the largest land-unit -- or should it die? If it dies the user could then go back and adjust their values in order to get something that does work. This way they know for sure what's happening. Whereas if it's a warning, it could scroll past and they don't see it. Either way is valid, it just depends on how you want it to work.

Note, also @negin513

bld/CLMBuildNamelist.pm
```diff
@@ -2341,6 +2359,17 @@ sub setup_logic_do_transient_pfts {
...
       $log->fatal_error($cannot_be_true);
    }
+
+   # if do_transient_pfts is .true. and any of these (n_dom_* or toosmall_*)
+   # are > 0 or collapse_urban = .true., then give fatal error
+   if (&value_is_true($nl->get_value($var))) {
+      if (&value_is_true($nl->get_value('collapse_urban'))) {
+         $log->fatal_error("$var cannot be combined with collapse_urban");
+      }
+      if ($n_dom_pfts > 0 || $n_dom_landunits > 0 || $toosmall_soil > 0 || $toosmall_crop > 0 || $toosmall_glacier > 0 || $toosmall_lake > 0 || $toosmall_wetland > 0 || $toosmall_urban > 0) {
```
ekluzek: Remove this restriction, since we envision using n_dom_* and toosmall_*'s together.

cime_config/testdefs/testmods_dirs/clm/rm_individual_lunits/user_nl_clm
```diff
@@ -0,0 +1,6 @@
+toosmall_soil = 5.0
```
ekluzek: I think this same test should also set n_dom_landunit and n_dom_pft so we have a system test that uses all of them together. And you might as well also test collapse_urban in this system test at the same time. This might mean the name of the test directory should be expanded.

bld/CLMBuildNamelist.pm
slevisconsulting: @ekluzek , I'm pretty sure this is fine. The logic translates to the following:
- If do_transient_pfts is .true. and ANY of these (n_dom_* or toosmall_*) are > 0  OR collapse_urban = .true., then give fatal error
- Same if do_transient_crops is .true.

slevisconsulting: It's not excluding coexistence of n_dom_* and toosmall_*.
It's excluding coexistence of those with *transient*.

ekluzek: Ahh, OK. Sorry I misread that. I thought there was another place where that was done.

src/main/surfrdUtilsMod.F90
slevisconsulting: @ekluzek , I'm adding this comment to the code:

Check that weight sums to 1 only if weight summed to 1 upon entering the subroutine. If weight did not sum to 1 upon entering the subroutine, then weight will not sum to 1 now, either. Currently the latter occurs when collapsing the urban landunits to the dominant urban landunit.

cime_config/testdefs/testmods_dirs/clm/rm_individual_lunits/user_nl_clm
slevisconsulting: Ok, combining all these into one test.

slevisconsulting: Good call! Triggers an error now. Looking into it.

slevisconsulting: Corrected. I will commit and push work this far and add the new unit test after.

src/main/surfrdUtilsMod.F90
barlage: Hmm, I would lean toward warning, mainly because I could see someone getting caught in an iteration loop trying to get a valid domain. Just need to make it clear in the documentation what the fallback procedure is. @slevisconsulting I don't suppose we have a count for the number of gridcells where this happens?

ekluzek: OK, let's go with the warning.

It's going to print it out each time this happens, so for every gridcell, as is. We could ask Sam to do a summary print at the bottom, but unless you get really fancy it would be for each task.  So it potentially could be a ream of output (although most tasks would go into cesm.log rather than lnd.log). So is a warning for each gridcell OK, or do you want us to do the work to do a summary? I'm thinking it wouldn't be worth it. But, I do acknowledge that the output this warning will log could be a lot....

src/main/test/surfrdUtils_test/test_surfrdUtils.pf
slevisconsulting: Done.

src/main/surfrdUtilsMod.F90
ekluzek: @slevisconsulting let me make sure I understand here. If you are collapsing urban landunits, wt_sum won't be equal to 1. upon entry or exit to this subroutine. At what point does that get rectified? When the model starts up wt_sum is going to sum to 1.0 right?

src/main/surfrdUtilsMod.F90
ekluzek: @barlage I think Sam should go ahead with this as this is done. We still want to hear your opinion on the fact that the log output from this warning could get huge. But, I'm thinking that we should move forward with the current solution. Doing a summary of the output could be a time-consuming tangent.

barlage: It's not worth any extra effort. I was just wondering if we could reduce output if a summary were available. I'm fine with the procedure as implemented.

ekluzek: You could simplify it a bit, by doing a summary of each processor. But, you'd still have separate output from each processor, and you are likely to use hundreds to thousands or processors so I'm not convinced that would buy you a lot. So I think it's best to start using it, and have people let us know how much output we get from it. If it is shown to be a problem, then we can implement a solution that makes the most sense based on what we see.

src/main/surfrdUtilsMod.F90
slevisconsulting: ...sorry for the delay in my response:

@ekluzek , yes, if you are collapsing urban landunits, wt_sum will be less than or equal to 1, because the urban landunits are only three of the model's landunits.

The code encounters this statement a second time if n_dom_landunits > 0 and < 9. This time wt_sum = 1 because now it's working with all the landunits.

One way to remove the `if` in front of the error check might be to move the error check outside of the subroutine call. Do you prefer that?

slevisconsulting: To be exact, I could see placing the error check without the `if` in subr. surfrd_get_data after each call:
- call collapse_to_dominant (if collapse_urban is true)
- call collapse_to_dominant (second call, this time for n_dom_landunits)
- call collapse_individual_lunits

slevisconsulting: @ekluzek , I will go ahead and try this since I haven't heard from you. We can always change back if you don't like the change.

ekluzek: Hey Sam. I looked into the code a bit more. I think maybe the thing to do would be to have a if else structure for the check_sums_equal_1 call. If the previous sum was 1, you call it normally, but otherwise you send it the previous sum as the optional argument "sumto". That way it's still always doing this check. In the end I think that's what we want, for the appropriate check to be done each time. So as you say another option is to pull the checkout out of the collapse subroutine and then you can make sure the right check is done afterwards. The bad thing there is that then if the collapse call is used somewhere else, you also need to bring the call to the check at the same time.

slevisconsulting: @ekluzek , we have the following conundrum...

When `collapse_urban = .true.` the wt_sum has a different value < 1 for each g, because the sum of urban is different in every grid cell.

This means that I need to call the error check within a `do g` loop. However, now I get the error:
`The shape matching rules of actual arguments and dummy arguments have been violated`

I'm not sure if this could be fixed by changing this...
`real(r8)        , intent(in) :: arr(lb:,:)   ! array to check`
to this...
`real(r8)        , intent(in) :: arr(lb:ub,:)   ! array to check

I could try it, but this would require changing the call arguments to include `ub` in all existing calls to this subroutine. So I wanted to run it by you before trying.

A different alternative would be to change `sumto` from scalar to an array.

ekluzek: It looks like sumto isn't used in any other calls (that I could see) of check_sums_equal_1, so I'd say change sumto to an array.

slevisconsulting: I found it here only:
`main/surfrdUtilsMod.F90:          call check_sums_equal_1(wt_cft, begg, 'wt_cft', subname//': irrigation', sumto=TotalSum)`
`main/surfrdUtilsMod.F90:       call check_sums_equal_1(wt_cft, begg, 'wt_cft', subname//': mergetoclmpft', sumto=TotalSum)`

...and maybe this will still work even with sumto as an array?

ekluzek: TotalSum would need to be an array as well. Which might not be hard to do..

ekluzek: The unit tests for it would also need to be updated...

slevisconsulting: Done. See latest push.

slevisconsulting:
Thank you! Once Negin's dev032 tag comes through, I will update and run tests.

billsacks:
@slevisconsulting @ekluzek - In discussion yesterday with @barlage and @negin513 , we were wondering what happens now if all landunits in a given grid cell have % cover less than their toosmall value. For example, what happens if a grid cell is 30% lake, 40% glacier and 30% wetland, and all the toosmall values are set to 50%? Is this handled gracefully now? We felt it should be. Our proposed solution would be to arbitrarily set the grid cell to 100% natural vegetation in this case, unless you have already come up with a different solution for this case.

This was connected with the discussion of whether you could eliminate a landunit entirely (e.g., eliminating urban) by setting toosmall to > 100%. It seems like, if we are handling the above case, then that would also handle this case, right?

ekluzek:
Yes, it has logic to handle that. The comment that describes it is "If all landunits got removed, go back and keep the largest landunit". So it keeps whatever it determines the largest landunit is. It will also print a warning in this case. So if you find that happens a lot, and you don't want it to, you could adjust your settings and try again.

billsacks:
Thanks, @ekluzek , that makes sense.

slevisconsulting:
@ekluzek , I think we're all set for the merge. All tests that are supposed to PASS, do.

I updated the ChangeLog and ChangeSum.

I placed baselines at their permanent locations.

Pls let me know if I'm forgetting anything.

ekluzek:
Sounds good! I'll check things over and do the merge tomorrow...

slevisconsulting:
I made the above reference by accident. I corrected at the other end, but I don't see how to remove the reference on this side.

ekluzek:
Yeah, don't worry about it...

Issue #274, opened 2018-02-13
Dynamic landunits: improve water and energy conservation


billsacks:
We'd like to improve some aspects of water and energy conservation with dynamic landunits. This includes:

- [ ] Treatment of the snow pack in a more rigorous way (needed for the new glacier surface mass balance definition that includes changes in the snow pack in SMB)

- [ ] Better handling of the virtual glacier ice in glacier landunits: currently this is included as real mass for conservation purposes. (This is not a conservation issue, but causes too-large fluxes for transitions to/from glaciers.)

billsacks:
We may also want to conserve water and energy more rigorously, as is done for C and N: i.e., rather than doing this all via adjustment fluxes, do at least some adjustment of individual state variables.

billsacks:
This part:

> Better handling of the virtual glacier ice in glacier landunits: currently this is included as real mass for conservation purposes. (This is not a conservation issue, but causes too-large fluxes for transitions to/from glaciers.)

is fixed in ctsm1.0.dev031, via #650.

PR #650, opened 2019-03-05
Reduce dynbal fluxes with dynamic glaciers

billsacks:
### Description of changes

Subtract virtual states to reduce dynbal fluxes for transient glaciers

Up until now, when computing the dynbal correction fluxes for transient
glacier columns, we have been (1) counting the mass and energy in the
ice column as if that is a real state, but on the other hand (2) NOT
accounting for the fact that glacier columns don't represent the soil
under glacier. These two issues work in opposite directions, but (1)
dominates, because there is much more ice (and energy, I think) in the
roughly 50 m of glacial ice than there is in a typical 50 m soil
column.

In discussing this issue with @whlipscomb, we came up with the idea of
subtracting some baseline value from each glacier column. I think that,
as long as we subtract the same baseline value throughout an entire
simulation for a given column, we will still conserve mass and energy
through dynamic landunit transitions.

So, here we subtract baseline values from glacier columns, accounting
for the two issues mentioned above: (1) we subtract the water and energy
in the glacier ice, because this is a virtual state in CTSM, and (2) we
add the water and energy from the vegetated column(s), to account for
the fact that we don't have an explicit representation of
soil-under-glacier (this carries the assumption that the
soil-under-glacier has the same state as the initial vegetated state in
that grid cell). We set these baselines in initialization, so they begin
equal to the cold start state. Water and ice in the glacial ice stay
fixed over the course of a simulation, so the cold start values should
be the same as the current values at any point in time. The heat content
of the glacial ice does change over time, but by subtracting this
baseline value, we should be able to reduce the dynbal sensible heat
fluxes.

In addition, this introduces a new namelist flag,
`reset_dynbal_baselines`, which allows the user to reset these baselines
at some desired point in the simulation. I think that, in general, this
resetting would break conservation. But as long as it is done before the
onset of transient glaciers, I think this should be okay. In this way,
the user can minimize dynbal fluxes even further, by resetting the
baselines after the system has spun up.

### Specific notes

- Partially addresses #274

- Changes answers significantly for runs with transient glaciers
  (substantial changes in dynbal water and energy fluxes)
- For other transient cases, possible roundoff-level changes for dynbal
  fluxes (if, e.g., a grid cell that has some static glacier has
  transient crop area: the total water and energy of the grid cell will
  change significantly, but by the same amount before and after the
  transition, leading to roundoff-level differences in fluxes)
- For all cases: Large changes in some diagnostic terms related to total
  water and energy

- New namelist variable: reset_dynbal_baselines

Testing performed, if any:
- So far, just manual testing
- Will run full test suite
- In addition, will run full test suite with minimal changes to achieve
  no greater than roundoff-level diffs from master, to confirm that the
  only answer changes came from the new subtraction of baselines for
  glacier columns

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -30,7 +30,7 @@ required = True
 local_path = cime
 protocol = git
 repo_url = https://github.com/ESMCI/cime
-tag = cime5.7.9
+tag = cime5.8.0
 required = True
 
 [externals_description]
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -30,7 +30,7 @@ required = True
 local_path = cime
 protocol = git
 repo_url = https://github.com/ESMCI/cime
-tag = cime5.8.0
+tag = ctsm/ctsm1.0/cime5.7.9/n01
 required = True
 
 [externals_description]
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -2288,6 +2288,10 @@ sub setup_logic_dynamic_subgrid {
    setup_logic_do_harvest($opts, $nl_flags, $definition, $defaults, $nl);
 
    add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'reset_dynbal_baselines');
+   if ( &value_is_true($nl->get_value('reset_dynbal_baselines')) &&
+        &remove_leading_and_trailing_quotes($nl_flags->{'clm_start_type'}) eq "branch") {
+      $log->fatal_error("reset_dynbal_baselines has no effect in a branch run");
+   }
 }
 
 sub setup_logic_do_transient_pfts {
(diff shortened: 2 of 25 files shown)


billsacks:
@ekluzek @Katetc @whlipscomb I wanted to give you all the chance to review (parts of) this, but feel free to remove yourself as a reviewer if you don't feel a need to review this. @Katetc and @whlipscomb one thing you could focus your review on is the clarity of the description of the new namelist variable in `namelist_definition_ctsm.xml`; I plan to put similar documentation in the CESM Land Ice github.io documentation soon.

I still have a bit of work to do on this PR:
- Some more manual testing to be comfortable with how things are working
- Adding a unit test
- Addressing FIXME notes that I have sprinkled in the code (mostly / entirely related to inline documentation of variables, etc.)

But since this is time-critical, I wanted to put this out for review now. I'll need to kick off testing by Thursday.

Issue #55, opened 2017-12-16
Balance check doesn't check for possible errors in all points


billsacks:
**Bill Sacks < sacks@ucar.edu > - 2017-01-26 09:44:05 -0700**
**Bugzilla Id:** [2415](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2415)
**Bugzilla CC:**  dlawren@ucar.edu,  l.vankampenhout@uu.nl,  oleson@ucar.edu,  rfisher@ucar.edu, 

Leo pointed this out, and I agree with his assessment:

I noticed that the following code from BalanceCheckMod does not check on the maximum error over all columns, merely checks the last column found. 
So there is a possibility that the abort is not triggered when it should have been. 

       ! Soil energy balance check

       found = .false.
       do c = bounds%begc,bounds%endc
          if (col%active(c)) then
             if (abs(errsoi_col(c)) > 1.0e-6_r8 ) then
                found = .true.
                indexc = c
             end if
          end if
       end do
       if ( found ) then
          write(iulog,*)'WARNING: BalanceCheck: soil balance error (W/m2)'
          write(iulog,*)'nstep         = ',nstep
          write(iulog,*)'errsoi_col    = ',errsoi_col(indexc)
          if (abs(errsoi_col(indexc)) > 1.e-4_r8 .and. (nstep > 2) ) then
             write(iulog,*)'clm model is stopping'
             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
          end if
       end if

We should change this code (and other code in BalanceCheck mod if there is other code that follows this pattern) to check for the maximum error across all points.

billsacks:
Fixed by ctsm1.0.dev032

PR #670, opened 2019-03-27
Fixing the balance check to check for possible errors in all columns/patches

negin513:
### Description of changes

Up until now, the BalanceCheckMod code in biogeophysics did not check all the possible errors over all columns or patches with the error threshold for aborting clm and it merely compared the last column/patch found (where warning threshold is met) with the error threshold. Therefore, there was always a possibility that the abort is not triggered when it should have been. @lvankampenhout and @billsacks pointed out this issue (Issue #55) for soil energy balance check. A similar issue is also happening for the other balance checks in `BalanceCheckMod.F90`. 

Below is the list of the balance checks with a similar issue:
1. Water balance check
2. Snow balance check
3. Solar radiation energy balance check
4. Longwave radiation energy balance check
5. Surface energy balance check
6. Soil energy balance check (Issue #55) 

The simple psuedo-code below illustrates the problems with all these balance checks. First loop compares the error over all columns with the `warning_threshhold` and save only the last column where `found= true`. The second block compares only the last column found with the `error_threshhold`. Hence there is always a possibility of an error higher than `error_threshhold`  that is simply skipped in the second comparison. 
```
! Sample balance check psuedo code

found = .false.
do c = bounds%begc,bounds%endc
      if (error(c) > warning_threshold ) then
         found = .true.
         indexc = c
      end if
end do

if ( found ) then
   write(iulog,*)'WARNING: BalanceCheck !!!!'
   if (error(c) > error_threshhold ) then
      write(iulog,*)'clm model is stopping'
      call endrun
   end if
end if
```

For remediating this problem, we used `MAXVAL` and `MAXLOC` to compare the largest error with the warning and error thresholds. 

### Specific notes

CTSM Issues Fixed (include github issue #): Fixes issue #55 

No. However there are possible cases that were previously working, but are no longer working, since in balance checks there might be errors larger than the error threshhold, which were previously skipped. 

Testing performed, if any:
* Manual testing making sure the code checks for the maximum error across all columns or patches with the error threshold. 
* Full `aux_clm` test suites on both Cheyenne and Hobart.

### DIFF
--- a/src/biogeophys/BalanceCheckMod.F90
+++ b/src/biogeophys/BalanceCheckMod.F90
@@ -699,10 +699,10 @@ contains
 
        ! Soil energy balance check
 
-       errsoi_col_max_val  =  maxval( abs(errsoi_col(bounds%begc:bounds%endc)) , mask = col%active(c))
+       errsoi_col_max_val  =  maxval( abs(errsoi_col(bounds%begc:bounds%endc)) , mask = col%active )
 
        if (errsoi_col_max_val > 1.0e-5_r8 ) then
-           indexc = maxloc( abs(errsoi_col(bounds%begc:bounds%endc)), 1 , mask = col%active(c) ) + bounds%begc -1
+           indexc = maxloc( abs(errsoi_col(bounds%begc:bounds%endc)), 1 , mask = col%active ) + bounds%begc -1
            write(iulog,*)'WARNING: BalanceCheck: soil balance error (W/m2)'
            write(iulog,*)'nstep         = ',nstep
            write(iulog,*)'errsoi_col    = ',errsoi_col(indexc)
--- a/src/biogeophys/BalanceCheckMod.F90
+++ b/src/biogeophys/BalanceCheckMod.F90
@@ -420,11 +420,11 @@ contains
 
        end do
        
-       errh2o_max_val = maxval( abs(errh2o(bounds%begc:bounds%endc)), mask = (errh2o(bounds%begc:bounds%endc) < spval) )
+       errh2o_max_val = maxval( abs(errh2o(bounds%begc:bounds%endc)), mask = (errh2o(bounds%begc:bounds%endc) /= spval) )
 
        if (errh2o_max_val > h2o_warning_thresh) then
 
-           indexc = maxloc( abs(errh2o(bounds%begc:bounds%endc)), 1 , mask = (errh2o(bounds%begc:bounds%endc) < spval) ) + bounds%begc -1
+           indexc = maxloc( abs(errh2o(bounds%begc:bounds%endc)), 1 , mask = (errh2o(bounds%begc:bounds%endc) /= spval) ) + bounds%begc -1
            write(iulog,*)'WARNING:  water balance error ',&
              ' nstep= ',nstep, &
              ' local indexc= ',indexc,&
@@ -516,15 +516,16 @@ contains
                 snow_sinks(c) = 0._r8
                 errh2osno(c) = 0._r8
              end if
-
+          else
+             errh2osno(c) = 0._r8
           end if
        end do
 
 
-       errh2osno_max_val = maxval( abs(errh2osno(bounds%begc:bounds%endc)),  mask = (errh2osno(bounds%begc:bounds%endc) < spval) )
+       errh2osno_max_val = maxval( abs(errh2osno(bounds%begc:bounds%endc)),  mask = (errh2osno(bounds%begc:bounds%endc) /= spval) )
        
        if (errh2osno_max_val > h2o_warning_thresh) then
-            indexc = maxloc( abs(errh2osno(bounds%begc:bounds%endc)), 1 ,  mask = (errh2osno(bounds%begc:bounds%endc) < spval) ) + bounds%begc -1
+            indexc = maxloc( abs(errh2osno(bounds%begc:bounds%endc)), 1 ,  mask = (errh2osno(bounds%begc:bounds%endc) /= spval) ) + bounds%begc -1
             write(iulog,*)'WARNING:  snow balance error '
             write(iulog,*)'nstep= ',nstep, &
                  ' local indexc= ',indexc, &
@@ -618,11 +619,11 @@ contains
 
        ! Solar radiation energy balance check
 
-       errsol_max_val = maxval( abs(errsol(bounds%begp:bounds%endp)), mask = (errsol(bounds%begp:bounds%endp) < spval) ) 
+       errsol_max_val = maxval( abs(errsol(bounds%begp:bounds%endp)), mask = (errsol(bounds%begp:bounds%endp) /= spval) ) 
 
        if  ((errsol_max_val > energy_warning_thresh) .and. (DAnstep > skip_steps)) then
 
-           indexp = maxloc( abs(errsol(bounds%begp:bounds%endp)), 1 , mask = (errsol(bounds%begp:bounds%endp) < spval) ) + bounds%begp -1
+           indexp = maxloc( abs(errsol(bounds%begp:bounds%endp)), 1 , mask = (errsol(bounds%begp:bounds%endp) /= spval) ) + bounds%begp -1
            indexg = patch%gridcell(indexp)
            !write(iulog,*)'indexg         = ',indexg
            write(iulog,*)'WARNING:: BalanceCheck, solar radiation balance error (W/m2)'
@@ -647,10 +648,10 @@ contains
        
        ! Longwave radiation energy balance check
 
-       errlon_max_val = maxval( abs(errlon(bounds%begp:bounds%endp)), mask = (errlon(bounds%begp:bounds%endp) < spval) )
+       errlon_max_val = maxval( abs(errlon(bounds%begp:bounds%endp)), mask = (errlon(bounds%begp:bounds%endp) /= spval) )
 
        if ((errlon_max_val > energy_warning_thresh) .and. (DAnstep > skip_steps)) then
-            indexp = maxloc( abs(errlon(bounds%begp:bounds%endp)), 1 , mask = (errlon(bounds%begp:bounds%endp) < spval) ) + bounds%begp -1
+            indexp = maxloc( abs(errlon(bounds%begp:bounds%endp)), 1 , mask = (errlon(bounds%begp:bounds%endp) /= spval) ) + bounds%begp -1
             write(iulog,*)'indexp         = ',indexp
             write(iulog,*)'WARNING: BalanceCheck: longwave energy balance error (W/m2)'
             write(iulog,*)'nstep        = ',nstep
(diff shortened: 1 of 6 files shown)


src/biogeophys/BalanceCheckMod.F90
```diff
@@ -630,118 +614,102 @@ subroutine BalanceCheck( bounds, &
...
-             call endrun(decomp_index=indexp, clmlevel=namep, msg=errmsg(sourcefile, __LINE__))
-          end if
-       end if
+       errsol_max_val = maxval( abs(errsol(bounds%begp:bounds%endp)), mask = (errsol(bounds%begp:bounds%endp) < spval) ) 
+
+       if  ((errsol_max_val > energy_warning_thresh) .and. (DAnstep > skip_steps)) then
+
+           indexp = maxloc( abs(errsol(bounds%begp:bounds%endp)), 1 , mask = (errsol(bounds%begp:bounds%endp) < spval) ) + bounds%begp -1
+           indexg = patch%gridcell(indexp)
+           !write(iulog,*)'indexg         = ',indexg
```
billsacks: Another commented-out line that should be removed

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -630,118 +614,102 @@ subroutine BalanceCheck( bounds, &
...
-          write(iulog,*)'errsoi_col    = ',errsoi_col(indexc)
-          if (abs(errsoi_col(indexc)) > 1.e-4_r8 .and. (DAnstep > skip_steps) ) then
-             write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
-          end if
-       end if
+       errsoi_col_max_val  =  maxval( abs(errsoi_col(bounds%begc:bounds%endc)), mask = (errsoi_col(bounds%begc:bounds%endc) < spval) )
+
+       if (errsoi_col_max_val > 1.0e-5_r8 ) then
+           indexc = maxloc( abs(errsoi_col(bounds%begc:bounds%endc)), 1 , mask = (errsoi_col(bounds%begc:bounds%endc) < spval) ) + bounds%begp -1
```
billsacks: `bounds%begp -1` should be `bounds%begc -1`

I didn't notice any other problems like this in my read-through of the changes, but since I found this one error, can you go back and double-check that all the indexing is correct for all variables - i.e., that you're using begp/endp vs. begc/endc in the correct places?

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -268,10 +268,22 @@ subroutine BalanceCheck( bounds, &
      real(r8) :: dtime                                  ! land model time step (sec)
      integer  :: nstep                                  ! time step number
      integer  :: DAnstep                                ! time step number since last Data Assimilation (DA)
-     logical  :: found                                  ! flag in search loop
+!     logical  :: found                                  ! flag in search loop
```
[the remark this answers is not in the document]
negin513: Thanks for the comment. The commented-out line is removed.

src/biogeophys/BalanceCheckMod.F90
negin513: Thanks for noticing this. I just removed it.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -407,78 +418,56 @@ subroutine BalanceCheck( bounds, &
...
-             write(iulog,*)'qflx_ice_runoff_xs    = ',qflx_ice_runoff_xs(indexc)*dtime
-             write(iulog,*)'qflx_glcice_dyn_water_flux = ', qflx_glcice_dyn_water_flux(indexc)*dtime
-             write(iulog,*)'qflx_snwcp_discarded_ice = ',qflx_snwcp_discarded_ice(indexc)*dtime
-             write(iulog,*)'qflx_snwcp_discarded_liq = ',qflx_snwcp_discarded_liq(indexc)*dtime
-             write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
-          end if
+       
+       errh2o_max_val = maxval(abs(errh2o(bounds%begc:bounds%endc)))
+       write(iulog,*)'errh2o_max_val      :', errh2o_max_val
```
billsacks: This write statement should be removed from the final version.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -527,56 +516,53 @@ subroutine BalanceCheck( bounds, &
...
-             write(iulog,*)'qflx_snwcp_discarded_ice = ',qflx_snwcp_discarded_ice(indexc)*dtime
-             write(iulog,*)'qflx_snwcp_discarded_liq = ',qflx_snwcp_discarded_liq(indexc)*dtime
-             write(iulog,*)'qflx_sl_top_soil   = ',qflx_sl_top_soil(indexc)*dtime
-             write(iulog,*)'clm model is stopping'
-
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
-          end if
+
+       errh2osno_max_val = maxval( abs(errh2osno(bounds%begc:bounds%endc)))
+       write(iulog,*)'errh2osno_max_val      :', errh2osno_max_val
```
billsacks: This write statement should also be removed.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -624,124 +610,115 @@ subroutine BalanceCheck( bounds, &
...
-             write(iulog,*)'forc_solai(1) = ',forc_solai(indexg,1)
-             write(iulog,*)'forc_solai(2) = ',forc_solai(indexg,2)
-             write(iulog,*)'forc_tot      = ',forc_solad(indexg,1)+forc_solad(indexg,2) &
-               +forc_solai(indexg,1)+forc_solai(indexg,2)
-             write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexp, clmlevel=namep, msg=errmsg(sourcefile, __LINE__))
-          end if
-       end if
+       errsol_max_val = maxval( abs(errsol(bounds%begp:bounds%endp)), mask = (errsol(bounds%begp:bounds%endp) /= spval) ) 
+       write(iulog,*)'errsol_max_val      :', errsol_max_val
```
billsacks: This write statement should also be removed.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -624,124 +610,115 @@ subroutine BalanceCheck( bounds, &
...
-       if ( found  .and. (DAnstep > skip_steps) ) then
-          write(iulog,*)'WARNING: BalanceCheck: longwave energy balance error (W/m2)' 
-          write(iulog,*)'nstep        = ',nstep 
-          write(iulog,*)'errlon       = ',errlon(indexp)
-          if (abs(errlon(indexp)) > 1.e-5_r8 ) then
-             write(iulog,*)'clm model is stopping - error is greater than 1e-5 (W/m2)'
-             call endrun(decomp_index=indexp, clmlevel=namep, msg=errmsg(sourcefile, __LINE__))
-          end if
+       errlon_max_val = maxval( abs(errlon(bounds%begp:bounds%endp)), mask = (errlon(bounds%begp:bounds%endp) /= spval) )
+       write(iulog,*)'errlon_max_val      :', errlon_max_val
```
billsacks: This write statement should also be removed.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -624,124 +610,115 @@ subroutine BalanceCheck( bounds, &
...
-             write(iulog,*)'fsa fsr = '        ,fsa(indexp),    fsr(indexp)
-             write(iulog,*)'fabd fabi = '      ,fabd(indexp,:), fabi(indexp,:)
-             write(iulog,*)'albd albi = '      ,albd(indexp,:), albi(indexp,:)
-             write(iulog,*)'ftii ftdd ftid = ' ,ftii(indexp,:), ftdd(indexp,:),ftid(indexp,:)
-             write(iulog,*)'elai esai = '      ,elai(indexp),   esai(indexp)      
-             write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexp, clmlevel=namep, msg=errmsg(sourcefile, __LINE__))
-          end if
+       errseb_max_val = maxval( abs(errseb(bounds%begp:bounds%endp)))
+       write(iulog,*)'errseb_max_val      :', errseb_max_val
```
billsacks: This write statement should also be removed.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -624,124 +610,115 @@ subroutine BalanceCheck( bounds, &
...
-          write(iulog,*)'nstep         = ',nstep
-          write(iulog,*)'errsoi_col    = ',errsoi_col(indexc)
-          if (abs(errsoi_col(indexc)) > 1.e-4_r8 .and. (DAnstep > skip_steps) ) then
-             write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
-          end if
-       end if
-
+       errsoi_col_max_val  =  maxval( abs(errsoi_col(bounds%begc:bounds%endc)) , mask = (errsoi_col(bounds%begc:bounds%endc) /= spval) )
+       write(iulog,*)'errsoi_col_max_val      :', errsoi_col_max_val
```
billsacks: This write statement should also be removed.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -624,124 +610,115 @@ subroutine BalanceCheck( bounds, &
...
-          write(iulog,*)'WARNING: BalanceCheck: soil balance error (W/m2)'
-          write(iulog,*)'nstep         = ',nstep
-          write(iulog,*)'errsoi_col    = ',errsoi_col(indexc)
-          if (abs(errsoi_col(indexc)) > 1.e-4_r8 .and. (DAnstep > skip_steps) ) then
-             write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
-          end if
-       end if
-
+       errsoi_col_max_val  =  maxval( abs(errsoi_col(bounds%begc:bounds%endc)) , mask = (errsoi_col(bounds%begc:bounds%endc) /= spval) )
```
billsacks: Mostly as a reminder for myself: We talked about changing this to use a mask of col%active rather than spval here, as well as a couple of lines down.

negin513: Correct! It uses `col%active` as the mask for `errsoi_col now`.

negin513: For some reason, using `col%active` as the mask for `errsoi_col` still gives out 1E+036 for maximum `errsoi_col` ( model stops) for the following six cases:

```
ERP_P36x2_D_Ld10.f10_f10_musgs.IHistClm50SpG.cheyenne_intel.clm-glcMEC_decrease
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm50Bgc.cheyenne_intel.clm-monthly
ERS_Ly3_P72x2.f10_f10_musgs.IHistClm50BgcCropG.cheyenne_intel.clm-cropMonthOutput
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm50Bgc.cheyenne_gnu.clm-monthly
```

They are all `P36x2` or `P72x2` and I think it might be some issue related to the threading.

billsacks: It looks like the problem is that you masked by `col%active`, without including bounds. Just like the other arrays, this column-level array needs explicit bounds, so the masking should be done by `col%active(bounds%begc:bounds%endc)`. Can you try making that change and rerunning those 6 tests? If things look good for those 6 tests, I'd be comfortable with your making the tag without needing to rerun the full test suite, if that would save you time. (We could just call Sam's attention to this, so that, in the unlikely event that he runs into balance check errors in his testing, he can come back to us.)

negin513: Thanks for your comment. It is corrected. I reran the full test suites and generated the baselines for this tag. Should I go ahead for the updates on the logs?  
I'd be happy to incorporate the changes in `run_sys_tests` in the PR, but there was one issue using gnu compiler, which I wanted to discuss with you.

PR #676, opened 2019-04-06
12 PFT FATES default file

rgknox:
### Description of changes

This set of changes updates the default FATES parameter file to be a 12 PFT file, instead of the previous 2-clone PFT file.  The wallclock was extended on one of the longer tests to accomodate the greater computational demand of FATES working with more cohorts and more diversity.

This is synchronized with: https://github.com/NGEET/fates/pull/514

### Specific notes

The 3 month fates hydro test does not complete.  

SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.G

Getting this test passing will be added as a new high-priority issue in FATES. I am certain that changes on the fates side in PR 514 did not generate a regression. The new parameter file imposes a much more demanding simulation condition, that is why it now fails.  My plan was to not add this test to the expected fails list on the fates test suite, as it should be listed as a high-priority critical issue to fix.

### Contributors

See: https://github.com/NGEET/fates/pull/514

### Are answers expected to change (and if so in what way)?

Yes, most definitely, different plants, different answers. Only FATES should change.

### Any User Interface Changes (namelist or namelist defaults changes)?

default file changed

Testing performed, if any:

fates test suite, non-baseline, cheyenne, intel:

/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.12pft-phenoff-st3-z0-orig2-C645e520-Fada3d90

### DIFF
--- a/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -356,7 +356,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- ================================================================== -->
 
 <!--<fates_paramfile>/glade/u/home/rgknox/ctsm/src/fates/parameter_files/fates_params_default_api.7.2.0.nc</fates_paramfile>-->
-<fates_paramfile>/glade/u/home/rgknox/ctsm/src/fates/parameter_files/fates_params_12blt_api.7.2.0.nc</fates_paramfile>
+<fates_paramfile>/glade/u/home/rgknox/ctsm/src/fates/parameter_files/fates_params_api.7.2.0.nc</fates_paramfile>
 <!--<fates_paramfile>lnd/clm2/paramdata/fates_params_default_2trop.c190205.nc</fates_paramfile> -->
 
 <!-- ========================================================================================  -->
--- a/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -355,9 +355,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- FATES default parameter file                                       -->
 <!-- ================================================================== -->
 
-<!--<fates_paramfile>/glade/u/home/rgknox/ctsm/src/fates/parameter_files/fates_params_default_api.7.2.0.nc</fates_paramfile>-->
-<fates_paramfile>/glade/u/home/rgknox/ctsm/src/fates/parameter_files/fates_params_api.7.2.0.nc</fates_paramfile>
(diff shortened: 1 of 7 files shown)


bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -355,7 +355,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- FATES default parameter file                                       -->
 <!-- ================================================================== -->
 
-<fates_paramfile>lnd/clm2/paramdata/fates_params_default_2trop.c190205.nc</fates_paramfile>
+<fates_paramfile>lnd/clm2/paramdata/fates_params_api.7.2.0.nc</fates_paramfile>
```
ekluzek: I love having the api version name in the params filename. But, I still think you want the number of PFTs in the name. And I think you want a creation date, or something like it, as you could still have multiple files for the same API right?

ekluzek:
Oh, I just read your comment about leaving this out of expected fails. IMHO if you commit a version that fails a test to fates_next_api, you should put it in expected fails -- even if you plan to fix it right away. Because, other's might run into this fail and now know why. The other way to handle it is to get it fixed before you commit to fates_next_api.

rgknox:
I can flag that failing test.  My fear was that I would forget about it failing if we were not constantly reminded with the FAIL message.  I'm labeling that particular failure as a high-priority on the fates site.

Will update the file name to contain number of PFTs and creation date.

ekluzek:
OK, I do see your point. My preference would still be to label it as an expected fail. But, if you really think that will cause you to ignore it -- feel free to override that preference.

rgknox:
I think with the issue created and the high-priority label, we've now got our bases covered.  Will set the fail to expected.

rgknox:
re-testing with the updated file name and fail entry.

ekluzek:
@rgknox go ahead and push the merge button when your testing is complete, since I've now looked it over and approved the changes. Then you won't be waiting on me to approve it.

rgknox:
ok, will do @ekluzek 

On a similar note, I reduced our Ly2 single site test to 13 months.   I was also curious, what is the protocol for moving a test to a different order in the test list.  For instance, the lm13 (previously Ly2) was our longest wall-time of all tests for the fates suite.  I would like that test to be first compiled and first submitted.  Is the protocol here to just copy and paste its entry in the xml file to the first location among fates test .... or do you guys recommend I use a special tool to do this?  In the past there was an xml change tool that we had used to change these entries.

ekluzek:
We don't have tools to edit the xml anymore. So you can only edit by hand. I haven't messed with the test order. There was a time that the order was sorted by testname, but I think you are right the order is based on the xml file. That just hasn't been something we've messed with on purpose.

rgknox:
updated external tag
re-tested with fresh manage_externals
imported parameter file
integrating

Issue #454, opened 2018-07-27
Allow namelist defaults to apply for phys versions greater than or less than a given version


billsacks:
As we support more and more versions, I feel like we'll (at least eventually) want the ability for the namelist defaults xml file to allow us to say things like: "For CLM versions <= clm5.0, use foo = 3; for CLM versions > clm5.0, use foo = 7". Currently, we need to explicitly list the defaults for every phys version, which is going to be hard to maintain and error-prone long-term.

I feel like it will be useful for this purpose to separate the phys version into two pieces: a name and a version number. The name will be something like "clm" or "nwp" (the former for the "climate" version of CTSM, the later for the numerical weather prediction version). The version will be numeric.

I'm then imagining we could have a namelist default entry that looks like:

```xml
<foo phys="clm" version="<= 5.0">3</foo>
<foo phys="clm" version="> 5.0">7</foo>
```

or

```xml
<foo phys="clm" version_le="5.0">3</foo>
<foo phys="clm" version_gt="5.0">7</foo>
```

billsacks:
Blocked by #585

ekluzek:
Another thing that ParamGen could solve.

PR #677, opened 2019-04-08
Add NWP compset and illustrate how it would set namelist defaults

billsacks:
This is just an initial PR for review and comments. This is not
completely functional yet: Changes are still needed in CLMBuildNamelist,
and some of the referenced namelist options do not yet exist in the
code.

### Specific notes

- New NWP compset, and some associated xml variables

Testing performed, if any:
- Just compared env_run.xml from I2000Clm50SpGs vs. I2000Nwp50SpGswpGs

### DIFF
--- a/bld/namelist_files/namelist_defaults_ctsm.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm.xml
@@ -124,8 +124,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <organic_frac_squared phys="clm4_5" >.true.</organic_frac_squared>
 <organic_frac_squared phys="clm5_0" >.false.</organic_frac_squared>
 
-<soil_layerstruct phys="clm5_0">20SL_8.5m</soil_layerstruct>
-<soil_layerstruct phys="clm4_5">10SL_3.5m</soil_layerstruct>
+<soil_layerstruct structure="fast">5SL_3m</soil_layerstruct>
+<soil_layerstruct structure="standard" phys="clm5_0">20SL_8.5m</soil_layerstruct>
+<soil_layerstruct structure="standard" phys="clm4_5">10SL_3.5m</soil_layerstruct>
 
 <use_bedrock phys="clm5_0" use_fates =".true.">.false.</use_bedrock>
 <use_bedrock phys="clm5_0" vichydro ="1"      >.false.</use_bedrock>
@@ -231,6 +232,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- Canopy fluxes namelist defaults -->
 <use_undercanopy_stability phys="clm5_0">.false.</use_undercanopy_stability>
 <use_undercanopy_stability phys="clm4_5">.true.</use_undercanopy_stability>
+
+<itmax_canopy_fluxes structure="standard">40</itmax_canopy_fluxes>
+<itmax_canopy_fluxes structure="fast">3</itmax_canopy_fluxes>
 
 <!-- Canopy hydrology namelist defaults -->
 <use_clm5_fpi phys="clm5_0">.true.</use_clm5_fpi>
@@ -447,9 +451,8 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <nnegcrit phys="clm4_5" use_cn=".true.">-6.d+1</nnegcrit>
 
 <!-- Plant hydraulic stress -->
-<use_hydrstress phys="clm4_5"                 >.false.</use_hydrstress>
-<use_hydrstress phys="clm5_0" use_fates=".true." >.false.</use_hydrstress>
-<use_hydrstress phys="clm5_0" use_fates=".false.">.true.</use_hydrstress>
+<use_hydrstress>.false.</use_hydrstress>
+<use_hydrstress phys="clm5_0" use_fates=".false." configuration="clm">.true.</use_hydrstress>
 <!--
 
      Initial condition files to use and or interpolate from
@@ -1137,6 +1140,15 @@ lnd/clm2/surfdata_map/surfdata_ne120np4_78pfts_CMIP6_simyr1850_c170824.nc</fsurd
 <create_crop_landunit     use_fates=".true." >.false.</create_crop_landunit>
 
 <run_zero_weight_urban>.false.</run_zero_weight_urban>
+
+<collapse_urban structure="standard">.false.</collapse_urban>
+<collapse_urban structure="fast">.true.</collapse_urban>
+
+<n_dom_landunits structure="standard">0</n_dom_landunits>
+<n_dom_landunits structure="fast">1</n_dom_landunits>
+
+<n_dom_pfts structure="standard">0</n_dom_pfts>
+<n_dom_pfts structure="fast">1</n_dom_pfts>
 
 <!-- =========================================  -->
 <!-- Performance issues                         -->
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -15,7 +15,7 @@
   <!-- Descriptions of all the different valid configurations for different model versions -->
   <description modifier_mode="1">
     <desc lnd="CLM45[%SP][%SP-VIC][%CN][%CNDV][%CN-CROP][%CNDV-CROP][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP]"                  >clm4.5:</desc>
-    <desc lnd="CLM50[%SP][%SP-VIC][%CN][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%BGC-CROP-CMIP6DECK][%BGC-CROP-CMIP6WACCMDECK]">clm5.0:</desc>
+    <desc lnd="CLM50[%SP][%SP-VIC][%CN][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%BGC-CROP-CMIP6DECK][%BGC-CROP-CMIP6WACCMDECK][%NWP-SP]">clm5.0:</desc>
     <desc option="SP"              >Satellite phenology:</desc>
     <desc option="CN"              >CN: Carbon Nitrogen model</desc>
     <desc option="CNDV"            >CNDV: CN with Dynamic Vegetation</desc>
@@ -30,6 +30,8 @@
     <desc option="BGCDV-CROP"      >BGC (vert. resol. CN and methane) with dynamic vegetation and prognostic crop:</desc>
     <desc option="BGC-CROP-CMIP6DECK">BGC (vert. resol. CN and methane) with prognostic crop, with modifications appropriate for CMIP6 DECK experiments:</desc>
     <desc option="BGC-CROP-CMIP6WACCMDECK">BGC (vert. resol. CN and methane) with prognostic crop, with modifications appropriate for CMIP6 WACCM DECK experiments:</desc>
+
+    <desc option="NWP-SP"          >NWP configuration with satellite phenology:</desc>
   </description>
 
   <entry id="COMP_LND">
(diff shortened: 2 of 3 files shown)


billsacks:
I want to make a few specific comments on this; I'll let you know when I'm done commenting and this is ready for review.

cime_config/config_component.xml
```diff
@@ -15,7 +15,7 @@
   <!-- Descriptions of all the different valid configurations for different model versions -->
   <description modifier_mode="1">
     <desc lnd="CLM45[%SP][%SP-VIC][%CN][%CNDV][%CN-CROP][%CNDV-CROP][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP]"                  >clm4.5:</desc>
-    <desc lnd="CLM50[%SP][%SP-VIC][%CN][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%BGC-CROP-CMIP6DECK][%BGC-CROP-CMIP6WACCMDECK]">clm5.0:</desc>
+    <desc lnd="CLM50[%SP][%SP-VIC][%CN][%BGC][%BGC-CROP][%FATES][%BGCDV][%BGCDV-CROP][%BGC-CROP-CMIP6DECK][%BGC-CROP-CMIP6WACCMDECK][%NWP-SP]">clm5.0:</desc>
```
billsacks: We need to make a choice here between combinatoric explosion (and the corresponding maintenance issues) vs. less error-checking at create_newcase time. The implementation here is the combinatoric explosion option, but I mainly chose that for now because it was slightly easier up-front, not because I'm necessarily in favor of that option.

With the current, combinatoric explosion implementation, we'll need to explicitly add support here for `%NWP-SP-VIC`, `%NWP-BGC`, `%NWP-BGC-CROP`, and many other options, assuming we want to support them. This will require an addition in the above line as well as a new line below for the description. We would have to do the same thing again to support a `%HYD` compset option that can be combined with these various other options.

The alternative would be to allow two (or more) `%` options on the compset name. So we'd have a compset like `CLM50%NWP%SP`. This would avoid the combinatoric problem. However, currently, I think the scripts would allow you to set up a case with `CLM50%NWP` without any BGC modifier (SP, BGC, etc.)... at least, I don't think this would abort at `create_newcase` time, but it might still be caught at `preview_namelists` time.

I guess my preference would be to see if we can have our cake and eat it, too: avoid the combinatoric explosion but see if we can come up with a fairly easy way of catching errors like `CLM50%NWP` without a BGC modifier.

What do others (especially @ekluzek ) think?

cime_config/config_compsets.xml
```diff
@@ -213,6 +213,12 @@
     <lname>1850_DATM%CPLHIST_CLM50%BGC_SICE_SOCN_MOSART_CISM2%NOEVOLVE_SWAV</lname>
   </compset>
 
+<!-- I CLM50 NWP Compsets -->
+
+  <compset>
+    <alias>I2000Nwp50SpGswpGs</alias>
```
billsacks: Other proposed compset names would be:

1. `I2000Nwp50SpNldas2Gs` - Uses NLDAS2 datm forcing

2. `I2000Nwp50SpNldas2RsGs` - Uses NLDAS2 datm forcing and stub ROF

billsacks:
At our meeting on Thursday, we discussed having the NWP and CLM configurations evolve together as much as possible. This decision is reflected here, where both leverage the same physics version number (currently "clm50", but I envision this being changed to "ctsm50"... though it's a bit confusing that that number is then unrelated to the dev tag numbering... that may be okay, but we may want to discuss that issue). Given that decision, I felt it would be most intuitive if the 50 appeared in the NWP compset name, too (both long name and alias), rather than using a compset name like "Nwp10" like we discussed Thursday.

After the Thursday meeting, as I started to think about how to implement what we discussed, I began to feel that it would be overly complicated, confusing and error-prone to try to have multiple numeric (version) options that could be mixed and matched. Specifically: We had talked about having separate version numbers for the structure vs. the physics, and I felt that would get messy and confusing.

What all this means is that evolution and backwards compatibility are tied to a single version number (currently at 50). If either the CLM or NWP configuration want a new option, while still supporting the previous option, this would be done via phys "ctsm51". Similarly, if we want to evolve the definition of, say, the "fast" structure option, we'd have a clause in namelist_defaults like:

```xml
<n_dom_pfts structure="standard">0</n_dom_pfts>
<n_dom_pfts structure="fast" phys="ctsm5_0">1</n_dom_pfts>
<n_dom_pfts structure="fast" phys="ctsm5_1">2</n_dom_pfts>
```

Finally, note that, eventually, I think the only place the name "clm" will appear in all of this is as one option for the "configuration". Other than that, references to "clm" in compset and variable names will eventually be replaced by "ctsm".

billsacks:
Okay, I'm done commenting and I welcome comments from others. I very much welcome others' thoughts on this proposal. (I'll admit that I have begun to develop some opinions, e.g., as reflected in my previous comment. So I'll say up-front that I might push back in certain areas, but please DO push me on things if you feel anything should be done differently here.)

cime_config/config_component.xml
ekluzek: Well, I think for now with just adding a single new option -- this is fine. 

I do like the concept of listing the valid configurations so it won't let you do an invalid option. I can see how we could trap this at `preview_namelist` time. But, I think you really want to catch it at `create_newcase` time. I think right now if we allowed more options, `CLM50%NWP` would use a default BGC option. So it would do something, but not necessarily what the user really wants. But, we could change it so that if you don't set it -- it aborts. To trap this at `create_newcase` time I think we need to add some more functionality into cime.

ekluzek:
I'm going to disagree with you about not having versioning for structure. I think we should in the long run anyway. But, I think the utility of doing so depends on how the scientists want to address this aspect. I think it will be useful to scientists to refer to structure versions independent of model physics.  And I think referring to things that way could be useful in scientific talks and papers. But, we need to hear from more scientists on if they agree with that idea or not. Going forward to describe an experiment for a given version of CTSM, you do have to list four things: CLM_STRUCTURE, CLM_CONFIGURATION, CLM_PHYSICS, and then anything you set that was non-default (eventually these will all become CTSM_*). So we are adding two new things to what you have to define. If you don't have versioning inside of CLM_STRUCTURE -- the versioning for that gets conflated with CLM_PHYSICS, and I'm not sure that's a good thing. But, even if you don't have versioning for CLM_STRUCTURE -- you still have to give it.

ekluzek:
Oh, and I was going to say, we can still start with this, and continue to talk about this in regard to moving forward. I'm not sure we can effectively talk about this inside this PR -- this probably needs more face to face discussion. And as I said the important thing is to hear how scientists want to think about this.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -124,8 +124,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <organic_frac_squared phys="clm4_5" >.true.</organic_frac_squared>
 <organic_frac_squared phys="clm5_0" >.false.</organic_frac_squared>
 
-<soil_layerstruct phys="clm5_0">20SL_8.5m</soil_layerstruct>
-<soil_layerstruct phys="clm4_5">10SL_3.5m</soil_layerstruct>
+<soil_layerstruct structure="fast">5SL_3m</soil_layerstruct>
+<soil_layerstruct structure="standard" phys="clm5_0">20SL_8.5m</soil_layerstruct>
+<soil_layerstruct structure="standard" phys="clm4_5">10SL_3.5m</soil_layerstruct>
```
ekluzek: This is where the CLM_PHYSICS version and the structure are conflated into one thing. And that could be OK. But, another way to do this would be that 10SL_3.5m is structure version standard4.5, and 20SL_8.5m is standard5.0. The compset would pick which structure version goes with which physics. But, then you can refer to combinations like -- "I'm running clm5.0 physics with standard4.5 structure". Rather than having to say -- "I'm running clm5.0 physics with standard structure and the 10SL_3.5m soil depth that was in clm4.5 physics". For previous structure, you might as well use the physics numbering that was used in the past, but it wouldn't have to align that way for moving forward -- unless that is useful.

billsacks:
I'd like to share some more thoughts on my rationale here.

If there is truly a need to be able to independently select the physics version number from the structure version number – e.g., using the default physics from version 5.0 but the default structure from version 4.5 – then I agree that we need separate version numbers for the two. It's not clear to me yet that that's really something that people are going to want to do often enough that we need to make it easy for them to do it. Why is independently selecting a structure version number more important, for example, than independently selecting a version number for some other subset of options? For example, someone might want to run the CLM5.0 configuration but with all snow physics backed out to CLM4.5, but I wouldn't suggest that we have separate version numbers for snow physics options to facilitate that.

If we do have separate structure version numbers, then I think:

(1) We would want to use the same numbering scheme for both physics and structure version numbers. My rationale here is: the point of having these version numbers is to make it easy for a user to revert to the default configuration from some earlier model version. This seems complicated and error-prone if we use a different numbering scheme for physics vs. structure version numbers.

(2) Part of my resistance to having structure version numbers was the idea that we'd have separate version numbers for "fast", "standard", "complex", etc. That seems unnecessary, and I think it's going to be easier to write and understand the namelist defaults file if we have the rule that a given flag sets either a version number or a categorical selector (e.g., "fast" vs. "standard"); not both. In addition, having separate version numbers for "fast", "standard", etc. will add complexity / duplication to the namelist defaults file in the case where some structural thing was the same for fast/standard/etc. before some version, but then diverged starting with a later version. So I'd argue that we should have a single physics version number and a single structure version number, along with a separate flag selecting "fast" vs. "standard" (etc.) structure.

With those two thoughts combined: If we wanted separate structure version numbers, then anywhere where some structural namelist default is controlled by a `phys=` attribute, we would change that to a `structure_version=` attribute. This doesn't add too much complexity, so I'd accept that if people feel this is truly an important capability.

However, if there isn't a real need for this, then I'd still rather not go there: This means that, whenever you want to change the default value of some parameter for a new model version, you need to ask whether this is a "physics" or "structural" parameter. In many cases this will be clear-cut, but what about things like number of snow layers? I fear this will waste our time with unnecessary discussions. In contrast, if you have a single version number, then there is no need to make this judgment call. (I realize that the same argument could be made about the new `structure` vs. `configuration` categories I have added. However, my impression is that we envision those being invoked less often than the version number, so I'm less concerned about those.) 

I wonder if part of the issue here is one of naming. It's confusing that structural things are versioned using a `phys` version number; @ekluzek (and others): would you be more in support of using a single version number if we renamed it to something generic like `version`?

dlawrenncar:
As I discussed with Bill yesterday, I am fine with not versioning the
structure and assuming, at least for out of the box configurations, that
the structure is tied to physics/configuration.  Users can always make
additional modifications on their own as needed for their research.

ekluzek:
I talked to both @dlawrenncar and @barlage to get their take on this. As above @dlawrenncar doesn't see a need for the default version of structure to have a version name. When you look at CLM's past there has been fewer changes in what we are terming "structure" than in model physics which has been more extensive. The number of soil layers has changed a few times, and there was an increase in the number of PFT's allowed in a grid cell. 

In talking with @barlage, the WRF and NOAH-MP default "structure" really hasn't changed in 20 years. So it's unlikely to change in the future. But, there is a strong need to have flexibility to match the soil layers that are currently being used. One concern I had with the current approach that ties structure to physics version is if the NWP and CLM structures would want to change at different times and hence not be tied to the physics version. The bottom line is that the default isn't going to change often enough for that to be a concern. And actually, the default isn't the important point, it's the flexibility to change it to what the user desires.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: OK, from my discussions with people conflating physics version and structure into one thing is OK, because default structure isn't going to change very often.

cime_config/config_component.xml
ekluzek: I'm going to resolve this since we are going to go with the current implementation for now.

billsacks:
@ekluzek thanks a lot for taking the time to think about this, talk with others and share your thoughts. I thought some about this point:

> if the NWP and CLM structures would want to change at different times and hence not be tied to the physics version

I think there are two sub-points here:

1. What if we want a change in structure without any change in the physics?

2. What if we want a change in NWP without any change in CLM (or vice versa)?

I think it would be very reasonable to make new versions more recently than we have in the past - say, once a year or more (as long as we do #454 to keep this maintainable). And it seems completely reasonable to say things like, "Version 5.1 changes defaults for NWP; CLM is identical to version 5.0" or, "Version 5.2 changes the default 'standard' structure, but physics remain the same as version 5.1". i.e., we can have minor version changes where some configurations change but others don't.

Also, addressing point (1), we should probably change the "phys" name to something more generic like "version".

cime_config/config_compsets.xml
billsacks: In discussion with @barlage - he would prefer `I2000Ctsm50NwpSpGswpGs`, and that makes sense to me (and similar changes for the two other compsets).

barlage:
I've spoke with @ekluzek and @billsacks separately on this PR and I believe I understand and agree with the proposed mods. Just a quick clarification, it is the WRF soil structure (with either the Noah or Noah-MP models) that hasn't changed for many years (at least 15). Besides those two, the other land models in WRF have different soil structures. 

My preference here is that the goal of the initial NWP configuration attempts to mimic the Noah-MP configuration in WRF (both structurally and approximating physics complexity) and we maintain the flexibility (which I think we have) to configure similar to other models (either in WRF or other systems) through namelist mods. 

That's why I'm hoping we don't need an NWP versioning and that there would be a single NWP configuration of CTSM versions (though these versions may need to not be so tightly coupled to CESM...maybe that's a separate discussion).

billsacks:
Okay, I have gotten the feedback I want here. I'm going to close this PR and I'll open a new one when this is closer to ready.

PR #684, opened 2019-04-16
Move some hard-coded parameters from code to params.nc file

### LINKED ISSUE #680: Pull some hard-coded parameters out to netcdf file

slevisconsulting:
### Description of changes
- Update params.nc file to contain the previously hard-coded parameters that Katie Dagon recommended moving out of the code (listed here: https://docs.google.com/spreadsheets/d/1cRmcB0xz7CjVP6ljUoCehJx49Usxsore5ma4mWcvQwM/edit#gid=0)
- Read the parameters in the modules where they are used

### Specific notes

@katiedagon @ekluzek

Fixes #680 

Ideally, no

/glade/p/cesm/cseg/inputdata/lnd/clm2/paramdata/clm5_params.c190420.nc
now contains 21 additional parameters.
When ready to run the full test suite, will need to introduce the same parameters to the clm4.5 params file (clm_params.c170913.nc) and update the file name.

Testing performed, if any:
Initial testing of reading parameter csoilc from the new params.nc file with this test:
./create_test ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom --compare ctsm1.0.dev033
gives BFB same answers


slevisconsulting:
@ekluzek I have completed the work for one of the 21 parameters and testing gave bfb. 

Could you (or @billsacks ) confirm that I'm on the right track before I do the work for the remaining 20 parameters? Should take less than 5 minutes to look at my mods.

slevisconsulting:
I followed the design used by ch4Mod and others and my test gave bfb.

In `readParamsMod` you will see this line:
`use CanopyFluxesMod, only : readParams_CanopyFluxes => readParams`
If we go with a different design, we may wish to refactor readParamsMod by renaming all the 'readParams' to the location-specific names, such as 
'CanopyFluxesReadParams'
'readCNGapMortParams'
'readCNMRespParams'
...and the rest in subr. readParameters.

ekluzek:
Ahh, yep, that's the way it was done for others as well. We should be consistent, until we have a reason to diverge. So yes, I think this good -- continue on!

billsacks:
I took a quick look, and this looks fine to me, too.

slevisconsulting:
In my latest commit, I was surprised to see

PASS ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom BASELINE ctsm1.0.dev033:

because I considered one of the code modifications as a change in the order of operations. In particular, where I replaced
`42.75_r8 / 1.0e-6_r8` with the new parameter cp25_yr2000 = 4.275e-05

In any case, it's convenient for my testing that this passed.

slevisconsulting:
Next I need to update the clm4.5 params.nc file and test.

Meanwhile @ekluzek and @billsacks the code is ready for review. It follows the template that you reviewed last time so, even if the mods seem lengthy, the same concept just repeats.

slevisconsulting:
In the Hobart test suite, all FATES cases FAIL because parameters `q10_mr` and `froz_q10` expect the "allpfts" dimension in the params.nc file. So @ekluzek and @billsacks what's your preference:

1) One of you would need to remove the "allpfts" dimension from FATES in FatesSynchronizedParamsMod.F90 because, as far as I know, I can't make changes to FATES

OR

2) I could put the "allpfts" dimension back in the params.nc files but now only for these two parameters.

billsacks:
I have no preference. Do you, @ekluzek ? I'm thinking maybe (2) is easier for now, and we let (1) happen on the fates side at some point, but you have a better sense of fates workflow & coordination.

slevisconsulting:
I will start with the easier option (2) and can always back out if you decide that you disagree @ekluzek .

slevisconsulting:
Hobart test suite now OK (ie only expected failures). 

As stated in 78521d1 above, I get bfb (here against baseline dev035) by overwriting 4 parameters as REAL that are ultimately becoming DOUBLE in the params.nc file. When I remove the overwrites I will get roundoff changes, as I can tell from writing the REAL and DOUBLE values of the 4 parameters to iulog.

slevisconsulting:
dev036 is now available; dev037 and dev038 are expected early next week, so I will wait to update and run the test suites once at that time.

ekluzek:
I agree with the course you did. It takes more coordination to get a change in fates. I'll add an issue to fates to remove it.

ekluzek:
Added an issue in fates about allpft, we'll see what they think about it...

https://github.com/NGEET/fates/issues/529

slevisconsulting:
Update:
- The merge to dev039+ (where + denotes that I had to `git merge escomp/master` to get everything) caused this PR's new CLM50 params.nc file to give BASELINE FAILures. Most likely culprit is the ncdump/ncgen sequence that I had used to reintroduce the allpfts dimension to FATES parameters froz_q10 and q10_mr.
- I have regenerated the new CLM50 params.nc file replacing ncdump/ncgen with a sequence of ncks commands.
- The single test that I was using during development now PASSes again, so I'm running the test-suites on cheyenne and hobart.
- Once they PASS, I will remove the 4 overwrites from the code that ensure BFB and will generate baselines for dev040.

src/biogeophys/CanopyHydrologyMod.F90
```diff
@@ -645,12 +675,12 @@ subroutine FracWet(numf, filter, canopystate_inst, waterstatebulk_inst, waterdia
...
-                dewmxi  = 1.0_r8/dewmx(p)
+                dewmxi  = 1.0_r8/params_inst%dewmx  ! TEMP for BFB; rm bef merge
                 fwet(p) = ((dewmxi/vegt)*h2ocan)**0.666666666666_r8
                 fwet(p) = min (fwet(p),maximum_leaf_wetted_fraction)   ! Check for maximum limit of fwet
                 if (snocan(p) > 0._r8) then
-                   dewmxi  = 1.0_r8/dewmx(p)
-                   fcansno(p) = ((dewmxi/(vegt*6.0_r8*10.0_r8))*snocan(p))**0.15_r8 ! must match snocanmx 
+                   dewmxi  = 1.0_r8/params_inst%dewmx  ! TEMP for BFB; rm bef merge
+                   fcansno(p) = ((dewmxi / (vegt * params_inst%sno_stor_max * 10.0_r8)) * snocan(p))**0.15_r8 ! must match snocanmx 
                    fcansno(p) = min (fcansno(p),1.0_r8)
```
slevisconsulting: @billsacks and @ekluzek 
I have spotted this piece of code that seems wasteful. We could remove dewmxi and change
`fwet(p) = ((dewmxi/vegt)*h2ocan)**0.666666666666_r8`
and
`fcansno(p) = ((dewmxi / (vegt * params_inst%sno_stor_max * 10.0_r8)) * snocan(p))**0.15_r8`
to 
`fwet(p) = (h2ocan / vegt)**0.666666666666_r8`
and
`fcansno(p) = (snocan(p) / (vegt * params_inst%sno_stor_max * 10.0_r8))**0.15_r8`

slevisconsulting: These changes would not be BFB. Should I add comments in the code for a future clean-up?

slevisconsulting:
The new CLM50 param.nc file seemed like it would work on cheyenne.
Unfortunately it generates BASELINE FAILures on hobart.
I have decided to return to the CLM50 param.nc file with the allpfts dimension NOT removed.

slevisconsulting:
The only FAILure I need to resolve on Cheyenne...
`FAIL FUNITCTSM_P1x1.f10_f10_musgs.I2000Clm50SpGs.cheyenne_intel RUN`
appears in `funit.log`
`/gpfs/fs1/work/slevis/git/hardcoded_params_out/src/biogeophys/WaterDiagnosticBulkType.F90(105): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [PARAMUTILMOD]
    use paramUtilMod, only: readNcdioScalar`
`--------^`
`/gpfs/fs1/work/slevis/git/hardcoded_params_out/src/biogeophys/WaterDiagnosticBulkType.F90(105): error #6580: Name in only-list does not exist.   [READNCDIOSCALAR]
    use paramUtilMod, only: readNcdioScalar`
`----------------------------^`

I have not encountered `funit.log` before, which makes me wonder whether this relates to "externals"

@ekluzek and @billsacks 
pls let me know, because I have not found an explanation for this error, yet...

slevisconsulting:
Aha, that last one came from the Unit Tests. I think I can probably figure it out...

billsacks:
@slevisconsulting I have opened a PR into your branch that fixes the unit tests. I have tested on my mac. It would be good if you could double-check this by merging my PR into your branch (you can just click the merge button in the PR interface if you're happy with it), then updating your branch on cheyenne and rerunning the FUNIT test.

Note that my changes only affect unit testing, so no need to rerun other tests in the test suite.

src/biogeophys/CanopyHydrologyMod.F90
billsacks: I'm confused: your new versions don't seem to have dewmx in them at all. Is that a mistake?

I'm guessing that this is written the way it is in an attempt to avoid divisions, since division is a lot more expensive than multiplication. The way it's written now, it probably doesn't accomplish anything. But if the inverse was taken outside of the loop, then it would have some performance benefit to keep this as a multiplication by dewmxi rather than a division by dewmx.

slevisconsulting: You're right, thank you for catching my omission. I have edited my suggested mods to now include params_inst%dewmx in the denominator.

So, yes, in fact I think it's more wasteful than necessary as written, in part because it calculates dewmxi twice in the same loop; also because it calculates dewmxi at all instead of using dewmx in a division that would happen regardless.

If we're attached to keeping dewmxi, then moving it out of the p loop would be more efficient, as you suggest.

slevisconsulting:
All tests PASS on hobart & cheyenne.
Now I will remove the four overwrites and rerun the tests to generate baselines for dev040.

doc/ChangeLog
```diff
@@ -94,24 +87,24 @@ Changes answers relative to baseline:
...
  ko25_coef 
  kc25_coef 
  frac_sat_soil_dsl_init 
-    were not explicitly specified as _r8 when hardwired in the code,
-    while they are now declared "double" in the params.nc files. I
+    that were not declared "_r8" when hardwired in the code,
+    while now they are declared "double" in the params.nc files. I
     showed that the change was no greater than roundoff by writing
     the new and old values to the lnd and cesm log files and confirming
     differences smaller than 1.e-6.
```
slevisconsulting: @ekluzek and @billsacks 
I wanted to confirm that you're ok with my statement here:
Differences start less than 1e-6 when the 4 parameters get read in.
Do these qualify as roundoff differences?

billsacks: Yes, though I'd probably clarify that they are single-precision roundoff-level differences. (Typical roundoff-level differences are double-precision roundoff-level, at the 1e-15 level rather than the 1e-6 level.)

src/biogeophys/CanopyHydrologyMod.F90
billsacks: I think this is something we can come back to later. I may do it in some upcoming cleanup I'll be doing of this module, or we can come back to this at a later time.

slevisconsulting:
@ekluzek and @billsacks 

The test-suites have finished successfully on cheyenne and hobart and baselines have been generated for dev040.

This PR is ready for final review and merge.

billsacks:
@slevisconsulting for the future - note the change I just made to the ChangeLog (we like keeping that section in explicitly, even if it's blank).

src/biogeophys/CanopyHydrologyMod.F90
```diff
@@ -294,8 +327,8 @@ subroutine CanopyHydrology(bounds, &
                       fpi = 0.25_r8*(1._r8 - exp(-0.5_r8*(elai(p) + esai(p))))
                    endif
 
-                   snocanmx = 60._r8*dewmx(p) * (elai(p) + esai(p))  ! 6*(LAI+SAI)
-                   liqcanmx = dewmx(p) * (elai(p) + esai(p))
+                   snocanmx = params_inst%sno_stor_max * (elai(p) + esai(p))  ! 6*(LAI+SAI)
+                   liqcanmx = params_inst%dewmx * (elai(p) + esai(p))
```
slevisconsulting: Hi @swensosc @olyson @dlawrenncar , this is a question about two lines of code in CanopyHydrologyMod. If none of you know the answer, pls add the right person to this conversation:

In this PR I changed snocanmx from varying by dewmx (=0.1) to varying by sno_stor_max (= 6).

When calculating fcansno further down, I replaced the 6 with sno_stor_max without removing dewmx from the calculation. Bill correctly suggests that this has the potential of introducing an inconsistency if we modify the values of dewmx and/or sno_stor_max.

The question is... Which is the preferred formula for snocanmx? The one that varies by sno_stor_max OR the one that varies by both dewmx and sno_stor_max?

src/biogeophys/CanopyHydrologyMod.F90
```diff
@@ -645,12 +674,12 @@ subroutine FracWet(numf, filter, canopystate_inst, waterstatebulk_inst, waterdia
...
                 vegt    = frac_veg_nosno(p)*(elai(p) + esai(p))
-                dewmxi  = 1.0_r8/dewmx(p)
+                dewmxi  = 1.0_r8/params_inst%dewmx  ! wasteful division
                 fwet(p) = ((dewmxi/vegt)*h2ocan)**0.666666666666_r8
                 fwet(p) = min (fwet(p),maximum_leaf_wetted_fraction)   ! Check for maximum limit of fwet
                 if (snocan(p) > 0._r8) then
-                   dewmxi  = 1.0_r8/dewmx(p)
-                   fcansno(p) = ((dewmxi/(vegt*6.0_r8*10.0_r8))*snocan(p))**0.15_r8 ! must match snocanmx 
+                   dewmxi  = 1.0_r8/params_inst%dewmx  ! wasteful division
+                   fcansno(p) = ((dewmxi / (vegt * params_inst%sno_stor_max * 10.0_r8)) * snocan(p))**0.15_r8 ! must match snocanmx
```
slevisconsulting: @swensosc @olyson @dlawrenncar , this is the second line that I referred to in my comment to you a few seconds ago.

swensosc:
if you want to do it robustly and consistently, each phase should have it's
own pft-specific value.  I would also recommend changing the variable
names.  1st, the parameters aren't 'max' values, b/c elai+esai doesn't
saturate at 1.  2nd, dewmx isn't even the right term, b/c it's not simply a
dew-related quantity.  I suggest something like:

snocanmx = params_inst%snow_canopy_storage_scalar(p) * (elai(p) + esai(p))
liqcanmx = params_inst%liq_canopy_storage_scalar(p) * (elai(p) + esai(p))

swensosc:
should fcansno just be:

fcansno(p) = ((snocan(p) / (vegt * params_inst%sno_stor_max)) * )**0.15_r8

w/ no need for dewmxi?

billsacks:
@swensosc your feeling in the last comment is mine, too: that's what led me to open https://github.com/ESCOMP/ctsm/issues/710, which led @slevisconsulting to ask these questions here.

Issue #452, opened 2018-07-26
Add an NWP compset


ekluzek:
We need a NWP compset that will turn on the ctsm-nwp1_0 physics option (see #451) as well as setup certain defaults (rivers and glaciers off).

@barlage @billsacks

billsacks:
This will be an SP compset.

Issue #456, opened 2018-07-27
What default options do we want for the initial NWP configuration?


billsacks:
@barlage @dlawrenncar @martynpclark - can you please use this issue to list the options we want for the initial NWP configuration? My impression is that we're starting with clm5_0 but then changing some options. What I know so far is that we want SP with a few things changed:

- Surface dataset with only one PFT per grid cell

- 4 soil layers

- turn off PHS

- turn off BVOCs (if they're on by default with SP)

What else do we want to change?

dlawrenncar:
I think it may be 2 PFTs, bare soil and the dominant PFT.  We will need to
make a decision about the river model.  This requires updated mapping
files, which we can have out-of-box for a CONUS configuration, but which
would need to be pulled into surface dataset generation workflow for the
more general runtime configurable grids that we will aim for eventually.

billsacks:
Other things:

- Note that Mike's dataset has no urban and lake

- Turn off urban heat stress indices (via calc_human_stress_indices)

ekluzek:
oldflag is essentially what's being used in NOAH-MP right now. So you want it as the out-of-the-box NWP option right now. And of course you want it to be an option that you can change. I also think that setting these means you no longer need STD_ELEV and SLOPE on the surface datasets, which is also good for NWP (since creating those maps takes so much CPU and memory).

Anyway, I think the bottom line is you want to be able to do both. To turn them on, and get them out of the box for NWP.

billsacks:
@ekluzek there are a lot of options in Noah-MP that we aren't going to support or use by default in our initial NWP configuration, which is why it isn't clear to me if we necessarily want these options initially. @barlage @dlawrenncar  and/or @swensosc can you confirm what you want in terms of oldfflag and subgridflag for the initial NWP configuration?

dlawrenncar:
I don't think we want the older version to be in the out-of-the-box NWP
configuration.  We believe that the newer parameterization is an
improvement so it should be out-of-the-box.  Users can revert to the older
version if they need to or want to for their research.

billsacks:
From discussion with @barlage 

- We'll also reduce the number of canopy iterations (to 3)

- No need to have a river model for now

- We _will_ support the NLDAS grid with the standard CLM5 configuration (so will have a standard surface dataset at that resolution).

- We will _not_ (for now) support the NWP option with any other grids (because we won't have the appropriate surface dataset)

PR #685, opened 2019-04-16
Add out-of-the-box NWP configuration

billsacks:
### Description of changes

Adds out-of-the-box NWP configuration.

The new NWP compset triggers two new XML variables: `CLM_CONFIGURATION`
and `CLM_STRUCTURE`. (These can be modified independently, as well. Note
that they will eventually be renamed to `CTSM_CONFIGURATION` and
`CTSM_STRUCTURE`.) These, in turn, control the default values for a
number of namelist settings:
- soil_layerstruct (using a new soil layer structure implemented by
  @barlage)
- itmax_canopy_fluxes (a new namelist flag implemented by
  @slevisconsulting)
- use_hydrstress
- collapse_urban
- n_dom_landunits
- n_dom_pfts

This follows from the initial implementation and discussion in
https://github.com/ESCOMP/ctsm/pull/677.

### Specific notes

- Mike Barlage (@barlage)
- Sam Levis (@slevisconsulting)

- Resolves ESCOMP/ctsm#451
- Resolves ESCOMP/ctsm#452

- New compset (I2000Ctsm50NwpSpGswpGs)
- New XML variables (CLM_CONFIGURATION, CLM_STRUCTURE)
- New namelist variable (itmax_canopy_fluxes)

Testing performed, if any: So far, just:
- `SMS_Ld1_D.f10_f10_musgs.I2000Ctsm50NwpSpGswpGs.cheyenne_intel.clm-default`
- Compared namelists in a case with the new NWP compset vs. a case with
  a similar, non-NWP compset.


bld/CLMBuildNamelist.pm
```diff
@@ -768,6 +777,31 @@ sub setup_cmdl_fates_mode {
...
+   # Error-check and set the 'configuration' and 'structure' namelist flags
+
+   my ($opts, $nl_flags, $definition, $defaults, $nl) = @_;
+
+   my $val;
+
+   my $var = "configuration";
+   $val = $opts->{$var};
+   if ($val eq "clm" || $val eq "nwp") {
+      $nl_flags->{$var} = $val;
```
billsacks: I was confused about the difference between opts and nl_flags. It looks like most/all opts are also added to nl_flags, then accessed from there; is that right? i.e., is what I do in setup_cmdl_configuration_and_structure reasonable?

ekluzek: $opts is treated read-only as it's the command line arguments. nl_flags are any flags or values worth saving as the processing goes on. So this looks correct to me.

billsacks:
I have two questions about possible additional settings, for @barlage , @dlawrenncar and/or @ekluzek 

1. According to #456 , we wanted to turn off urban heat stress indices (via calc_human_stress_indices). I think that note was made before the default was changed: now the default for this flag is "FAST". Do we still want to turn these completely off for the NWP configuration, by default? If so, would you call this a "configuration=nwp" or "structure=fast" option?

2. According to #456 , we wanted to "turn off BVOCs (if they're on by default with SP)". Is this still something wanted for the NWP configuration? If so, can someone please point me to the exact setting(s)?

ekluzek:
My sense would be that calc_human_stress_indices='NONE" would be for structure="fast". The only reason to do it for NWP is for speed. But, there's nothing about NWP itself that says you couldn't run human stress indices, as a matter of fact it makes sense that you would.

Turning off BVOC's means turning off MEGAN, which is a namelist option to build-namelist (-no-megan which would be done as part of CLM_BLDNML_OPTS). Or you could change the namelist defaults file where megan is either set to 0 or 1.

barlage:
Something else to consider for the NWM configuration: nlevsno set to 3 or similar.

billsacks:
@ekluzek @barlage this is ready for final review. Note that this PR just contains the changes for the nwp configuration - not for the nldas grid. I'll combine the two for my final testing.

Here are lnd_in diffs between a case set up with compset `I2000Clm50SpGs` (marked with `<`) and one set up with compset `I2000Ctsm50NwpSpGswpGs` (marked with '>'):

```diff
5c5
<  collapse_urban = .false.
---
>  collapse_urban = .true.
16c16
<  h2osno_max = 10000.0
---
>  h2osno_max = 5000.0
21,22c21,22
<  n_dom_landunits = 0
<  n_dom_pfts = 0
---
>  n_dom_landunits = 1
>  n_dom_pfts = 1
24c24
<  nlevsno = 12
---
>  nlevsno = 5
28c28
<  soil_layerstruct = '20SL_8.5m'
---
>  soil_layerstruct = '5SL_3m'
45c45
<  use_hydrstress = .true.
---
>  use_hydrstress = .false.
143c143
<  itmax_canopy_fluxes = 40
---
>  itmax_canopy_fluxes = 3
203c203
<  stomatalcond_method = 'Medlyn2011'
---
>  stomatalcond_method = 'Ball-Berry1987'
```

barlage:
@billsacks these initial lnd_in diffs look good to me.

bld/CLMBuildNamelist.pm
```diff
@@ -768,6 +777,31 @@ sub setup_cmdl_fates_mode {
...
+sub setup_cmdl_configuration_and_structure {
+   # Error-check and set the 'configuration' and 'structure' namelist flags
+
+   my ($opts, $nl_flags, $definition, $defaults, $nl) = @_;
+
+   my $val;
+
+   my $var = "configuration";
+   $val = $opts->{$var};
+   if ($val eq "clm" || $val eq "nwp") {
```
ekluzek: @billsacks, when you invoke warnings perl doesn't like the use of variable that isn't defined (or explicitly set to undef). Sp it's better to do it this way...

`if ( defined($val) && ($val eq "clm" || $val eq "nwp") ) {`

There are other examples of that throughout.

ekluzek: Same thing below with structure

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -284,9 +289,14 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
...
-<nlevsno phys="clm5_0" >12</nlevsno>
+<nlevsno phys="clm5_0" structure="standard">12</nlevsno>
+<nlevsno phys="clm5_0" structure="fast">5</nlevsno>
 <nlevsno phys="clm4_5" >5</nlevsno>
-<h2osno_max phys="clm5_0" >10000.0</h2osno_max>
+<!-- h2osno_max is more 'configuration' than 'structure'. But since it's
+     tied to nlevsno, we're controlling it via 'structure', like
+     nlevsno, so that defaults for the two remain consistent. -->
+<h2osno_max phys="clm5_0" structure="standard">10000.0</h2osno_max>
+<h2osno_max phys="clm5_0" structure="fast">5000.0</h2osno_max>
```
ekluzek: This is maybe a wider question a nit picky one. But, something I do try to do is to keep important columns to line up. So in this case I like to have the numbers line up by adding extra spaces...

``` xml
<h2osno_max phys="clm5_0" structure="standard">10000.0</h2osno_max>
<h2osno_max phys="clm5_0" structure="fast" .       >5000.0</h2osno_max>
```

To me this makes it easier to read the numbers. But, not everyone agrees with this and for example pylint actually calls this kind of lining up by column noncompliance with the standard.

So I mainly bring it up for discussion purposes. I'd like to decide on our views on this. We could try to always do it, never do it, leave it as is, or let individuals do it differently.

bld/CLMBuildNamelist.pm
billsacks: Done

bld/namelist_files/namelist_defaults_ctsm.xml
billsacks: Done here and elsewhere in my changes in this file.

In this case I agree with you.

As a general rule, I absolutely agree that doing this enhances readability. It carries some cost to add, though for the typical things we align in the Fortran code, I have macros that let me highlight a block and run a single command to vertically align everything properly. If I didn't have those macros, I would argue against it, and I am hesitant to set this as a coding standard unless we provide macros like this for all common editors.

On the other hand, I find that it has a pretty high maintenance cost, for two reasons:

(1) It takes more time to add code or move code around when you need to vertically align everything around it. A very common case is adding a new variable declaration, associate line, etc., with a variable name that is longer than anything else currently there. Then you have to realign all the lines around it.

(2) Even more significantly, this realignment of existing code is a common cause of merge conflicts, which are the bane of my existence. True, these whitespace conflicts are easy enough to resolve once you figure out what happened, but I find it can waste significant time to figure out what happened and be confident in the merge resolution nonetheless.

All in all, I support vertical alignment as a general guideline but not as a hard-and-fast rule. I tend to vertically align most new code that I add, but often don't realign existing blocks of code when I add a new variable, mostly for reason (2): the merge cost does not seem worth the benefit in that case.

A quick google search turned up similar feelings: https://stackoverflow.com/questions/101958/code-formatting-is-lining-up-similar-lines-ok and https://softwareengineering.stackexchange.com/questions/30029/vertical-alignment-yea-or-nay

billsacks:
Thanks for your reviews, @barlage and @ekluzek . Erik, I have addressed your points.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: Good discussion. We should talk more Monday...

ekluzek:
Awesome. Looks good.

PR #687, opened 2019-04-17
Update fates version on release branch

ekluzek:
### Description of changes

Update fates to a more recent "stable" version

### Specific notes

Contributors other than yourself, if any: rgknox ckoven

Are answers expected to change (and if so in what way)? Only for FATES

Any User Interface Changes (namelist or namelist defaults changes)? new fates options

Testing performed, if any: Ran fates test list on cheyenne


ekluzek:
There's a simple build error on hobart so that tests fail on hobart. I had run the Fates test list previously, so either machine/compiler updates interfered, or I didn't run it on this scaled back version of Fates, or it just fails on the non-Fates tests.

Issue #199, opened 2017-12-29
Clean up relationship between h2ocan, snocan and liqcan


billsacks:
_From @billsacks on March 9, 2017 20:39_

Currently, it appears that snocan and liqcan are only set if `snowveg_flag` is ON or ON_RAD; it looks like they are set to 0 otherwise. This tripped me up: I assumed that we would always have h2ocan = (snocan + liqcan), regardless of the value of snowveg_flag.

Furthermore, the updating of h2ocan, snocan and liqcan currently duplicates some code - e.g., with some updates being applied to h2ocan if snowveg_flag is OFF, but to snocan and/or liqcan if snowveg_flag is ON.

I would find it easier to read and modify the code if we always maintained h2ocan = (snocan + liqcan), and if h2ocan were just a diagnostic quantity computed as that sum - so that all fluxes got applied directly to snocan and/or liqcan. That is, anywhere where h2ocan is currently adjusted, we should instead adjust snocan and/or liqcan, and then later re-compute h2ocan as (snocan + liqcan). If snowveg_flag is OFF, I think this would mean applying all fluxes to liqcan.

Caveat: I haven't come to totally understand the relevant code, so I may be misunderstanding something.

_Copied from original issue: NCAR/CLM#1_

billsacks:
_From @JustinPerket on May 30, 2017 23:14_

Hi Bill, I just got onto this github to write this area up in the tech note, so dunno if this issue is still open. If it's still relevant, your understanding and recommendation sounds good to me. Here's a path to implement them:

### 1. Replace uses of ho2can

It looks like ho2can is only explicitly utilized in the code in a couple of places, and can be safely replaced with liqcan+snocan:
- calculation of rpp in CanopyFluxes
- calculation of ecidif and qflx_evap_veg(p)  in CanopyFluxes
- use in water balance check in BalanceCheckMod and endwb in HydrologyDrainage. Averaging for h2ocan in clm_drv_patch2col would have to be replaced with averaging for liqcan and snocan

### 2. Clean up snow vegetation code to remove explicit ho2can updates

**It seems that this is still currently in CanopyFluxes:**

```fortran
         ! Update dew accumulation (kg/m2)                                                                       
         h2ocan(p) = max(0._r8,h2ocan(p)+(qflx_tran_veg(p)-qflx_evap_veg(p))*dtime)

         if (snowveg_on .or. snowveg_onrad) then
[...]

PR #689, opened 2019-04-20
Replace h2ocan with liqcan+snocan: First stage

billsacks:
### Description of changes

This PR is the first step towards cleaning up the relationship between h2ocan, liqcan and snocan. This replaces uses of `h2ocan_patch(p)` with `(liqcan_patch(p) + snocan_patch(p))`, generally via a temporary variable, `h2ocan_new = liqcan_patch(p) + snocan_patch(p)`. In each part of the code where I have made this replacement, I have put in temporary checks to ensure that these two values of h2ocan (`h2ocan_patch(p) vs. liqcan_patch(p)+snocan_patch(p)`) are the same within roundoff. I have run the full aux_clm test suite on these changes to ensure that the two expressions are always the same within roundoff.

This does change answers, because in general, h2ocan_patch differs at the roundoff level from liqcan_patch+snocan_patch. These roundoff-level differences can lead to divergent code behavior. The most obvious example of this is in the check of h2ocan > 0 in subroutine FracWet. I have found that, due to roundoff errors, it is possible for h2ocan_patch > 0 by roundoff, while (liqcan_patch + snocan_patch) <= 0 by roundoff. (Presumably the reverse could also be true, though I haven't seen that – but I haven't spent much time looking for it.)

My spot-checks of differences from baselines have shown some pretty big differences for instantaneous values or daily averages. See the two attached screen shots, one showing differences in ground evaporation (FGEV) for a one-day average from test `SMS_D_Ld1.f09_g17.I1850Clm45BgcCruGs.cheyenne_intel.clm-default`, and one showing differences in sensible heat flux for a single time step from test `SMS_Ld1.f09_g17.I1850Clm50Bgc.cheyenne_intel.clm-drydepnomegan`.
4C972227-7A13-424B-84FB-5D9AC6DE4CDF
F3C37595-0769-4F2E-B0DA-DD6E3F068FDF

I have NOT done any climatological comparisons (e.g., with the diagnostics package) to see if there are any changes in the mean state.

### Specific notes

- Next step towards resolving #199 

YES: see above

Testing performed, if any:
Full aux_clm test suite

### DIFF
--- a/src/biogeophys/CanopyFluxesMod.F90
+++ b/src/biogeophys/CanopyFluxesMod.F90
@@ -357,7 +357,9 @@ contains
     integer  :: jtop(bounds%begc:bounds%endc)            ! lbning
     integer  :: filterc_tmp(bounds%endp-bounds%begp+1)   ! temporary variable
     integer  :: ft                                       ! plant functional type index
-    real(r8) :: temprootr                 
+    ! FIXME(wjs, 2019-04-19) rename this to simply h2ocan
+    real(r8) :: h2ocan_new                               ! total canopy water (mm H2O)
+    real(r8) :: temprootr
     real(r8) :: dt_veg_temp(bounds%begp:bounds%endp)
     integer  :: iv
     logical  :: is_end_day                               ! is end of current day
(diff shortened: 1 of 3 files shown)


billsacks:
@dlawrenncar @olyson @swensosc - I have assigned you all as reviewers, but I don't think you all need to review it. However, I would love sign-off from at least one of you on these changes. My next steps here are to do some code cleanup, but I wanted to open this PR so you could review the answer-changing pieces as well as the temporary code I put in place to demonstrate that any answer changes originate from only roundoff-level differences between old and new terms.

A specific question is: Should we do a longer run with these changes to verify that there are no significant science changes? Or would you like to hold off on doing this, combining it with what I'm sure will be some other probably-roundoff-level changes I'll be making over the next few weeks? Or do you feel this is unnecessary to test, given the in-code checks I have done?

olyson:
I took a look at this, it looks ok.  I don't see any other instances where replacement is needed.
I could maintain a control climatology (e.g., for a present day 2deg run) to evaluate changes like these against. I personally don't see the need to run anything for just these changes because of the in-code checks.

Issue #699, opened 2019-04-26
Canopy liquid and snow capacities not enforced consistently?


billsacks:
I'm not sure if this is actually a bug, but it looks weird to me, and I'd like to at least get someone to sign off on this:

In CanopyHydrology, there is this code for enforcing canopy liquid and snow capacities (corresponding to equations 2.7.8 and 2.7.9 in https://escomp.github.io/ctsm-docs/doc/build/html/tech_note/Hydrology/CLM50_Tech_Note_Hydrology.html):

https://github.com/ESCOMP/ctsm/blob/89ff666b6d73011c966df52fbfc1d203b1287a2c/src/biogeophys/CanopyHydrologyMod.F90#L318-L330

What I don't understand is why there is a conditional on temperature. It looks to me like, if it's above freezing, snow is allowed to accumulate beyond the specified capacity, and if it's below freezing, liquid is allowed to accumulate beyond the specified capacity. Maybe that's dealt with elsewhere (and in any case, the problem probably isn't too large, or at least probably doesn't persist for two many time steps) but this seems weird to me: I would have expected both snow and liquid capacities to be enforced regardless of the temperature. It seems like the tech note makes no mention of temperatures with respect to enforcing the canopy liquid and snow capacities.

@olyson @swensosc @JustinPerket are any of you able to comment on this?

olyson:
I don't really understand the need for this either.  There's no paper on this and Justin is unlikely to respond. Seems like it could be removed and tested to see what effect it has.  I'm happy to do the testing if this is the way we want to go.

billsacks:
Thanks, @olyson . I just talked with @swensosc and it sounds like he also thinks this conditional shouldn't be here - i.e., this block of code should compute both excess liquid and excess snow runoff, regardless of temperature. @olyson if you'd be willing to look at the scientific impact, that would be great. I haven't talked with Dave about it yet, so I'll leave it to you whether you want to check with him first.

olyson:
I'll just test it with an I2000 case and see what happens.

olyson:
Dave suggested creating a tag with this change and folding it into the CTSM testing that I'll be doing, if you think that is acceptable.
On the subject of CTSM testing, Dave and I talked and thought that we would re-run a section near the end of a control historical simulation with the tag in question and examine our regular diagnostics. 
[...]

PR #703, opened 2019-04-30
Remove excess canopy liquid/snow regardless of temperature

billsacks:
### Description of changes

Previously: if it was above freezing, snow was allowed to accumulate
beyond the specified capacity, and if it was below freezing, liquid was
allowed to accumulate beyond the specified capacity. Sean Swenson, Keith
Oleson and Dave Lawrence agreed that this seems wrong.

This commit changes the behavior so that both liquid and snow canopy
capacities are enforced regardless of temperature.

### Specific notes

- Change supported by @swensosc @olyson @dlawrenncar 

Resolves ESCOMP/ctsm#699

We expect answer changes, though probably fairly small. @olyson will investigate this after it comes to master.

Testing performed, if any: none yet

### DIFF
--- a/src/biogeophys/CanopyHydrologyMod.F90
+++ b/src/biogeophys/CanopyHydrologyMod.F90
@@ -315,18 +315,16 @@ contains
                    qflx_snotempunload(p) = 0._r8
                    snounload(p)=0._r8
 
-                   if (forc_t(c) > tfrz) then ! Above freezing (Use t_veg?)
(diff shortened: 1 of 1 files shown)


billsacks:
@olyson @swensosc can one or both of you please review this to confirm that this is the change you were expecting? I recommend viewing the diffs with "No whitespace" (click on the "No whitespace" button when viewing the diffs on GitHub).

PR #751, opened 2019-06-14
Update fates_next_api to release 5.0.21

rgknox:
### Description of changes

@ekluzek brought changes from fates_next_api into release tag 5.0.21  (NOW 5.0.30).  This PR essentially does the reverse, and now updates fates_next_api to be in-line with that tag.   Most of the heavy lifting was already done by bringing fates_next_api into that tag (thanks @ekluzek )

### Specific notes

Contributors other than yourself, if any: @ekluzek , @glemieux 

For FATES simulations, this will bring in myriad updates and improvements into the CLM5 side of things, so I would imagine that any non-prescribed simulations will have different results.

### Testing performed, if any:

https://github.com/ESCOMP/CTSM/pull/751#issuecomment-583048661

Standard regression testing: TBD


ekluzek:
@rgknox awesome. I think this is great. The one thing this brings to my mind is if fates_next_api should be based off of CTSM master -- or the CLM5.0 release branch as it is here. I think it might make sense for it to be CLM5.0 as that's the most tested and least experimental and it's what all the CMIP6 experiments are being done with.

The reason to have it point to CTSM master though is to get all the latest improvements in. For example, the FUN bug was fixed on master and NOT on release-clm5.0. Possibly the way to handle that might be to cherry pick individual changes that go on master, but not release-clm5.0 if they should go into fates_next_api.

So is pointing to release-clm5.0 the way to go for fates_next_api? We should still pull this one in even if we think the long term plan should be to point to master as it's such a long term advancement.

rgknox:
Actually I was wondering that myself @ekluzek .  I thought I would start off bringing in that release tag because it was the simplest (ie reverse of what you did).  But, I think your argument makes sense, and these bug fixes are important to get in!

ekluzek:
OK, are you here next week? Could we maybe talk a bit about this and decide on the longer term plan? If your not here next week we could schedule a telecon for later in the week or in the weeks afterward...

rgknox:
I updated my comment above BTW.  I will not be there next week (I'm sad about that now, I'm already geting FOMO and it hasn't even started).  Lets have a call-in.  This is definitely a question that we want stake-holder feedback on.

rgknox:
@ekluzek , in the mean time, is there a document or note somewhere that gives more specifics on the difference between the clm5 release branch and master?  Or is the difference really simple, in that the clm5 release branch is a lagged version of master?

ekluzek:
@rgknox there isn't a document that cleanly tells the difference between the two. I suppose the only document is the release change log versus the ChangeLog on master. So...

https://github.com/ESCOMP/ctsm/blob/release-clm5.0/doc/release-clm5.0.ChangeLog

vs.

https://github.com/ESCOMP/ctsm/blob/master/doc/ChangeLog

So right now the release branch is ctsm1.0.dev025 plus a list of things (all of which either have or will go onto master).

But, the release-clm5.0 branch is the stable branch for the CESM2.1.x release series. It includes updates for future scenario SSP's for CMIP6 and some of the bug fixes that are deemed critical by the CESM co-chairs. Whereas MASTER includes ALL of our latest developments for the CESM2.2.x series and beyond going into CESM3. So some of those are bleeding edge developments. Some are things that are going into both master AND the release-clm5.0 branches. Some are going to be bug fixes (like the FUN bug fix) that the co-chairs don't want on the CESM2.1.x release branch. The things that have gone into master and not release are things like to support water isotopes and to support the Numerical Weather Prediction (NWP) community.

ekluzek:
@rgknox is this going to go to fates_next_api soon? It looks like it's all approved and everything.

I'm actually working on another fates update on release-clm5.0.

rgknox:
I've updated this to merge in tag 5.0.29.  
Starting testing now.

rgknox:
looks like an unexpected test is failing: >ERP_D_P32x2_Ld3.f19_g16.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef RUN time=77

I will take a look and check back in

rgknox:
hmm, nothing simple is jumping out at me. Here is the offending directory:

> /glade/scratch/rgknox/clmed-tests/fates-sci-1.31.1_api.8.1.0-clm5.0.29-C8dbc45cb-F17aaa879.fates.cheyenne.intel/ERP_D_P32x2_Ld3.f19_g16.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef.G.20200106_124019_lw3ssi

line 644 in subgridAveMod.F90

rgknox:
a ran a 60 year smoke test with the f10 grid, using our default pft set.
https://drive.google.com/file/d/1KtFlEvX3EJp8L7PGqriKXEJBu_ilQSBx/view?usp=sharing
The C4 grasses really dominate, but I think the simulation indicates the models are still working together.
I suppose to track down the problem with the thread-change test, I may just attempt the test after merging in previous release tags, to get a sense of when this test stopped working (method of bisection?)

ekluzek:
The fix for the threading test should be the same as the one I had to do for histFileMod.F90 on the release branch.

Getting this PR in would be a good update to have in, because of the updated externals that fix an issue with using Python3. Python2 is no longer supported so we should make sure we don't require using it.

ekluzek:
For the threading fix, see this commit: https://github.com/ESCOMP/CTSM/commit/2887c561e9972bca166fd0dfc300657505a678c1

rgknox:
Updated this branch to 5.0.30 and ran a 100 year simulation at BCI panama, compared with base (fates_next_api).  Results look surprisingly, exactly the same.  I was expecting that over the last year(s) there would be more process change on the CLM side?
https://drive.google.com/file/d/1HXGRCRj6ZtUOiRVNliwhR2PlfICVNXwj/view?usp=sharing

Anyway, will run tests on cheyenne now.

rgknox:
Ran a 60 year test on the f10 grid, I removed the grass pfts and left our default set of 9 trees and bushes:
https://drive.google.com/file/d/1ixE7MzBFzI2Fa6QN9LfCzT68uHvlWWCp/view?usp=sharing
I ran this test mostly as a big smoke test to see if the simulation generated biomass patterns with the same level of reasonableness as we did before, and it does.  I will run the standard regression suite now.

rgknox:
Tests are expected PASS, did not try b4b, do not expect it:
/glade/scratch/rgknox/clmed-tests/fates-sci.1.33.0_api.8.1.0-clm5.0.30-C0962ed4d-Ffbbcafd0.fates.cheyenne.intel

@glemieux , @ekluzek , thoughts, time to merge up?!

ekluzek:
Sounds good to me. Time to hit the green button!

glemieux:
Nice!  Looks good.

PR #759, opened 2019-07-06
Soil layer definition clean-up and user-defined option

### LINKED ISSUE #279: Clean up soil vertical layers definition
**Moving a comment I made 2015-10-09 to here**

I was looking at the new initVerticalMod on the branch that @bandre-ucar  is about to bring to the trunk, because it ties in with the initInterp vertical interpolation stuff I'm working on. In doing so, I noticed some duplicated code / logic in setting up the vertical structure: It seems like there are basically two options:

1. Fundamentally set zsoi, then set dzsoi and zisoi based on that

2. Fundamentally set dzsoi, then set zsoi and zisoi based on that

However, the code to set zsoi based on dzsoi (or vice versa), and the code to set zisoi based on the others is duplicated for each of the 4(?) options. To me, having this near-duplicated code obscures the real differences between the different methods.

I feel this should look like:

```
if (method 1) then
   code to set zsoi...
   call set_dzsoi_from_zsoi(dzsoi, zsoi)
else if (method 2) then
   code to set dzsoi...
   call set_zsoi_from_dszoi(zsoi, dzsoi)
else if ...
...
end if

call set_zisoi_from_zsoi(zisoi, zsoi)
```

See also #728

### LINKED ISSUE #728: Allow users to specify soil layer structure more flexibly
Rather than (or maybe in addition to?) having a fixed, predefined set of possible soil layer structures, we'd like to let the user specify the soil layer structure arbitrarily via a namelist setting. The user would probably specify a vector of layer thicknesses, though the details of the "user interface" still need to be worked out.

This has been raised as a high priority need by @barlage and Stan Benjamin.

See also #279 , #729

slevisconsulting:
I currently envision the user-defined option getting folded into the clean-up work because I do not expect extensive code changes for either; of course I'm open to different outcomes.

### Description of changes
1) Clean-up following @swensosc and @billsacks suggestions in issue #279

2) Initial proposal for a user-defined option of soil layer structure:
- Use existing namelist variable soil_layerstruct
- Instead of setting to an existing option (e.g. '5SL_3m', etc.),
  set soil_layerstruct = 'user:aa,bbbb,cccc,dddd,eeee,ffff,gggg,...'
  where the string is declared of len=256 and
        'user:' tells the code to expect a user defined structure,
        aa gets assigned to nlevsoi and nlevgrnd in the code,
        bbbb gets assigned to dzsoi(1) in the code,
        cccc gets assigned to dzsoi(2) in the code,
        and so on to dzsoi(nlevgrnd).

This user-defined option should likely be tested with a new (or modified existing) test in the test-suite.

### Specific notes

@billsacks @dlawrenncar @swensosc @ekluzek 

fixes #279
fixes #728

For NWP cases, because had to change nlevsoi from 5 to 4 to work around the current limitation that nlevsoi must be less than nlevgrnd for the model to not abort.

New user-defined option to specify soil layer structure.

Testing performed, if any:
This far I have only tried one test (it PASSed) to make sure that a pre-existing soil layer structure works as it did before the clean-up:
./create_test ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom -c /glade/p/cgd/tss/ctsm_baselines/ctsm1.0.dev049


src/main/initVerticalMod.F90
```diff
@@ -257,24 +251,50 @@ subroutine initVertical(bounds, glc_behavior, snow_depth, thick_wall, thick_roof
...
+       else if (soil_layerstruct == '5SL_3m') then
+          dzsoi(1) = 0.1_r8
+          dzsoi(2) = 0.3_r8
+          dzsoi(3) = 0.6_r8
+          dzsoi(4) = 1.0_r8
+          dzsoi(5) = 1.0_r8
+       else if (soil_layerstruct(1:5) == 'user:') then
+          do j = 1, nlevgrnd
+             ! read string indices 8 to 9, 11 to 12, 14 to 15, and so on
+             read(soil_layerstruct(3*(j+2)-1:3*(j+2)),*) dzsoi(j)
```
slevisconsulting: I need to change this to read 4 characters per dzsoi(j) since the units are meters and 4 characters will allow 
`min(dzsoi(j)) = .001 m`

billsacks:
@slevisconsulting thanks a lot for starting to work on this - this is going to be a big help!

As for the user interface: I feel it would be better to use a new vector-valued namelist variable (or multiple variables, if necessary) for this, rather than trying to do this in the existing string variable. I feel this will be more intuitive for users, make error-checking easier, and it will be easier for us to write and maintain the code.

I think the first thing to decide is: for our out-of-the-box cases, do we want to continue to use a string that gives some shorthand name for a hard-coded layer structure, or do we want to switch to always having a list of dzsoi values in the namelist? The latter will be easier (and argues even more strongly for doing this with a new variable that is a vector of reals rather than maintaining the old string value), but my suggestion should be doable even if we go with the former. I'll elaborate on how I envision the former working:

We have two namelist variables; a string (say, `soil_layerstruct_predefined`) and a vector of reals (say, `soil_layerstruct_userdefined`). For a given run, exactly one of them would appear in lnd_in. The hard-coded defaults are 'UNSET' and a vector of spvals (or some other value that would never appear in practice, like -999). If both are specified in user_nl_clm, that's an error (caught by build-namelist). The Fortran code would then look like this:

```Fortran
  if (soil_layerstruct_userdefined(1) == spval .and. soil_layerstruct_predefined == 'UNSET') then
     call endrun("Must set one")
  else if (soil_layerstruct_userdefined(1) /= spval .and. soil_layerstruct_predefined /= 'UNSET') then
     call endrun("Can't set both")
  end if

  if (soil_layerstruct_userdefined(1) /= spval) then
     nlevgrnd = ...  ! set based on last value that is NOT spval
     ! Here, set various things based on the userdefined values
  else
     ! Here, set various things based on the predefined string
  end if

  ! Here, set things common to both methods
```

Does that make sense, or do you see problems with that approach?

slevisconsulting:
@billsacks this makes sense to me.

slevisconsulting:
Early tests have revealed a couple of preexisting bugs. I will outline them here. Let me know your preferred way to proceed.

This test originally **PASS**ed:
`ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom -c /glade/p/cgd/tss/ctsm_baselines/ctsm1.0.dev049`

But with `soil_layerstruct = '5SL_3m'` the test **FAIL**ed with this error
Subscript #2 of the array DZ has value 6 which is greater than the upper bound of 5
in LakeTemperatureMod.F90 in a loop with use_lch4 = .true.
I replaced dz with dz_lake in line 960 and got past this error.

Next, the same test **FAIL**ed with
Subscript #2 of the array SOM_ADV_COEF has value 11 which is > than the upper bound of 10
in SoilBiogeochemLittVertTranspMod.F90
I included "+1" in the next four statements in SoilBiogeochemStateType.F90
`allocate(this%som_adv_coef_col    (begc:endc,1:nlevdecomp_full+1))
allocate(this%som_diffus_coef_col (begc:endc,1:nlevdecomp_full+1))
this%som_adv_coef_col(c,1:nlevdecomp_full+1)    = 0._r8
this%som_diffus_coef_col(c,1:nlevdecomp_full+1) = 0._r8`
and got past this error.

Next, the same test **FAIL**ed with
Subscript #1 of the array ZSOI has value 6 which is greater than the upper bound of 5
in SoilBiogeochemLittVertTranspMod.F90
The indexing problem is pervasive here and manifests when nlevgrnd = nlevsoi because the code here assumes nlevgrnd > nlevsoi. This is more complex to resolve, and we can install error checks that inform users of this requirement.

billsacks:
So it sounds like the 5-layer soil can't be used with BGC currently? I can't tell from your description if the fixes you put in for (1) and (2) are correct fixes (as far as you can tell) or just temporary workarounds to get the test to pass; also, would those fixes change answers for standard cases?

Maybe we can discuss this further at the meeting tomorrow?

slevisconsulting:
I will respond. We can also discuss further at tomorrow's meeting:
1) Yes, the 5-layer soil cannot be used with BGC currently.
2) My first fix (dz --> dz_lake) is the correct, as far as I can tell, fix of the first bug; it changes answers from dev049.
3) My second fix is a partial fix (correct, I think) to the larger problem that I described with SoilBiogeochemLittVertTranspMod.F90. Due to the complexity of the larger problem, I backed out the second fix and re-tested the 5-layer soil with a 2-line change that worked:
`nlevgrnd = 6` 
`dzsoi(6) = 1`
...and a 1-line change that also worked; here I also confirmed that (as I expected) answers change from dev049:
`nlevsoi = 4`

billsacks:
From today's CTSM software meeting:

In the short-term, to fix the problem @slevisconsulting ran into: Set nlevsoi = 4 for the NWP case. We should then rename `5SL_3m` to `4SL_2m`.

What do we want to do for out-of-the-box cases? Use string or have the new vector of values? For now, maybe stick with string; we may decide to change that later.

The user interface for setting a user-defined soil layer structure should be:
- Specify a vector of real values giving dz for each layer; the length of this vector gives nlevgrnd
- If a user specifies that, then they also need to set a second namelist value that gives the max number of soil layers (i.e., nlevsoi). For now, we require that this be <= (length of dz vector) - 1. (Eventually the requirement will just be that it be <= (length of dz vector).)

slevisconsulting:
Testing to date:
./create_test ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom -c /glade/p/cgd/tss/ctsm_baselines/ctsm1.0.dev049
**PASS when I remove the dz → dz_lake fix in LakeTemperatureMod.F90, else not bfb w dev049**

Rerun w/o comparing to baseline, with the dz → dz_lake fix, and with soil_layerstruct_predefined = '5SL_3m'  **PASS**

Rerun with soil_layerstruct_userdefined = 0.01,0.02,0.04,0.08,0.16,0.32,0.64,1.28,2.56,5.12 and with nlevsoinl = 9  **PASS**

billsacks:
@slevisconsulting - Thanks for your updates! I haven't looked closely yet, but at a glance this looks good.

I was thinking about what testing we could / should have for this. I can think of three possibilities:

(1) A few unit tests of the initialization logic. This would be cleanest if we could combine the logic for initializating nlevsoi and nlevgrnd with the other logic in initVertical, but I'm not sure if that's feasible. Otherwise, the unit tests could call clm_varpar_init followed by clm_varcon_init (to allocate variables) followed by initVertical. Ideally, to test all of the logic, we'd also do the namelist reads, but I think that's awkward right now in a unit test, so instead the unit tests could start by setting the appropriate variables that are typically read from namelists, and then call the above routines.

(2) Introduce a new test type that runs two cases: one with a predefined soil layer structure and one with a user-defined soil layer structure that exactly matches the predefined one; ensure the two are bit-for-bit. This should be quite easy to do. (Caveat: the namelist settings probably need to look like 0.1d0 in order to give double precision that matches the user-defined settings.)

(3) Make sure this new option is covered by some system test, but don't actually use the new test type of (2).

I feel like (1) would be ideal, both to give confidence that the settings are being made correctly now and in the future and to reduce our current over-reliance on system tests. However, it would also take the most work to put in place. With (1), I personally wouldn't feel a need for an additional system test. If we *do* have a system test, (2) would give the most confidence, but it would also contribute to our system test bloat. Overall, I don't have strong feelings on what should be done, though it seems we should probably have *some* testing of this.

slevisconsulting:
Thank you for these suggestions @billsacks. I'm drawn to option (3), partly because it's easiest for me to implement and partly because I suspect that we could go from (3) to (2) rather easily (with your guidance because right now I would not know how).

Regarding the "system test bloat" I wonder whether one of the NWP tests could merge with my "rm_indiv_lunits_and_collapse_to_dom" test at some point. I don't think they need to be 2 tests.

When I'm ready to run the test-suite, I will plan on running without the #760 bug-fix to confirm bfb. Then I will run again with the bug-fix.

billsacks:
> Regarding the "system test bloat" I wonder whether one of the NWP tests could merge with my "rm_indiv_lunits_and_collapse_to_dom" test at some point. I don't think they need to be 2 tests.

Good point. Do you want to go ahead and make this change, folded in with this tag or some other? I see that there are currently 3 NWP tests:

```
aux_clm: ERP_P36x2_D_Ld5.f10_f10_musgs.I2000Ctsm50NwpSpGswpGs.cheyenne_intel.clm-default                      # Include a debug ERP test of the NWP configuration.
aux_clm: SMS_Ld1.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasGs.cheyenne_gnu.clm-default                      # Include a short smoke test covering the nldas2 grid and the I2000Ctsm50NwpSpNldasGs compset, which uses NLDAS datm forcing.
aux_clm: SMS_Ld1.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasRsGs.cheyenne_gnu.clm-default                    # Include a short smoke test covering the nldas2 grid and the I2000Ctsm50NwpSpNldasRsGs compset, which uses NLDAS datm forcing.
```

I'd like to keep at least one that uses the default, out-of-the-box NWP configuration, but one of the above could be changed to include `rm_indiv_lunits_and_collapse_to_dom`. NWP already includes some aspects of `rm_indiv_lunits_and_collapse_to_dom`, but not all. I don't have a good sense of whether the aspects of `rm_indiv_lunits_and_collapse_to_dom` that are *not* included by NWP out-of-the-box might already be sufficiently covered by unit test; if so, we could remove `rm_indiv_lunits_and_collapse_to_dom` entirely. I'll leave that up to you.

slevisconsulting:
@billsacks I'm revisiting this question:
> What do we want to do for out-of-the-box cases? Use string or have the new vector of values? For now, maybe stick with string; we may decide to change that later.

...because I've encountered the following conundrum in the test suite about keeping answers bfb:

The code used to say...
`<soil_layerstruct structure="fast"                  >5SL_3m</soil_layerstruct>
<soil_layerstruct structure="standard" phys="clm5_0">20SL_8.5m</soil_layerstruct>
<soil_layerstruct structure="standard" phys="clm4_5">10SL_3.5m</soil_layerstruct>
`
and now says...
`<soil_layerstruct_predefined structure="fast">4SL_2m</soil_layerstruct_predefined>
<soil_layerstruct_predefined structure="standard" phys="clm5_0">UNSET</soil_layerstruct_predefined>
<soil_layerstruct_predefined structure="standard" phys="clm4_5">10SL_3.5m</soil_layerstruct_predefined>
`

I went to `.../cime_config/testdefs/testmods_dirs/clm` to add `soil_layerstruct_predefined` definitions in all the user_nl_clm files and remembered that **many (most?) clm4.5 tests share the same user_nl_clm as their clm5.0 counterparts.** So adding `soil_layerstruct_predefined = '20SL_8.5m'` would change clm4.5 answers. Replacing these strings with vectors of values will allow different defaults for clm5 and clm4.5 using the code quoted above. We may still lose bfb because replacing (e.g.) this formula 
`dzsoi(j) = dzsoi(nlevsoi) + (((j - nlevsoi) * 25._r8)**1.5_r8) / 100._r8`
with a vector of values may not be exact.

I'm interested in your (and others') opinion.

billsacks:
I may be confused here: why did you need to change the namelist_defaults for structure="standard"? The feeling from last week's meeting was that we should stick with a string for the out-of-the-box cases.

slevisconsulting:
I changed it to UNSET to ensure that the user was aware that they had to set at least one, either the predefined or the userdefined option.

billsacks:
I don't think that's what we want: we want the out-of-the-box behavior to be the same as before - otherwise users will always have to explicitly set this, which will be a pain.

billsacks:
Ah, but I think I see how my earlier suggestions may have been wrong, then. Maybe we need some logic in build-namelist that sets the predefined option to UNSET if the user-defined one is set???

slevisconsulting:
Ok, I'll do **that**.

slevisconsulting:
To combine
`ERP_P36x2_D_Ld5.f10_f10_musgs.I2000Ctsm50NwpSpGswpGs.cheyenne_intel.clm-default`
with
`ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom`
I propose replacing the two with this test
`ERP_P36x2_D_Ld5.f10_f10_musgs.I2000Ctsm50NwpBgcCropGs.cheyenne_intel.clm-default`
What does Gswp do? Should I keep it?

A diff between lnd_in files from the NWP and the rm_indiv... tests tells me that we'd be testing everything from the latter in the new test.

billsacks:
Thanks, your proposal for combining the tests mostly seems mostly good. I like the idea of introducing a test of NWP with BgcCrop. However:

- It seems like your proposed test won't test the toosmall options; do you feel that's okay?

- The one issue I see with your suggestion is that it leaves compset `I2000Ctsm50NwpSpGswpGs` untested. I think we like to have at least one test of each out-of-the-box compset. So as much as I like the idea of combining tests: I think for now you should actually leave the current NWP tests as they are and add a new NWP test like the one you suggest.

The Gswp says to use the GSWP3 datm forcing data. I think we decided to be explicit about the forcing data in the new NWP compsets. So yes, you should leave that in the compset alias, and make sure to use the GSWP3 datm forcing in the compset long name.

slevisconsulting:
> * It seems like your proposed test won't test the toosmall options; do you feel that's okay?

That seems ok because the unit tests check a wide range of toosmall options.

> * [...] add a new NWP test like the one you suggest.

Ok, setting that up now.

slevisconsulting:
So I replaced...
`ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom`
with…
`ERP_P36x2_D_Ld5.f10_f10_musgs.I2000Ctsm50NwpBgcCropGswpGs.cheyenne_intel.clm-default`   
and got **PASS**

@billsacks do you recommend ERP_P36x2 here, or could we get away with ERS?

slevisconsulting:
Also, I have now added this new test:
`ERS_D_Ld5.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-soil_layerstruct`
where user_nl_clm contains:
`nlevsoinl = 4`
`soil_layerstruct_userdefined = 0.1d0,0.3d0,0.6d0,1.0d0,1.0d0`
and I have manually confirmed that it's bfb with a user_nl_clm that sets:
`soil_layerstruct_predefined = '4SL_2m'`

@billsacks I need help setting up the automated confirmation that the two `user_nl_clm`s give bfb same answers.

billsacks:
> do you recommend ERP_P36x2 here, or could we get away with ERS?

As a general rule, we add ERP (and often ERP_P36x2) tests rather than just ERS tests. The reason is that their cost and time isn't much greater (at least for short tests), and they kill multiple birds with one stone: In addition to testing exact restart (like ERS), they also test for sensitivity to processor count (which catches more problems than you might imagine, including forgetting to do an mpi_broadcast on a namelist item) and threading issues if you use _P36x2. So I'd recommend at least ERP, and possibly ERP_P36x2, for your new tests.

slevisconsulting:
The Hobart test-suite PASSed.
I will run the Cheyenne test-suite after I update to dev051 when that becomes available.

slevisconsulting:
Confirmed that modified new test still PASSes.

src/biogeophys/SoilStateInitTimeConstMod.F90
```diff
@@ -389,7 +389,7 @@ subroutine SoilStateInitTimeConst(bounds, soilstate_inst, nlfilename)
           do lev = 1,nlevgrnd
              ! DML - this if statement could probably be removed and just the
              ! top part used for all soil layer structures
-             if ( soil_layerstruct /= '10SL_3.5m' )then ! apply soil texture from 10 layer input dataset 
+             if ( soil_layerstruct_predefined /= '10SL_3.5m' )then ! apply soil texture from 10 layer input dataset
```
billsacks: Similar comment as for SoilHydrologyInitTimeConst

src/main/initVerticalMod.F90
```diff
@@ -213,41 +215,37 @@ subroutine initVertical(bounds, glc_behavior, snow_depth, thick_wall, thick_roof
...
-          dzsoi(j)= 0.5_r8*(zsoi(j+1)-zsoi(j-1))
-       enddo
-       dzsoi(nlevgrnd) = zsoi(nlevgrnd)-zsoi(nlevgrnd-1)
-
-       zisoi(0) = 0._r8
-       do j = 1, nlevgrnd-1
-          zisoi(j) = 0.5_r8*(zsoi(j)+zsoi(j+1))         !interface depths
-       enddo
-       zisoi(nlevgrnd) = zsoi(nlevgrnd) + 0.5_r8*dzsoi(nlevgrnd)
+    if (soil_layerstruct_predefined == '10SL_3.5m' .or. soil_layerstruct_predefined == '23SL_3.5m') then
```
billsacks: Do you know if all of the predefined options are covered by at least one system test? If there is an option that is NOT covered by a system test, it would be good to run a one-time system test comparing your branch against master. (i.e., set up a test on master and your branch, change user_nl in both, run them and ensure bit-for-bit; a 5-day SMS test would be sufficient.)

src/main/initVerticalMod.F90
```diff
@@ -257,24 +255,50 @@ subroutine initVertical(bounds, glc_behavior, snow_depth, thick_wall, thick_roof
...
+          dzsoi(2) = 0.3_r8
+          dzsoi(3) = 0.6_r8
+          dzsoi(4) = 1.0_r8
+          dzsoi(5) = 1.0_r8
+       else if (soil_layerstruct_userdefined(1) /= rundef) then
+          do j = 1, nlevgrnd
+             ! read dzsoi from user-entered namelist vector
+             dzsoi(j) = soil_layerstruct_userdefined(j)
+          end do
+       end if  ! thickness-based options
```
billsacks: It would be good to have a couple of extra conditions in this big if/else if block, basically mirroring some of the checking you do in clm_varctl:
- Before checking userdefined, you should have a block that checks `else if (soil_layerstruct_predefined /= 'UNSET')`, and abort with a message "Unrecognized soil layer structure"
- Finally, you should have an `else` which would be entered if soil_layerstruct_predefined is UNSET and soil_layerstruct_userdefined(1) is undefined: abort in this case, too.

I know you already have checking along these lines in clm_varctl, but the extra checking here would help catch the possibility that the two get out of sync.

(A more ideal situation would be if the code in this module could be combined with the code in clm_varctl, but I'm suspecting that may be difficult because of when each of them needs to be called.)

billsacks:
> I need help setting up the automated confirmation that the two `user_nl_clm`s give bfb same answers.

Okay, I'm going to try to explain this rather than just doing it, so that you can take a stab at it yourself and learn the process for next time.

First, as a general overview: what you're going to do is introduce a new test type, so rather than having an SMS, ERS or ERP test (for example), you're going to have something like a LSOILSTRUCTUD test (I just made that up to stand for: Land SOILSTRUCT User-Defined; maybe you can come up with something better). So, to be clear, this test will *not* test exact restart or changing processor counts, but I think that's okay in this case. (I'm not really concerned about testing exact restart for this change. As I mentioned above, changing processor counts can catch a variety of errors, but I think this new test type will pick up most of those errors, too - given that we already have ERP tests that exercise the given predefined soil layer structure that you'll be comparing against.)

Also, rather than pointing to a testmods directory that turns on the new user-defined soil layer structure option, you're going to point to a testmods that just contains what's in common between the two runs. So I'm thinking you'll want to just use clm-default for this test.

Then there are two things you need to do:

1. Introduce a new block in cime_config/config_tests.xml. This is pretty boiler-plate. You can follow the example of LII:

https://github.com/ESCOMP/ctsm/blob/05fca10c266f66c4de4f338a492c8dddc46e648f/cime_config/config_tests.xml#L30-L38

2. Introduce a new (small) python file that specifies what's unique to this test. This will go in cime_config/SystemTests, and the name of the file must be a lowercase version of the test name. Again, you can follow the template of lii.py. You'll want to change the documentation, the string arguments (brief descriptions) in the __init__ method, and – importantly – the contents that you append to the user_nl files in each case. From looking at the lciso.py test, it looks like you can append multiple lines to a user_nl file via a comma-separated list. This will then create a test with two cases, one with the extra user_nl contents as defined in `_case_one_setup`, and one with the extra user_nl contents as defined in `_case_two_setup`. So I imagine that `_case_one_setup` will set a predefined soil layer structure (like `4SL_2m`), and `_case_two_setup` will set a user-defined structure along with the number of soil layers.

When I'm introducing a new test like this, I always like to confirm that it will fail in the cases where it should fail. Once you have this working, I highly recommend making some temporary change that should cause the test to fail - for example, changing one of the values in the user-defined soil layer structure so that it intentionally differs from `4SL_2m`, and confirming that the test gives a FAIL result in the COMPARE_xxx_yyy phase.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -125,9 +125,12 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
...
 <organic_frac_squared phys="clm5_0" >.false.</organic_frac_squared>
 
-<soil_layerstruct structure="fast"                  >5SL_3m</soil_layerstruct>
-<soil_layerstruct structure="standard" phys="clm5_0">20SL_8.5m</soil_layerstruct>
-<soil_layerstruct structure="standard" phys="clm4_5">10SL_3.5m</soil_layerstruct>
+<soil_layerstruct_predefined structure="fast">4SL_2m</soil_layerstruct_predefined>
+<soil_layerstruct_predefined structure="standard" phys="clm5_0">20SL_8.5m</soil_layerstruct_predefined>
+<soil_layerstruct_predefined structure="standard" phys="clm4_5">10SL_3.5m</soil_layerstruct_predefined>
+
+<soil_layerstruct_userdefined>-9999999._r8</soil_layerstruct_userdefined>
```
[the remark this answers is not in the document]
slevisconsulting: Makes sense.
Deleted.

bld/CLMBuildNamelist.pm
```diff
@@ -2089,10 +2089,24 @@ sub setup_logic_soilstate {
...
+  my $var2 = "soil_layerstruct_userdefined";
+  my $soil_layerstruct_predefined = $nl->get_value($var1);
+  my $soil_layerstruct_userdefined = $nl->get_value($var2);
+
+  if (defined($soil_layerstruct_userdefined)) {
+    if (defined($soil_layerstruct_predefined)) {
+      $log->fatal_error("You have set both soil_layerstruct_userdefined and soil_layerstruct_predefined in your namelist; model cannot determine which to use");
+    }
+  } else {
+    if (not defined($soil_layerstruct_predefined)) {
```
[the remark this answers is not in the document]
slevisconsulting: This also makes sense.
Removed the `if (not defined...`

bld/CLMBuildNamelist.pm
```diff
@@ -2089,10 +2089,24 @@ sub setup_logic_soilstate {
...
-              'structure'=>$nl_flags->{'structure'});
   add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'use_bedrock',
               'use_fates'=>$nl_flags->{'use_fates'}, 'vichydro'=>$nl_flags->{'vichydro'} );
+
+  my $var1 = "soil_layerstruct_predefined";
+  my $var2 = "soil_layerstruct_userdefined";
+  my $soil_layerstruct_predefined = $nl->get_value($var1);
+  my $soil_layerstruct_userdefined = $nl->get_value($var2);
+
+  if (defined($soil_layerstruct_userdefined)) {
```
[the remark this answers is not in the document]
slevisconsulting: Done.

As a result of adding this logic, I reintroduced a modified version of the `if (not defined...` logic discussed in your previous comment. I will commit and push my mods this far, to show you what I mean.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -114,13 +114,25 @@ If TRUE, square the organic fraction when it's used (as was done in CLM4.5)
...
+4SL_2m       = 4 layer soil column, 2m of soil, 0 bedrock layers
+</entry>
+
+<entry id="soil_layerstruct_userdefined" type="real(99)" category="clm_physics"
+       group="clm_inparm" valid_values="" >
+User-defined vector of dzsoi. The length of this vector determines nlevgrnd. When the user sets this vector, they have to set nlevsoinl in the namelist, too; nlevsoinl must be less than nlevgrnd in this version of the model, even though ideally nlevsoinl could also equal nlevgrnd.
+<default>Default: rundef</default>
+</entry>
+
+<entry id="nlevsoinl" type="integer" category="clm_physics"
```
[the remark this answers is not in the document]
slevisconsulting: I chose `soil_layerstruct_userdefined_nlevsoi`.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -1667,6 +1667,18 @@
 
     </machines>
   </test>
+  <test name="ERP_P36x2_D_Ld5" grid="f10_f10_musgs" compset="I2000Ctsm50NwpBgcCropGswpGs" testmods="clm/default">
```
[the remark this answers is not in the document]
slevisconsulting: Added. 
Is it surprising that this test worked and PASSed before I had added the compset here? (I hope not.)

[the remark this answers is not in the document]
billsacks: Yes, very surprising. cime issue incoming....

src/biogeophys/SoilHydrologyInitTimeConstMod.F90
```diff
@@ -211,7 +211,7 @@ subroutine SoilHydrologyInitTimeConst(bounds, soilhydrology_inst)
                 ! do nothing
              else
                 do lev = 1,nlevgrnd
-                   if ( soil_layerstruct /= '10SL_3.5m' )then
+                   if ( soil_layerstruct_predefined /= '10SL_3.5m' )then
```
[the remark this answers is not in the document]
slevisconsulting: I see a few clues and have come to no conclusions so will need input from you guys:
1) The same if-statement a few lines up makes reference to VIC, which suggests that VIC cannot use the predefined `10SL_3.5m` option:
`if ( soil_layerstruct_predefined /= '10SL_3.5m' ) write(iulog,*) 'Setting clay, sand, organic, in Soil Hydrology for VIC'`
2) Next, the if-statement that you point out here repeats the code of the second instance that you point out in SoilStateInitTimeConstMod.F90. This seems problematic for code maintenance. Replace repeating code with a subroutine or function?
3) In the second instance, DML (is that you @dlawrenncar ?) inserted the comment:
`! DML - this if statement could probably be removed and just the top part used for all soil layer structures`
I'm not sure if this means eliminating the else `! duplicate clay and sand values from 10th soil layer` section or something different...
4) Last clue, the predefined `10SL_3.5m` option is clm4.5's default currently.

src/main/initVerticalMod.F90
slevisconsulting: All but one, as far as I can tell.
10SL... gets tested in clm4.5 tests
23SL... gets tested in vrtlay tests
20SL... gets tested in clm5 tests
4SL... gets tested in nwp tests
49SL... DOES NOT GET TESTED
I can do that.

src/main/initVerticalMod.F90
slevisconsulting: The if-block here and the if-block in clm_varpar are a bit different. The former distinguishes by calc_method (node-based vs. thickness-based). The latter distinguishes by whether user-defined and/or predefined have valid entries. However, the former is preceded by the if-block that sets calc_method. This one already mostly handles everything:
- assigns `node-based` for two predefined cases
- assigns `thickness-based` for three predefined and all user-defined cases
- aborts if predefined and user-defined cases were not valid
For completeness, I am adding error checks to this if-block in case both predefined and user-defined are set.

In the if-block that you brought up, I'm adding an error check in case we slip past the if-block that sets calc_method without a valid calc_method.

src/main/initVerticalMod.F90
slevisconsulting: I tried
./create_test ERI_D_Ld9.ne30_g16.I2000Clm50BgcCruGs.cheyenne_intel.clm-vrtlay
with soil_layerstruct_predefined changed to '49SL_10m'  and got **FAIL**  with an abort in soilbiogeochemlittverttranspmod, which may be revealing a bug here, but I'll need to come back to this later.

I went back and tried the test unchanged (soil_layerstruct_predefined = '23SL_3.5m’) and got roundoff DIFF in three variables from dev050. I hadn't expected diffs given my changes. I guess I need to back out changes affecting 23SL_3.5m and try again...

slevisconsulting: Oops, I was supposed to compare to dev051. Trying again.

src/biogeophys/SoilHydrologyInitTimeConstMod.F90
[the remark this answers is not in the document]
billsacks: @swensosc - I also wanted to ping you here. Do you know anything about the above block of code (i.e., the logic that does something different for 10SL_3.5m, maybe related to VIC), and how it might need to be generalized now that we allow user-defined soil layer structures?

@slevisconsulting if this tag gets close to complete, and this is the last thing to resolve, I'm okay with this being left unresolved for now, until we're able to all meet and discuss it.

swensosc:
it could be due to the fact that the surface datasets are still created
based on the original vertical discretization, and then within the code it
is interpolated to other vertical resolutions?

src/main/initVerticalMod.F90
billsacks: Since the point of this is just to make sure the soil layer structure is set up identically to before, you could try a non-BGC (i.e., SP) test with 49SL_10m and see if that is bfb. If that still fails, I'd be comfortable enough if you just print out all of the relevant variables in initialization, before (i.e., on master) and after your changes, and verify that these variables (dzsoi, zsoi, nlevsoi, etc.) are identical in the log files (if you haven't already done so).

billsacks:
> it could be due to the fact that the surface datasets are still created based on the original vertical discretization, and then within the code it is interpolated to other vertical resolutions?

Ah, good point @swensosc : that looks like the right explanation. Thanks! Yes, it looks like the code assumes that PCT_SAND, PCT_CLAY and ORGANIC on the file have 10 layers, and that these layers are defined according to formulas given in this subroutine - and so, if we're also using a matching 10-layer soil structure in the code, then we don't need to do the nearest-neighbor vertical interpolation.

So it seems like we could either:

(1) Keep the logic basically as it is here, distinguishing between whether or not we want to interpolate. To handle general soil layer structures - giving the same answer whether someone uses a predefined structure vs. a user-defined structure that is exactly the same - it seems like it would be best if we generalized the logic a bit here. i.e., rather than checking if we're using 10SL_3.5m, it would be better if, before the column loop, we checked if the soil layer structure in the code matches the soil layer structure assumed on the file (if nlevsoi matches nlevsoifl and if the dzsoi values all match); if so, set some variable like 'soil_structure_matches_file' to true; if not, set that to false.

(2) Use the general-purpose logic always. At a glance, it looks like the general-purpose logic should do just about the right thing if running with 10SL_3.5m, so we could simplify the code by just always using the general-purpose logic. There are three caveats here:

(Caveat a) One difference I see is the squaring of this term when we're not doing interpolation:
https://github.com/ESCOMP/ctsm/blob/842185f8ce5da306c7ad1a7955cffca827239c48/src/biogeophys/SoilHydrologyInitTimeConstMod.F90#L237
Is this difference really intentional? This seems problematic to me: why would we do this when the soil layer structure matches that on the file, but not otherwise?

(Caveat b) I haven't looked super-closely at the code, so may be missing something else that causes things to work differently in the general-purpose case.

(Caveat c) There would be a small performance hit in initialization if we happen to be using the same soil layer structure in the code as on the file, but I'm not too concerned about this. (If it turns out that this part of the code takes a significant amount of time in initialization, a significant optimization would be moving the logic matching soil layers in the model with soil layers on the file to outside the column loop, since this matching is independent of column. i.e., build an array 1..nlevsoi that gives the level on the file for each level in the model. But this probably isn't necessary.)

I'd probably go for option (2), as long as Caveat (a) is reconciled (which may require a separate, answer-changing tag) and then we can show that it's bit-for-bit.

(Side note: it would be better if dzsoi and/or other variables defining the soil layer structure on the file actually appeared on the file rather than being assumed in the code here. But that cleanup / generalization can happen later.)

src/main/initVerticalMod.F90
slevisconsulting: > Oops, I was supposed to compare to dev051. Trying again.

Ok, this PASSed. Now waiting for cheyenne's compute nodes to return to svc...

slevisconsulting:
> (Caveat a) One difference I see is the squaring of this term when we're not doing interpolation:
> https://github.com/ESCOMP/ctsm/blob/842185f8ce5da306c7ad1a7955cffca827239c48/src/biogeophys/SoilHydrologyInitTimeConstMod.F90#L237
> 
> 
> Is this difference really intentional? This seems problematic to me: why would we do this when the soil layer structure matches that on the file, but not otherwise?

I have a vague recollection that this was a conscious choice made a long time ago by the CLM group (at NCAR? the LMWG? I don't remember) to bump up om_frac. The fact that this is now only used with clm4.5 and older makes me think that the need to bump up om_frac has gone away...

billsacks:
> I have a vague recollection that this was a conscious choice made a long time ago by the CLM group (at NCAR? the LMWG? I don't remember) to bump up om_frac. The fact that this is now only used with clm4.5 and older makes me think that the need to bump up om_frac has gone away...

Okay. If this is the explanation, then we probably want to handle this in a different way, rather than tying it in with the logic here. e.g., having a separate namelist variable so that for clm4_5 we bump up om_frac regardless of the soil layer structure you're using. Does that make sense? Should we check with @swensosc and @dlawrenncar on this?

ekluzek:
Yeah, the square was a bug. Since simulations were run with CLM4_5, we left it like is was, but had a namelist option to do the right thing for newer versions. See the bug report here: http://bugz.cgd.ucar.edu/show_bug.cgi?id=1765

slevisconsulting:
> Yeah, the square was a bug. Since simulations were run with CLM4_5, we left it like is was, but had a namelist option to do the right thing for newer versions. See the bug report here: http://bugz.cgd.ucar.edu/show_bug.cgi?id=1765

Interesting. Not sure where my memory stems from :-)

So do we still need clm4.5 to remain unchanged because simulations were run a long time ago?

slevisconsulting:
Out of curiosity I tried
`grep organic_frac_squared /glade/scratch/slevis/tests_0518-234729ch/*lm45*/run/lnd_in`
and got
`organic_frac_squared = .true.`
for all.

ekluzek:
If this is adding a bunch of complexity -- it might be good to change it for all future versions. @dlawrenncar and @swensosc @olyson -- what do you think? Is it OK to change answers for clm4_5 to clean this up?

cime_config/testdefs/testmods_dirs/clm/soil_layerstruct/README
```diff
@@ -1,5 +1,10 @@
 This test sets soil_layerstruct_userdefined to the dzsoi values of the
 predefined case 4SL_2m and checks for bfb same answers.
 
+This is done by
+(a) listing the test in (1) testlist_clm.xml (as all tests) and
+                        (2) config_tests.xml
+(b) creating the file .../cime_config/SystemTests/soilstructud.py
+named after the test in lower case
```
slevisconsulting: @billsacks I summarized here what I did to create the new test type. Do you see anything wrong at this point? I have not tested, yet, because cheyenne's compute nodes are down.

olyson:
If my understanding is correct that this is being done on master then I don't see why we couldn't clean this up.  Any new or re-do simulations I personally would run with CLM4.5 would be with a release tag.

slevisconsulting:
> If my understanding is correct that this is being done on master then I don't see why we couldn't clean this up. Any new or re-do simulations I personally would run with CLM4.5 would be with a release tag.

I think the proposal then is to remove:
1) The bug that squares om_frac for the 10SL option (answer-changing modification in a separate tag before the current PR @billsacks ?)
2) The organic_frac_squared namelist option intended as special treatment of the 10SL option (after (1) is done, this should be bfb and can be done in the current PR)
3) The if-blocks resulting in special treatment of the 10SL option (after (1) is done, this should be bfb and can be done in the current PR)

@billsacks let me know when it's decided that we're going ahead with this and whether you agree with the order I listed above.

billsacks:
@slevisconsulting I agree with your proposal if everyone is okay with this change for clm4_5. This may need final approval from @dlawrenncar , though. In the meantime, at a glance, I think you could maintain the clm4_5 behavior bit-for-bit by keeping this namelist option and applying it regardless of whether we're using 10SL_3.5m. i.e., after setting om_frac and other variables in a generic way (using the same code regardless of whether we're using 10SL_3.5m), then check organic_frac_squared; if it is true, then square the term (again, regardless of whether we're using 10SL_3.5m). Since 10SL_3.5m was the typical soil layer structure for clm4_5, this should maintain the clm4_5 behavior for default cases.

I'm not positive about this, though: I haven't examined the code closely.

Also, I realized that there are two similar blocks of code: one in SoilStateInitTimeConst and one in SoilHydrologyInitTimeConst. They may actually be basically identical (in which case they should be combined); I think one is used for VIC and one in general??? At a glance, we could do the same organic_frac_squared check in SoilHydrologyInitTimeConst (for VIC?), even though it isn't currently there - though again, I haven't looked closely.

To be clear on this: I'm fine with simplifying things by removing this organic_frac_squared check if we get definitive approval, but I'm just suggesting this alternative path forward to maintain answers if you want to move ahead with things before getting this approval.

slevisconsulting:
I have performed the following tests:
1) @billsacks recommended that I submit this test from my branch and again from my master with `soil_layerstruct_predefined = '49SL_10m'` to manually confirm bfb:
`ERP_D_Ld5.f10_f10_musgs.I2000Clm50Vic.cheyenne_intel.clm-vrtlay`
**PASS** because my master in test-id d2vv1z is bfb with my branch in test-id taa26m
2) I followed @billsacks instructions for creating a new test type that automatically confirms bfb between a case using soil_layerstruct_predefined and a case using soil_layerstruct_userdefined with dzsoi values identical to those of the former case:
`./create_test SOILSTRUCTUD_Ld5.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-soil_layerstruct`  **PASS**
3) om_frac clean-up following Bill’s recommendation unfortunately FAILed when compared to dev052:
`ERP_P36x2_D_Ld5.f10_f10_musgs.IHistClm45BgcCruGs.cheyenne_intel.clm-decStart`  **DIFF**
`ERP_D_Ld5.f10_f10_musgs.I2000Clm50Vic.cheyenne_intel.clm-vrtlay`  **PASS**
I will revert to changes that I hope will pass in the next round of testing.

cime_config/testdefs/testmods_dirs/clm/soil_layerstruct/README
billsacks: Great - it looks like you did everything that's needed here. You should now change the testmods directory to simply clm-default, since the `soil_layerstruct` testmods directory no longer does anything beyond default.

Also: Have you done the step of ensuring that the test fails if you introduce a deliberate error (either in the code or in the test's python script – e.g., changing the dz for one of the soil layers)? I view that as an important step when introducing a new test (either a new system test or a new unit test).

slevisconsulting: Ok, done. Thank you for the reminder to perform this sanity check. I ran it and got:
FAIL SOILSTRUCTUD_Ld5.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-default COMPARE_base_4SL_2m

slevisconsulting:
@billsacks I will try simplifying repetitive sections in the code as we talked about, but this far I'm seeing changing answers...

cime_config/testdefs/testmods_dirs/clm/soil_layerstruct/README
billsacks: Perfect, thanks.

billsacks:
> @billsacks I will try simplifying repetitive sections in the code as we talked about, but this far I'm seeing changing answers...

Okay. If it's not too much trouble to isolate which changes change answers, that would be helpful, then we can decide what to do from there.

slevisconsulting:
Status of testing...

I used this single test to breeze through the most recent commits:
./create_test ERP_D_Ld5.f10_f10_musgs.I2000Clm50Vic.cheyenne_intel.clm-vrtlay -p P93300041 -c /glade/p/cgd/tss/ctsm_baselines/ctsm1.0.dev052 &  PASS

After the last commit (f62c5d0) I ran test-suites on hobart and cheyenne and got DIFF from dev052 predominantly in the clm4.5 tests (likely all of them).

I did `git checkout 938f125` and ran both test-suites again.
- Hobart PASSed!
- Cheyenne had a short list of FAILs:
`SMS_Ld1.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasRsGs.cheyenne_gnu.clm-default` RUN  `ERP_P72x2_Lm36.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-clm50cropIrrigMonth_interp` RUN
`ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic` RUN
`ERS_Ly5_P72x1.f10_f10_musgs.IHistClm45BgcCrop.cheyenne_intel.clm-cropMonthOutput` RUN
`SMS_Ld1.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasGs.cheyenne_gnu.clm-default` BASELINE ctsm1.0.dev052: DIFF
I've started by investigating the DIFF. I reran without the _r8 chg in initVerticalMod.F90 and still get DIFF. Next I will look for the last commit when this test worked.

Another issue on Cheyenne is the **long list of tests that FAIL while building** (it's happened twice now). I suspect that the job that submits them is running out of time, not the individual tests. How do I modify that the requested time? I received this job failure (and the previous one) by email:
Post job file processing error; job 7332387.chadmin1.ib0.cheyenne.ucar.edu on host r14i7n12

slevisconsulting:
I resubmitted the cheyenne test suite in `git checkout 938f125` and only one test remained PENDing. I will resubmit it. If it PASSES, then we should be good!

slevisconsulting:
@billsacks this is ready to merge, and I'm open to further review if you prefer.

billsacks:
> Also on Cheyenne I got a **long list of FAILures while building** (it's happened twice now). I suspect that the job that submits them is running out of time, not the individual tests. How do I modify the requested time? I received this job failure (and the previous one) by email:
> Post job file processing error; job 7332387.chadmin1.ib0.cheyenne.ucar.edu on host r14i7n12

Your explanation is likely correct. The default wallclock time for the job that builds the tests is 6 hours. This is set in `python/ctsm/machine_defaults.py`, line 46. Why don't you go ahead and change that to 12 hours (the queue maximum); feel free to commit that change to this branch.

billsacks:
> Why don't you go ahead and change that to 12 hours (the queue maximum); feel free to commit that change to this branch.

Actually, seeing that the changes on this branch are basically complete, and may not need additional testing (I'm doing a final review now): we can just make this change in a future tag.

src/biogeophys/SoilStateInitTimeConstMod.F90
```diff
@@ -358,48 +364,22 @@ subroutine SoilStateInitTimeConst(bounds, soilstate_inst, nlfilename)
...
-             soilstate_inst%csol_col(c,lev)   = spval
-          end do
-
        else
 
           do lev = 1,nlevgrnd
              ! DML - this if statement could probably be removed and just the
              ! top part used for all soil layer structures
-             if ( soil_layerstruct /= '10SL_3.5m' )then ! apply soil texture from 10 layer input dataset 
+             if (.not. organic_frac_squared) then
```
billsacks: Your rework of `organic_fraq_squared` doesn't look quite right to me: `organic_frac_squared` .true. now seems to assume that we're using the same soil layer structure as on the file. Also, did you confirm that `organic_frac_squared` .false. gives the same answers as master when using 10SL_3.5m, since now you'd be using the interpolation code for that case (and my sense from the earlier discussion was that this combo isn't tested in our test suite)?

slevisconsulting:
> I have performed the following tests:
> 
> 1. @billsacks recommended that I submit this test from my branch and again from my master with `soil_layerstruct_predefined = '49SL_10m'` to manually confirm bfb:
>    `ERP_D_Ld5.f10_f10_musgs.I2000Clm50Vic.cheyenne_intel.clm-vrtlay`
>    **PASS** because my master in test-id d2vv1z is bfb with my branch in test-id taa26m
> 2. I followed @billsacks instructions for creating a new test type that automatically confirms bfb between a case using soil_layerstruct_predefined and a case using soil_layerstruct_userdefined with dzsoi values identical to those of the former case:
>    `./create_test SOILSTRUCTUD_Ld5.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-soil_layerstruct`  **PASS**
> 3. om_frac clean-up following Bill’s recommendation unfortunately FAILed when compared to dev052:
>    `ERP_P36x2_D_Ld5.f10_f10_musgs.IHistClm45BgcCruGs.cheyenne_intel.clm-decStart`  **DIFF**
>    `ERP_D_Ld5.f10_f10_musgs.I2000Clm50Vic.cheyenne_intel.clm-vrtlay`  **PASS**
>    I will revert to changes that I hope will pass in the next round of testing.

@billsacks I found it: Point 3 in the above [post](https://github.com/ESCOMP/ctsm/pull/759/#issuecomment-514841290) is where I said that your recommendation gave different answers relative to dev052. In any case, let me know what you find.

billsacks:
@slevisconsulting I think I have fixed the above issues over in #771 . However, I'd be okay with your bringing this to master before #771 is fully reviewed; if so, I would just ask that you change the `if (.not. organic_frac_squared)` conditional back to `if ( soil_layerstruct /= '10SL_3.5m' )then ! apply soil texture from 10 layer input dataset`

slevisconsulting:
@billsacks I think your suggestion makes sense, hence my most recent commit(s). Do you feel this change warrants running the test suites again?

billsacks:
@slevisconsulting - If you feel confident in these changes, I don't feel you need to rerun the full test suite, but please at least run one clm50 and one clm45 test with the final version to ensure that they pass and are bit-for-bit.

slevisconsulting:
> @slevisconsulting - If you feel confident in these changes, I don't feel you need to rerun the full test suite, but please at least run one clm50 and one clm45 test with the final version to ensure that they pass and are bit-for-bit.

Will do.

slevisconsulting:
git pull  ! to get latest push on hobart; then ran these tests

ERP_Ld5_P48x1.f10_f10_musgs.I2000Clm50BgcCruGs.hobart_nag.clm-reduceOutput -c ctsm1.0.dev052  **PASS**
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm45FatesGs.hobart_nag.clm-FatesColdDef -c ctsm1.0.dev052  **PASS**

PR #761, opened 2019-07-13
dz --> dz_lake bug-fix in LakeTemperatureMod.F90 line 960

### LINKED ISSUE #760: dz used instead of dz_lake in LakeTemperatureMod.F90 in a loop with use_lch4 = .true.
### Brief summary of bug

I'm pretty sure that dz should have been dz_lake in this loop, though I'm not an expert on this section of the code.

### General bug information

**CTSM version you are using:** ctsm1.0.dev049-4-gebe59970

**Does this bug cause significantly incorrect results in the model's science?** In use_lch4 = .true. cases.; if we were to turn on biogeochemistry with the NWP configuration, the bug would trigger the error shown below.

**Configurations affected:** use_lch4 = .true. with (for example) soil_layerstruct = '5SL_3m'

### Details of bug

While performing testing in PR #759, the bug triggered this error:

Subscript 2 of the array DZ has value 10 which is greater than the upper bound of 9
65:Image              PC                Routine            Line        Source   
65:cesm.exe  00000000019A2477  laketemperaturemo  960  LakeTemperatureMod.F90

when I ran this test
`ERS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-rm_indiv_lunits_and_collapse_to_dom`
modified to have a soil layer profile of 9 (and later 5) soil layers.
So I believe that the bug triggers the error when the number of soil layers is less than the number of lake layers.

### How I fixed the bug

I wish to know if reviewers agree with my fix:

Change this comment from
`! The CH4 will diffuse directly from the top soil layer to the atmosphere, so`
to
`! The CH4 will diffuse directly from the top lake layer to the atmosphere, so`

Change line 960 (in my version) from
`lakeresist(c) = lakeresist(c) + dz(c,j)/kme(c,j) ! dz/eddy or molecular diffusivity`
to
`lakeresist(c) = lakeresist(c) + dz_lake(c,j)/kme(c,j) ! dz/eddy or molecular diffusivity`

@billsacks recommended I ask @olyson and @dlawrenncar first.

slevisconsulting:
### Description of changes
As described in the title plus a comment correction.

### Specific notes

@billsacks @dlawrenncar 

fixes #760 

Only in diagnostic variable WTGQ

Testing performed, if any:
Cheyenne test suite FAILs with DIFFs from dev049 baseline. 
The following commands...
./summarize_cprnc_diffs -basedir /glade/scratch/slevis/tests_0712-173856ch -testid '*'
vi cprnc.summary.\*.by_varname
...confirm that only WTGQ is different from dev049. 
I will repeat on Hobart.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,114 @@
 ===============================================================
+Tag name:  ctsm1.0.dev050
+Originator(s):  slevis (Samuel Levis,SLevisConsulting LLC,303-665-1310)
+Date:  Mon Jul 15 11:40:09 MDT 2019
+One-line Summary: dz --> dz_lake bug-fix in LakeTemperatureMod.F90 line 960
+
+Purpose of changes
+------------------
+
+ Bug-fix to prevent the model from aborting when running with fewer soil
+ layers than lake layers; not to imply that this was not a bug when the
+ model wasn't aborting. It was.
+
+
+Bugs fixed or introduced
+------------------------
+
+Issues fixed (include CTSM Issue #): #760
+
+Known bugs found since the previous tag (include github issue ID): #759 bug causing model to abort when nlevsoi = nlevgrnd
+
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+
+Does this tag change answers significantly for any of the following physics configurations?
+(Details of any changes will be given in the "Answer changes" section below.)
+
+    [Put an [X] in the box for any configuration with significant answer changes.]
+
+[ ] clm5_0
+
+[ ] ctsm5_0-nwp
+
+[ ] clm4_5
+
+Notes of particular relevance for users
+---------------------------------------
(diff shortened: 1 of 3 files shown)


slevisconsulting:
Ran Hobart's test suite. Then...
`./summarize_cprnc_diffs -basedir /glade/scratch/slevis/tests_0712-173856ch -testid '*'`
`vi cprnc.summary.*.by_varname`
...confirmed that only WTGQ is different from dev049.

@billsacks I generated the baselines for dev050 on cheyenne and hobart while running the above mentioned tests. I will update the /doc directory next.

billsacks:
@slevisconsulting thanks a lot for this. It looks to me like this is ready to go. Is tat right? If so, I'll merge it.

slevisconsulting:
Yes, that's right.

PR #766, opened 2019-07-24
interface changes for fates-parteh (nutrients)

rgknox:
### Description of changes

This set of changes has modifications that support multiple elements (nutrients) in FATES.  There will be more changes in the future that will focus specifically on CLM side hypotheses, and how they couple with fates, however the current set of changes just focuses on the following:
1) update/correction to how the registering of history variables is performed
2) slight modifications to how soil layer/decomposition layers are passed to fates
3) changed name of FATES to CLM litter fluxes
4) modifications to the history file to allow for a second dimension that is length 1.

Note item 4, and this may have impacts on non-fates code.

### Specific notes

This is syncronized with:  https://github.com/NGEET/fates/pull/524

These changes should not change any model states not associated with FATES.  These changes could potentially enable other modules to make use of 2D history output, where the second dimension is length 1.

Any User Interface Changes (namelist or namelist defaults changes)?     no

Testing performed, if any:

Testing is TBD

@bishtgautam , I will be submitting very similar changes to ELM in the near future, including changes to the history.

### DIFF
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -1229,7 +1229,7 @@ contains
            call get_clump_bounds(nc, bounds_clump)
 
            do s = 1,this%fates(nc)%nsites
-              call init_site_vars(this%fates(nc)%sites(s))
+              call init_site_vars(this%fates(nc)%sites(s),this%fates(nc)%bc_in(s) )
               call zero_site(this%fates(nc)%sites(s))
            end do
            
@@ -1306,7 +1306,7 @@ contains
         end if
      end do
      !$OMP END PARALLEL DO
-
+     
    end subroutine init_coldstart
 
    ! ======================================================================================
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -2329,10 +2329,11 @@ contains
    use FatesIODimensionsMod, only : fates_bounds_type
    use FatesInterfaceMod, only : nlevsclass, nlevage
    use FatesInterfaceMod, only : nlevheight
-   use EDtypesMod, only : nfsc, ncwd
-   use EDtypesMod, only : nlevleaf, nclmax
+   use EDtypesMod,        only : nfsc
+   use FatesLitterMod,    only : ncwd
+   use EDtypesMod,        only : nlevleaf, nclmax
    use FatesInterfaceMod, only : numpft_fates => numpft
-   use clm_varpar, only : nlevgrnd
+   use clm_varpar,        only : nlevgrnd
 
    implicit none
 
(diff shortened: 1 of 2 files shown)


src/utils/clmfates_interfaceMod.F90
```diff
@@ -2075,151 +2063,29 @@ subroutine init_history_io(this,bounds_proc)
...
            dim2name = this%fates_hist%dim_bounds(d_index)%name
            call hist_addfld2d(fname=trim(vname),units=trim(vunits),         & ! <--- addfld2d
                               type2d=trim(dim2name),                        & ! <--- type2d
                               avgflag=trim(vavgflag),long_name=trim(vlong), &
                               ptr_patch=this%fates_hist%hvars(ivar)%r82d,    & 
-                              default=trim(vdefault),                       &
-                              set_lake=0._r8,set_urb=0._r8)
+                              default=trim(vdefault))
            
-        case(patch_size_pft_r8)
```
ekluzek: Why does so many of these case options go away?

src/utils/clmfates_interfaceMod.F90
```diff
@@ -123,7 +123,7 @@ module CLMFatesInterfaceMod
    use EDCanopyStructureMod  , only : canopy_summarization, update_hlm_dynamics
    use FatesPlantRespPhotosynthMod, only : FatesPlantRespPhotosynthDrive
    use EDAccumulateFluxesMod , only : AccumulateFluxes_ED
-   use EDPhysiologyMod       , only : flux_into_litter_pools
+   use EDPhysiologyMod       , only : FluxIntoLitterPools
```
ekluzek: Is this a general name change that is taking place? From all lowercase separated by underscores to mixed case without underscores?

rgknox: In general, I'd like us in FATES to use camel case for all procedures (which I believe is consistent with CTSM).  I suppose it would be better to do a complete audit of the procedures and change them all in one go.

src/utils/clmfates_interfaceMod.F90
rgknox: After looking at the code,  all of these options actually had the exact same arguments (or arguments that could be defined as variables in the loop).  They are still there, but lumped together into 1 case select statement:

https://github.com/ESCOMP/ctsm/pull/766/files/946519e0e5cfe70e3b094d07d166e32357436cf1#diff-d5aa89e69afc455e4e84461df1cfa3b6R2076

src/utils/clmfates_interfaceMod.F90
ekluzek: If you don't do it in one go, you'll likely be inconsistent. But, at least need to make it clear that your goal is to move to camel case for procedure names.

PR #768, opened 2019-07-24
Prescribed soil moisture

ekluzek:
### Description of changes\

Changes to run with observed prescribed soil moisture streams rather than prognosing soil moisture. So soil moisture is overridden with the values from the input streams file.

### Specific notes

Are answers expected to change (and if so in what way)? Only when this is turned on (default off)

  New soil_moisture_streams namelist group and items:  use_soil_moisture_streams, soilm_tintalgo,
 stream_fldfilename_soilm,
 stream_year_first_soilm,
 stream_year_last_soilm,
stream_year_align_soilm

Testing performed, if any: Ran some tests with and without:

SMS_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default

@swensosc and @olyson are running some longer simulations to make sure it's reading in the data correctly and responds as expected.


ekluzek:
Note, a problem that I'm annoyed by with this is that it requires that the input soilm stream be on the same grid as you are running on. But, so far I'm not seeing a way for it to ensure that this is the case. So it will happily run and give you strange results even if the grids don't match.

ekluzek:
It's seeming likely that a fix for the grid mismatch problem will require a fix to cime in this issue: https://github.com/ESMCI/cime/issues/3184

ekluzek:
I'm going to just close this PR, because development for this has gone on the release-clm5.0 branch, and now that version just needs to come to master, rather than this version.

Issue #503, opened 2018-09-06
Clean up CanopyHydrology


billsacks:
Before implementing water isotopes in CanopyHydrology (#360 ), there is some general cleanup that needs to be done.

First, we should probably separate this into at least three routines:
- Snow cover fraction
- Snow initialization
- Canopy Hydrology (though it does more than that)

We'll separate flux calculations from state updates so that we can make tracer update calls at the appropriate points.

For now, we'll keep state updates in this module, done at the point where they are currently done (though in separate subroutines). Later, we might merge some state updates (which would change answers).

billsacks:
**Refactoring snow cover fraction calculations in CanopyHydrology**

We'll keep in place the original fsca formulation (n&y 07) - probably giving it a better name than just origfflag. Note that that is the form used in Noah-MP.

Note that, the way things are currently structured, it does the new frac_sno calculation, then possibly overwrites it with the old.

And note that there are order differences for the new vs. old: New needs to compute `frac_sno`, then update `snow_depth` based on that; old needs to first update `snow_depth`, then compute `frac_sno` based on that. So we should have a routine that updates both `frac_sno` and `snow_depth`, and let individual parameterizations do that in whatever order they want.

Then this will be lumped with the FSCA block above it:

https://github.com/ESCOMP/ctsm/blob/5295f61ed21bd6b9ff0d7770d064b1328f030058/src/biogeophys/CanopyHydrologyMod.F90#L548-L554

whereas this will be lumped with the "original fsca formulation" below it:

https://github.com/ESCOMP/ctsm/blob/5295f61ed21bd6b9ff0d7770d064b1328f030058/src/biogeophys/CanopyHydrologyMod.F90#L555-L558

i.e., it is currently assumed that subgridflag==1 is combined with oldfflag==0, and vice versa (see also #502 ).

So we'll end up with code that looks like this:

```Fortran
select case (snow_cover_frac_method)
case (method_snow_cover_frac_clm5)
   call snow_cover_frac_clm5(..., frac_sno, snow_depth)
case (method_snow_cover_frac_niu)
   call snow_cover_frac_niu(..., frac_sno, snow_depth)
end select

subroutine snow_cover_frac_clm5(..., frac_sno, snow_depth)
  ! First, compute frac_sno

  ! Hmmmm, need to determine what to do about the lun%urbpoi check in the current code.
  ! (I feel like we had to deal with something similar for soil hydrology - i.e., urban
[...]

Issue #502, opened 2018-09-06
oldfflag can NOT be used with subgridflag==1, and rename subgridflag


ekluzek:
### Brief summary of bug

In looking at the code in a ctsm meeting we noticed that oldflag can NOT be used with subgridflag==1.

### General bug information

**CTSM version you are using:** ctsm1.0.dev010

**Does this bug cause significantly incorrect results in the model's science?** Yes

Used together results wouldn't make sense.

**Configurations affected:** Clm45 case turning subgridflag=1 and oldfflag=1

### Details of bug

This configuration will probably run, but give screwy results, because snow_depth will be set by frac_snow of the other method, and then frac_snow will be overwritten by the oldffflag method. We just decided this based on code inspection, without doing a simulation.

We also decided that oldfflag is a flag that should be used for the NWP case.

billsacks:
@ekluzek I'm relabeling this as an enhancement because this is really just about adding an error-check on user inputs. My feeling is that "bug - impacts science" should really be restricted to areas where the model is giving scientifically incorrect answers for some relatively common configuration. That, in my mind, labels a serious scientific issue that needs to be addressed, and I feel we should try not to dilute that label too much.

billsacks:
While doing this, we should rename subgridflag to use_subgrid_fluxes, as @swensosc suggested.

billsacks:
@ekluzek you self-assigned this, but I'd be happy to take this on in the course of doing my CanopyHydrology cleanup. So I'll reassign it to myself.

billsacks:
I noticed that none of the error checks in CLMBuildNamelist involving subgridflag were working as intended: subgridflag wasn't being added to the default namelist, so the error checks only kicked in if the user explicitly specified subgridflag in user_nl_clm. I have fixed that in d8d2a2d8 by calling add_default for use_subgrid_fluxes. However, it looks like the same problem (impotent error checks) might exist for other variables checked in `setup_logic_hydrology_switches` (I'm not fixing those for now).

PR #769, opened 2019-07-27
Modularize snow cover fraction method

billsacks:
### Description of changes

This PR moves the calculation of `frac_sno` - and the related updates of
`snow_depth` - into a new set of classes, with one class for each
parameterization (Niu & Yang 2007 and the CLM5 parameterization).

Previously, the code always calculated `frac_sno` the new way, but then
possibly overwrote it if using the older Niu & Yang method. The new code
cleans this up, only doing the calculations that are needed for each
method.

In addition, other code that is specific to one of the two methods is
now moved to a home that makes this dependence on method explicit. This
includes the addition of `newsnow` to `int_snow`: previously, `int_snow` was
always updated using an equation specific to the newer CLM5
parameterization of `frac_sno`, which was not appropriate if using the
Niu & Yang 2007 parameterization; this doesn't make a difference
currently, since `int_snow` is only referenced if using the CLM5
parameterization, but this clears up some confusion. Also, time-constant
parameters read from namelist or the netCDF parameter file now reside in
the appropriate class rather than being more global.

This PR also renames two namelist options to increase clarity:
- `subgridflag` is renamed to `use_subgrid_fluxes`, and is now a logical
- `oldfflag` is renamed to `snow_cover_fraction_method`, and is now a
  string

### Specific notes

Contributors other than yourself, if any: @swensosc provided significant
feedback

- Resolves ESCOMP/ctsm#502
- Resolves ESCOMP/ctsm#503

See notes above

Testing performed, if any:
So far, just these two tests on my mac: both pass and are bit-for-bit:
`SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default`
`SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm45BgcCropQianRsGs.bishorn_gnu.clm-oldhyd`


cime_config/config_compsets.xml
```diff
@@ -143,6 +143,10 @@
     <alias>I2000Clm50BgcCropQianRsGs</alias>
     <lname>2000_DATM%QIA_CLM50%BGC-CROP_SICE_SOCN_SROF_SGLC_SWAV</lname>
   </compset>
+  <compset>
+    <alias>I2000Clm45BgcCropQianRsGs</alias>
```
billsacks: I added this to facilitate quick-turnaround testing on my mac (and elsewhere)

cime_config/testdefs/testmods_dirs/clm/oldhyd/include_user_mods
```diff
@@ -0,0 +1 @@
+../default
```
billsacks: I changed this testmod a bit so it would actually produce output

src/main/clm_instMod.F90
```diff
@@ -214,6 +220,9 @@ subroutine clm_instInit(bounds)
     begc = bounds%begc; endc = bounds%endc 
     begl = bounds%begl; endl = bounds%endl
 
+    call getfil (paramfile, locfn, 0)
+    call ncd_pio_openfile (params_ncid, trim(locfn), 0)
```
billsacks: Note that I open the param file here then pass the ncid into the init function below, rather than opening the file there. There were two reasons for doing it this way:

1. I envision needing to have other parameterizations read from the param file in their init methods, and I was concerned about the performance implications of repeatedly opening and closing the file.

2. This would lend itself better to unit testing, if we ever want to pass in a fake ncid for use in unit tests.

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,767 @@
...
+    character(len=*), parameter             :: subname = 'InitNY07'
+    !-----------------------------------------------------------------------
+
+    if (use_subgrid_fluxes) then
+       write(iulog,*) 'ERROR: Attempt to use N&Y07 snow cover fraction parameterization with use_subgrid_fluxes.'
+       write(iulog,*) 'These two options are incompatible.'
+       call endrun('N&Y07 snow cover fraction parameterization incompatible with use_subgrid_fluxes')
+    end if
+
+    call this%ReadParamsNY07( &
```
billsacks: Note that I call the ReadParams method from inside the Init method (and similarly for the CLM5 parameterization), rather than from readParamsMod. This is similar to how we read namelist values for OO classes, and feels like a more modular way of dealing with parameter reads - as opposed to having a God-like readParamsMod that knows all of the parameterizations that need to read parameters. This also keeps all of the initialization of the class together in one place, rather than having it partially initialized by the Init method and partly by readParamsMod. A final consideration in designing it this way was the possibility that there could be a parameterization that doesn't need to read any parameters: If we wanted the parameter read to be handled by readParamsMod, we'd need to have a public readParams method on this class, which could be confusing if some classes don't actually need any parameters.

@ekluzek and @slevisconsulting I'd suggest that we move to a design like this moving forward for any object-oriented classes that need to read parameters.

billsacks:
@swensosc If you have time, I'd love to get your review of these changes, with a particular focus on:
- A general scientific review: look at new code in SnowCoverFractionMod, and can also look at the code it replaced in SnowHydrologyMod:BulkDiag_NewSnowDiagnostics
- Naming of the methods: are ny07 and clm5 the best names here? I suspect not, especially for "clm5"; what would be a better name for this, which we could use both in the code and as the namelist option? Similarly, can you suggest better documentation in namelist_definition?

@ekluzek If you have time to give this a quick look through, I'd welcome any comments you have, but it probably makes sense to leave the detailed look at the Fortran to @swensosc .

Both @ekluzek and @swensosc One final question I'd like your feelings on is: In the past, I have had a separate file for each class; this would mean one small file for the abstract base class that defines the interface, one small file for the factory method, one file for the ny07 implementation and one file for the clm5 implementation (e.g., see the four files with names starting with "SoilWaterRetentionCurve"). However, I was starting to feel like this may be leading to too much proliferation of small files, and for relatively small parameterizations like this, maybe it's better to pack all of this into one file; I tried this here. I'm on the fence about which is better. Please let me know if you have a preference. Note that, with separate files, I could remove the suffixes on the subroutine names (e.g., the "Clm5" suffix on a bunch of methods) - but I could also keep them if you prefer having uniquely named subroutines.

billsacks:
With the latest changes, the full test suite is passing & bit-for-bit.

In addition, I compared against master (with changes on master to support
the new tests, including changing oldhyd to inherit from default, and
having an oldhyd_monthly that inherits from monthly rather than default):
- ERP_D_P36x2_Ld3.f10_f10_musgs.I2000Clm45BgcCrop.cheyenne_gnu.clm-no_subgrid_fluxes:
bit-for-bit
- SMS_D_Ld1_P48x1.f10_f10_musgs.I2000Clm45BgcCrop.hobart_nag.clm-oldhyd:
bit-for-bit
- SMS_Ly3.f10_f10_musgs.I2000Clm45BgcCrop.cheyenne_intel.clm-oldhyd_monthly:
bit-for-bit

So I feel this branch is ready after it has been reviewed.

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,785 @@
...
+             if(snow_depth(c) > 0.0_r8)  then
+                frac_sno(c) = tanh(snow_depth(c) / (2.5_r8 * this%zlnd * &
+                     (min(800._r8,newsnow(c)/snow_depth(c))/100._r8)**1._r8) )
+             end if
+          else
+             snow_depth(c) = 0._r8
+             frac_sno(c) = 0._r8
+          end if
+       end if ! end of h2osno_total > 0
+    end do
```
swensosc: I wonder if this could be written in such a way to reduce redundancy?  Like this maybe:
  do fc = 1, num_c
       c = filter_c(fc)

! initialize frac_sno
    frac_sno(c) = 0._r8
! reset snow_depth if no h2osno    
    if (h2osno_total(c) == 0.0) snow_depth(c) = 0._r8
! update snow depth with new snowfall
   snow_depth(c) = snow_depth(c) + newsnow(c) / bifall(c)
! calculate frac_sno

          if(h2osno_total(c) < 1.0_r8)  then
             frac_sno(c)=min(frac_sno(c),h2osno_total(c))
	  else
	     if(snow_depth(c) > 0.0_r8)  then
             frac_sno(c) = tanh(snow_depth(c) / (2.5_r8 * this%zlnd * &
                  (min(800._r8,(h2osno_total(c)+ newsnow(c))/snow_depth(c))/100._r8)**1._r8) )
             end if
          end if
enddo

one thing I don't like is this though: if (h2osno_total(c) == 0.0) snow_depth(c) = 0._r8
it seems like a function doing the snow_depth reset should be called any time h2osno is 
updated, and then it wouldn't need to be done here.

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,785 @@
...
+             frac_sno(c) = this%FracSnowDuringMelt( &
+                  c            = c, &
+                  h2osno_total = h2osno_total(c), &
+                  int_snow     = int_snow(c))
+          end if
+
+          ! update fsca by new snow event, add to previous fsca
+          if (newsnow(c) > 0._r8) then
+             fsno_new = 1._r8 - (1._r8 - tanh(this%accum_factor * newsnow(c))) * (1._r8 - frac_sno(c))
+             frac_sno(c) = fsno_new
```
swensosc: is there a benefit to keeeping this fsno_new variable?

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,785 @@
...
+          else
+             ! for uniform snow cover
+             snow_depth(c)=snow_depth(c)+newsnow(c)/bifall(c)
+          end if
+
+       else ! h2osno_total == 0
+          ! initialize frac_sno and snow_depth when no snow present initially
+          if (newsnow(c) > 0._r8) then
+             z_avg = newsnow(c)/bifall(c)
+             fmelt=newsnow(c)
```
swensosc: can fmelt be removed?

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,785 @@
+module SnowCoverFractionMod
```
swensosc: It feels like most of this module is infrastructure.  Is there any way to make the 'science' code more front-and-center?

billsacks: I have a few ideas, and would be interested to hear what you think of them, or if you have other thoughts:

1. I could move the definition of the abstract snow_cover_fraction_type and the whole abstract interface block near the top of this module into its own module. This is what I have done for other object-oriented parts of the code; the downside is just having more files. That would remove about 85 lines from the start of the file.

2. I could also move the factory method (CreateAndInitScfMethod) into its own module. Again, this is what I have done for other object-oriented parts of the code; the downside is again having more files. (Unfortunately, this cannot be combined with the module for the abstract type, because that introduces a circular dependency.) This would remove about 50 lines of code. Alternatively, I could move this method to the bottom of the file.

3. I could move the various science routines to above the infrastructurey routines that are just used in initialization. If I do this, I'd like to know if you'd prefer:

(a) The different parameterizations each get their own file

(b) The different parameterizations stay together in a single file, with all subroutines for a given parameterization grouped together (as they are now), but just change the order of these routines within each grouping so that the science routines come first.

(c) The different parameterizations stay together in a single file, with two sets of groupings: first we have all of the science routines for NY07 followed by all of the science routines for SL12; then we have all of the initialization routines for NY07 followed by all of the initialization routines for SL12.

(d) The different parameterizations stay together in a single file, with the groupings flipped around: first we have UpdateSnowDepthAndFracNY07 followed by UpdateSnowDepthAndFracSL12; then we have AddNewsnowToIntsnowNY07 followed by AddNewsnowToIntsnowSL12; etc.

I am totally happy to do any of these, so please let me know what you'd find most helpful.

src/biogeophys/SnowCoverFractionMod.F90
billsacks: Good point. I'll take a stab at something like this. I'm not sure that the code you wrote is exactly right as is (particularly for the initial setting of frac_sno in the h2osno_total < 1 condition), but I see what you're getting at, and I can try for something similar.

The `h2osno_total` < 1 block makes this messier, and in looking more closely, I'm not sure it's really correct: although I don't understand what it's trying to do, it feels to me like this should be looking at (h2osno_total+newsnow), both in the condition and in the min, and that a similar min should be applied in the case where h2osno_total==0 but newsnow > 0. But I'm reluctant to make that change without really understanding it, so unless you give me guidance otherwise, I'll plan to keep the current code logic when doing the restructuring, even if this makes things less clean and consistent than they could be.

I agree that it feels like the snow_depth reset should happen somewhere else, but determining all the places that would need to happen would take some time; for now, I'll probably just add a note to look back at that.

src/biogeophys/SnowCoverFractionMod.F90
billsacks: No benefit; I'll remove it: thanks.

src/biogeophys/SnowCoverFractionMod.F90
billsacks: Good point; I'll remove it.

src/biogeophys/SnowCoverFractionMod.F90
ekluzek: I think personally I like the idea of an abstract interface for the infrastructure, and then a ny07 implementation and clm5 (still not sure I like calling it the clm5 version) implementation. I think it's OK to have the different implementations in different files, and maybe better actually. The science is then pretty much all front and center in either of the implementations. I picture it can be useful to compare the different implementations, but then having them in different files is actually helpful. Since, the implementations are quite different it seems unlikely that you'd have shared code. And if you do you could probably put in in a base class.

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,785 @@
...
+
+       class(snow_cover_fraction_type), intent(inout) :: this
+       type(bounds_type), intent(in) :: bounds
+       type(column_type), intent(in) :: col
+       type(glc_behavior_type), intent(in) :: glc_behavior
+       character(len=*), intent(in) :: NLFilename ! Namelist filename
+       type(file_desc_t), intent(inout) :: params_ncid ! pio netCDF file id for parameter file
+     end subroutine Init_Interface
+
+     subroutine UpdateSnowDepthAndFrac_Interface(this, bounds, num_c, filter_c, &
```
ekluzek: Personally I still like to have at least a short description of the subroutine

src/biogeophys/SnowCoverFractionMod.F90
```diff
@@ -0,0 +1,785 @@
...
+          if(h2osno_total(c) < 1.0_r8)  then
+             frac_sno(c)=min(frac_sno(c),h2osno_total(c))
+          end if
+
+       else !h2osno_total == 0
+          ! initialize frac_sno and snow_depth when no snow present initially
+          if (newsnow(c) > 0._r8) then
+             snow_depth(c) = newsnow(c) / bifall(c)
+             if(snow_depth(c) > 0.0_r8)  then
+                frac_sno(c) = tanh(snow_depth(c) / (2.5_r8 * this%zlnd * &
```
ekluzek: There are a couple magic numbers that could be made parameters here.

bld/CLMBuildNamelist.pm
```diff
@@ -1682,7 +1682,12 @@ sub process_namelist_inline_logic {
   #######################################################################
   # namelist groups: clm_hydrology1_inparm and clm_soilhydrology_inparm #
   #######################################################################
-  setup_logic_hydrology_switches($nl);
+  setup_logic_hydrology_switches($opts, $nl_flags, $definition, $defaults, $nl);
+
+  #######################################################################
+  # namelist group: scf_clm5_inparm                                     #
+  #######################################################################
+  setup_logic_scf_clm5($opts, $nl_flags, $definition, $defaults, $nl);
```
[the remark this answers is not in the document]
billsacks: Sorry that I wasn't more clear about this: I used "clm5" somewhat as a placeholder just because that was the tentative name used in the discussion in #503 , and I was waiting for a better name. @swensosc has now suggested the name "sl12" (for Swenson & Lawrence 2012), and I will switch the references to clm5 to sl12 throughout this PR.

cime_config/SystemTests/lwiso.py
```diff
@@ -0,0 +1,44 @@
+"""
```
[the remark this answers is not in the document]
billsacks: Yes, this is unrelated.

src/biogeophys/SnowCoverFractionMod.F90
billsacks: Thanks for your thoughts @ekluzek . You raise good points about the advantages of having separate files. So if I understand correctly, it sounds like your preferences on (3) would be (a) (which I think would also imply doing (1) & (2)) then (b). I'll talk more with @swensosc about this next week to see if he agrees.

src/biogeophys/SnowCoverFractionMod.F90
billsacks: I'd rely on scientists (@swensosc 	or others) to provide meaningful names for the parameters.

src/biogeophys/SnowCoverFractionMod.F90
billsacks: Done in 3daf3066

src/biogeophys/SnowCoverFractionMod.F90
billsacks: Fixed in 24ffe4e6. I have run a 3-year test (`SMS_Ly3.f10_f10_musgs.I2000Clm45BgcCrop.cheyenne_intel.clm-oldhyd_monthly`) and verified that it is bit-for-bit with master (despite the theoretical potential for this commit to change answers, as noted in the log message for the commit).

billsacks: @swensosc it might be good if you could look at the end result of 24ffe4e6 to make sure it (a) looks correct, and (b) accomplishes what you want.

swensosc:
this looks better to me; cleaner code w/o redundancy.  I don't know the
reason for the last if block, so I agree that it should be left there for
now.

PR #770, opened 2019-07-30
fixes to transpiration boundary condition to FATES-HYDRO

rgknox:
### Description of changes

During canopy_fluxes(), only patches within the exposed vegetation filter are updated with valid transpiration rates.  We had previously been passing in all values to the fates patches, for use in FATES-HYDRO.  However, this caused fates patches that were not exposed to use uninitialized transpiration rates, instead of zero.  This PR applies a fix so that transpiration is zero in FATES patches by default, and updated only in patches that are in the exposed filter.

### Specific notes

This is NOT a backwards incompatible api change, and is NOT synced with a FATES side change. 

https://github.com/NGEET/fates/issues/559

Yes, this may change answers, but only for obscure cases when FATES-HYDRO is active.

Testing performed, if any:

TBD

### DIFF
--- a/src/biogeophys/CanopyFluxesMod.F90
+++ b/src/biogeophys/CanopyFluxesMod.F90
@@ -1266,7 +1266,7 @@ contains
          
          
          call clm_fates%wrap_accumulatefluxes(nc,fn,filterp(1:fn))
-         call clm_fates%wrap_hydraulics_drive(bounds,nc,soilstate_inst, &
+         call clm_fates%wrap_hydraulics_drive(bounds,nc,fn,filterp(1:fn),soilstate_inst, &
                waterstate_inst,waterflux_inst,solarabs_inst,energyflux_inst)
 
       else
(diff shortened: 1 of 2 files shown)


rgknox:
I was curious why my test simulations were generating a simulation where I had a FATES patch that was not in the exposed vegetation filter. I was indeed able to confirm that this was a patch that did not have any cohorts on it, and therefore did not have any exposed vegetation.  I will go ahead and test, and then integrate pending tests.

rgknox:
re-tested, all PASS (all b4b with base except for the test with the large set of history variables, which makes sense to me).

/glade/scratch/rgknox/clmed-tests/fates-sci-1.31.1_api.8.1.0-C097f4239-F17aaa879.fates.cheyenne.intel

glemieux:
> re-tested, all PASS (all b4b with base except for the test with the large set of history variables, which makes sense to me).
> 
> /glade/scratch/rgknox/clmed-tests/fates-sci-1.31.1_api.8.1.0-C097f4239-F17aaa879.fates.cheyenne.intel

Looked at the AllVars non B4B and the particular baseline for CTSM it was tested against and I think this is just that the CTSM baseline didn't have PR 820 in it's commit history.  The RMS differences are all in the variables having to do with `hlm_model_day`.

PR #788, opened 2019-08-15
Soil texture interpolation bug-fix

### LINKED ISSUE #772: Interpolation of clay, sand and om_frac can use the wrong layer
### Brief summary of bug

It looks like the vertical interpolation of clay, sand and om_frac from the surface dataset to the model's soil layer structure will sometimes use a different soil layer (off by 1) from what should ideally be used.

### General bug information

**CTSM version you are using:** ctsm1.0.dev052

Note that I have made some fixes to this interpolation in #771 , but these don't address this fundamental issue.

**Does this bug cause significantly incorrect results in the model's science?** Incorrect, yes, but probably not *significantly* incorrect.

**Configurations affected:** Potentially all configurations that use a different soil layer structure than 10SL_3.5m (so, by default, all CLM5 configurations). I haven't checked what is actually affected in practice.

### Details of bug

This refers to the logic here:

https://github.com/ESCOMP/ctsm/blob/842185f8ce5da306c7ad1a7955cffca827239c48/src/biogeophys/SoilStateInitTimeConstMod.F90#L398-L404

I think it's incorrect to compare zisoi in the model with zisoi on the file. Instead, I think we should be comparing zsoi in the model with zisoi on the file - asking if the *node center* of a given layer in the model falls in between two interface depths on the file.

A simple example illustrating the problem with the current code is if the zisoi values on the file are shifted by roundoff from the zisoi values in the model; e.g.:

zisoi = 0.1, 0.2, 0.3, 0.4
zisoifl = 0.099999999, 0.199999999, 0.299999999, 0.399999999

In this case, I'm pretty sure that we would put the file's values from level 3 into the model's level 2, etc. - whereas I think we'd want to put the file's values from level 2 into the model's level 2, etc. in this case: In addition to generally feeling that we should be using the model's node center (zsoi rather than zisoi) in this comparison, it also feels wrong to have an algorithm that is so sensitive to roundoff-level differences like this.

slevisconsulting:
### Description of changes
Referring to 
https://github.com/ESCOMP/ctsm/blob/842185f8ce5da306c7ad1a7955cffca827239c48/src/biogeophys/SoilStateInitTimeConstMod.F90#L398-L404

Changes as recommended by @billsacks in #772:
"... we should be comparing zsoi in the model with zisoi on the file - asking if the node center of a given layer in the model falls in between two interface depths on the file."

### Specific notes

CTSM Issues Fixed (include github issue #): Fixes #772 

Yes. Quoting from #772 "... the vertical interpolation of clay, sand and om_frac from the surface dataset to the model's soil layer structure will sometimes use a different soil layer (off by 1) from what should ideally be used."

No

Testing performed, if any:
None

### DIFF
--- a/src/biogeophys/SoilStateInitTimeConstMod.F90
+++ b/src/biogeophys/SoilStateInitTimeConstMod.F90
@@ -158,6 +158,7 @@ contains
     integer            :: begp, endp
     integer            :: begc, endc
     integer            :: begg, endg
+    integer :: found  ! flag that equals 0 if not found and 1 if found
     !-----------------------------------------------------------------------
 
     begp = bounds%begp; endp= bounds%endp
@@ -371,20 +372,26 @@ contains
                 sand = sand3d(g,1)
                 om_frac = organic3d(g,1)/organic_max
              else if (lev <= nlevsoi) then
+                found = 0  ! reset value
+                ! Search within the dataset's range of zisoifl values
                 do j = 1,nlevsoifl-1
-                   ! NOTE(wjs, 2019-08-01) It appears that the code currently doesn't set
-                   ! clay, sand and om_frac explicitly under some conditions. It probably
-                   ! should. My understanding is that currently things work okay, though,
-                   ! because clay, sand and om_frac will remain set at their previous
-                   ! values, which is probably reasonable enough. See also
-                   ! <https://github.com/ESCOMP/ctsm/pull/771#discussion_r309509596>.
-                   if (zsoi(lev) > zisoifl(j) .AND. zsoi(lev) <= zisoifl(j+1)) then
+                   if (zisoi(lev) > zisoifl(j) .AND. zisoi(lev) <= zisoifl(j+1)) then
                       clay = clay3d(g,j+1)
                       sand = sand3d(g,j+1)
                       om_frac = organic3d(g,j+1)/organic_max
+                      found = 1
                    endif
                 end do
-             else
+                if (found == 0) then  ! go outside dataset's range of zisoifl
+                   do j = 1,nlevsoifl-1
+                      if (zisoi(lev) <= zisoifl(j) .or. zisoi(lev) > zisoifl(j+1)) then
+                         clay = clay3d(g,j+1)
+                         sand = sand3d(g,j+1)
+                         om_frac = organic3d(g,j+1)/organic_max
+                      end if
+                   end do
+                end if
+             else  ! if lev > nlevsoi
                 clay = clay3d(g,nlevsoifl)
                 sand = sand3d(g,nlevsoifl)
                 om_frac = 0._r8
--- a/src/biogeophys/SoilStateInitTimeConstMod.F90
+++ b/src/biogeophys/SoilStateInitTimeConstMod.F90
@@ -376,14 +376,10 @@ contains
              else if (lev <= nlevsoi) then
                 found = 0  ! reset value
                 ! For remaining model soil levels, search within the dataset's
-                ! range of zisoifl values. Look for model interface depths
-                ! that are between the dataset's interface depths. IN ONE OF
-                ! THE NEXT FEW COMMITS I WILL REPLACE the first occurrence of
-                ! "interface" with "node" in this comment and in the three
-                ! corresponding if-statements below. At that time clean up this
-                ! comment, too.
+                ! range of zisoifl values. Look for model node depths
+                ! that are between the dataset's interface depths.
                 do j = 1,nlevsoifl-1
-                   if (zisoi(lev) > zisoifl(j) .AND. zisoi(lev) <= zisoifl(j+1)) then
+                   if (zsoi(lev) > zisoifl(j) .AND. zsoi(lev) <= zisoifl(j+1)) then
                       clay = clay3d(g,j+1)
                       sand = sand3d(g,j+1)
                       om_frac = organic3d(g,j+1)/organic_max
@@ -394,7 +390,7 @@ contains
                 ! If not found, search below the dataset's range of zisoifl
                 ! depths
                 if (found == 0) then
-                   if (zisoi(lev) > zisoifl(nlevsoifl)) then
+                   if (zsoi(lev) > zisoifl(nlevsoifl)) then
                       clay = clay3d(g,nlevsoifl)
                       sand = sand3d(g,nlevsoifl)
                       om_frac = organic3d(g,nlevsoifl)/organic_max
@@ -403,7 +399,7 @@ contains
                 end if
                 ! If still not found, search closer to the surface again
                 if (found == 0) then
-                   if (zisoi(lev) <= zisoifl(1)) then
+                   if (zsoi(lev) <= zisoifl(1)) then
                       clay = clay3d(g,1)
                       sand = sand3d(g,1)
                       om_frac = organic3d(g,1)/organic_max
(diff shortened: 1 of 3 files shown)


slevisconsulting:
@billsacks my committed change addresses #772; did you have additional changes in mind for this PR?

billsacks:
Thanks @slevisconsulting ! I don't think I gave this careful thought before now, but what you've done looks correct to me. (I wasn't sure if more changes would be needed to fix this, but if you've given this some thought and it seems right to you, then this sounds good.) Two questions about possible additional changes:

1. Would it make sense to also remove the block above the one you changed:

```Fortran
             if (lev .eq. 1) then
                clay = clay3d(g,1)
                sand = sand3d(g,1)
                om_frac = organic3d(g,1)/organic_max
```

instead incorporating that into the general-purpose logic? It looks to me like this block of code assumes that layer one in the model falls within layer one on the file, but that isn't necessarily the case in practice. At a glance, I'm thinking you may be able to remove this if you start the `do j = 1,nlevsoifl-1` loop at 0 rather than 1, but I haven't thought this through carefully.

2. Would it also make sense to address the issue you raised in https://github.com/ESCOMP/ctsm/pull/771#discussion_r309490747?

It seems like addressing those two, together what you've already done, would make this interpolation correct. I don't feel like those absolutely need to be fixed at the same time, but maybe it makes sense to go ahead and fix this all at once?

Finally, for testing, it would be good to ensure that Clm45 cases don't change answers... I *think* that should be the case: do you agree? Also, given that we expect answer changes, can you document in this PR what verification you have done that the new logic works correctly (e.g., temporarily adding write statements to the code in one case and ensuring that the correct levels are being used - or whatever)?

slevisconsulting:
1) Ran a clm45 test and confirmed bfb with dev055.
2) Ran a clm5 test and confirmed diff. answers relative to dev055. 
This test sets `soil_layerstruct_predefined = '23SL_3.5m'`
I found that lev == 23 falls in the limbo scenario of `zsoi(23) > zisoifl(10)` where clay, sand, and om_frac simply keep their most recently set values.

I think a fix to the limbo cases should be bfb with dev055, so I will run that through the test suites first. Then I will put back the answer-changing mods.

slevisconsulting:
Regarding lev == 1:
I agree that CLM's lev == 1 may not always correspond to the dataset's j == 1, so technically we should merge that case with the second group of cases. However, we may prefer to leave as is if the goal was to make the model's top-most layer always have the texture properties of the dataset's top-most layer regardless of depth. This I will let @swensosc @dlawrenncar (any others?) decide.

slevisconsulting:
Cheyenne test-suite PASS

src/biogeophys/SoilStateInitTimeConstMod.F90
```diff
@@ -371,20 +372,26 @@ subroutine SoilStateInitTimeConst(bounds, soilstate_inst, nlfilename)
...
              else if (lev <= nlevsoi) then
+                found = 0  ! reset value
+                ! Search within the dataset's range of zisoifl values
                 do j = 1,nlevsoifl-1
-                   ! NOTE(wjs, 2019-08-01) It appears that the code currently doesn't set
-                   ! clay, sand and om_frac explicitly under some conditions. It probably
-                   ! should. My understanding is that currently things work okay, though,
-                   ! because clay, sand and om_frac will remain set at their previous
-                   ! values, which is probably reasonable enough. See also
-                   ! <https://github.com/ESCOMP/ctsm/pull/771#discussion_r309509596>.
```
slevisconsulting: Temporarily reverted the next line to the way it was before this PR, in order to expect bfb from all tests.

slevisconsulting:
Hobart test-suite also PASS

dlawrenncar:
I agree with this.  CLM lev==1 should correspond to dataset level == 1

slevisconsulting:
By the way, before the most recent commit, I reran the test-suites to confirm bfb   **PASS**

slevisconsulting:
Since the last commit, I have rerun a clm45 test and confirmed bfb with dev055.
I have also ran a clm5 test with write statements comparing the old vs. new behavior, e.g.:
OLD `lev == 1 uses j == 1` but
`lev, j, nlevsoi, nlevsoifl = 2 2 20 10`
`zisoi(2), zisoifl(2), zisoifl(3) =   5.9999999999999998E-02   4.5091787175931458E-02   9.0561820418344652E-02
NEW `lev == 1 uses j == 1` but
`lev, j, nlevsoi, nlevsoifl = 2 1 20 10`
`zsoi(2), zisoifl(1), zisoifl(2) =   4.0000000000000001E-02   1.7512817916255204E-02   4.5091787175931458E-02`
thereby confirming that the new mapping is correct.

Write statements in tests with nlevsoi = 4 and nlevsoi = 49 confirm that soil textures get assigned explicitly now to all model soil levels, rather than leaving some levels with the most recent previous assignment.

As such, I consider this PR is ready for review @billsacks . In the meantime I will prepare the ChangeLog. Pls let me know if I should update to dev056 and complete final testing or if there are imminent tags that I should wait for.

swensosc:
I think that "the model's top-most layer always have the texture properties
of the dataset's top-most layer regardless of depth" is the correct way to
do it.

slevisconsulting:
> (1) I haven't looked super closely, so may be missing something, and (2) I'm fine with it the way it is, I just wanted to suggest an alternative for reference. So no need to change things unless you see this and think, "Yes, that would be a lot better."
> 
> This tag should go next, so go ahead and kick off final testing whenever you're ready.

I tested the alternate order with one clm45 test and one clm50 test. It appears to work the same, with the benefit of fewer `if (found...` statements.

Now I have submitted the test suites and will be in touch when they complete.

billsacks:
Thanks a lot - looks good!

slevisconsulting:
One unexpected test FAILs on Hobart:
SMS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCropGs.hobart_nag.clm-tracer_consistency
where lnd.log says that we're at nstep = 52
and cesm.log says
`ERROR in CompareBulkToTracer: tracer does not agree with bulk water`
`Called from: after downscale_forcings`
`Variable: ice1_grc`
@billsacks my first reaction is that I'm not responsible for this one. Let me know if you think I should investigate it further.

billsacks:
That failure is really puzzling. I have confirmed that the test passes on master, and I also get this same failure when running it from your branch. The changes you made should have no bearing on this test - particularly not the way it's failing. As a sanity check, I also ran from your branch but with the changes to SoilStateInitTimeConstMod.F90 backed out, and it passes (confirming that the problem is _not_ some rogue changes on your branch).

I suspected a memory issue, so I tried running a version of the test with `_P96x1`, but that failed in the same way.

This test passes from your branch on my mac (with gnu) and on hobart intel. I'm suspecting some compiler-specific issue, and rather than trying to get to the bottom of this, I think it's best if we just move this test to intel. I'll open an issue documenting this... but will probably just immediately close it as a wontfix.

So, once you give me the go-ahead (rest of testing looks good and you're done writing the ChangeLog), I'll make that tweak to the test list then merge this branch to master. I have already run the new version of the test (on intel rather than nag) and generated baselines.

slevisconsulting:
@billsacks thank you for following up regarding the failing "nag" test.

Cheyenne testing: PASS

I updated the ChangeLog, so I'm ready to hand this PR over to you, unless you think of anything else.

Issue #648, opened 2019-02-26
Convert SCRIP grid files used by mkmapdata.sh to ESMF unstructured format


ekluzek:
mkmapdata.sh currently uses SCRIP grid files to generate the mapping files. With the update in #643, we will want to update these files into ESMF UGRID format. UGRID format contains the connectivity information that helps with running in parallel. There are two tasks to do that.

First is to convert the current SCRIP files into ESMF format using the standard ESMF tool. ESMF_Scrip2Unstruct that is created with ESMF distributions. 

The second is to update the files in namelist_defaults_ctsm_tools.xml from SCRIP grid format to ESMF_UGRID. I think the first step would be to add new fields that parallel the following type of fields in the file:

`<scripgriddata hgrid="0.9x1.25"  lmask="nomask" >lnd/clm2/mappingdata/grids/0.9x1.25_c110307.nc</scripgriddata>`

with a new name something like "esmfgriddata".

ekluzek:
The format should be ESMF unstructured mesh and NOT UGRID.

ekluzek:
@bekozi @slevisconsulting @negin513 note this is something we talked about today, but we said we just needed SCRIP grid files. We do need the destination grids to be SCRIP grid files since gen_domain requires it. But, the source grids could be in a different format (and note the 1km file already is in UGRID format). This is still a possible performance optimization that we could take advantage of.

slevisconsulting:
@ekluzek I'm writing a script to run ESMF_Scrip2Unstruct, and I thought I'd let you know in case you already had a script to share.

ekluzek:
No I don't. @mvertens did this step for some of the SCRIP files, but I'm not sure how many she converted. So you might look to see if you can find the files she made.  I'm assuming they would be somewhere near the existing SCRIP grid files.

slevisconsulting:
@mvertens I see that you made such files for all but the 0.25x0.25 resolution. This is the raw dataset resolution of the LAI, PFT, harvest, and soil color variables.

Do you happen to still have a copy to share of the script that you used to convert the SCRIP files to unstructured mesh?

mvertens:
Unfortunatley, I no longer have that script. It would be useful frankly to
regenerate all of these files since I did not do a careful validation of
the conversion and the conversion script has been updated since then.

ekluzek:
With #1663 this won't be needed.

Issue #643, opened 2019-02-26
Integrate Ben's chunking in the new mksurfdat toolchain


slevisconsulting:
Modify mkmapdata to include calls to OCGIS.

bekozi:
I wanted to get a task list down here for the chunked regridding (see this [board](https://github.com/NCPP/ocgis/projects/1) for more detailed issues). I ordered the tasks by priority.

 * [x] ~~Merge branch for running demo code to the ocgis master.~~ _Done._
 * [ ] Support ocgis installation on Cheyenne.
 * [ ] Support ESMF Unstruct output in ocgis. Currently, the format may only be read.
 * [ ] Track global indices following a spatial subset. Needed to construct a weight file with correct indexing when using a spatial subset intermediary as the source grid.
 * [ ] Verbose output for chunked regridding CLI.

Ping @rsdunlapiv

slevisconsulting:
My current understanding:

1) The ocgis call should replace this command in mkmapdata.sh:

cmd="$mpirun $ESMF_REGRID --ignore_unmapped -s ${INGRID[nfile]} "

By "ocgis call" we mean parts of Ben's bash script that does the spatial subsetting and weight generation:
https://github.com/NCPP/ocgis/blob/i488-full-regrid/examples/do-chunked-rwg-with-ss.sh

I would like an expert's opinion on which parts of Ben's script MUST end up in mkmapdata.sh.

2) I would like an expert's tour of mkmapdata.sh, to help ensure correct translation into python. 

At the Monday meeting we can discuss when to have the tour of mkmapdata.sh to address (1) and (2).

Are (1) and (2) in the correct order here?

billsacks:
From today's discussion: Next step is to just play around with using ocgis, to make sure it will work for our use cases with reasonable memory, time, etc. We can worry about getting a robustly engineered solution as a following step. Can start by just seeing what the cmds are that are done by mkmapdata.

slevisconsulting:
Keeping @billsacks and @ekluzek in the loop:

@negin513 @bekozi and I met today. Negin and Ben worked in Negin’s environment (not using conda) to get Ben's script running. They started troubleshooting from the same error that I got in my attempts: “...requires PIO…”

Now Negin has a script that works with Ben’s source and destination datasets.

Next Negin and Ben tried running the script with a couple of CTSM’s default source and destination datasets and failed. Ben will troubleshoot and let us know in the next day or two.

bekozi:
[...]

PR #815, opened 2019-10-05
Integrate @bekozi's ocgis tool in mkmapdata.sh

### LINKED ISSUE #286: For mksurfdata_map: use maps with no source masking, applying mask separately
@swensosc suggested this 2015-10-19, and it seems like a good idea to me: When mapping files from their raw data grid to CLM resolutions, we could use maps with no source masking, and then apply the mask in a separate step.

We currently have a LOT of mapping files from the mksurfdata_map raw data files to the CLM grids. Much of the reason we have so many is that we have a separate set of mapping files for each raw data mask - e.g., even if many of the raw data files are at the same 3' resolution, we need different mapping files for the different masks.

Sean pointed out that we should be able to use mapping files without masks, and then tweak the mapping algorithms to apply the source masks separately. I think we do things like that in other parts of CESM (e.g., in the coupler?). This would greatly reduce the number of mapping files we need to maintain. Furthermore, if a raw dataset is updated, and this update involves changing the mask, you wouldn't need to remake mapping files. (This was Sean's original motivation, as he is updating the lake dataset in this way.) Instead, mksurfdata_map would simply read the mask off of the (updated) raw data file.

### LINKED ISSUE #645: mksurfdat toolchain: portability across platforms
Make sure the tool build can work on a wide range of systems.

### LINKED ISSUE #806: Add initial sub-setting of regional source grids as part of the mkmapdata process
This relates to #643 and is dependent on it.

In order for OCGIS to work reasonably for regional grids, a first step of the process needs to run OCGIS to sub-set the grids to just the area needed for the destination grid. This will improve performance of the process, especially in terms of memory.

@bekozi has this as an option in OCGIS, but he needs to get it working to preserve the global indices, because we'll need that for mksurfdata_map. So we are also dependent on that issue in OCGIS (https://github.com/NCPP/ocgis/issues/494).

@negin513 @slevisconsulting @bekozi

slevisconsulting:
### Description of changes
Modify mkmapdata.sh to run with @bekozi's ocgis tool because it allows for
- chunking of the data while regridding to reduce memory requirements and
- subsetting of the source files when regridding to regional destination files

### Specific notes

@bekozi @negin513 @ekluzek @billsacks 

Begins to address #643 #644  
Fixes #806 
I could see it possibly fixing #648 

I expect roundoff differences in output weight (aka map) files from testing this far.
I use roundoff differences as a minimum requirement for my ocgis tests to pass.

None once #645 is resolved

Testing performed, if any:
- Manual testing, i.e. run default mkmapdata.sh and new mkmapdata.sh and compare the S variable in the weight files. For regional destination files, must also compare the i and j index vectors.
- Testing will become simpler (fewer weight files to compare) once #286 is resolved


billsacks:
@slevisconsulting looking at your PR comment, it looks like GitHub is trying to be too clever, and thinks that this PR resolves some issues that aren't really resolved. It looks like this is truly supposed to close #806, but the way you reference #645 and #286 makes GitHub think that this PR closes those two issues, too. Can you please reword your PR comment to avoid this? For example, maybe use passive voice (e.g., "once #645 is resolved").

slevisconsulting:
@bekozi I've reached the point where I need the weight files from ocgis to contain the same metadata as NCAR's weight files in order to test running them through our script `mksurfdata_map`

If you look in one of the NCAR weight files that I've shared with you in the past, you'll see what I mean. Pls let me know if you need anything from me. Thanks!

bekozi:
@slevisconsulting Would you mind pointing me towards one (Cheyenne is fine)? I just want to make sure we are on the same page.

slevisconsulting:
@bekozi some of the data in the NCAR weight files is redundant now, so I will first confirm what's really needed and then point you to a file.

slevisconsulting:
@bekozi I take it back. Removing things that I thought might be redundant caused me problems, so let's keep everything. Here's a weight file with everything in it:
`/glade/p/cesm/cseg/inputdata/lnd/clm2/mappingdata/maps/0.9x1.25/map_0.25x0.25_nomask_to_0.9x1.25_nomask_aave_da_c191016.nc`

bekozi:
Got it. You only need the global attributes added, correct?

slevisconsulting:
As far as I can tell, the code that makes the surface datasets needs all the variables and dimensions that you see in there, where
- yc and xc are the lats and lons at grid cell centers
- yv and xv are the lats and lons at grid cell vertices
- suffixes a and b refer to the SRC and DST resolutions

In terms of the global attributes, some now already make it in my ocgis-made weight files. See here:
`/glade/work/slevis/git/mksurfdata_maps_wo_src_masks/tools/mkmapdata/map_files_ocgis/0.25x0.25/map_0.25x0.25_nomask_to_0.9x1.25_nomask_aave_da_c191111.nc`
I do not know if we need the other global attributes for the code to run...

bekozi:
ESMF is moving away from including center coordinates and other auxiliary variables in the weight files. The concurrent weight file write routine (used by ESMPy) also does not support them. The reason for the removal is the increasing size of grids and associated coordinate arrays. If this is a deal breaker, I can definitely investigate getting them output in the concurrent routine.

I'll add the additional global attributes no problem!

slevisconsulting:
> **ESMF is moving away from including center coordinates and other auxiliary variables in the weight files.** The concurrent weight file write routine (used by ESMPy) also does not support them. The reason for the removal is the increasing size of grids and associated coordinate arrays. If this is a deal breaker, I can definitely investigate getting them output in the concurrent routine.

@ekluzek I will let you decide how to approach this.

ekluzek:
> > **ESMF is moving away from including center coordinates and other auxiliary variables in the weight files.** The concurrent weight file write routine (used by ESMPy) also does not support them. The reason for the removal is the increasing size of grids and associated coordinate arrays. If this is a deal breaker, I can definitely investigate getting them output in the concurrent routine.
> 
> @ekluzek I will let you decide how to approach this.

Thanks for pointing this out @slevisconsulting. I'm checking with Jim Edwards and Mariana V. about this

tools/mkmapdata/mkmapdata.sh
```diff
@@ -446,17 +429,9 @@ if [ "$interactive" = "NO" ]; then
    mpirun=$MPIEXEC
    echo "Running in batch mode"
 else
-   mpirun=""
+   mpirun="mpirun -np 1"
```
ekluzek: It would be better for the serial case to NOT be required to run with MPI. Is always running with MPI a requirement of OCGIS?

tools/mkmapdata/mkmapdata.sh
```diff
@@ -288,7 +280,7 @@ else
 fi
 
 # Set timestamp for names below 
-CDATE="c"`date +%y%m%d`
+CDATE="c"`date -d "0 days" +%y%m%d`
```
ekluzek: See comment about this in the other PR. Is this for setting of daylight savings time? If so I especially think we shouldn't mess with it.

tools/mkmapdata/mkmapdata.sh
```diff
@@ -356,13 +349,18 @@ case $hostname in
...
   fi
-  module load esmf-${esmfvers}-ncdfio-${mpi}-O
-  if [ -z "$ESMFBIN_PATH" ]; then
-     ESMFBIN_PATH=`grep ESMF_APPSDIR $ESMFMKFILE | awk -F= '{print $2}'`
-  fi
   if [ -z "$MPIEXEC" ]; then
      MPIEXEC="mpiexec_mpt -np $REGRID_PROC"
   fi
+
+  module load ncarenv/1.3
```
ekluzek: Is it critical to use specific versions of software packages? This can be something difficult to stay up to date with.

tools/mkmapdata/mkmapdata.sh
```diff
@@ -509,17 +484,28 @@ until ((nfile>${#INGRID[*]})); do
       echo "Skipping creation of ${OUTFILE[nfile]} as fast mode is on so skipping large files in NetCDF4 format"
    else
 
-      cmd="$mpirun $ESMF_REGRID --ignore_unmapped -s ${INGRID[nfile]} "
-      cmd="$cmd -d $GRIDFILE -m conserve -w ${OUTFILE[nfile]}"
-      if [ $type = "regional" ]; then
-        cmd="$cmd --dst_regional"
+      PATH_PREFIX=/glade/work/slevis
```
ekluzek: We should have this outside of a particular user's space. On cheyenne it could be under: /glade/tss/

tools/mkmapdata/mkmapdata.sh
```diff
@@ -509,17 +484,28 @@ until ((nfile>${#INGRID[*]})); do
...
+      WD=${PATH_PREFIX}/git/mkmapdata_ocgis/tools/mkmapdata
+      SS_PATH=${WD}/subsets/spatial_subset.nc
+      rm -rf ${SS_PATH}
+      CHUNKDIR=${WD}/chunking
+      rm -rf ${CHUNKDIR}
+      OCLI_FILE=${PATH_PREFIX}/git_ocgis/ocgis/src/ocgis/ocli.py
+      CRWG="python ${OCLI_FILE} chunked-rwg"
```
ekluzek: Some of these names are a bit short and cryptic. I get WD for Working Directory. And I suppose from context SS_PATH is Spatial Subset path. But, I'm not sure what CRWG means? What does the "chunked -rwg" option mean?

tools/mkmapdata/mkmapdata.sh
bekozi: It is not a requirement for ocgis. ESMF must be built with MPI support for IO operations...

ekluzek: Thanks for that @bekozi. On cheyenne there are versions of ESMF that are built without MPI (the "uni" versions). So we used them for the serial cases. Can ocgis choose to point to a version of ESMF that is built without MPI?

tools/mkmapdata/mkmapdata.sh
bekozi: I added this variable for the development script. It allows switching the CLI executable easily. For production code, I expect you could just use `ocli chunked-rwg` (chunked-rwg is the CLI subcommand).

ekluzek: I'm mainly complaining about the names being too short to be meaningful and obvious (at least to me). So if it could be a little longer and more descriptive that would be useful. Or adding a comment that describes what it's doing and what its purpose is. But, typically it's better to just have a longer more descriptive name.

tools/mkmapdata/mkmapdata.sh
bekozi: I think we will encounter a problem in the weight file write routine since it uses PIO v1 which ESMF does not wrap with mpiuni. This could be another reason to fall back on the other weight file write routine which, while not writing concurrently, uses netcdf directly so supports mpiuni (I'm 99% sure this is the case). I'll confirm this with the ESMF group.

tools/mkmapdata/mkmapdata.sh
bekozi: :+1: Names are indeed a bit cryptic.

tools/mkmapdata/mkmapdata.sh
ekluzek: OK, thanks for pointing that out. If we need to require MPI, I think that's OK. But, because it's so much simpler to run without MPI, I like to have it as an option. It's often easier to setup the non-MPI version on a machine first, rather than having to also bring in the extra complexity for MPI. Practically, I'm not sure you can run the regridding for a non-MPI case at least for any sizable grid. It would be hopeful (and helpful) that you could run a single point grid without MPI though for example though.

tools/mkmapdata/mkmapdata.sh
slevisconsulting: Adding comment explaining the use of the `date -d` flag, as discussed [here](https://github.com/ESCOMP/CTSM/pull/823/files#r359095927).

tools/mkmapdata/mkmapdata.sh
slevisconsulting: This is the combination of module loads that @bekozi determined we needed for ocgis to run. I can test by trial-and-error whether removing the version numbers will continue to work, unless @bekozi advises against it.

slevisconsulting: @bekozi, before attempting to remove specific version numbers by trial-and-error, I returned to my previously working script on cheyenne:
`/glade/work/slevis/ocgis_work/subset_1x1_20191213.sh`
and found that it now fails with this error in the PET0.ESMF_LogFile:
`20191229 114147.072 ERROR  PET0 ESMF_Array.F90:3962 ESMF_ArrayWrite Unable to write to file  - Internal subroutine call returned Error`
`20191229 114147.073 ERROR  PET0 ESMF_IOScrip.F90:367 ESMF_OutputWeight File Unable to write to file  - Internal subroutine call returned Error`
`20191229 114147.073 ERROR  PET0 ESMF_IOScrip.F90:136 ESMF_SparseMatrix Write Unable to write to file  - Internal subroutine call returned Error`
`20191229 114147.073 ERROR  PET0 ESMF_Field_C.F90:1269 f_esmf_regridstorefile Unable to write to file  - Internal subroutine call returned Error`
It still writes out the same subset file as before:
`/glade/work/slevis/ocgis_work/subsets/spatial_subset_191209.nc`
but fails to generate the weight file that it had previously:
`/glade/work/slevis/ocgis_work/weights/weights_191209_after_subsetg_1x1.nc`

I tried two things that didn't help. 
In `/glade/work/slevis/git_ocgis/ocgis` and 
in `/glade/work/slevis/git_esmf/esmf` I typed
`git fetch`
`git pull`
to get caught up with the latest codes.

What else am I missing?

tools/mkmapdata/mkmapdata.sh
slevisconsulting: For now I changed these path assignments to be based off of `$dir` because this is already assigned at the top of the script. This way I avoid using `slevis` in the path.

I am open to changing to something like `/glade/tss/` later.

tools/mkmapdata/mkmapdata.sh
slevisconsulting: Renamed CRWG to CLI_EXECUTABLE.
Renamed SS_PATH to SUBSETS_PATH.
Removed PATH_PREFIX and WD.

tools/mkmapdata/mkmapdata.sh
slevisconsulting: Turns out the job fails when the weight file already exists. I'm surprised that I had not run into this before :-)  Thanks for troubleshooting with me, @bekozi. Reminder for you to put a more explanatory error message in the code about this.

slevisconsulting:
This error came up while running my ocgis script on izumi. I confirmed that the same SRC/DST combination leads to the same error on cheyenne:
`ValueError: wrap action combination not supported`

@bekozi this is the script that gives the error on cheyenne:
`/glade/work/slevis/ocgis_work/do-this-20200101_subsetting.sh`

Thanks, again, for setting up the call today!

slevisconsulting:
@bekozi sorry... I renamed the file to `no_subset_20200102.sh` to make it more descriptive.

tools/mkmapdata/mkmapdata.sh
slevisconsulting: Replaced specific module loads with generic module loads now.

bekozi:
@slevisconsulting The two issues we identified yesterday are fixed in `ocgis/master`:
1. Wrap action no supported (https://github.com/NCPP/ocgis/issues/509) -- The wrapped state was using an incorrect variable when identifying the appropriate action. The test script runs to completion. _Two_ variables needed swapping and only one was switched the previous time we encountered this issue. :scream:
2. Check for weight file presence before overwriting (https://github.com/NCPP/ocgis/issues/510).

tools/mkmapdata/mkmapdata.sh
bekozi: I'll talk with @rokuingh about adding another option for writing weight files in ESMPy. This should allow us to handle the `mpiuni` case. We will also be able to maintain the center coordinates in the weight file for the interim. An option to switch between the write routines will be available. I created an issue on the ocgis side (https://github.com/NCPP/ocgis/issues/511) to track this capability.

I did want to amend my comment on ESMF weight file center coordinates from earlier. The center coordinates in the ESMF weight file are _not_ repeated for each source/destination index combination. The center coordinates are copied over from the source/destination grids. This can be inefficient for larger grids, but not as inefficient as my earlier comment would make it appear.

Ping @oehmke

slevisconsulting:
Thank you, @bekozi , I pulled your mods and the test script runs.

tools/mkmapdata/mkmapdata.sh
bekozi: @rokuingh is planning to work on these two features (verbose weight file and arbitrary sequence indices for factor lists) hopefully next week.

bekozi:
A quick update. @rokuingh pushed an initial implementation of the [weight file with auxiliary variables](https://github.com/esmf-org/esmf/tree/ESMPy-filemode) and [arbitrary sequence indices](https://github.com/esmf-org/esmf/tree/ESMPy-arbindices) to ESMF feature branches. We're working on getting these merged.

slevisconsulting:
> A quick update. @rokuingh pushed an initial implementation of the [weight file with auxiliary variables](https://github.com/esmf-org/esmf/tree/ESMPy-filemode) and [arbitrary sequence indices](https://github.com/esmf-org/esmf/tree/ESMPy-arbindices) to ESMF feature branches. We're working on getting these merged.

@bekozi 
Hi! I hope you're well!

@billsacks @ekluzek @negin513 @dlawrenncar pinged me for an update, and I realized that I hadn't seen a follow-up on this work. Is it all merged now?

bekozi:
@slevisconsulting I'm doing pretty well thanks for asking. I hope it's the same for you!

My last update for the issue is here: https://github.com/NCPP/ocgis/issues/494#issuecomment-629268072. I should have pinged you on it. @rokuingh and I got the arbitrary sequence indices for grids working in both ESMF and ocgis. The mesh is still outstanding. We are aiming to do it the right way, and once the MOAB mesh in ESMF is implemented fully, we can change the sequence indices in-memory without rebuilding the mesh. I know this has taken awhile. It ended up being more complex than I had expected. Given the success with the grid, I expect once the new mesh capabilities are available, we can move on it pretty quickly. Is this holding you up in anyway? The mesh changes are getting close, but it's difficult to give you a reliable time estimate...

slevisconsulting:
> My last update for the issue is here: [NCPP/ocgis#494 (comment)](https://github.com/NCPP/ocgis/issues/494#issuecomment-629268072). I should have pinged you on it. @rokuingh and I got the arbitrary sequence indices for grids working in both ESMF and ocgis. The mesh is still outstanding. We are aiming to do it the right way, and once the MOAB mesh in ESMF is implemented fully, we can change the sequence indices in-memory without rebuilding the mesh. I know this has taken awhile. It ended up being more complex than I had expected. Given the success with the grid, I expect once the new mesh capabilities are available, we can move on it pretty quickly. Is this holding you up in anyway? The mesh changes are getting close, but it's difficult to give you a reliable time estimate...

Thank you @bekozi for this update. One more piece of work discussed [above](https://github.com/ESCOMP/CTSM/pull/815#issuecomment-590402902) was the "weight file with auxiliary variables." Is that one done?

I will let somebody from the NCAR group respond to your question about timing.

bekozi:
> One more piece of work discussed above was the "weight file with auxiliary variables." Is that one done?

It is!: https://github.com/NCPP/ocgis/issues/511

The argument `--weightfilemode WITHAUX` will write a verbose weight file. @ekluzek and I went back and forth on the history string and arrived at something he liked. The argument gets passed down to ESMPy which also required a feature improvement. Hence, a newer beta snapshot of ESMF/ESMPy is required.

slevisconsulting:
Great, thank you!

dlawrenncar:
I'm not able to follow all the technical details enough to even know what
part of our project is or is not being held up.  I suggest that we discuss
at a forthcoming WRF-CTSM meeting.

bld/namelist_files/namelist_defaults_ctsm_tools.xml
```diff
@@ -126,6 +165,32 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
...
+<unstructdata hgrid="3x3min" lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_3x3min_nomask_c200129.nc</unstructdata>
+<unstructdata_type hgrid="3x3min" lmask="nomask">ESMFMESH</unstructdata_type>
+<unstructdata hgrid="0.9x1.25" lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.9x1.25_nomask_c200129.nc</unstructdata>
+<unstructdata_type hgrid="0.9x1.25" lmask="nomask">ESMFMESH</unstructdata_type>
+<unstructdata hgrid="1km-merge-10min" lmask="HYDRO1K-merge-nomask"
+   >lnd/clm2/mappingdata/grids/UGRID_1km-merge-10min_HYDRO1K-merge-nomask_c130402.nc</unstructdata>
+<unstructdata_type hgrid="1km-merge-10min" lmask="HYDRO1K-merge-nomask">UGRID</unstructdata_type>
+
```
slevisconsulting, line 193: I kept the above as placeholders. I'm ok with removing if that's preferred.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -1159,6 +1159,11 @@ Toggle to turn on on diagnostic Snow Radiative Effect
 SCRIP format grid data file
 </entry>
 
+<entry id="unstructdata" type="char*256" category="mkmapdata"
+       input_pathname="abs" group="clmexp" valid_values="" >
+UNSTRUCT format grid data file
+</entry>
+
```
slevisconsulting, line 1166: Placeholder. I'm ok with removing if that's preferred.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -1172,13 +1177,25 @@ Flag to pass to the ESMF mapping utility, telling it what kind of grid
 file this is (SCRIP or UGRID).
 </entry>
 
+<entry id="unstructdata_type" type="char*256" category="mkmapdata"
+       group="clmexp" valid_values="ESMFMESH" >
+Flag to pass to the ESMF mapping utility, telling it what kind of grid
+file this is.
+</entry>
+
```
slevisconsulting, line 1185: Placeholder. I'm ok with removing if that's preferred.

tools/mkmapdata/mkmapdata.sh
```diff
@@ -269,11 +270,9 @@ if [ "$phys" = "clm4_5" ]; then
            "3x3min_LandScan2004" \
            "3x3min_MODIS-wCsp" \
            "3x3min_USGS"       \
-           "5x5min_nomask"     \
            "5x5min_IGBP-GSDP"  \
            "5x5min_ISRIC-WISE" \
            "5x5min_ORNL-Soil" \
-           "10x10min_nomask"   \
```
slevisconsulting, line 276: Oops, I missed two lines. I will put them back in the next commit.

slevisconsulting:
Update:
I have run `qsub regridbatch` and regenerated all the map_files **except for the ones pointing to our highest resolution SRC files, i.e. _3x3min_ and _1km-merge-10min_HYDRO1K-merge-nomask_.**

The latter SRC files return this error:
NetCDF Error: ./chunking/esmf_weights_1.nc : NetCDF: One or more variable sizes violate format constraints

I request 10 chunks and end up with 1 for these SRC files. As a test I requested 1 chunk but the error remains the same.

@bekozi suggested pointing to older esmf builds, so I tried changing my ESMFMKFILE to
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b23/mpt/2.21/intel/19.0.5/lib/libO/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b23/mpt/2.21/intel/19.0.5/lib/libg/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b21/mpt/2.21/intel/19.0.5/lib/libO/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b21/mpt/2.21/intel/19.0.5/lib/libg/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b17/mpt/2.21/intel/19.0.5/lib/libO/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b17/mpt/2.21/intel/19.0.5/lib/libg/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b17/mpt/2.19/intel/19.0.5/lib/libO/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b16/mpt/2.21/intel/19.0.5/lib/libO/Linux.intel.64.mpt.default/esmf.mk`
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b16/mpt/2.21/intel/19.0.5/lib/libg/Linux.intel.64.mpt.default/esmf.mk`
All cases gave the same error.

@bekozi will be taking time off from today until 8/17.

slevisconsulting:
Latest:
Turns out that all along the problem with the 3x3min and 1km SRC files has been the new option (--weightfilemode WITHAUX). Somehow I had not tested this correctly before.

So, back to using the newer esmf build 
`/glade/p/cesmdata/cseg/PROGS/esmf/8.1.0b24/mpt/2.21/intel/19.0.5/lib/libO/Linux.intel.64.mpt.default/esmf.mk`
the 3x3min and 1km SRC files generate weight files, except when I include the new WITHAUX option.

The WITHAUX option is what will allow us to generate fsurdat files from the map_files.

slevisconsulting:
Update:
@bekozi has pointed me to a new esmf.mk file. The 3x3min SRC files work now and I am generating corresponding map_files.

The 1km SRC files still give errors and @bekozi is investigating.

slevisconsulting:
Good news:
@bekozi has a combination of ocgis/esmf that appears to work with the 1km SRC files now. Ben and I need to clarify how to get this working more reliably. Then I will proceed with the generation of map_files for 1km SRC files.

slevisconsulting:
**Update**
I have been running two separate regridbatch jobs:
- one to generate the map_3x3min_* files 
- the other to generate the map_1km-* files

There are 34 DST files per SRC file:
- the 3x3min SRC files are almost done
- the 1km SRC files are a third of the way
- other SRC files are done

I have found that the 1km SRC files work with 8 ncpus/mpiprocs but not with 12, 16, 20, 36, or 72.
@bekozi did some of his own testing and gave me this assessment:
"From what I can tell, this is memory-related on the ESMF side. [...] what is the overall goal regarding the flexibility of ocli for the high resolution grid combinations? My sense is that the approach works well for most grid combos, it's just these really high resolution ones that take more tuning. I don't think we have the time (at least I don't right now) to make this completely general. Is there a proc count / chunk count that you are targeting for all regridding operations? If there is one, I can look into those settings in more depth."

I will let @ekluzek respond to Ben's questions.

slevisconsulting:
At Sep 24 meeting we decided to move fwd with the same testing that @ekluzek and @billsacks had proposed for #823 .

The first test-suite is `/glade/work/slevis/git/mkmapdata_ocgis/test/tools/test_driver.sh`, which I compared to the ctsm5.1.dev002 baseline. As far as I can tell this test-suite passed. I attach the std out file to share the results that I see.

[run.txt](https://github.com/ESCOMP/CTSM/files/5314948/run.txt)

ekluzek:
OK, note the relevant tests are passing. If you look at  /glade/work/slevis/git/mkmapdata_ocgis/test/tools/td.2615.status you'll see the summary of the tests. They all pass, except the PTCLM tests, but they don't matter for this work. Likely you would see the same for running the baseline version. But, the problem with those tests don't matter for this work.

slevisconsulting:
> [...] They all pass, except the PTCLM tests, but they don't matter for this work. Likely you would see the same for running the baseline version.

Thanks @ekluzek and you are right that tests 31-35 fail in the baseline, as well.

slevisconsulting:
Ben removed the cause of blank `row`, `col`, `S` variables in the map_files when running ocgis through mkmapdata.sh

I have now generated the first surface dataset using map_files that I generated with ocgis in mkmapdata.sh.

Although the surface dataset seems good, the corresponding log file shows an error-checking issue where all variable areas (input and output) are zero. I have posted this [here](https://github.com/NCPP/ocgis/issues/506#issuecomment-707321029).

ekluzek:
The work on parallel mksurfdata probably means this work will be closed without coming in.

ekluzek:
With PE #1663 this is something that we just aren't going to do.

PR #820, opened 2019-10-11
Fates next api uninitialized hlm_model_day fix

glemieux:
### Specific notes
This change fixes an uninitialized variable `hlm_model_day`.  It was found while trying to run in debug mode on Lawrencium, which builds with the intel compiler option `-ftrapuv` which initializes the variable to `NaN`.  FatesHistoryInterfaceMod.F90 tries to reference the variable and as such, causes a floating invalid error.  See also discussion from https://github.com/ESMCI/cime/issues/3205.

Contributors other than yourself, if any: @rgknox

CTSM Issues Fixed (include github issue #): https://github.com/NGEET/fates/issues/548

Due to the structure of the `init_coldstart` procedure the model day will be set to one during the initialization (which is technically day zero), resulting in the `cleafoff` (and related variables) being off by one day compare to the baseline for the initialization step only.  Howver, the `cleafoff`values for the timesteps greater than zero all match.

Testing performed, if any:
Comparision testing against CTSM baseline hash.  All expected PASS except `ERP_Ld9.f45_f45_mg37.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesAllVars` due to changes mentioned above.
CTSM baseline hash: 9c12e402
Output location: 
```
/glade/u/home/glemieux/scratch/clmed-tests/hlm-uninit-fix.fates.cheyenne.intel.Cbd70a67d-F393b9054
```


src/utils/clmfates_interfaceMod.F90
```diff
@@ -197,6 +197,11 @@ module CLMFatesInterfaceMod
    ! developer will at least question its usage (RGK)
    private :: hlm_bounds_to_fates_bounds
 
+   ! The GetAndSetFatesTimeis function is used to get the current time from the CIME
```
ekluzek: Small request to change use of CIME to CLM. The functions being used are the CLM specific ones, not the CIME ones.

glemieux: Thanks for the correction and clarification @ekluzek.

PR #825, opened 2019-10-20
Correct frac_sno = 0 for lakes on old initial conditions files

### LINKED ISSUE #783: frac_sno is always 0 for lake points
### Brief summary of bug

`frac_sno` appears to always be 0 for lake columns. This appears to cause incorrect results in various other lake variables.

### General bug information

**CTSM version you are using:** ctsm1.0.dev055

### Details of bug

I initially noticed this issue by examining LakeHydrology and seeing that `frac_sno` is never set there. I then confirmed that these three out-of-the-box initial conditions files have `frac_sno` identically 0 for all lake columns: `clmi.B1850.0161-01-01.0.9x1.25_gx1v7_simyr1850_c190111.nc`, `clmi.I1850Clm50BgcCrop-ciso.1366-01-01.0.9x1.25_gx1v7_simyr1850_c190116.nc` and `clmi.I2000Clm50BgcCrop.2011-01-01.1.9x2.5_gx1v7_gl4_simyr2000_c180715.nc`.

I then added this code (which isn't exactly what we want in the end) in order to see if these incorrect `frac_sno` values for lake have any effect on the simulation:

```Fortran
       if (snow_depth(c) > 0.0_r8) then
          frac_sno(c) = 1._r8
       else
          frac_sno(c) = 0._r8
       end if
```

This led to answer changes in many variables in a short test run. I have not yet investigated whether these differences are significant scientifically.

@swensosc thinks we should use the same snow cover fraction method for lakes as is used for non-lakes. I will plan to make that change. However, this will warrant a closer investigation to see the impact of this change.

billsacks:
### Description of changes

Due to #783 , frac_sno used to be 0 for all lake points. That was mostly fixed, but the issue is still present on initial conditions files that were generated prior to that fix (which includes most or all of our out-of-the-box restart files). This causes frac_sno to be 0 for lake points at the start of the simulation, even if there is a snow pack present. Currently this doesn't cause significant problems, but I'd like to change the calculation of frac_sno_eff so that it is 0 if frac_sno is 0 - and then this frac_sno problem becomes an issue (causing divide by 0 in at least one place).

I struggled with how to fix this problem. My initial attempts involved setting frac_sno to 1 for lake points that had frac_sno = 0 on the restart file despite having either (attempt 1) snow_depth > 0 or (attempt 2) h2osno_total > 0. However, both of these attempts led to failures in ERI tests. Presumably there are legitimate cases where frac_sno can be 0 even when snow_depth and/or h2osno_total are > 0, so trying to fix this issue for already-fixed restart files can change answers.

This led me to introduce some metadata on the restart file saying whether this is a new enough restart file to already have the fix in place, and if so, to avoid reapplying the fix. I first thought I'd use the existing version number metadata on the restart file, but I came to feel that that would be less robust due to (1) the need to parse this version number, and more importantly (2) if you haven't fetched recent tags in your local clone (which is often the case for me), then the version number makes it look like you're using an older version of the code than you really are. So I settled on putting a new piece of global metadata on the file specifically saying whether this restart file is new enough to fix this particular issue.

### Specific notes

CTSM Issues Fixed:
- Fixes #783 

Are answers expected to change (and if so in what way)? YES: Answers will change over lake points due to changing initial frac_sno

Testing performed, if any: None yet (due to cheyenne being down)

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,167 @@
 ===============================================================
+Tag name:  ctsm1.0.dev074
+Originator(s):  sacks (Bill Sacks)
+Date:  Wed Oct 23 19:25:21 MDT 2019
+One-line Summary: For lakes: when reading finidat, set frac_sno=1 if h2osno_total > 0
+
+Purpose of changes
+------------------
+
+Due to ESCOMP/ctsm#783, frac_sno used to be 0 for all lake points. That
+was mostly fixed, but the issue is still present on initial conditions
+files that were generated prior to that fix (which includes most or all
+of our out-of-the-box restart files). This causes frac_sno to be 0 for
+lake points at the start of the simulation, even if there is a snow pack
+present. Currently this doesn't cause significant problems, but I'd like
+to change the calculation of frac_sno_eff so that it is 0 if frac_sno is
+0 - and then this frac_sno problem becomes an issue (causing divide by 0
+in at least one place).
+
+However, further problems were introduced if I tried to always apply
+this correction: It appears that sometimes the restart file legitimately
+has frac_sno == 0 for lake columns with non-zero snow cover. To avoid
+changing answers for newer restart files, I am writing metadata to the
+restart file documenting whether the fix has already been applied on
+that restart file, and if so, we avoid reapplying the fix.
+
+Getting this restart metadata correct was tricky, especially when
+running init_interp. I have introduced a new module to encapsulate this
+complexity. This can be used in general for writing / reading metadata
+on which issues have been fixed on a given restart file.
+
+
+Bugs fixed or introduced
+------------------------
+
+Issues fixed (include CTSM Issue #): [If none, remove this line]
+- Resolves ESCOMP/ctsm#783
+
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+
+Does this tag change answers significantly for any of the following physics configurations?
+(Details of any changes will be given in the "Answer changes" section below.)
+
+    [Put an [X] in the box for any configuration with significant answer changes.]
+
+[ ] clm5_0
+
+[ ] ctsm5_0-nwp
+
+[ ] clm4_5
+
+Notes of particular relevance for users
+---------------------------------------
+
+Caveats for users (e.g., need to interpolate initial conditions):
+- The setting of initial lake frac_sno to 1 isn't exactly correct, but
+  it's better than the old value of 0.
+
+Changes to CTSM's user interface (e.g., new/renamed XML or namelist variables): none
+
+Changes made to namelist defaults (e.g., changed parameter values): none
+
+Changes to the datasets (e.g., parameter, surface or initial files): none
+
+Substantial timing or memory changes: none
+- I didn't check the PFS test, but the only change that could have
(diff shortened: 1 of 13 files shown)


billsacks:
@ekluzek I'd be happy to hear any thoughts you have on the mechanism I put in place for indicating whether the backwards compatibility fix needs to be applied - i.e., putting a new piece of global metadata on the restart file.

billsacks:
I have confirmed that this test passes now (it failed before introducing the latest set of changes, with the new metadata on the restart file): `ERI_Ld9_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default`.

billsacks:
@ekluzek I realized that this doesn't work right when doing an interpolated start. I'm going to rework the implementation. I hope to have something ready for review in a couple of hours.

billsacks:
@ekluzek this is now ready for review.

billsacks:
@ekluzek gave his approval for the broad ideas in this PR. Since I have other upcoming changes that depend on this, he gave his okay for me to merge this to master. As I told him earlier, I am still very open to revising specifics of the implementation in a following tag / PR once he has had a chance to look through this implementation in more detail.

billsacks:
I had vague recollections that putting a netcdf file into redefine mode could take a long time, so I wanted to confirm that the changes to init_interp didn't have a large impact on performance. I have confirmed that they do NOT impact the total init_interp time significantly by running `SMS_Ln1.f09_g17.I2000Clm50SpGs.cheyenne_intel`, pointing to an finidat file where I added a global attribute (`issue_fixed_742 = 42`) that got copied during interpolation, with the following diffs to time this:

```diff
diff --git a/src/init_interp/initInterp.F90 b/src/init_interp/initInterp.F90
index 220c4593..cf81e4f3 100644
--- a/src/init_interp/initInterp.F90
+++ b/src/init_interp/initInterp.F90
@@ -28,6 +28,7 @@ module initInterpMod
   use ncdio_utils    , only: find_var_on_file
   use ncdio_pio
   use pio
+  use perf_mod
 
   implicit none
   private
@@ -224,6 +225,8 @@ contains
     type(interp_2dvar_type) :: var2d_i, var2d_o  ! holds metadata for 2-d variables
     !--------------------------------------------------------------------
 
+    call t_startf('init_interp')
+
     if (masterproc) then
        write (iulog,*) '**** Mapping clm initial data from input ',trim(filei),&
             '  to output ',trim(fileo),' ****'
@@ -236,7 +239,9 @@ contains
     call ncd_pio_openfile (ncidi, trim(filei) , 0)
     call ncd_pio_openfile (ncido, trim(fileo),  ncd_write)
 
+    call t_startf('copy_metadata')
     call copy_metadata(ncidi, ncido)
+    call t_stopf('copy_metadata')
 
     call check_interp_non_ciso_to_ciso(ncidi)
 
@@ -698,6 +703,8 @@ contains
        write (iulog,*) ' Successfully created initial condition file mapped from input IC file'
     end if
 
+    call t_stopf('init_interp')
+
   end subroutine initInterp
 
   !-----------------------------------------------------------------------
```

Looking at the max time for any processor: the time of the full `init_interp` was 36.297 sec; the time of the `copy_metadata` call was only 0.008 sec.

PR #841, opened 2019-11-15
Some answer changes for release branch

### LINKED ISSUE #175: Set stream_year_first_urbantv/last for CLM4.5
**Erik Kluzek < erik > - 2017-08-25 16:27:07 -0600**
**Bugzilla Id:** [2505](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2505)
**Bugzilla CC:**  dll,  oleson,  rfisher, 

Danica and I noticed that stream_year_first_urbantv/stream_year_last_urbantv was being set to 2000 for 1850 spinup cases. It turns out it doesn't matter for CLM4.5 cases as the file it points to is static, but for consistency we should set it for CLM4.5 and not just CLM5.0 because it looks like the case is setup wrong as it is.

So we should set it for the CLM4.5 cases as well as CLM50 just so it looks correct.

### LINKED ISSUE #817: Have clm4_5 use the same ndep file for 1850 as clm5_0
For 1850, clm5_0 is using:

fndep_clm_WACCM6_CMIP6piControl001_y21-50avg_1850monthly_0.95x1.25_c180802.nc

which is from the PI-control.

While clm4_5 is using the 1850 from the historical period...

fndep_clm_hist_b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensmean_1849-2015_monthly_0.9x1.25_c180926.nc

1849 and 1850 are supposed to be and also the same as the PI-control case. And I've verified this is the case.

But, I do see a difference in answers when I run with one versus the other. It does look like it gets the right times from the file in both cases. So I don't have an explanation for the difference. It might just be a small rounding difference that happens in the temporal or spatial interpolation. If so it probably doesn't matter.

### LINKED ISSUE #832: When soil moisture prescribed, if input H2OSOI is missing abort, but have an option to override
It's possible for the prescribed soil moisture stream to have a point that's missing that's needed for the simulation being done (because the surface dataset being used is slightly different in vegetation cover from the dataset that was used to create the input H2OSOI stream). In talking with @swensosc we thought the best solution was to abort when this is found. But, add a namelist option that can override the abort and allow that point to use the model's H2OSOI.

@olyson @dlawrenncar

### LINKED ISSUE #833: Linear time interpolation of soil moisture streams causes some points to be just less than spval
### Brief summary of bug

With linear time interpolation of soil moisture streams, there are some points that end up being just smaller than 1.e36 (spval), but greater than 1.e35.

### General bug information

**CTSM version you are using:** release-clm5.0.27-52-gc8a0ccb0

**Does this bug cause significantly incorrect results in the model's science?** Yes (for those points)

**Configurations affected:** Only for use of soil moisture streams

### Details of bug

 use_soil_moisture_streams = .true.
 soilm_tintalgo = 'linear'

ekluzek:
### Description of changes

 answer changes needed. Move prescribed soil moisture dataset being used.

For soil moisture streams apply over both nat-veg and crop.

This is answer changing.

Point to $CSMDATA/lnd/clm2/prescribed_data/LFMIP-pdLC-SST.H2OSOI.0.9x1.25.20levsoi.natveg.1980-2014.MONS_climo.c190716.nc

### Specific notes

CTSM Issues Fixed (include github issue #): #175, #817, #832 #833

Fixes #817 -- Have clm4_5 use the same ndep file as clm5_0 for 1850
Fixes #175 -- Have clm4_5 use the same settings for urbantv years as clm5_0
Fixes #832 -- don't override with missing value, let the model set it
Fixes #833 -- time interpolation of soilm causes bad values for some missing points

prescribed soil moisture changes, so that grid points that are missing on the input dataset, just run CLM normally, time interpolation might show that more points are missing, and prescribed soil moisture is also done for crops.

For clm4_5, the same ndep file is used for 1850 as for clm5_0, and same settings for urbantv are now used as well (rather than present day).

Any User Interface Changes (namelist or namelist defaults changes)? Yes namelist defaults

clm4_5 defaults for ndep and urbantv years mirror clm5_0

Testing performed, if any: regular, running standard testing on cheyenne and izumi

### DIFF
--- a/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -1098,7 +1098,7 @@ lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_16pfts_Irrig_CMIP6_simyr18
 <stream_year_last_soilm   >1997</stream_year_last_soilm>
 <model_year_align_soilm   >1997</model_year_align_soilm>
 
-<stream_fldfilename_soilm hgrid="0.9x1.25">lnd/clm2/prescribed_data/LFMIP-pdLC-SST.H2OSOI.0.9x1.25.20levsoi.natveg.climo1980-2014.MONS_c190709.nc</stream_fldfilename_soilm>
+<stream_fldfilename_soilm hgrid="0.9x1.25">lnd/clm2/prescribed_data/LFMIP-pdLC-SST.H2OSOI.0.9x1.25.20levsoi.natveg.1980-2014.MONS_climo.c190716.nc</stream_fldfilename_soilm>
 
 <soilm_tintalgo>linear</soilm_tintalgo>
 <soilm_offset  >0</soilm_offset>
--- a/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -1211,23 +1211,39 @@ lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_16pfts_Irrig_CMIP6_simyr18
 <stream_year_last_urbantv  phys="clm5_0" sim_year_range="1850-2100" >2106</stream_year_last_urbantv>
 <model_year_align_urbantv  phys="clm5_0" sim_year_range="1850-2100" >2015</model_year_align_urbantv>
 
-<stream_year_first_urbantv phys="clm4_5"  >2000</stream_year_first_urbantv>
-<stream_year_last_urbantv  phys="clm4_5"  >2000</stream_year_last_urbantv>
+<stream_year_first_urbantv phys="clm4_5" sim_year_range="1850-2100" >2015</stream_year_first_urbantv>
+<stream_year_last_urbantv  phys="clm4_5" sim_year_range="1850-2100" >2106</stream_year_last_urbantv>
+<model_year_align_urbantv  phys="clm4_5" sim_year_range="1850-2100" >2015</model_year_align_urbantv>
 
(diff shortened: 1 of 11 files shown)


ekluzek:
Testing has gone as expected. Here are the tests that show different answers from release-clm5.0.28:

SMS_D.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-crop
SMS_D_Ld1.f19_g17.I1850Clm45Cn.izumi_gnu.clm-default
SMS_D_Ld5.f10_f10_musgs.I1850Clm45BgcCrop.izumi_nag.clm-crop
ERI_D_Ld9.f09_g16.I1850Clm45BgcCruGs.cheyenne_intel.clm-default
ERP_D_Ld5.f10_f10_musgs.IHistClm45SpGs.cheyenne_intel.clm-decStart
ERP_P36x2_D_Ld5.f09_g17.I2000Clm45Sp.cheyenne_gnu.clm-prescribed
ERP_P36x2_D_Ld5.f09_g17.I2000Clm45Sp.cheyenne_intel.clm-prescribed
ERP_P36x2_D_Ld5.f10_f10_musgs.I1850Clm45BgcCrop.cheyenne_intel.clm-crop
ERP_P36x2_D_Ld5.f10_f10_musgs.I1850Clm45BgcCru.cheyenne_intel.clm-default
ERP_P36x2_D_Ld5.f10_f10_musgs.I1850Clm45Cn.cheyenne_intel.clm-default
SMS_Ld5.f10_f10_musgs.I1850Clm45BgcCrop.cheyenne_gnu.clm-crop
SMS_Ld5.f10_f10_musgs.I1850Clm45BgcCrop.cheyenne_intel.clm-crop

PR #846, opened 2019-11-19
Update to fates sci.1.30.0_api.8.0.0

ekluzek:
### Description of changes

Update to latest version of FATES. Also now pointing to the public NGEET/fates github 
repository rather than NCAR/fates-release.

### Specific notes

Contributors other than yourself, if any: NGEET fates team @rgknox and others

Are answers expected to change (and if so in what way)? Yes, only for runs with FATES

  fates params file is updated

Testing performed, if any: running fates tests will run aux_clm as well.


ekluzek:
This test is failing SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro, but it looks like it's because of https://github.com/NGEET/fates/issues/559. So I'm marking it as an expected fail. @rgknox does that sound correct?

rgknox:
yes, thanks for checking @ekluzek , the fates hydro run tests are currently not passing

PR #849, opened 2019-12-04
Fix link for new forums

billsacks:
I noticed the previous link was broken. @ekluzek how do you feel about merging this directly to master?


ekluzek:
I think it's fine to merge directly to master. I do think we should point to the land/river specific forum. And it seems we should wait until they do the global subforum change they are going to do, in case that changes the name. Also note that they do have an alias for bb.cgd.ucar.edu to xenforo.cgd.ucar.edu.

PR #883, opened 2020-01-10
Bring changes on release-clm5.0 branch to master

### LINKED ISSUE #150: Remove logic for missing_crops and old format rawdata PFT datasets in mksurfdata_map
**Erik Kluzek < erik > - 2016-11-29 11:10:22 -0700**
**Bugzilla Id:** [2380](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2380)
**Bugzilla CC:**  dlawren,  lawrence,  rfisher,  sacks, 

There's some logic in mksurfdat.F90 and mkpftMod.F90 for mksurfdata_map that allows PFT raw datasets without all of the prognostic crop CFT's to be used. For a transient case the first year has to have crop, and then it uses that as a constant for future years.

Since, we are moving to raw datasets that always contain prognostic crop, we can remove this logic as a simplification. We'll leave it in place, until all the raw datasets are in place and available and we can use crop datasets for non-crop simulations (with create_croplandunit=.true.).

Also see...

https://trello.com/c/uWWZq88m/252-collapse-all-crops-to-generic-for-sp-or-non-crop-bgc-simulations

### LINKED ISSUE #175: Set stream_year_first_urbantv/last for CLM4.5

### LINKED ISSUE #262: hirespft option for mksurfdata.pl doesn't work
Hui Tang reports that the -hirespft option doesn't work in mksurfdata.pl for CLM5.0.000. This is something we've recognized, but we thought we'd get this data eventually. We need Peter L. to produce high resolution data for present day for this. We've talked about this a few times, but haven't made it happen.

So we should decided on what the plan is for this. And also get back to Hui on what it is.

### LINKED ISSUE #276: Add some tests to the test suites Chris Fischer runs
In a discussion with @mvertens , @jedwards4b , @fischer-ncar @cacraigucar, we decided that Chris will _not_ regularly run the `aux_clm` test suite. Instead, we will add some tests to the prealpha and prebeta test suites if we feel there are additional tests that could catch cime issues that currently slip through the cracks.

In addition, we decided that there should be a new test category, `aux_cime_baselines`. This will likely be a subset of the prealpha test suite. This should be a _small_ test suite (ideally no more than 3 or 4 tests defined by each component) whose purpose is to make sure that changes in cime have not changed answers relative to baselines for important configurations - i.e., that no answer changes have accidentally crept into cime. We want to keep this test suite small (even if that means not quite complete coverage) because it will be run frequently. It should be sufficient to have a few short SMS tests on a single machine/compiler (`cheyenne_intel`) - maybe covering a few different time periods and/or datm modes?

- [ ] Review prealpha tests defined by ctsm for completeness. @ekluzek have you already done this recently? Also, make sure that all prealpha tests are also in aux_clm (or have very similar tests in aux_clm).

- [ ] Review prebeta tests defined by ctsm for completeness. @ekluzek have you already done this recently? Also, make sure that all prebeta tests are also in aux_clm (or have very similar tests in aux_clm).

- [x] Define a new set of tests in ctsm's testlist xml file - `aux_cime_baselines` - as per the above notes.

- [ ] Make sure that all `aux_cime_baselines` tests have similar tests in `prealpha` (see https://github.com/ESCOMP/ctsm/issues/276#issuecomment-442652686)

cc @ekluzek @bandre-ucar

### LINKED ISSUE #533: Add -no-anthro option to mksurfdata_map
We need an easy way to turn off urban in mksurfdata_map.

### LINKED ISSUE #544: Transient cases coupled to CAM are matching the wrong IC file (2000 instead of 1850)
### Brief summary of bug

SMS.f19_g17.BHISTWs.cheyenne_intel.allactive-defaultio is matching a 2000 file instead of an 1850 one.

### General bug information

**CTSM version you are using:** release-clm5.0.09

**Does this bug cause significantly incorrect results in the model's science?** Wrong initial conditions (so sort of)

**Configurations affected:** LND_TUNING_MODE=clm5_0_cam6_0 (and I think clm4_5_cam6_0)

### Details of bug

It's wrongly matching this file rather than the 1850 one...

``` xml
<!-- Present day crop spinup at 2-degree with irrigation on -->
<finidat hgrid="1.9x2.5"    maxpft="79"  mask="gx1v7" use_cn=".true." 
use_cndv=".false." use_fates=".false."
          ic_ymd="20110101" use_nitrif_denitrif=".true." 
use_vertsoilc=".true." sim_year="2000"
          ic_tod="0" glc_nec="10" use_crop=".true." irrigate=".true."
 >lnd/clm2/initdata_map/clmi.I2000Clm50BgcCrop.2011-01-01.1.9x2.5_gx1v7_gl4_simyr2000_c180715.nc
</finidat>
```

It dies for cases that are on f19_g17_gl4 because it thinks it can run without interpolation and it can't. Other cases would run with interpolation, but from the wrong initial conditions.

Dies since it isn't setup to run with interpolation

### LINKED ISSUE #545: Antarctica ice shelves are being treated as wetlands rather than glaciers

### LINKED ISSUE #547: Add conus_30_x8 grid as valid option for CTSM and mksurfdata_map
Add the conus_30_x8 SCRIP grid file and mapping files into the XML database, so that conus_30_x8 grids can be created. Also add in the surfdata files that have been created in as well. To get this fully functioning, we would also need to get a cime update to introduce the grid into the coupler.

### LINKED ISSUE #553: More robust mksurfdata_map logic for determining where to put wetlands
As noted in #545 , the logic determining where to put wetlands is tied to the mask on the pft raw data file:

https://github.com/ESCOMP/ctsm/blob/a34404419aaa81f5021602963222c50daa10c0f6/tools/mksurfdata_map/src/mksurfdat.F90#L707-L724

I think a more robust way to do this would be: rather than just determining pctland_pft, as is currently done, instead determine a pctland field as the max of the pctland from each pct raw dataset. So, we'd start by setting pctland equal to 0. Then each routine that remaps a pct field (pct pft/cft, pct glacier, pct lake and pct urban) would accept pctland as an inout argument. Each of these routines would compute a my_pctland field, similarly to how pctland_pft is currently computed, and then have code like:

```Fortran
do i = 1, n
   pctland(i) = max(pctland(i), my_pctland(i))
end do
```

Then we could still turn areas to wetland that are outside the landmask - but we'd just do this for areas outside the union of all PCT fields' landmasks, rather than using only the pctpft landmask.

@dlawrenncar @lawrencepj1 @ekluzek @mvertens does this seem reasonable to you? I don't think we should do this for the cmip6 runs (it could change answers in more than just the Antarctic ice shelves), but I'm considering this as a long-term solution.

### LINKED ISSUE #557: Add no-anthro compset and options for CLM
We need to add an option for paleo work for no-anthro turning off urban and any urban related effects. One part of this is to use surface datasets from mksurfdata that have urban turned off (so see #533)

use_crop=.false.
fsurdat='file without urban, GDP=0, potential-veg,all crops zeroed'
urban_hac='OFF'
calc_human_stress_indices='NONE'
stream_year_first_urbantv= 'Set to non-urban dataset'
stream_fldfilename_popdens='zero population dataset'
pot_hmn_ign_counts_alpha=0.0
do_harvest=.false.
stream_fldfilename_ndep='remove human component of ndep

If carbon isotopes on
atm_c14_filename=''without anthro'
atm_c13_filename='without anthro'

For datm:

prescribed aerosols without human affect
CO2 at 1850 level if CO2_TIME_SERIES is on

### LINKED ISSUE #589: Existence of content in the lai_streams namelist makes it confusing to users
This came up with Charles Bardeen and some others. They were confused about the content of the lai_streams namelist as it wasn't being used in their case (SP case with use_lai_streams=F)

```
&lai_streams
 lai_mapalgo = 'bilinear'
 model_year_align_lai = 2001
 stream_fldfilename_lai = '/glade/p/cesmdata/cseg/inputdata/lnd/clm2/lai_streams/MODISPFTLAI_0.5x0.5_c140711.nc'
 stream_year_first_lai = 2001
 stream_year_last_lai = 2013
/
```

The fact that it has values filled in makes it seem like this is important, when in fact it is not. It would be be better to not fill in the namelist, unless use_lai_streams=T.

### LINKED ISSUE #608: Odd (blocky) spatial patterns in generic crop leafc for BGC simulation when initialized from BGC-Crop
### Brief summary of bug

A user reports odd (blocky) spatial patterns in leafc for generic crop pft for an I2000CLM50BGC case.
This is a BGC case initialized from a BGC-Crop initial file.

### General bug information

**Model version: cesm2.0.0 (release-clm5.0.01)

**Does this bug cause significantly incorrect results in the model's science? Yes

**Configurations affected: I2000CLM50BGC with out-of-the-box spunup initial conditions and accelerated spinup on

### Details of bug

Here is the initial report from the user:
"So when I start a simulations from an finidat file from the inputdata repository there appear some strange lines in the leaf carbon for rain-fed crop at the first time step. Most of these lines disappear after a couple of years. However, some persist east of Europe and in Central Africa. I found out so far, that the lines are different, when I use a different finidat file, but there always seems to be some of them."

I've reproduced the user's case and gotten the same behavior.  Some of the blocky patterns go away after several simulation years, but there are large areas of dead generic crop.

If I do a cold start I don't see this.  I think this is happening because there are actually some valid values for C3 generic crop in the initial file (even though it is an initial file from a prognostic crop simulation), about 47 gridcells.  It looks like these are being interpolated to the C3 crop pft based on the closest valid values, which is why things look blocky.  One value is being interpolated to a bunch of gridcells.  I can see this pattern in the interpolated initial file as well ("finidat_interp_dest").  I've tried setting all of the leafc for C3 crop in the initial file to zero and the pattern in the interpolated initial file goes away. I'm sure there are other variables that will look blocky though and that may cause a blocky pattern in leafc to reappear.

I've noticed that generic crops are not included in the code logic for reseeding (CNVegCarbonStateType.F90).  When I change this to include reseeding the generic crops and run for several years, most of the crops are now alive.  However, there are a few remaining areas of dead crop (China and South Africa).  Presumably this is because the generic crop gridcells that are being interpolated from have no carbon/nitrogen and reseeding is not sufficient to allow growth.  I will try to determine if that is actually the case.  I'm also going to create a branch with the change to reseeding and issue a pull request for that change because I don't see why we wouldn't try to reseed generic crops.

The presence of generic crops in the BGC-CROP initial file is incorrect.  However, if the initial file had no generic crop to interpolate from what would happen to a BGC simulation with only generic crops when interpolating from this file? 

I've reproduced the bug with:

/glade/work/oleson/release-cesm2.0.0/cime/scripts/clm5.0_SeSCs_spinup_test_181121

The initial file in question is:

clmi.IGM2000GSWP3CLM50BGCCROPIRR.2011-01-01.1.9x2.5_gx1v6_gl5_simyr2000_c170419.nc

### LINKED ISSUE #621: Fully remove CMIP5 rcp's (for clm40,clm45,clm50) and implement CMIP6 ssp_rcp's for clm45/clm50
The CMIP5 future scenarios were addressed as RCP#.#. The CMIP6 future scenarios are SSP (Shared Socioeconomic Pathway) and RCP combinations SSPn-#.#. In order to not have to support both options I'm going to remove the options for RCP's and only implement SSP-RCP combinations which only applies to CLM45/CLM50. So this will remove the CMIP5 RCP's for clm40.

This will start on the release-clm5.0 branch, but then be moved to master. By, not supporting RCP's for CLM40, it will make the change easier to come to master. As well as removing extra complexity on the release branch.

### LINKED ISSUE #624: QSNOEVAP history output incorrect

### LINKED ISSUE #629: Remove unused fields from surface datasets
There are some unused fields on the surface datasets. These should be removed in order to save disk space:

PCT_GLC_MEC_GIC
PCT_GLC_MEC_ICESHEET
PCT_GLC_GIC
PCT_GLC_ICESHEET
F0
P3
ZWT0

There are also a bunch of fields that could be eliminated for cases where anthropocentric influences are removed:

URBAN_REGION_ID
CONST_HARVEST_VH1
CONST_HARVEST_VH2
CONST_FERTNITRO_CFT
CONST_HARVEST_SH1
CONST_HARVEST_SH2
CONST_HARVEST_SH3
CONST_GRAZING
gdp
abm
NLEV_IMPROAD
CV_IMPROAD
CV_WALL
CV_ROOF
TK_IMPROAD
TK_WALL
TK_ROOF
ALB_WALL_DIF
ALB_ROOF_DIF
ALB_WALL_DIR
ALB_ROOF_DIR
ALB_IMPROAD_DIF
ALB_IMPROAD_DIF
ALB_PERROAD_DIF
ALB_PERROAD_DIF
WTROAD_PERV
WTLUNIT_ROOF
WIND_HGT_CANYON
T_BUILDING_MIN
THICK_WALL
THICK_ROOF
HT_ROOF
EM_WALL
EM_ROOF
EM_PERROAD
EM_IMPROAD
CANYON_HWR
UNREPRESENTED_CFT_LULCC
UNREPRESENTED_PFT_LULCC

And finally there's several fields only used for VIC (which comes from f09 tunings)

binfl
Dsmax
Ds
Ws

### LINKED ISSUE #653: To save disk on fsurdat files use zwt0 on finindated file
Right now for clm45 and the zwt inversion zwt0 and others are read from the fsurdat file. But this is also on the ch4inversion file so doesn't need to be.

### LINKED ISSUE #673: Ice shelf wetland fix in mksurfdata_map can lead to glacier+lake > 100% on surface datasets
### Brief summary of bug

Creating a conus_30_x8 surface dataset and then trying to run with it doesn't work.

### General bug information

**CTSM version you are using:**release-clm5.0.20

**Configurations affected:** conus_30_x8

Setup a case for ne30...

SMS_Ld1_D.ne30_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-default

That case will fail because it needs a new ne30np4 surface dataset. Pointing to a new one it then works.

Here's some changes to get it setup for the new grid..

``` csh
./xmlchange LND2GLC_FMAPNAME="$DIN_LOC_ROOT/cpl/gridmaps/ne0np4CONUS.ne30x8/map_ne0CONUSne30x8_TO_gland4km_aave.190227.nc"
./xmlchange LND2GLC_SMAPNAME="$DIN_LOC_ROOT/cpl/gridmaps/ne0np4CONUS.ne30x8/map_ne0CONUSne30x8_TO_gland4km_blin.190227.nc"
./xmlchange GLC2LND_SMAPNAME="$DIN_LOC_ROOT/cpl/gridmaps/gland4km/map_gland4km_TO_ne0CONUSne30x8_aave.190227.nc"
./xmlchange GLC2LND_FMAPNAME="$DIN_LOC_ROOT/cpl/gridmaps/gland4km/map_gland4km_TO_ne0CONUSne30x8_aave.190227.nc"
./xmlchange LND2ROF_FMAPNAME="$DIN_LOC_ROOT/lnd/clm2/mappingdata/maps/ne0np4CONUS.ne30x8/map_ne0CONUSne30x8_TO_0.5x0.5_nomask_aave.190227.nc"
./xmlchange ROF2LND_FMAPNAME="$DIN_LOC_ROOT/lnd/clm2/mappingdata/maps/ne0np4CONUS.ne30x8/map_0.5x0.5_nomask_TO_ne0CONUSne30x8_aave.190227.nc"
./xmlchange ATM_DOMAIN_FILE="domain.lnd.conus_30_x8_gx1v6.170119.nc"
./xmlchange LND_DOMAIN_FILE="domain.lnd.conus_30_x8_gx1v6.170119.nc"
./xmlchange ATM_DOMAIN_PATH="$DIN_LOC_ROOT/atm/datm7/domain.clm"
./xmlchange LND_DOMAIN_PATH="$DIN_LOC_ROOT/atm/datm7/domain.clm"
```

Add this to user_nl_clm...

  fsurdat = '/gpfs/fs1/scratch/erik/release-clm5.0.20/tools/mksurfdata_map/surfdata_conus_30_x8_hist_78pfts_CMIP6_simyr1850_c190402.nc'

Since, it dies early on, there probably aren't problems with the above mapping files to cause the problem.

Here's the error from the cesm.log file:

>  check_weights ERROR: at g =        38823 total PFT weight is 
>    1.45000000000000      active_only =  F
>  check_weights ERROR: at g =        38823 total col weight is 
>    1.45000000000000      active_only =  F
>  check_weights ERROR: at g =        38823 total lunit weight is 
>    1.45000000000000      active_only =  F
>   
>  If you are seeing this message at the beginning of a run with
>  use_init_interp = .true. and init_interp_method = "use_finidat_areas",
>  and you are seeing weights less than 1, then a likely cause is:
>  For the above-mentioned grid cell(s):
>  The matching input grid cell had some non-zero-weight subgrid type
>  that is not present in memory in the new run.
>   
>  ENDRUN:
>  ERROR: ERROR in subgridWeightsMod.F90 at line 684
> Image              PC                Routine            Line        Source             
> cesm.exe           0000000004E9528D  Unknown               Unknown  Unknown
> cesm.exe           00000000041A7901  shr_abort_mod_mp_         114  shr_abort_mod.F90
> cesm.exe           00000000041A7782  shr_abort_mod_mp_          61  shr_abort_mod.F90
> cesm.exe           0000000000873EC4  abortutils_mp_end          50  abortutils.F90
> cesm.exe           0000000000D7DC5E  subgridweightsmod         684  subgridWeightsMod.F90
> cesm.exe           0000000000CEDF74  reweightmod_mp_re          55  reweightMod.F90
> cesm.exe           00000000008ADD8E  clm_initializemod         227  clm_initializeMod.F90

### LINKED ISSUE #707: Missing mapping files for hirespft
Mapping files are missing for hirespft for these resolutions: 0.125x0.125,ne4np4,ne16np4,ne120np4,ne240np4.

This is in release-clm5.0.22

### LINKED ISSUE #716: 2-degree WACCM CMIP6DECK needs to match a no-ciso user-mod
### Brief summary of bug

A compset with WACCM and CMIP6DECK for 2-degree should match a WACCM CMIP6 user-mod without Carbon isotopes turned on.

### General bug information

**CTSM version you are using:** release-clm5.0.22

**Configurations affected:** WACCM at 2-degree with CMIP6DECK

### Details of bug

The matching for CLM_USER_MODS, currently for non-WACCM cases with CMIP6DECK on will match either regular or the "nociso" user-mod case at 2-degree. This same logic needs to be added to the CMIP6WACCMDECK compsets so that nociso is used for 2-degree and the regular user-mod for anything else.

Here's an example test case that fails: SMS_D_Ln9.f19_g17.BWmaCO2x4cmip6.cheyenne_intel.allactive-default

ERROR: Cannot interpolate from a run without c13/c14 to a run with c13/c14

### LINKED ISSUE #734: Isotope historical files are in NetCDF-4 format need them in NetCDF-3 or NetCDF-5
### Brief summary of bug

@jedwards4b found that two files are in NetCDF-4 format and can't be read on the Korean system he is working in...

> "I'm trying to run f09_g17.B1850cmip6 on the korean system using the 2.1.1 release candidate and am having trouble reading the files
> /share/CESM/cesm_input/lnd/clm2/isotopes/atm_delta_C13_CMIP6_1850-2015_yearly_v2.0_c171012.nc
> 
> /share/CESM/cesm_input/lnd/clm2/isotopes/atm_delta_C14_CMIP6_3x1_global_1850-2015_yearly_v2.0_c171012.nc
> 
> 
> 
> It appears that the files were created using netcdf4 and pnetcdf is having trouble reading them.   Using nccopy I made cdf5 versions of the files and that works, 

### General bug information

**CTSM version you are using:** release-clm5.0.24

**Configurations affected:** Some external machines

### Details of bug

Some external machines seem to have trouble reading in NetCDF-4 format files. This works OK on local machines. NERSC machines sometimes has this problem in addition to the Korean system. From Jim "The problem seems to be related to cray systems, but I haven't figured out exactly why." So the best way to handle it now is to change the file format to either NetCDF-3 or NetCDF-5.

### LINKED ISSUE #741: C13_HR, C13_NBP, FPI values result in numeric conversion not representable error
### Brief summary of bug

C13_HR, C13_NBP, FPI values result in numeric conversion not representable error

### General bug information

**CTSM version you are using:** relcisofix.n01_release-clm5.0.20

**Configurations affected:** --compset BSSP585_BPRPcmip6 --res f09_g17

### Details of bug

Original message from Keith Lindsay:

I'm running the esm-ssp585 CESM2 CMIP6 experiment.
The model is aborting when writing the monthly clm history files for 2046-01 with the error message
NetCDF: Numeric conversion not representable

Suspecting that the problem is that CLM is trying to write an r8 value that is outside of the range of r4 into a netCDF float(=r4),
I've rerun the last segment with hist_ndens=1, to get r8 output.

Sure enough, the run went past where the original run aborted, and there are 3 variables with values outside the range of r4 values.

All of the out of range values are at the same location:
(i,j)=(40,168) (lon,lat)=(50E,68.3N)
The variables and excessive values are:
C13_HR: -6.8144E+46
C13_NBP: 6.8144E+46
FPI: -1.4824+100

The largest r4 value is ~3.4028E+38
The following vars are also unusually large, but not outside r4 limits
C14_HR:-2.9915E+36
C14_NBP: 2.9915E+36

Values for HR and NBP at the same location are -1.837E-41 and -9.452E-18 respectively.
pretty small

Any suggestions on what I should do to further diagnose this and/or work around it?

The original case is
CASE=b.e21.BSSP585_BPRPcmip6.f09_g17.CMIP6-esm-ssp585.001
CASEROOT=/glade/work/cmip6/cases/C4MIP/b.e21.BSSP585_BPRPcmip6.f09_g17.CMIP6-esm-ssp585.001
RUNDIR=/glade/scratch/cmip6/b.e21.BSSP585_BPRPcmip6.f09_g17.CMIP6-esm-ssp585.001/run

My clone, where I changed hist_ndens and reran, is
CASE=b.e21.BSSP585_BPRPcmip6.f09_g17.CMIP6-esm-ssp585.001
CASEROOT=/glade/scratch/klindsay/temp_caseroots/b.e21.BSSP585_BPRPcmip6.f09_g17.CMIP6-esm-ssp585.001
RUNDIR=/glade/scratch/klindsay/b.e21.BSSP585_BPRPcmip6.f09_g17.CMIP6-esm-ssp585.001/run

### LINKED ISSUE #746: Some CLM45 configurations use CLM50 initial conditions files
### Brief summary of bug

I noticed that some of our CLM45 tests use CLM50 initial conditions files. I'm not sure if this is actually a bug, but I wanted to bring it to @ekluzek 's attention.

### General bug information

**CTSM version you are using:** ctsm1.0.dev045

(Actually I'm using a branch off of that, but I haven't changed anything that should change this behavior.)

**Does this bug cause significantly incorrect results in the model's science?** Unsure

**Configurations affected:**

This shows up in the following tests:

```
ERP_P36x2_D_Ld5.f10_f10_musgs.I2000Clm45Sp.cheyenne_intel.clm-default
ERS_Ld5_Mmpi-serial.1x1_vancouverCAN.I1PtClm45SpGs.cheyenne_intel.clm-default
SMS_D_Ld1_Mmpi-serial.1x1_vancouverCAN.I1PtClm45SpGs.hobart_nag.clm-default
SMS_D_Ld1_P48x1.f10_f10_musgs.I2000Clm45BgcCrop.hobart_nag.clm-oldhyd
```

### Details of bug

@ekluzek is this a bug or intended behavior?

### LINKED ISSUE #786: Allow use_init_interp in a branch case even though it doesn't do anything
If you try to run a branch case with `use_init_interp = .true.`, the model dies at runtime with the message, `ERROR: Can only set use_init_interp if finidat is set`.

On the surface, this check that `use_init_interp` can only be set if finidat is set seems reasonable. However, it causes problems in practice if you set up a branch case using a compset and resolution for which use_init_interp are set automatically (via CLM_NAMELIST_OPTS) (note that that automatic setting is done in some CESM2.1 release configurations in order to have some compset/resolution combinations automatically give a hybrid case with use_init_interp = .true., to interpolate from the out-of-the-box refcase). This was a problem in the recent CESM tutorial.

We should reduce the error checking on `use_init_interp`, so that it can be set to `.true.` even if `finidat` is not set. In this case, it should just be ignored, and a warning message should be printed to the log file stating that we're ignoring `use_init_interp` in this branch run.

I reproduced this on my mac as follows:

First, I created a startup case:

```
./create_newcase --case /Users/sacks/projects/scratch/check_branch_baseline_0813a --compset I2000Clm50BgcCropQianRsGs --res f10_f10_musgs --run-unsupported

./xmlchange NTASKS=4
./xmlchange STOP_N=1
```

Then I created a branch:

```
./create_newcase --case /Users/sacks/projects/scratch/check_branch_0813a --compset I2000Clm50BgcCropQianRsGs --res f10_f10_musgs --run-unsupported

./xmlchange NTASKS=4
./xmlchange RUN_TYPE=branch
./xmlchange RUN_REFCASE=check_branch_baseline_0813a
./xmlchange RUN_REFDATE=2000-01-02
./xmlchange STOP_N=1
./xmlchange GET_REFCASE=FALSE
./xmlchange CLM_NAMELIST_OPTS=use_init_interp=.true.
```

And I prestaged the restart files from the first case into the second case's run directory.

### LINKED ISSUE #791: prescribed soil moisture (and LAI) streams are not threadsafe
### Brief summary of bug

Streams in cime for cime5.6.20 are not threadsafe. So any calls to streams needs to be done outside a clumps loop.

@olyson @swensosc 

### General bug information

**CTSM version you are using:** release-clm5.0.27

**Configurations affected:** 

With threading on and  use_soil_moisture_streams = .true.

### Details of bug

My test case is: ERP_P72x2_D_Ld3.f09_g17.I2000Clm50SpGs.cheyenne_intel.clm-prescribed

Note, the non-threaded version of above works fine.

Looks like the two important things are using the prescribed soil moisture streams, and running threaded.

The call in clm_driver looks like this...

``` fortran   !============================================================================

    !$OMP PARALLEL DO PRIVATE (nc,bounds_clump)
    do nc = 1,nclumps
       call get_clump_bounds(nc, bounds_clump)

       if (use_soil_moisture_streams) then
          call t_startf('prescribed_sm')
          call PrescribedSoilMoistureInterp(bounds_clump, soilstate_inst, &
               waterstate_inst)
          call t_stopf('prescribed_sm')
       endif


```

If it's moved outside the clump loop and run over bounds_proc, it should work fine.

> 21:MPT ERROR: shared memory sequence number error received 46073 expected 61394
> 26:MPT ERROR: Assertion failed at req.h:347: "0 <= ref_count"
> 21:MPT ERROR: shared memory sequence number error received 61394 expected 61395
> 0:forrtl: severe (151): allocatable array is already allocated
> 0:Image              PC                Routine            Line        Source             
> 0:cesm.exe           00000000043657BC  Unknown               Unknown  Unknown
> 0:cesm.exe           0000000003916CDD  shr_stream_mod_mp        1542  shr_stream_mod.F90
> 0:cesm.exe           00000000038FE5AB  shr_stream_mod_mp        1123  shr_stream_mod.F90
> 0:cesm.exe           00000000036EB10A  shr_dmodel_mod_mp         622  shr_dmodel_mod.F90
> 0:cesm.exe           0000000003891136  shr_strdata_mod_m         743  shr_strdata_mod.F90
> 0:cesm.exe           00000000022B7CC7  soilmoisturestrea         244  SoilMoistureStreamMod.F90
> 0:cesm.exe           0000000000882A6F  clm_driver_mp_clm         339  clm_driver.F90
> 0:libiomp5.so        00002B58773B6D13  __kmp_invoke_micr     Unknown  Unknown
> 0:libiomp5.so        00002B5877386B17  Unknown               Unknown  Unknown
> 0:libiomp5.so        00002B58773861C5  Unknown               Unknown  Unknown
> 0:libiomp5.so        00002B58773B7193  Unknown               Unknown  Unknown
> 0:libdplace.so.0.0.  00002B58731BA1AD  Unknown               Unknown  Unknown
> 0:libpthread.so.0    00002B58781A4734  Unknown               Unknown  Unknown
> 0:libc.so.6          00002B5878EA6D3D  clone                 Unknown  Unknown
> -1:MPT ERROR: MPI_COMM_WORLD rank 0 has terminated without calling MPI_Finalize()
> -1:	aborting job
>

### LINKED ISSUE #804: Allow mksurfdata to prescribe vegetation fractions, but not set the entire gridcell to vegetated-only
We'd like to be able to prescribe the vegetation fractions for a grid, but instead of assuming that all grid cells are 100% vegetated to use the PFT landmask to prescribe where those PFT's will be set. Right now when you prescribe the vegetation fractions with pft_frc and pft_idx it assumes that all grid-cells are 100% vegetated (so all other land-use types are ignored).

This is useful for the Delft LGM project. And may be useful in other instances as well. This can be straightforwardly done, by allowing all_veg to be set independent of use_input_pft. And when use_input_pft is on, but all_veg is not to still read some of the surface dataset, but NOT for setting PFT and CFT percentages.

This is dependent on #150.

### LINKED ISSUE #809: mkmapdata.sh is using an old version of intel compiler on cheyenne and needs to be updated
### Brief summary of bug

This happened with the last cheyenne upgrade, but we didn't notice it.  But, mkmapdata.sh has a module load for intel/17 and needs to be updated to use intel/18 as intel/17 isn't on the machine anymore.

### General bug information

**CTSM version you are using:** release-clm5.0

**Configurations affected:** cheyenne for running mkmapdata.sh
### Details of bug

This is on both master and release-clm5.0

``` bash
  intelvers=17.0.1
  module load intel/$intelvers
```

Note, this is similar to #655, but the change to remove intel-17 must have happened after that time (May/2019).

### LINKED ISSUE #817: Have clm4_5 use the same ndep file for 1850 as clm5_0

### LINKED ISSUE #831: Add namelist tint_algo namelist settings for input streams so can easily check that output agrees with input file
There are several streams options in CTSM: lai, ndep, ch4inundated, soilm, lightng, popdens, and urbantv. It would be good if all of these had a namelist option for tintalgo, so that you could easily check that the data being used by the model is the same as the input file as should be expected. Right now only soilm has such a namelist item.

### LINKED ISSUE #836: Add izumi as a place that can run the tools tests
Tools tests only work on cheyenne and hobart, so should be extended to izumi.

### LINKED ISSUE #871: By default turn off resolution checking in CTSM
The CTSM build-namelist by default turns on the option "-chk_res" which then checks that the input resolution is a valid one that CTSM knows about. This is set in the namelist_defaults_overall.xml file. Since, it's now more common to run at unsupported resolutions, we want to make this default off. The user could still turn it on if they want to. This also leaves in the place the supported resolution list that we find useful for running the CTSM inputfile tool chain. That toolchain already has a way to run an unsupported resolution, so the list of supported resolutions is still useful in that context. If someone provides an unsupported resolution, the namelist will still fail when it doesn't find the needed input surface dataset.

@barlage @negin513

### LINKED ISSUE #873: Does release-clm5.0.12 need to be moved to master?
@ekluzek - in the card for moving release changes to master (https://github.com/ESCOMP/CTSM/projects/6#card-13680676), you don't mention needing to move release-clm5.0.12 to master. But it looks like at least the fix for #550 is on the release branch but not on master, which makes me wonder if there are other changes that came into release-clm5.0.12 but not to master.

ekluzek:
### Description of changes

Brings changes on release-clm5.0 branch to master. Most of release-clm5.0.13 to release-clm5.0.30.

### Specific notes

  Fixes #873 -- Release clm5.0.12 changes
  Fixes #871 -- don't check for valid resolution for CESM cases by default
  Fixes #836 -- add izumi as a place that can run tools tests
  Fixes #831 -- add namelist tint_algo settings for input streams
  Fixes #817 -- Have clm4_5 use the same ndep file as clm5_0 for 1850
  Fixes #809 -- mkmapdata using an old version of intel
  Fixes #804 -- mksurfdata option
  Fixes  #791 -- fix threading for prescribed LAI and soil-moisture
  Fixes #786 -- allow use_init_interp for a branch
  Fixes #746 -- Some CLM45 configurations use CLM50 initial conditions files
  Fixes #741 -- C13_HR, C13_NBP, FPI values result in numeric conversion not representable error (changes answers)
  Fixes #734 -- Isotope historical files off of NetCDF-4
  Fixes #716 -- 2 degree CMIP6WACCMDECK needs Carbon isotopes off
  Fixes #707 -- Missing mapping files for hirespft
  Fixes #673 -- Ice shelf wetland fix in mksurfdata_map can lead to glacier+lake > 100% on surface datasets (changes surface dataset creation)
  Fixes #653 -- To save disk on fsurdat files use zwt0 on finindated file
  Fixes #629 -- Remove unused fields from surface datasets
  Fixes #624 -- QSNOEVAP history output incorrect
  Fixes #621 -- Fully remove rcp in place of only ssp_rcp
  Fixes #608 -- Odd (blocky) spatial patterns in generic crop leafc for BGC simulation when initialized from BGC-Crop
  Fixes #589 -- Existence of content in the lai_streams namelist makes it confusing to users
  Fixes #557 -- no-anthro compset and options
  Fixes #553 -- More robust mksurfdata_map logic for determining where to put wetlands
  Fixes #547 -- Add conus_30_x8 grid as valid option for CTSM and mksurfdata_map
  Fixes #545 -- Antarctica ice shelves are being treated as wetlands rather than glaciers
  Fixes #544 -- Transient cases are matching wrong IC file
  Fixes #533 -- Add -no-anthro option to mksurfdata
  Fixes #276 -- Tests for pre_alpha/pre_beta
  Fixes #262 -- hirespft option to mksurfdata.pl
 Fixes  #175 -- Set stream_year_first_urbantv/last for CLM4.5
  Fixes #150 -- mksurfdata code cleanup

Some of the namelist defaults change.

Testing performed, if any: will run regular testing


billsacks:
@ekluzek should release-clm5.0.12 also be moved to master? See #873

ekluzek:
Yes, I'm still figuring that one out. I will figure it out before this PR goes in.

ekluzek:
With some help from @billsacks I've got the last two tests working on cheyenne. Fates is still a problem as I had to turn off 2D fields to get it to run. That essentially means Fates can't be used at all on master. I'm now running the test list on izumi. This is up to date with dev084. So I also need to update to the latest. Looks like there a few files in conflict, but they shouldn't be a problem to fix.

ekluzek:
I got two finidat files in place 
that now get the last tests to pass...

ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic			
LII_D_Ld3.f19_g17_gl4.I2000Clm50BgcCrop.cheyenne_intel.clm-glcMEC_spunup_1way			
SMS_Ld2_D.f09_g17.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic_interp			
SMS_Lm1.f10_f10_musgs.I1850Clm50BgcCropCmip6waccm.cheyenne_gnu.clm-basic

ekluzek:
Regular testing has passed as expected.

Issue #235, opened 2018-01-29
Change pointer-based arrays to allocatable arrays where possible


billsacks:
Pointers are supposed to be worse for performance. We should assess the
impact of using pointers in all of our derived types, and (assuming this
has a significant impact) change the code to use allocatable arrays
instead of pointers.

* [ ] Assess impact of pointers on performance
    * Put timers around some representative part of the code
    * Check current timing
    * Change all pointers to allocatable arrays in that part of the code
        * May need to temporarily get rid of some other code that
          requires pointers, such as adding the relevant fields to the
          history field list)
        * Note that the allocatable arrays will still need the target
          attribute, because we do have a need to keep pointers to some
          of these things (e.g., for history fields)
    * Re-check timing
    
* [ ] Change infrastructure code that currently requires pointers to
  no longer require pointers
    * In cases where we need the ability to point to the passed-in
      array, the dummy argument can have the "target" attribute rather
      than the "pointer" attribute
    * In many cases, the infrastructure code that accepted pointers also
      made use of the lower bounds of the dummy variable - since
      pointers maintain their lower bounds when passed to a subroutine,
      whereas other arrays do not. In cases where the lower bound is
      needed (and not 1), this will now need to be passed in explicitly.
        * Note that we have plans to change all lower bounds for
          gridcell, landunit, column, etc. indices to be 1. If we do
          that first, this task will become easier - though we may still
          need to pass in the lower bounds in a few cases for level
          dimensions.
      
* [ ] Change arrays to allocatable with the pointer attribute

billsacks:
Based on investigations in https://github.com/billsacks/prototypes-fortran_vectorization, pointers (without the "contiguous" attribute) *do* prevent ifx from vectorizing code; allocatables or pointers with the contiguous attribute allow vectorization. For nvfortran, the "contiguous" attribute doesn't seem to have an impact. This does provide some argument for changing pointers to allocatables.

However, ifx and nvfortran exhibited conflicting results in terms of the effect of the "associate" statement: For ifx, "associate" prevents profitable vectorization for any dynamically-sized arrays; for nvfortran, in contrast, "associate" allows vectorization where it otherwise wasn't being done before (this may be a bug).

[...]

PR #900, opened 2020-01-30
[DO NOT MERGE] Demonstrate a method for handling ciso terms in C truncation

### LINKED ISSUE #811: Some bounds assertions aren't done when threading is on
In these two blocks of code:

https://github.com/ESCOMP/CTSM/blob/1720bc754a75bbf3cab67424c63c292a2b6e45eb/src/biogeochem/CNPrecisionControlMod.F90#L438-L447

https://github.com/ESCOMP/CTSM/blob/1720bc754a75bbf3cab67424c63c292a2b6e45eb/src/biogeochem/CNPrecisionControlMod.F90#L525-L534

the bounds of subroutine arguments are not being checked if threading is on. It looks like this came in in `clm4_5_9_r184`. Presumably that was needed at some point, but I hope we can remove this now: if this is needed, I think that would suggest a threading bug.

@ekluzek I'm assigning this to you since you made `clm4_5_9_r184`, but it probably isn't high priority. The fix should be simple (removing the `#ifdefs`) as long as it doesn't cause problems.

billsacks:
This PR is for discussion only. When @negin513 was trying to fix #811 , she ran into problems which I guess point out why the code referenced in #811 was done that way in the first place. The basic problem is that the c13 / c14 arrays aren't always allocated, and so we can't use the standard method for passing sub-arrays into subroutines (e.g., `foo(bounds%begp:bounds%endp)`).

My first thought was that we could define variables like `c13_begp` and `c13_endp`; these would be set equal to `bounds%begp` and `bounds%endp` if `use_c13` is true, and 0 otherwise. So if `use_c13` is false, we would pass the slice `foo(0:0)`. But I'm not sure if that will work on all compilers (or any compiler, for that matter).

Then I thought about how I am handling things like this with water isotopes: Basically, first I do some operation on the bulk, possibly returning a logical array or filter specifying the points that need to be operated on for each isotope. Then I handle the isotopes in a separate call, passing in the already-computed logical array or filter. This is what I have done here.

I don't love the extra code needed for each call to TruncateCandNStates, but I do like that: (1) it allows us to handle the c13 and c14 arrays like everything else in CTSM, including bounds checking; (2) it doesn't require c13 and c14 arrays to be pointers (important if we ever want to tackle #235 ); (3) it cleans up the body of TruncateCandNStates.

This is only partially implemented and won't compile yet. Needs are:

(1) Changing all of the calls to TruncateCandNStates to use the new pattern

(2) Applying the same pattern to TruncateCStates - in both the implementation of this routine and all calls to it

But before I went further with this, I wanted to run it by @ekluzek and @negin513 . Does this seem like a good solution to both of you? If so, I'd ask @negin513 to finish the implementation.

@ekluzek - I'm a little nervous about rewriting this module like this without knowing if this code is actually triggered typically. Erik, do you know if this code is triggered in typical runs? In particular, can we expect it to be triggered in our ciso tests? Ideally, I'd want a test that triggers truncation for at least one point for each and every call to TruncateCandNStates and TruncateCStates to ensure that all of the code has been refactored properly. That may not be feasible, but I'd at least hope for one variable that triggers truncation in one of the ciso tests; then that test demonstrating bit-for-bit (which would show that the basic logic in TruncateCandNStates is good) together with a very careful code review (focusing on whether the correct variables are used for each call to TruncateCandNStates and TruncateCStates) could give me enough confidence that the refactoring was done properly.

@negin513 if @ekluzek doesn't know the answer to that question, then I'll ask you to investigate which of these calls actually have truncation occurring in our typical ciso tests. If none do, then we should either run a long test (e.g., 10-year f10 case, if that triggers these) with comparison against baselines, or somehow force some of these truncations to be triggered.

### DIFF
--- a/src/biogeochem/CNPrecisionControlMod.F90
+++ b/src/biogeochem/CNPrecisionControlMod.F90
@@ -120,6 +120,8 @@ contains
     ! !LOCAL VARIABLES:
     integer :: p,j,k                             ! indices
     integer :: fp                                ! filter indices
+    integer :: num_truncatep                     ! number of points in filter_truncatep
+    integer :: filter_truncatep(bounds%endp-bounds%begp+1) ! filter for points that need truncation
     real(r8):: pc(bounds%begp:bounds%endp)       ! truncation terms for patch-level corrections Carbon
     real(r8):: pn(bounds%begp:bounds%endp)       ! truncation terms for patch-level corrections nitrogen
     real(r8):: pc13(bounds%begp:bounds%endp)     ! truncation terms for patch-level corrections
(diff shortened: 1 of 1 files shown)


billsacks:
Closing, as this has been incorporated in #923

PR #901, opened 2020-01-30
FATES: Condition the zeroing of qflx bc_in upon having hydro on

rgknox:
### Description of changes

This is a bug fix. The boundary condition to fates for tranpiration flux is only allocated when the hydraulics module is active.  This variable was being zero'd for cases where the array was not allocated.

### Specific notes

@serbinsh and  @Qianyuxuan found the error.

FATES issue fixed:  https://github.com/NGEET/fates/issues/605

Are answers expected to change (and if so in what way)?  No, but I don't know how the tests did not catch this!

No

Testing performed, if any:

TBD

### DIFF
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -1407,9 +1407,9 @@ contains
 
         ! set transpiration input boundary condition to zero. The exposed
         ! vegetation filter may not even call every patch.
-
-        this%fates(nc)%bc_in(s)%qflx_transp_pa(:) = 0._r8
-
+        if (use_fates_planthydro) then
+            this%fates(nc)%bc_in(s)%qflx_transp_pa(:) = 0._r8
+        end if
 
      end do
   end subroutine prep_canopyfluxes


serbinsh:
@rgknox I am running a test case for the Brazil 1x1 - the fix allowed the model to run. Is there a specific test I can run for this PR?

rgknox:
@serbinsh thanks for the help btw.  We will probably just run the standard regression suite (if that).  But @glemieux and I plan to pull this in pretty soon.   I suppose just let us know if in your science runs you encounter problems.

glemieux:
On a related note, @rgknox and I chatted about adding regression tests (as a separate PR) using other compliers to see if we can catch stuff that might pass through the intel compiler.

serbinsh:
@glemieux @rgknox I think additional tests would be good. We would happy to be guinea pigs here to try out tests for GNU, etc.

glemieux:
I kicked off the regression tests with the intel compiler before heading out of the office yesterday.  All expected PASS.

glemieux:
Verified that the gnu test pass on this PR branch and fail on the fates_next_api branch.

PR #907, opened 2020-01-31
Adjusting regression test definitions

glemieux:
### Description of changes
Updating regression test definition

### Specific notes
This came up in the context of PR #901 (we apparently didn't test on the gnu compiler, only intel).  We thought it might be good to add a smoke test to the test definition list.  Also, we realized that we were testing the highest resolution grid on clm4.5 and Ryan thought we should update to clm5.  Finally, I removed all calls to `edison` which is now retired (in all categories, not just `fates`).

Contributors other than yourself, if any: @rgknox @serbinsh 

Are answers expected to change (and if so in what way)? No.  We will now have an additional set of test outputs for gnu, and the `f09` resolution will use clm5 so a new baseline will need to be generated.

Testing performed, if any: TBD


cime_config/testdefs/testlist_clm.xml
```diff
@@ -1390,7 +1389,6 @@
   <test name="SMS_Ld1" grid="f19_g17" compset="I2000Clm50BgcCruGs" testmods="clm/default">
     <machines>
       <machine name="cheyenne" compiler="intel" category="prebeta"/>
-      <machine name="edison" compiler="intel" category="prebeta"/>
```
billsacks: Here and in a couple of other places: you removed an edison prebeta test. I'm thinking you just meant to update the fates test list, not our prebeta test list?

glemieux: @billsacks I purposely did that assuming it would be fine since edison is retired.  Is it better practice not to change since it's across categories/branches?

glemieux:
@rgknox it looks like the added test `SMS_Lm6.f45` is failing on RUN currently.  I'm kind of at a loss as to why.  My output is here:
```
/glade/u/home/glemieux/scratch/clmed-tests/fates-sci-1.31.1_api.8.1.0-C07d9ae60-F17aaa879.fates.cheyenne.gnu
```

olyson:
I just had a similar problem, so thought I'd let you know what I see in SMS_Lm6.f45_f45_mg37.I2000Clm50FatesCruGs.cheyenne_gnu.clm-Fates.C.20200131_155554_1pjt2v.test.o686914:

=>> PBS: job killed: walltime 1228 exceeded limit 1200
ERROR: Job killed due to receiving signal 15 (SIGTERM)

glemieux:
Thanks @olyson.  This is the first time I've had something fail due to exceeding wall-time.  Is it typical for an equivalent test to have different run times depending on the compiler?

olyson:
You should probably ask a real software engineer :)
Could it just be system variability?
I've also noticed some file system hangs on cheyenne today.

cime_config/testdefs/testlist_clm.xml
billsacks: Our CESM testing uses the edison tests for testing on cori (confusingly).

billsacks:
> Is it typical for an equivalent test to have different run times depending on the compiler?

Yes this is typical.

Variability in run length is also typical - especially if there are file system issues.

I tend to give tests at least 2-3x the expected wallclock time.

cime_config/testdefs/testlist_clm.xml
glemieux: Ha, thanks for the clarification.  I'll make the changes.

glemieux:
Switched out the f45 smoke test for the 1x1 brazil.  Test result are here:
```
/glade/u/home/glemieux/scratch/clmed-tests/fates-sci-1.31.1_api.8.1.0-C7675ac17-F17aaa879.fates.cheyenne.gnu
```
Note I pre-merged fates_next_api on my testing branch to pull in #901 fix.  I didn't commit that pre-merge to my remote branch though which is why the commit hashes differ.

rgknox:
> Looks good.
> Edison is gone.
> I like the 3 cheyenne/gnu tests we have.

Noting that "edison" is used for cori, didn't know that, thanks @billsacks

ekluzek:
I put this in as a "Work in Progress" feel free to modify if that isn't right.

glemieux:
Huh, I thought we had already merged this one.  It looks like it's good to go since @rgknox and @billsacks approved.  Guess it just fell down the queue.  I'll resolve the merge conflicts.

glemieux:
Retesting in progress.

glemieux:
All expected PASS:
```
/glade/u/home/glemieux/scratch/clmed-tests/ctsm-pr907-C903397a4-F0059286f.fates.cheyenne.gnu
/glade/u/home/glemieux/scratch/clmed-tests/ctsm-pr907-C903397a4-F0059286f.fates.cheyenne.intel
```
@rgknox I'll run an update baseline post merge.  Would you do the honors?

PR #923, opened 2020-02-10
Resolve Issue #811: Bounds assertion for ciso when threading is on

negin513:
### Description of changes
This PR request resolves issue #811 . 
The bounds of arguments in subroutines `TruncateCStates` and `TruncateCandNStates` 
are not getting checked when threading is on for c13 and c14. 
Removing `#ifndef _OPENMP` was causing problems since c13 and c14 are not always allocated. 
@billsacks  suggested using a similar method that is used for handling water isotopes here.
 Based on this method the isotopes (i.e. c13 and c14) are being handled separately from the bulk and in another subroutine. Here we use `TruncateAdditional` for c13 and c14 only when they are allocated ( `if (use_c13)` or ` if (use_c14)`). 

### Specific notes

CTSM Issues Fixed (include github issue #):  #811 

Any User Interface Changes (namelist or namelist defaults changes)? None

Testing performed, if any:
Tested `clm_short` on Cheyenne. I will run standard testing ( `aux_clm`) for Cheyenne and Izumi.

### DIFF
--- a/cime_config/testdefs/testlist_clm.xml
+++ b/cime_config/testdefs/testlist_clm.xml
@@ -432,7 +432,7 @@
       <machine name="cheyenne" compiler="intel" category="aux_clm"/>
     </machines>
     <options>
-      <option name="wallclock">0:20:00</option>
+      <option name="wallclock">00:20:00</option>
       <option name="comment"  >Do a test with RTM and flooding on as that also impacts CLM code</option>
     </options>
   </test>
@@ -740,7 +740,7 @@
     <machines>
       <machine name="cheyenne" compiler="intel" category="aux_clm">
         <options>
-          <option name="wallclock">0:30:00</option>
+          <option name="wallclock">00:30:00</option>
           <option name="comment">Want an ERP test covering some non-default irrigation options. Long enough so that we're likely to exercise the various groundwater irrigation code.</option>
         </options>
       </machine>
@@ -1646,7 +1646,7 @@
     <machines>
       <machine name="izumi" compiler="nag" category="aux_clm">
         <options>
-          <option name="wallclock">0:20:00</option>
+          <option name="wallclock">00:20:00</option>
           <option name="comment">Debug test covering some non-default irrigation options.</option>
         </options>
       </machine>
@@ -1657,7 +1657,7 @@
 
       <machine name="izumi" compiler="intel" category="aux_clm">
         <options>
-          <option name="wallclock">0:30:00</option>
+          <option name="wallclock">00:30:00</option>
           <option name="comment">Include a tracer consistency check in debug mode.</option>
         </options>
       </machine>
@@ -1669,7 +1669,7 @@
 
       <machine name="cheyenne" compiler="intel" category="aux_clm">
         <options>
-          <option name="wallclock">0:30:00</option>
+          <option name="wallclock">00:30:00</option>
           <option name="comment">A debug ERP test of the NWP configuration with active BGC and CROP.</option>
         </options>
       </machine>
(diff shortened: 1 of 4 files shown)


negin513:
I've tested a couple of carbon isotope tests to make sure that `TruncateCStates` and `TruncateCandNStates` are being called and the inner loop conditionals also triggered:
`ERP_Ld3.f09_g17.I1850Clm50BgcCropCru.cheyenne_intel.clm-ciso` 
`ERP_P36x2_D_Ld5.f10_f10_musgs.I1850Clm50Bgc.cheyenne_intel.clm-ciso`

Both of the above tests use `TruncateCStates` and `TruncateCandNStates` to fill in `filter_truncatep` to use with `TruncateAdditional`. 
Both of these tests show bit-for-bit results comparing to the baselines.

billsacks:
One more question... sorry, I should have asked this before: Have you checked if there are any points where, in TruncateAdditional, a non-zero value is changed to a zero value? That's what we really want to confirm (because, if the "truncation" is on values that are already 0, that doesn't tell us if the filters are set up right).

i.e., these diffs

```diff
diff --git a/src/biogeochem/CNPrecisionControlMod.F90 b/src/biogeochem/CNPrecisionControlMod.F90
index 2617dce9..fa598dc9 100644
--- a/src/biogeochem/CNPrecisionControlMod.F90
+++ b/src/biogeochem/CNPrecisionControlMod.F90
@@ -839,6 +839,9 @@ subroutine TruncateAdditional( bounds, num_truncatep, filter_truncatep, &
 
    do fp = 1, num_truncatep
       p = filter_truncatep(fp)
+      if (state_patch(p) /= 0._r8) then
+         write(iulog,*) 'TruncateAdditional truncating non-zero value: ', state_patch(p)
+      end if
       truncation_patch(p) = truncation_patch(p) + state_patch(p)
       state_patch(p) = 0._r8
    end do
```

(Note that I do *not* have a masterproc conditional around that: it may be that this only happens rarely, in which case we want to see it happening regardless of whether it's on the master proc or some other proc.)

negin513:
Yes! For example test case `ERP_Ld3.f09_g17.I1850Clm50BgcCropCru.cheyenne_intel.clm-ciso` has the following in its output:

```
 TruncateAdditional truncating non-zero value:   1.564893370052320E-019
 TruncateAdditional truncating non-zero value:  -2.778268066994105E-019
 TruncateAdditional truncating non-zero value:   5.251604272976662E-019
 TruncateAdditional truncating non-zero value:  -2.317482143687766E-018
 TruncateAdditional truncating non-zero value:   3.737477822618029E-016
 TruncateAdditional truncating non-zero value:  -6.437450399132683E-020
 TruncateAdditional truncating non-zero value:   8.381647117973250E-030
 TruncateAdditional truncating non-zero value:   6.508102468073347E-030
 TruncateAdditional truncating non-zero value:   1.755215514116751E-029
 TruncateAdditional truncating non-zero value:  -1.814380082008327E-028
 TruncateAdditional truncating non-zero value:   3.459731923931080E-026
 TruncateAdditional truncating non-zero value:  -1.597443333072549E-029
...
```
The state_patch(p) values that are being truncated  are too low (e-16 to e-30). Would that be okay?

negin513:
Also, I 've noticed in `TruncateNStates` , the code is not checking if ` nitrogen_patch(p)` less than the `nnegcrit`. In fact, the warning and error messages for this are commented out:

https://github.com/ESCOMP/CTSM/blob/6c9c8b5d8ea000f6ff26d5b094eec116ad034981/src/biogeochem/CNPrecisionControlMod.F90#L802

If that `if` statement is redundant should we remove it for performance purposes?

billsacks:
Thanks for doing this extra check, @negin513 . I feel this is now ready to go.

negin513:
@billsacks I think this PR should be ready to be merged.

Issue #922, opened 2020-02-07
Turn on anomaly forcing for ISSP compsets


ekluzek:
Automatically turn on the anomaly forcing when ISSP compsets are being used. This should go on both release-clm5.0 and master.

The primary change is in cime, which should go on maint-5.6 as well as master.

ESMCI/cime/#3391

Periodically changes in cime from maint-5.6 are automatically moved to master.

wwieder:
link to #987

ekluzek:
I don't think we should do this on the release-clm5.0 branch anymore. But, we do want it on main development still.

wwieder:
Yes, I think we want this feature available on the development branch.
Thanks Erik

PR #949, opened 2020-03-26
Surface dataset creation for 2100-2300 SSP extensions (SSP1-2.6, SSP5-3.4, SSP5-8.5)

### LINKED ISSUE #921: PGI build fails on izumi for release branch
### Brief summary of bug

The build is failing for PGI on izumi

### General bug information

**CTSM version you are using:** release-clm5.0.30
**Does this bug cause significantly incorrect results in the model's science?** N

**Configurations affected:** any build on izumi with PGI

### Details of bug

SMS_D.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-crop

Here's the fix:

``` fortran
diff --git a/src/utils/clmfates_paraminterfaceMod.F90 b/src/utils/clmfates_paraminterfaceMod.F90
index 9a3ce58f..2f63389a 100644
--- a/src/utils/clmfates_paraminterfaceMod.F90
+++ b/src/utils/clmfates_paraminterfaceMod.F90
@@ -2,6 +2,7 @@ module CLMFatesParamInterfaceMod
   ! NOTE(bja, 2017-01) this code can not go into the main clm-fates
   ! interface module because of circular dependancies with pftvarcon.
 
+  use shr_kind_mod, only : r8 => shr_kind_r8
   use FatesGlobals, only : fates_log
   
   implicit none
@@ -172,7 +173,6 @@ end subroutine GetUsedDimensionSizes
  subroutine ParametersFromNetCDF(filename, is_host_file, fates_params)
 
-   use shr_kind_mod, only: r8 => shr_kind_r8
    use abortutils, only : endrun
    use fileutils  , only : getfil
    use ncdio_pio  , only : file_desc_t, ncd_pio_closefile, ncd_pio_openfile

```

> PGF90-S-0087-Non-constant expression where constant expression required (/fs/cgd/data0/erik/ctsm_relanswerchangesfor5029/src/utils/clmfates_paraminterfaceMod.F90: 197)
> PGF90-S-0081-Illegal selector - KIND parameter has unknown value for data type  (/fs/cgd/data0/erik/ctsm_relanswerchangesfor5029/src/utils/clmfates_paraminterfaceMod.F90: 197)
>   0 inform,   0 warnings,   2 severes, 0 fatal for parametersfromnetcdf
> gmake: *** [clmfates_paraminterfaceMod.o] Error 2
> gmake: *** Waiting for unfinished jobs....
>   Timing stats:

### LINKED ISSUE #933: Add some notes about creating landuse.timeseries files with mksurfdata_map
The README, users-guide, and mksurfdata.pl help don't give enough instruction on using year 1850-2100 and such for transient landuse timeseries files. Add some more notes on doing this and also add a note to the -dynpft option of mksurfdata.pl, to say that it's NOT normally used.

### LINKED ISSUE #946: Have CTSM give a better error message when ndep files are missing especially for the case when coupled to CAM
### Brief summary of bug

Dummy ndep files are required even for WACCM simulations that don't need them.

### General bug information

**CTSM version you are using:** release-clm5.0.30

**Configurations affected:** Compsets with CTSM and WACCM

### Details of bug

If a WACCM compset doesn't find a ndep file, the case will fail to build, or run preview_namelists. This is even though when WACCM is being run it doesn't actually 
use the ndep file. It would be better for documentation to NOT list a file that isn't actually being used for the case. When run with WACCM the nitrogen-deposition is coming from CAM.

The code can check for this by making sure it either gets n-dep from CAM or from the ndep file. If it doesn't get it from either it should fail.

The case that failed was a SSP5-3.4 case (since we don't have a ndep file for it). It would also fail for the non Tier-I SSP scenarios where we don't have ndep files. And note for I compsets it will fail when it can't find appropriate presaero files.

It would also be helpful to make the error from build-namelist to be more helpful.

> ERROR: Command /glade/u/home/cmip6/cesm_tags/release-cesm2.1.2/components/clm/bld/build-namelist failed rc=255
> out=CLM adding use_case 1850-2100_SSP5-3.4_transient defaults for var 'sim_year' with val '1850'
> CLM adding use_case 1850-2100_SSP5-3.4_transient defaults for var 'sim_year_range' with val '1850-2100'
> CLM adding use_case 1850-2100_SSP5-3.4_transient defaults for var 'ssp_rcp' with val 'SSP5-3.4'
> CLM adding use_case 1850-2100_SSP5-3.4_transient defaults for var 'use_case_desc' with val 'Simulate transient land-use, aerosol deposition, and Nitrogen deposition changes from 1850 to current day with historical data, and then to 2100 with the CMIP6 SSP5-3.4 scenario'
> err=ERROR : CLM build-namelist::CLMBuildNamelist::add_default() : No default value found for stream_fldfilename_ndep.
>             Are defaults provided for this resolution and land mask?

ekluzek:
### Description of changes

Bring in changes to add the extensions of 2100-2300 to mksurfdata_map
creation for SSP1-2.6, SSP5-.8.5, SSP5-3.4.

Fix the pgi build on izumi.
Bring in anomaly forcing script to create af forcing
that can be used by datm
Update manage_externals for python 3.8 bug.

### Specific notes

Contributors other than yourself, if any: @lawrencepj1 

CTSM Issues Fixed (include github issue #): #921 #922 #933 #946

Fixes #921 
Sets up the script to use for #922 
Fixes #933 
Fixes #946

If the ndep file is missing the error message is better, and -ignore_warnings will
turn it off (in the case of running with WACCM when 

Testing performed, if any: starting standard testing


ekluzek:
Fates tests on izumi were showing differences, but it looks like the release-clm5.0.30 baselines weren't correct, so I'm redoing those.

Issue #239, opened 2018-01-30
Move documentation source to master


billsacks:
Currently, both the source and build of the documentation resides on the gh-pages branch. We should move the source to master, adopting a workflow like https://github.com/esmci/cime/wiki/Working-with-Sphinx-and-reStructuredText

billsacks:
**Proposed workflow, copied from a post that I made here two days ago: https://github.com/orgs/ESCOMP/teams/ctsm-write**

@bertinia and I just had a long discussion to try to come up with a plan for hosting the CTSM documentation in a way that would allow us to host separate frozen versions of the documentation (e.g., corresponding to the CLM5.0 release) alongside the latest dev version. Please let us know if you have any thoughts on this.

**Summary of the plan:** Documentation source (rst files) will live in the main ctsm repo. Built versions of the documentation will live in a separate repository, named `ctsm-docs`, and so will appear at `escomp.github.io/ctsm-docs`. The top-level landing page will give you a list of available versions ('Development', 'CLM5.0 release', etc.) with links. Then the built documentation for individual releases will be stored in subdirectories of the ctsm-docs repo. So, for example, we'll have `escomp.github.io/ctsm-docs/development/` and `escomp.github.io/ctsm-docs/release-clm5.0/`.

**More details, discussion and rationale**

@dlawrenncar and @olyson expressed a preference for having the documentation hosted in the same place as the code, which argues for having it hosted on a github.io site rather than (e.g.) readthedocs. The problem this raised was how to keep multiple versions of the documentation.

@bertinia and I considered https://robpol86.github.io/sphinxcontrib-versioning/v2.2.1/index.html . This uses the same general idea of having a subdirectory for each version of the documentation, but adds some automation to this and gives a nice list of versions in the sidebar of the html page. However, we ran into some problems with this tool, and it seems like this project has stop being maintained, so we gave up on this, at least for now. (We could always adopt this later if it starts being maintained again.)

The rationale for having a separate repository for the documentation builds is: If this were in the main repo, people would get these builds whenever they clone the repo. Even a single build doubles or triples the size of the repository. As we start supporting more side-by-side builds, the issue of repository bloat will grow. Since most people won't want or need these doc builds, I thought it would be better to isolate these builds to their own repository rather than having them on an orphan branch in the main repository.

[...]

PR #954, opened 2020-03-30
Bring documentation source to master

billsacks:
### Description of changes

1. Bring documentation source to master: Pulls in the source from
   https://github.com/escomp/ctsm-docs. This is important so that
   documentation can remain in sync with changes in the model
   code. Images are stored here using git-lfs (Git Large File
   Storage). I also made some minor fixes to get the pdf build of the
   tech note working.
   
2. Use a different documentation theme that supports a version dropdown
   menu, and add the code needed to support this versioning on the
   documentation web pages. At a high level, the way the versioned
   documentation works is to have separate subdirectories in the
   gh-pages branch of the ctsm-docs repository for each version of the
   documentation we want to support. There is then a bit of JavaScript
   code which uses a json file in the gh-pages branch to determine which
   versions exist and how these should be named in the dropdown
   menu. Most of these changes were borrowed from
   https://github.com/ESMCI/cime/pull/3439, which in turn borrowed from
   https://github.com/ESCOMP/CISM-wrapper/pull/23, which in turn was a
   slight modification of an implementation provided by @mnlevy1981 for
   the MARBL documentation, which in turn borrowed from an
   implementation put together by Unidata (credit where credit is due).
   
   I am not aware of out-of-the-box support for a version pull-down in
   out-of-the-box sphinx themes (though the last time I looked was in
   Fall, 2018, so there may be something available now). However,
   support for a version dropdown exists in an open PR in the sphinx
   readthedocs theme repository: readthedocs/sphinx_rtd_theme#438. I
   have pushed this branch to a new repository in ESMCI
   (https://github.com/ESMCI/sphinx_rtd_theme) to facilitate long-term
   maintenance of this branch in case it disappears from the official
   sphinx_rtd_theme repository. I have also cherry-picked a commit onto
   that branch, which is needed to fix search functionality in sphinx1.8
   (from readthedocs/sphinx_rtd_theme#672) (which is another reason for
   maintaining our own copy of this branch). The branch in this
   repository is now named version-dropdown-with-fixes (branching off of
   the version-dropdown branch in the sphinx_rtd_theme repository). In
   the long-term, I am a little concerned about using this theme that
   isn't showing any signs of being merged to the main branch of the
   readthedocs theme, but this has been working for us in other projects
   for the last 2 years, so I feel this is a reasonable approach in the
   short-medium term.

### Specific notes

- Resolves ESCOMP/ctsm#239


billsacks:
@ekluzek - as I mentioned by email, this would be a beast to review, and I don't think the PR itself really requires a detailed review since most of the changes are simply pulling in the documentation source verbatim from the ctsm-docs repo, and it's more useful to review the resulting documentation build (https://billsacks.github.io/ctsm-docs).

However, if you want to review the interesting pieces of the PR, I suggest focusing on these commits:
- dd76d2a66cae1151980ac1430956fde2be5337e4
- 13a3883b3c629c6b72de951568e31db46e51b423
- 50e7e62f2bf58245baee54c4b53b83feca9566c5
- fb4e647ea8b95bec413669fdb99bb917c839b8ff
- 885a37e38fb83e9dd10b96bdbb7835d0fb88e222
- d1ea77c6cb1b70fbaa6504bceef9e97fa0ec7581

billsacks:
I have updated the wiki page on the documentation:

https://github.com/ESCOMP/CTSM/wiki/Directions-for-editing-CLM-documentation-on-github-and-sphinx

@ekluzek as far as I'm concerned this is ready to come to master. But please let me know if you'd like me to wait for you do any further review before I merge it.

doc/source/conf.py
```diff
@@ -0,0 +1,185 @@
...
+project = u'ctsm'
+copyright = u'2020, UCAR'
+author = u''
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = u'master'
```
ekluzek, line 65: I'm thinking 

version = u'CTSM1'
release = u'CTSM master'

doc/source/users_guide/substitutions.rst
```diff
@@ -0,0 +1,5 @@
+.. |cesmrelease| replace:: CESM2.0.0
```
ekluzek, line 1: "CESM2.0.0" should be "CESM2.2 beta"

doc/source/conf.py
```diff
@@ -0,0 +1,185 @@
...
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'ctsm'
+copyright = u'2020, UCAR'
+author = u''
```
ekluzek, line 58: I think author should probably stay as

"Erik Kluzek, Bill Sacks, Ben Andre, Alice Bertini"

since we are the ones that worked with the bulk of the content. Or it could set set to 

"CTSM software team"

billsacks: I disagree with this. I removed the authorship information because I felt it did not give credit to the appropriate people. There are many people who should get credit, as carefully laid out in https://escomp.github.io/ctsm-docs/versions/master/html/tech_note/Introduction/CLM50_Tech_Note_Introduction.html for the tech note. Furthermore, I'm not sure whether this authorship information actually gets used anywhere (it may, but if it does, I didn't see how at a glance), so it didn't feel worth putting a lot of effort into coming up with an accurate authorship list (especially given that this would be hard to maintain accurately).

billsacks:
@ekluzek I have addressed your comments about the versions. As noted above, I disagree about the authorship. If you feel strongly about this, I'd request that we bring this to master as it is now (so our collaborators can start working on their documentation) and then have a broader discussion on this point - though I'd like to see a demonstration that it actually makes a difference in the built documentation before putting any more thought into it. I'm not sure what the pergro jpg files are that you're referring to. Note that I removed the entire old UsersGuide directory, which included two pergro image files. Are there others that you still see somewhere?

billsacks:
@ekluzek - also: once you approve this, I'll do a final rebuild of the documentation.

ekluzek:
OK this is good. You are right about author not actually showing up. Note, that there are three conf.py files: one global one, one for the tech note and one for the UG. I was thinking about the authorship of the UG and NOT the Tech Note, which obviously has a ton of authors. And the authorship of each is completely different.

But, you are also right that it appears that the author replacement string isn't actually used anywhere. So it doesn't actually matter.

billsacks:
@ekluzek thanks for your comment. There used to be three different conf.py files, but that's no longer the case: now there is just one. I removed the separate conf.py files for the tech note and user's guide for two reasons:

1. It was going to be a pain to try to keep these three files in sync moving forward

2. It didn't seem like these tech note and user's guide conf.py files were actually used... at least, not consistently. From what I could tell, if you built the full documentation, the top-level conf.py file was used for both the tech note and user's guide (so the lower-level conf.py files were being ignored). The tech note's conf.py file was being used for the pdf build of the tech note, but I have gotten things working so that that can use the top-level conf.py file, too, and give the same results as before.

Issue #163, opened 2017-12-16
Have CLM listen to the driver for number of threads rather than OMP_NUM_THREADS


ekluzek:
**Erik Kluzek < erik > - 2017-02-10 18:10:02 -0700**
**Bugzilla Id:** [2421](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2421)
**Bugzilla CC:**  dlawren,  mvertens,  rfisher, 

This is CLM's part of Issue 864 for cime, where the components don't listen to what the driver sets the number of threads for them -- but use the env variable OMP_NUM_THREADS

https://github.com/ESMCI/cime/issues/864

This can be a problem if another component is threaded, and you don't want CLM to be. It'll end up being threaded anyway.

ekluzek:
**Erik Kluzek < erik > - 2017-02-10 18:20:45 -0700**

This is a problem both in controlMod and decompMod. The seq_comm_getnthreads interface to seq_comm_mod needs to be queried in lnd_comp_mct and the result saved in CLM.

PR #967, opened 2020-04-03
get actual thread count instead of max

jedwards4b:
### Specific notes

CTSM Issues Fixed (include github issue #): #163 

Testing performed, if any:  Tested SMS_D_Ld5_Vnuopc.f10_f10_musgs.I1850Clm45BgcCru.cheyenne_intel.clm-ciso
with several different pe layouts. 

### DIFF
--- a/src/cpl/mct/lnd_comp_mct.F90
+++ b/src/cpl/mct/lnd_comp_mct.F90
@@ -28,6 +28,7 @@ module lnd_comp_mct
   private :: lnd_domain_mct    ! set the land model domain information
   private :: lnd_handle_resume ! handle pause/resume signals from the coupler
   !---------------------------------------------------------------------------
+  integer, private      :: LNDID	     ! Land identifyer
 
 contains
 
@@ -59,13 +60,14 @@ contains
     use seq_infodata_mod , only : seq_infodata_type, seq_infodata_GetData, seq_infodata_PutData, &
                                   seq_infodata_start_type_start, seq_infodata_start_type_cont,   &
                                   seq_infodata_start_type_brnch
-    use seq_comm_mct     , only : seq_comm_suffix, seq_comm_inst, seq_comm_name
+    use seq_comm_mct     , only : seq_comm_suffix, seq_comm_inst, seq_comm_name, seq_comm_setptrs
     use seq_flds_mod     , only : seq_flds_x2l_fields, seq_flds_l2x_fields
     use spmdMod          , only : masterproc, spmd_init
     use clm_varctl       , only : nsrStartup, nsrContinue, nsrBranch
     use clm_cpl_indices  , only : clm_cpl_indices_set
     use mct_mod          , only : mct_aVect_init, mct_aVect_zero, mct_gsMap_lsize
     use ESMF
+!$  use omp_lib         , only : omp_set_num_threads
     !
     ! !ARGUMENTS:
     type(ESMF_Clock),           intent(inout) :: EClock           ! Input synchronization clock
@@ -74,7 +76,6 @@ contains
     character(len=*), optional, intent(in)    :: NLFilename       ! Namelist filename to read
     !
     ! !LOCAL VARIABLES:
-    integer                          :: LNDID	     ! Land identifyer
     integer                          :: mpicom_lnd   ! MPI communicator
     type(mct_gsMap),         pointer :: GSMap_lnd    ! Land model MCT GS map
     type(mct_gGrid),         pointer :: dom_l        ! Land model domain
@@ -83,6 +84,7 @@ contains
     integer  :: g,i,j                                ! indices
     integer  :: dtime_sync                           ! coupling time-step from the input synchronization clock
     integer  :: dtime_clm                            ! clm time-step
+    integer  :: nthreads
     logical  :: exists                               ! true if file exists
     logical  :: atm_aero                             ! Flag if aerosol data sent from atm model
     real(r8) :: scmlat                               ! single-column latitude
@@ -115,6 +117,9 @@ contains
     call seq_cdata_setptrs(cdata_l, ID=LNDID, mpicom=mpicom_lnd, &
          gsMap=GSMap_lnd, dom=dom_l, infodata=infodata)
 
+    call seq_comm_setptrs(LNDID, nthreads=nthreads)
+!$  call omp_set_num_threads(nthreads)
+
     ! Determine attriute vector indices
 
     call clm_cpl_indices_set()
@@ -122,6 +127,8 @@ contains
     ! Initialize clm MPI communicator
 
     call spmd_init( mpicom_lnd, LNDID )
+
+
 
 #if (defined _MEMTRACE)
     if(masterproc) then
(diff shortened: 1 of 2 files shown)


ekluzek:
@jedwards4b a few questions about this. So OMP_MAX_THREADS will get the maximum threads being run on a given task (even if those threads are only being run in a different component)? While num_threads gets the threads being used for that component -- even if it's outside of an OMP loop?

billsacks:
@ekluzek note that the `omp_get_num_threads` call is inside a parallel region - it shouldn't matter if it's in a loop or not.

ekluzek:
I'm mainly thinking about decompMod.F90 which has some _OPENMP \#ifdefs as well. It's using OMP_MAX_THREADS to indicate the max number of threads on the task. And it's assuming that num_threads when called outside of a threaded loop will be 1. It sounds like that might not be right.

jedwards4b:
@ekluzek that's correct, this will work as long as omp_set_num_threads is called from the cap to set the number of threads for the component, I am doing this in the nuopc cap, I'll look at the mct cap now and make sure it's set.

src/main/controlMod.F90
```diff
@@ -294,7 +292,11 @@ subroutine control_init( )
     ! Set clumps per procoessor
 
 #if (defined _OPENMP)
-    clump_pproc = omp_get_max_threads()
+!$omp parallel
+!$omp critical
```
billsacks, line 296: One minor change: I'll probably change this to `omp master` rather than `omp critical`. It shouldn't matter in practice, but that makes it more clear to the reader that this only needs to be done by a single thread.

billsacks:
@jedwards4b just to understand the overall design here - partly thinking about how this will work with LILAC: Is the idea that the driver will somehow limit the number of available threads before calling CTSM's init method? (I'd like to add a bit of documentation about this somewhere.)

jedwards4b:
Yes, it should happen in the cap, I'm working on the mct cap now and will push back an update.

ekluzek:
OK, the MCT cap is NOT setting the number of threads. It's relying on the driver to do that for it. Thanks for adding that in also.

But, that also means that decompMod.F90 needs some changes for it to work right. We could possibly just remove the OPENMP logic from it. The thing it was trying to do was to tell if it was being called from outside a threaded region. Is there a way to do that?

billsacks:
> I'm mainly thinking about decompMod.F90 which has some _OPENMP #ifdefs as well. It's using OMP_MAX_THREADS to indicate the max number of threads on the task. And it's assuming that num_threads when called outside of a threaded loop will be 1. It sounds like that might not be right.

Ah, you're right @ekluzek - I do see a problem with this logic, although I can't tell if it's the same problem you're thinking of: The current logic will wrongly abort if the whole system is running with multiple threads but CTSM is only running with 1 thread. I think the way to fix that is to change this

```Fortran
     if ( OMP_GET_NUM_THREADS() == 1 .and. OMP_GET_MAX_THREADS() > 1 )then
```

to

```Fortran
     if ( OMP_GET_NUM_THREADS() == 1 .and. clump_pproc > 1 )then
```

billsacks:
> The thing it was trying to do was to tell if it was being called from outside a threaded region. Is there a way to do that?

@ekluzek - it seems to me like the logic in my above comment accomplishes this. Do you agree? `omp_get_num_threads` is 1 if you're outside a threaded region. But it's now possible that it will be 1 even inside a threaded region, if CTSM is only running with one thread. The check of `clump_pproc` should stop us from mistakenly aborting in that case, I think. What do you think?

billsacks:
Correction to my above logic: @ekluzek pointed out that `clump_pproc` can be overwritten in the namelist, so my proposal wasn't robust. What we settled on is that we'll add a variable, `ctsm_num_threads`. This will be set to `omp_get_num_threads()` where @jedwards4b is currently setting `clump_pproc`. Then we'll set `clump_pproc = ctsm_num_threads`. Then, in decompMod, we can check:

```Fortran
     if ( OMP_GET_NUM_THREADS() == 1 .and. ctsm_num_threads > 1 )then
```

jedwards4b:
I've updated the mct cap with this change.

billsacks:
Thanks @jedwards4b . Your change looks good to me. So is this ready to go as far as you're concerned? Did you want any changes to the esmf cap?

src/cpl/mct/lnd_comp_mct.F90
```diff
@@ -83,6 +84,7 @@ subroutine lnd_init_mct( EClock, cdata_l, x2l_l, l2x_l, NLFilename )
     integer  :: g,i,j                                ! indices
     integer  :: dtime_sync                           ! coupling time-step from the input synchronization clock
     integer  :: dtime_clm                            ! clm time-step
+    integer  :: nthreads
```
ekluzek, line 87: Could you add a comment that this is the number of OpenMP for the land model

src/cpl/mct/lnd_comp_mct.F90
```diff
@@ -115,6 +117,9 @@ subroutine lnd_init_mct( EClock, cdata_l, x2l_l, l2x_l, NLFilename )
     call seq_cdata_setptrs(cdata_l, ID=LNDID, mpicom=mpicom_lnd, &
          gsMap=GSMap_lnd, dom=dom_l, infodata=infodata)
 
+    call seq_comm_setptrs(LNDID, nthreads=nthreads)
+!$  call omp_set_num_threads(nthreads)
```
ekluzek, line 121: @jedwards4b you are using what I'd say is a a more obscure syntax for the OMP operations here. I think I'd prefer going with 

#ifdef _OPENMP

which I'd say is more explicit. I'd say change it throughout this file. @billsacks do you have an opinion on this? Without the more explicit #ifdef I think it needs a comment to explain what it is.

billsacks: I don't have strong feelings on it. I had never before seen `!$` used outside of an actual openMP directive, but from a bit of googling it looks like it's a standard way of doing conditional compilation with OpenMP, and I see how it makes sense given that the same form is used in OpenMP directives. It keeps the code cleaner at the cost of being more obscure.

Thinking about it more, though, I guess I'd agree with you @ekluzek that it's probably best to stick with an `#ifdef` here.

For this and your other comment, I'd propose that you or I just do the necessary small changes so @jedwards4b can move on to other things.

jedwards4b:
@billsacks I have the nuopc cap changes on a branch that we will bring in with the next nuopc update.   I would be happy to make the other changes you and @ekluzek have suggested, but it sounds like you already have a sandbox, so I'll assume you'll make them.

billsacks:
@jedwards4b and @ekluzek - I was reading some documentation on `omp_get_max_threads` and playing around with it. Based on this, I think that all we need in this PR is the change @jedwards4b made to the mct cap: After you call `omp_set_num_threads`, then `omp_get_max_threads` returns the number that you set in `omp_set_num_threads`. So we can continue to set the clumps per proc by examining `omp_get_max_threads` in controlMod. Furthermore, I don't think we need any changes in decompMod, because `omp_get_num_threads` will only be > 1 if the land is running with multiple threads.

I'll make these changes and test them.

ekluzek:
Ahh OK perfect.

billsacks:
Closing this in favor of #970 . Please add any comments there.

PR #970, opened 2020-04-04
For mct, run land with a subset of global threads if desired

### LINKED ISSUE #163: Have CLM listen to the driver for number of threads rather than OMP_NUM_THREADS

billsacks:
### Description of changes

Listen to the land thread number set in the mct driver rather than running on the full global number of threads.

This started from @jedwards4b 's PR #967 and supersedes that one. See comments there for more details.

However, I'm wondering if this is really the right way to do it: It feels to me like a better solution would be to set the number of threads in the driver rather than requiring all components to do this in their caps. The mct driver already has code to do this in `component_mod`:

```Fortran
             if (drv_threading) call seq_comm_setnthreads(comp(1)%nthreads_compid)
```

but maybe the `drv_threading` conditional needs to be removed for this to work right.

### Specific notes

- Resolves #163 

### DIFF
--- a/src/cpl/mct/lnd_comp_mct.F90
+++ b/src/cpl/mct/lnd_comp_mct.F90
@@ -28,6 +28,7 @@ module lnd_comp_mct
   private :: lnd_domain_mct    ! set the land model domain information
   private :: lnd_handle_resume ! handle pause/resume signals from the coupler
   !---------------------------------------------------------------------------
+  integer, private      :: LNDID	     ! Land identifyer
 
 contains
 
@@ -59,13 +60,14 @@ contains
     use seq_infodata_mod , only : seq_infodata_type, seq_infodata_GetData, seq_infodata_PutData, &
                                   seq_infodata_start_type_start, seq_infodata_start_type_cont,   &
                                   seq_infodata_start_type_brnch
-    use seq_comm_mct     , only : seq_comm_suffix, seq_comm_inst, seq_comm_name
+    use seq_comm_mct     , only : seq_comm_suffix, seq_comm_inst, seq_comm_name, seq_comm_setptrs
     use seq_flds_mod     , only : seq_flds_x2l_fields, seq_flds_l2x_fields
     use spmdMod          , only : masterproc, spmd_init
     use clm_varctl       , only : nsrStartup, nsrContinue, nsrBranch
     use clm_cpl_indices  , only : clm_cpl_indices_set
     use mct_mod          , only : mct_aVect_init, mct_aVect_zero, mct_gsMap_lsize
     use ESMF
+!$  use omp_lib         , only : omp_set_num_threads
     !
     ! !ARGUMENTS:
     type(ESMF_Clock),           intent(inout) :: EClock           ! Input synchronization clock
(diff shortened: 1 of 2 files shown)


billsacks:
@jedwards4b and @ekluzek - as I mentioned above: After implementing this, I began to question the need for doing this in the component caps. It feels like a cleaner and more robust solution would be to have the driver itself set the number of threads before and after each call to the appropriate `comp_run` routine. As I mentioned above, it looks like there is sort of already code to do this in `component_run`, but it may not be working right because it's in a `drv_threading` conditional.

@jedwards4b what do you think about that? Is there really a need for every component to implement this logic, or can this be done in the driver?

billsacks:
By the way, my general design motivation is: If there's something that needs to be done for every component, and if the driver knows enough to do it itself, then it should be done by the driver, rather than requiring every component to have duplicated logic in its cap.

src/cpl/mct/lnd_comp_mct.F90
```diff
@@ -276,6 +289,10 @@ subroutine lnd_init_mct( EClock, cdata_l, x2l_l, l2x_l, NLFilename )
     endif
 #endif
 
+#ifdef _OPENMP
```
jedwards4b, line 292: This one shouldn't really be necessary, but it won't hurt either.

jedwards4b:
I think that it can be done in the driver.

billsacks:
Thanks for the quick reply @jedwards4b . I guess I'll leave this open until it can be demonstrated to work just in the driver, without any changes necessary in CTSM or other component caps.

ekluzek:
> By the way, my general design motivation is: If there's something that needs to be done for every component, and if the driver knows enough to do it itself, then it should be done by the driver, rather than requiring every component to have duplicated logic in its cap.

That makes sense as a principle. So you are saying that we should modify CIME/CMEPS so that they set the component threads in the driver for each generic component call. That would mean that the caps for each component won't have to set the threads, and wouldn't have to change if they aren't already setting the number of threads. And like you say there wouldn't be the duplicated code for setting threads in each component.

The one caveat I thought of for that, would be for components that can't handle threading, or can only handle certain number of threads. But, even then the component without OpenMP just would never invoke any threaded loops. And a component that had restrictions on the number of threads could check for that by checking the number of threads at initialization. So it's still OK. And a component that didn't have OpenMP could still add a check for openmp threads and die if it's greater than one.

So it seems like handling this in cime/cmeps would be the right thing to do.

billsacks:
Based on input from @jedwards4b this is no longer needed.

PR #976, opened 2020-04-07
Bring documentation source to clm5.0 release branch

billsacks:
### Description of changes

Basically the same as #954 but now for the release branch.

The only changes I made from #954 are changing the version information, in 98b3812a.

Built documentation can be viewed here:
https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/

### Specific notes


doc/source/conf.py
```diff
@@ -0,0 +1,185 @@
...
+author = u''
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = u'CLM5.0'
+# The full version, including alpha/beta/rc tags.
+release = u'release-clm5.0.32'
```
billsacks, line 67: I wasn't sure what to do about `release`. What I currently have (release-clm5.0.32) follows what used to be in place in the ctsm-docs repo, but this doesn't seem ideal in terms of maintainability.

ekluzek: Seems like instead of a specific release tag, we should point to the release branch. When there is a CESM release we might change it to the specific tag for that release.

so now it would be...

release-clm5.0

When a CESM release happens we'd update it to the specific version in that release. Or you could have it point to the latest release version that was in a CESM release which would be release-clm5.0.30.

ekluzek:
I figure I'll bring this in without much testing. So I'll create a softlink for release-clm5.0.34 baselines to point to release-clm5.0.33.

billsacks:
@ekluzek thanks for your feedback. I have updated the release as you suggested. Yeah, I just brought this into master without any testing and made symlinks as you say. That said, I assigned this to you so you could bring it in whenever is most convenient.

If you want to convince yourself that this doesn't touch anything outside of documentation:

```
$ git diff --name-status release-clm5.0... | grep -v '\Wdoc/'
A	.gitattributes
A	.lfsconfig
```

ekluzek:
Hmmm. I didn't notice this when it came to master, but some of the substitutions aren't happening properly. So some things are coming out as |version| rather than substituted like they should be.

billsacks:
> Hmmm. I didn't notice this when it came to master, but some of the substitutions aren't happening properly. So some things are coming out as |version| rather than substituted like they should be.

I'm not sure which ones you're looking at, but at least some of these are pre-existing issues. I pushed an older version of the gh-pages branch (from the build you did, @ekluzek , Jan 2, 2020) to my fork; it can be viewed here: https://escomp.github.io/ctsm-docs/versions/master/html/users_guide/overview/introduction.html

ekluzek:
Ahh, yes, that's the same things.

Issue #983, opened 2020-04-19
Fire method renames and add additional layers


ekluzek:
This is to take the non-answer changing part of #982 to make a simple bit-for-bit tag to make a better class structure to extend for a FATESfireData class...

The current hierarchy is...

C?NFireMethodMod
CNFireBaseMod
(and then the three classes that extend that of CNFireLi2014Mod, CNFireLi2016Mod, and CNFireNoFireMod)

Rename this to...

FireMethod
FireBaseData

(FATESFireData and CNFireNoFire would extend from here)

CNLiFireBase -- would be added to extend for the two LiFire class extensions
(CNFireLi2014Mod, CNFireLi2016Mod would then extend from this point). The instance of the fire method for CN would remain within CNVegetationFacade, but a new instance for FATES would be created in clmfates_interface. So in the end only one of those would be created either the FATES series or the CN/BGC series.

FireMethod and FireBaseData would be moved to the :"main" directory from biogeochem subdirectory.

Also CNFireFactoryMod.F90 would be renamed FireFactory and moved to the main directory as well. There would be too different methods in FireFactory that either create the CN/BGC versions or another version that creates the FATES fire data versions.

ekluzek:
We went over this today with @slevisconsulting @negin513 and @glemieux. There's agreement that this change makes sense. It will make the names more clear and distinguish the parts used for FATES versus the parts used for CN/BGC.

We also think the right ordering for this is for @slevisconsulting to bring in his changes as they are to fates_next_api branch. I will bring these changes to master. At some point we will merge those changes from fates_next_api into master and resolve the conflicts between the two. Then master will be able to be moved to fates_next_api and both branches will have the same renames.

ekluzek:
The changes in ctsm1.0.dev095 have an impact here. In there CNFireBaseMod becomes an abstract type, and a new interface was added need_lightning_and_popdens, to replace the need_lightning_and_popdens object data.

Allocation of the object data is also done this way

`allocate(cnfire_li2014_type :: cnfire_method)`

instead of...

`allocate(cnfire_method, source=cnfire_li2014_type())`

So here there will be a method for each of lightning and popdens.

ekluzek:
[...]

Issue #982, opened 2020-04-19
Extend CNFireMethodMod with a FATESFireData class


ekluzek:
In order to give FATES access to the Fire data (such as lightning and population density), we'd like to extend the CNFireMethodMod abstract class with a FATESFireData class. This also extends the CNFireBaseMod class so that it will get access to the hdm and pop-density data.

This will add an additional option to the fire_method namelist item I propose to be called "fateslnfm". It could only be set if use_fates is set. We should also add the current default option of "fatesnofiredata" so that FATES wouldn't use any additional data.

FATESFireData class would extend CNFireBaseMod class. It wouldn't use the CNFireArea, CNFireFluxes nor CNFireReadNML, methods, so should probably set them to a stub that dies if called. It would also save the data for when read in, and handle the accumulator methods to average it over the daily interval.

Initially the hierarchy of

CNFireMethodMod
CNFireBaseMod
(and then the three classes that extend that of CNFireLi2014Mod, CNFireLi2016Mod, and CNFireNoFireMod)

would just be extended with the FATESFireData class.

Some renaming would be helpful though, I've put that in a separate issue #983.

For CN-Fire the fire instance would be handled just as it is now. Which means it's attached to the CNVegetationFacade, which is just as it should be. When FATES is active though FireMethod would be a type added to the FATES interface as new data. The FireFactory would be called for FATES. The FireFactory would just need to make sure that the No-Fire and Li-fire options are only used when use_cn is on, and fatesnofiredata and fateslnfm are only allowed when fates is on.

@rgknox @slevisconsulting

ekluzek:
I think the way to handle it is for cnfire_method_type to be instantiated inside of clmfates_interface in the clm_fates instance of hlm_fates_interface_type.

Here's some rough code that sketches out how to do that below. I show passing in a control variable called fates_fire_data that would signal the level of fire data being sent down to FATES, but I don't show how it comes into play. That would also need to be fleshed out. I do think that making a CNFireMethod creation type just for FATES makes sense though. That would be part of the CNFireFactoryMod and would be specific to the FATES fire data levels.

``` fortran
diff --git a/src/utils/clmfates_interfaceMod.F90 b/src/utils/clmfates_interfaceMod.F90
index df05801e..2a7cfd70 100644
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -130,6 +130,7 @@ module CLMFatesInterfaceMod
[...]

Issue #667, opened 2019-03-24
Exact restart problem with Fates


ekluzek:
### Brief summary of bug

Activation of planthydro in Fates on fates_next_api seems to have an exact restart issue.

### General bug information

**CTSM version you are using:** clm5.0.dev008-90-g04e931e

**Does this bug cause significantly incorrect results in the model's science?** No?

**Configurations affected:**

ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro

The test passes when use_fates_planthydro=.false.

### Important output or errors that show the problem

```
cases/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.GC.clm5d8-fna-n1chintelf> tail  /glade/scratch/erik/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.GC.clm5d8-fna-n1chintelf/run/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.GC.clm5d8-fna-n1chintelf.cpl.hi.0001-01-06-00000.nc.base.cprnc.out
  
SUMMARY of cprnc:
 A total number of    214 fields were compared
          of which     13 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
 A total number of      0 fields could not be analyzed
 A total number of      0 fields on file 1 were not found on file2.
  diff_test: the two files seem to be DIFFERENT 
  
cases/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.GC.clm5d8-fna-n1chintelf> grep RMS /glade/scratch/erik/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.GC.clm5d8-fna-n1chintelf/run/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro.GC.clm5d8-fna-n1chintelf.cpl.hi.0001-01-06-00000.nc.base.cprnc.out
 RMS l2x_Sl_tref                      1.3476E-07            NORMALIZED  4.5143E-10
[...]

PR #991, opened 2020-04-24
Extend CNFireMethodMod with a FATESFireData class

slevisconsulting:
### Description of changes
Allow FATES access to CTSM's infrastructure that reads lightning and
human population data sets for use in SPITFIRE.

Following @ekluzek's detailed recommendations in #982 #983 .

### Specific notes

@ekluzek @rgknox @ckoven @lmkueppers @jkshuman

#982 #983 
Also relates directly to issue https://github.com/NGEET/fates/issues/562
and to pull request https://github.com/NGEET/fates/pull/635

Not in non-FATES simulations.
Not in FATES simulations using `use_fates_spitfire < 2`
Yes in FATES simulations using `use_fates_spitfire >= 2`

Changing `use_fates_spitfire` from logical to integer with valid values of:
0 for no fire
1 for spitfire with default lightning constant
2 for spitfire with Li et al lightning data
3 for spitfire with successful ignitions data
as recommended by @rgknox [here](https://github.com/NGEET/fates/issues/562#issuecomment-615394863)

Testing performed, if any:
Code builds. No further testing, yet.


src/utils/clmfates_interfaceMod.F90
```diff
@@ -583,8 +596,12 @@ subroutine dynamics_driv(this, nc, bounds_clump,      &
       type(soilbiogeochem_carbonflux_type), intent(inout) :: soilbiogeochem_carbonflux_inst
       type(frictionvel_type)  , intent(inout)        :: frictionvel_inst
 
+      type(fates_fire_data_type) :: fates_fire_data_inst
+      type(fates_fire_no_data_type) :: fates_fire_no_data_inst
```
slevisconsulting, line 600: Here (and possibly elsewhere) I entered `fates_fire_data_inst` and `fates_fire_no_data_inst` via `use` statements (see a few lines up) rather than as arguments through the call. Will this cause trouble?

src/main/clm_initializeMod.F90
```diff
@@ -293,8 +293,12 @@ subroutine initialize2( )
     use controlMod            , only : NLFilename
     use clm_instMod           , only : clm_fates
     use BalanceCheckMod       , only : BalanceCheckInit
+    use FATESFireNoDataMod, only: fates_fire_no_data_type
+    use FATESFireDataMod, only: fates_fire_data_type
     !
     ! !ARGUMENTS    
+    type(fates_fire_data_type) :: fates_fire_data_inst
```
ekluzek, line 300: Just use the generic type rather than the specific one

src/main/clm_initializeMod.F90
```diff
@@ -471,6 +475,10 @@ subroutine initialize2( )
        if ( use_c13 .and. use_c13_timeseries ) then
           call C13_init_TimeSeries()
        end if
+    ! use_fates_spitfire is assigned an integer value in the namelist
+    ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
+    else if (use_fates_spitfire > 1) then
+       call bgc_vegetation_inst%Init2(bounds_proc, NLFilename)
```
ekluzek, line 481: Don't call into bgc_veg you want to use the instance that's tied to FATES

src/main/clm_initializeMod.F90
```diff
@@ -606,6 +614,14 @@ subroutine initialize2( )
        call crop_inst%initAccVars(bounds_proc)
     end if
 
+    ! use_fates_spitfire is assigned an integer value in the namelist
+    ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
+    if (use_fates_spitfire > 1) then
```
ekluzek, line 619: Just call the generic one rather than the specific one

src/main/clm_instMod.F90
```diff
@@ -187,9 +188,13 @@ subroutine clm_instInit(bounds)
     use SoilWaterRetentionCurveFactoryMod  , only : create_soil_water_retention_curve
     use decompMod                          , only : get_proc_bounds
     use BalanceCheckMod                    , only : GetBalanceCheckSkipSteps
+    use FATESFireNoDataMod, only: fates_fire_no_data_type
+    use FATESFireDataMod, only: fates_fire_data_type
     !
     ! !ARGUMENTS    
     type(bounds_type), intent(in) :: bounds  ! processor bounds
+    type(fates_fire_data_type) :: fates_fire_data_inst
```
ekluzek, line 196: Same here, you just need to know the generic type rather than the specific one

src/main/clm_instMod.F90
```diff
@@ -433,6 +438,14 @@ subroutine clm_instInit(bounds)
        call crop_inst%InitAccBuffer(bounds)
     end if
 
+    ! use_fates_spitfire is assigned an integer value in the namelist
+    ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
+    if (use_fates_spitfire > 1) then
```
ekluzek, line 443: And same here, you give it the generic type

src/main/clm_initializeMod.F90
ekluzek: You point to fates_fire_data_method in clmfates_interface

src/main/clm_driver.F90
```diff
@@ -395,6 +399,10 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
        end if
        call bgc_vegetation_inst%InterpFileInputs(bounds_proc)
        call t_stopf('bgc_interp')
+    ! use_fates_spitfire is assigned an integer value in the namelist
+    ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
+    else if (use_fates_spitfire > 1) then
+       call bgc_vegetation_inst%InterpFileInputs(bounds_proc)
```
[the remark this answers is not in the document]
ekluzek: You point to fates_fire_data_method in clmfates_interface

src/utils/clmfates_interfaceMod.F90
ekluzek: You actually don't want these at all. You will reference them through the generic type of fates_fire_data_method, so you don't have to know which specific type it is. Nor do you have to declare each one.

bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -1134,14 +1134,19 @@ lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_16pfts_Irrig_CMIP6_simyr18
 <light_res    use_cn=".true." fire_method="nofire" >none</light_res>
 <light_res    use_cn=".true." phys="clm4_5"        >94x192</light_res>
 <light_res    use_cn=".true." phys="clm5_0"        >94x192</light_res>
+<light_res    use_fates=".true." phys="clm5_0"     >360x720</light_res>
```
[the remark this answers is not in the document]
slevisconsulting: Ok, added.

bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -1134,14 +1134,19 @@ lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_16pfts_Irrig_CMIP6_simyr18
...
 <stream_year_last_lightng  use_cn=".true."   >0001</stream_year_last_lightng>
+<stream_year_first_lightng use_fates=".true.">0001</stream_year_first_lightng>
+<stream_year_last_lightng  use_fates=".true.">0001</stream_year_last_lightng>
 
 <stream_fldfilename_lightng hgrid="94x192"    use_cn=".true."   >atm/datm7/NASA_LIS/clmforc.Li_2012_climo1995-2011.T62.lnfm_Total_c140423.nc</stream_fldfilename_lightng>
 <stream_fldfilename_lightng hgrid="360x720"   use_cn=".true."   >atm/datm7/NASA_LIS/clmforc.Li_2016_climo1995-2013.360x720.lnfm_Total_c160825.nc</stream_fldfilename_lightng>
+<stream_fldfilename_lightng use_fates=".true."   >atm/datm7/NASA_LIS/clmforc.Li_2016_climo1995-2013.360x720.lnfm_Total_c160825.nc</stream_fldfilename_lightng>
 
 <lightngmapalgo use_cn=".true."                               >bilinear</lightngmapalgo>
+<lightngmapalgo use_fates=".true."                            >nn</lightngmapalgo>
```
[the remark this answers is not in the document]
slevisconsulting: I used nearest neighbor for the lightning regridding because @lmkueppers' group runs mostly local and regional simulations at resolutions that are (much) higher than the input data. I will let @lmkueppers and others offer feedback on this.

bld/namelist_files/namelist_definition_clm4_5.xml
```diff
@@ -618,9 +618,9 @@ Toggle to turn on the FATES model
 Switch deciding which nutrient model to use in FATES.
 </entry>
 
-<entry id="use_fates_spitfire" type="logical" category="physics"
-        group="clm_inparm" valid_values="" value=".false.">
-Toggle to turn on spitfire module for modeling fire (only relevant if FATES is being used).
+<entry id="use_fates_spitfire" type="integer" category="physics"
+        group="clm_inparm" valid_values="0,1,2,3" value=".false.">
+Turn on spitfire module for modeling fire (only relevant if FATES is being used) by setting > 0. Values of 1, 2, and 3 correspond to no external inputs, external lightning data, and external confirmed ignition data, respectively.
```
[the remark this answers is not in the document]
slevisconsulting: Done.

src/biogeochem/FATESFireNoDataMod.F90
```diff
@@ -0,0 +1,177 @@
...
+
+contains
+
+  !------------------------------------------------------------------------
+  type(fates_fire_no_data_type) function constructor()
+    !
+    ! !DESCRIPTION:
+    ! Creates an object of type fates_fire_no_data_type
+    ! !ARGUMENTS:
+    constructor%need_lightning_and_popdens = .true.
```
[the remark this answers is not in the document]
slevisconsulting: Good point.

src/biogeochem/FATESFireNoDataMod.F90
```diff
@@ -0,0 +1,177 @@
...
+  !
+  implicit none
+  private
+  !
+  ! !PUBLIC TYPES:
+  public :: fates_fire_no_data_type
+  !
+  type, extends(cnfire_base_type) :: fates_fire_no_data_type
+      ! !PRIVATE MEMBER DATA:
+      real(r8), public, pointer :: lnfm24(:)  ! Daily avg lightning by grid cell (#/km2/hr)
```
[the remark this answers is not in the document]
slevisconsulting: Let me explain my thought process. The NoData method has two options:
1) use_fates_spitfire = 0  i.e. no fire
2) use_fates_spitfire = 1  i.e. fire with no external inputs for lightning etc.

In case (2) the model uses `EDParamsMod, only: ED_val_nignitions`. Although this is a single value of global lightning, I decided to run it through the same steps as the 3D lightning and end up with lnfm24, to allow users the flexibility of calculating time-varying lightning in the code.

If everyone is opposed to this logic, I could skip the lnfm24 calculations in the NoData method and place `ED_val_nignitions` directly in the calculation of lightning24 in `clmfates_interfaceMod.F90`.

src/biogeochem/FATESFireNoDataMod.F90
```diff
@@ -0,0 +1,177 @@
...
+  type(fates_fire_no_data_type) function constructor()
+    !
+    ! !DESCRIPTION:
+    ! Creates an object of type fates_fire_no_data_type
+    ! !ARGUMENTS:
+    constructor%need_lightning_and_popdens = .true.
+  end function constructor
+
+  !-----------------------------------------------------------------------
+  subroutine InitAccBuffer (this, bounds)
```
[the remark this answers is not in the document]
slevisconsulting: See my response [here](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415340963).

If the group agrees with you, then yes, and do we make them "empty" or remove them from the module entirely?

src/biogeochem/FATESFireNoDataMod.F90
```diff
@@ -0,0 +1,177 @@
...
+    endif
+    this%lnfm24(:) = spval
+    call init_accum_field (name='lnfm24', units='strikes/km2/hr', &
+         desc='24hr average of lightning strikes',  accum_type='runmean', &
+         accum_period=-1, subgrid_type='gridcell', numlev=1, init_value=0._r8)
+
+  end subroutine InitAccBuffer
+
+  !-----------------------------------------------------------------------
+  subroutine InitAccVars(this, bounds)
```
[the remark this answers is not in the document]
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415346630).

src/biogeochem/FATESFireNoDataMod.F90
```diff
@@ -0,0 +1,177 @@
...
+
+    call extract_accum_field ('lnfm24', rbufslg, nstep)
+    this%lnfm24(begg:endg) = rbufslg(begg:endg)
+
+    deallocate(rbufslg)
+
+  end subroutine InitAccVars
+
+  !-----------------------------------------------------------------------
+  subroutine UpdateAccVars (this, bounds)
```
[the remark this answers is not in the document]
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415346630).

src/main/clm_driver.F90
```diff
@@ -109,11 +109,15 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
...
-    use clm_varctl           , only : use_lai_streams
+    use clm_varctl, only: use_lai_streams, use_fates_spitfire
     use SatellitePhenologyMod, only : lai_advance
+    use FATESFireNoDataMod, only: fates_fire_no_data_type
+    use FATESFireDataMod, only: fates_fire_data_type
     !
     ! !ARGUMENTS:
     implicit none
+    type(fates_fire_data_type) :: fates_fire_data_inst
+    type(fates_fire_no_data_type) :: fates_fire_no_data_inst
```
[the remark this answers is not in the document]
slevisconsulting: Done. I'm replacing with
```suggestion
+    use CNFireMethodMod, only: cnfire_method_type
[...]
+    type(cnfire_method_type) :: cnfire_method_inst
```
I hope I understood correctly.

src/main/clm_driver.F90
```diff
@@ -1133,6 +1141,14 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
                temperature_inst%t_ref2m_patch, temperature_inst%t_soisno_col)
        end if
 
+       ! use_fates_spitfire is assigned an integer value in the namelist
+       ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
+       if (use_fates_spitfire > 1) then
+          call fates_fire_data_inst%UpdateAccVars(bounds_proc)
```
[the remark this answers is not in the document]
slevisconsulting: Done. I'm replacing with
```suggestion
+       if (use_fates_spitfire > 0) then
+          call cnfire_method_inst%UpdateAccVars(bounds_proc)
```

src/main/clm_initializeMod.F90
```diff
@@ -293,8 +293,12 @@ subroutine initialize2( )
     use controlMod            , only : NLFilename
     use clm_instMod           , only : clm_fates
     use BalanceCheckMod       , only : BalanceCheckInit
+    use FATESFireNoDataMod, only: fates_fire_no_data_type
```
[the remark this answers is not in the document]
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415423650).

src/main/clm_initializeMod.F90
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415423650).

src/main/clm_initializeMod.F90
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415424002).

src/main/clm_instMod.F90
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415424002).

src/main/clm_instMod.F90
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415423650)

src/utils/clmfates_interfaceMod.F90
slevisconsulting: Same [comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r415423650)

src/main/clm_driver.F90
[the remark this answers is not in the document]
slevisconsulting: In this case I don't understand. Unless I also need to make a copy of
`subroutine InterpFileInputs`
in clmfates_interface, which seems redundant to me.

src/main/clm_initializeMod.F90
slevisconsulting: In this case I don't understand. Unless I also need to make a copy of
`subroutine Init2`
in clmfates_interface, which seems redundant to me.

slevisconsulting:
@ekluzek pls see my responses and revisions. If I am on the wrong track, let's definitely have a meeting.

src/biogeochem/FATESFireNoDataMod.F90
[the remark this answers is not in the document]
ekluzek: OK I see. That does make the calculation more expensive because it's over an array rather than a single value. I see the point about making the code in FATES identical though. But, I don't think keeping it as an array here really helps to compare the two options. What would be useful is to make sure the constant in FATES is in same units as the lightning data here. Then you can just visually compare the lightning data with that single value. And the lightning data map can help guide you as what the single constant should be. If lnfm24 is being written out to the history file I could see how it could be useful to compare the two methods that way. But, still one output is a constant the the other varying in space and time, so you really just need to know the value of the constant to do that comparison.

So I would vote to remove the 3D data here, and just use a single constant. Just make sure the single constant is in the same units as lnfm24 so that you can easily compare them.

src/biogeochem/FATESFireNoDataMod.F90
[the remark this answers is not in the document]
ekluzek: If you do decide to keep lnfm24, a better way to do this would be to have these methods in the NoData class and then extend it with the DataClass. That way they can both share the same accumulation averaging methods without having to duplicate them.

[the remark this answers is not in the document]
ekluzek: If you do decide to keep them, they need to be made empty, because they will be called, they just want do anything. The calling interface makes them required, but in the No-Data case they don't do anything.

src/main/clm_driver.F90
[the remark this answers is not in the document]
ekluzek: The only thing you actually need from InterpFileInputs is the FATESFireData Interpolation. So there really isn't any duplication of anything in InterpFileInputs. It also HAS to be different than the Vegetation-Facade version -- because the instance of the object is in a different place for FATES than for BGC. The BGC instance isn't actually ever instantiated, so you can't actually use it -- you HAVE to use the version in clmfates_interface.

The reason to create a InterpFileInputs in clmfates_interface is so that the contents of clmfates_interface aren't exposed outside of it. It is a good design goal to keep clmfates_interface content local to itself, so I'd support that method. Technically you could call the FATESFireDdata interpolation call directly here if you exposed the clmfates_interface FATESFireMethod data with a use statement -- but that breaks keeping FATES interface data module to clmfates_interface. So I do recommend having the interface in clmfates_interface.

src/main/clm_initializeMod.F90
ekluzek: The reasoning is actually similar to above, but even stronger here, because there are things in the BGC VegetationFacade that you do NOT want for the FATES case. The only think you want is the FATES Fire data Init, you don't want DNPhenologyInit nor SoilBiogeochemCompetitionInit (because it must already be called separately for FATES).

src/biogeochem/FATESFireNoDataMod.F90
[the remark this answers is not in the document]
lmkueppers: It would be useful to have @jkshuman comment on this. 
I'm not sure I understand the alternatives. 
It seems that someone could create a global constant value in the format of the 3D lightening if they wanted to run it through that way.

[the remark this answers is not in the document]
ekluzek: That's a good point @lmkueppers. That's a totally valid and easy thing to do, and you could do that to show that you get the same results with either method without having to keep the 3D data around. So yeah, I vote to remove it from here.

bld/namelist_files/namelist_defaults_clm4_5.xml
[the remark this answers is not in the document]
lmkueppers: I don't think this matters so much for point scale simulations. 
At regional scale, there may be alternate resolution datasets to the current lightening one, so regridding seems preferable. 
@jkshuman is running at continental scales with fire, and we presumably will run FATES globally with fire on at some point.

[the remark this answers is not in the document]
jkshuman: I will defer to @ekluzek on this point.

src/main/clm_instMod.F90
```diff
@@ -440,10 +438,8 @@ subroutine clm_instInit(bounds)
 
     ! use_fates_spitfire is assigned an integer value in the namelist
     ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
-    if (use_fates_spitfire > 1) then
-       call fates_fire_data_inst%InitAccBuffer(bounds)
-    else
-       call fates_fire_no_data_inst%InitAccBuffer(bounds)
+    if (use_fates_spitfire > 0) then
```
ekluzek, line 441: I actually suggest that you keep the accumulation interfaces in both the FATESFireNoData and FATESfireData classes, so that you don't have to have this if statement. But, if you keep the if statement you don't have to have empty methods for the FATESFireNoData class. So it can be done either way...

src/main/clm_initializeMod.F90
```diff
@@ -616,10 +614,8 @@ subroutine initialize2( )
 
     ! use_fates_spitfire is assigned an integer value in the namelist
     ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
-    if (use_fates_spitfire > 1) then
-       call fates_fire_data_inst%initAccVars(bounds_proc)
-    else
-       call fates_fire_no_data_inst%initAccVars(bounds_proc)
+    if (use_fates_spitfire > 0) then
```
ekluzek, line 617: See note below on the if statement.

src/biogeochem/FATESFireNoDataMod.F90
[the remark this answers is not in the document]
ekluzek: I added some discussion of this point to the FATES side. I think that's the crux of how this should be handled. Once that's determined it will likely dictate the solution here. So let's see what FATES developers want to see on that side.

bld/namelist_files/namelist_defaults_clm4_5.xml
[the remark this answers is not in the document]
ekluzek: OK change it to bilinear for FATES. Scientists can change it to nn on their own if it's appropriate for what they are doing. Likely that would just be if they have actual lightning data from some nearby source.

src/main/clm_driver.F90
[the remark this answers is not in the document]
ekluzek: Yes!

src/biogeochem/FATESFireNoDataMod.F90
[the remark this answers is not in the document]
jkshuman: I agree with removing the 3D. @ekluzek also makes a good point about printing out the lightning data.

bld/namelist_files/namelist_defaults_clm4_5.xml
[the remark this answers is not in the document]
slevisconsulting: Ok, done.

src/main/clm_instMod.F90
slevisconsulting: Ok, I'm keeping the acc interfaces in both (empty in the NoData case) and removing the if statement.

src/main/clm_initializeMod.F90
slevisconsulting: Ok, I'm keeping the acc interfaces in both the FATESFireNoData and FATESfireData classes (empty in FATESFireNoData) and removing the if statement.

src/biogeochem/FATESFireNoDataMod.F90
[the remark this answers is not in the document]
slevisconsulting: @ekluzek wrote:
> You don't need the lightning data for the noData method.

Based on the conversation this far, I have gone ahead and removed `lnfm24` from the NoData case and made the corresponding accumulation procedures empty.

Our next choice is this. I vote for option (1) and explain my reasoning below:
1) An if statement in `clmfates_interfaceMod.F90`
```suggestion
if (use_fates_spitfire > 1) then
   this%fates(nc)%bc_in(s)%lightning24 = cnfire_method_inst%lnfm24(g) * 24._r8  ! #/km2/hr to #/km2/day
else
   this%fates(nc)%bc_in(s)%lightning24 = ED_val_nignitions / days_per_year  ! #/km2/yr to #/km2/day
end if```
with no if statement in `SFMainMod.F90` (SPITFIRE)
```currentSite%NF = bc_in%lightning24 * cg_strikes```

OR
2) An if statement in each, as follows:
```suggestion
if (use_fates_spitfire > 1) then
   this%fates(nc)%bc_in(s)%lightning24 = cnfire_method_inst%lnfm24(g)
else
   this%fates(nc)%bc_in(s)%lightning24 = 0._r8
end if```

```suggestion
if (use_fates_spitfire > 1) then
   currentSite%NF = bc_in%lightning24 * 24._r8 * cg_strikes  ! #/km2/hr to #/km2/day
else
   currentSite%NF = ED_val_nignitions * years_per_day * cg_strikes  ! #/km2/yr to #/km2/day
end if```

I vote for (1) because:
- SPITFIRE's code becomes simpler with no if statement and no units conversions
- `clmfates_interfaceMod.F90` converts units and passes ignitions to FATES under all circumstances, i.e. we have uniform treatment of all ignition options

My next commit will include option (1). Of course I am happy to change to option (2) if people prefer that.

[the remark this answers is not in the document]
Comment (slevisconsulting) [reply src/biogeochem/FATESFireNoDataMod.F90]: > I agree with removing the 3D. @ekluzek also makes a good point about printing out the lightning data.

3D removed.

Regarding sending the variable lnfm to history:
The capability already exists in `subroutine lnfm_init` both in the implementation that we're working on now (in CNFireBaseMod.F90), as well as in my previous implementation (in SFMainMod.F90). All you need to do is change the variable to `active` because it is set to `default=inactive`.

File: src/biogeochem/FATESFireNoDataMod.F90 (hunk 13 above)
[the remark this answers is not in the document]
Comment (slevisconsulting) [reply src/biogeochem/FATESFireNoDataMod.F90:59]: Made empty.

File: src/main/clm_initializeMod.F90 (hunk 3 above)
Comment (slevisconsulting) [reply src/main/clm_initializeMod.F90:479]: @ekluzek pls let me know if I misunderstood your explanation in the next commit that I'm about to push.

File: src/main/clm_driver.F90 (hunk 7 above)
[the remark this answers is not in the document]
Comment (slevisconsulting) [reply src/main/clm_driver.F90:403]: @ekluzek pls let me know if I misunderstood your explanation in the next commit that I'm about to push.

File: src/main/clm_initializeMod.F90 (hunk 3 above)
Comment (slevisconsulting) [reply src/main/clm_initializeMod.F90]: I added the subroutines `Init2` and `InterpFileInputs` to both `FATESFireDataMod` and `FATESFireNoDataMod`. I made the latter empty for consistency with the other subroutines, although the NoData versions do not get called because of the if statement `else if (use_fates_spitfire > 1)`. I see two options:
1) Delete the NoData copies of `Init2` and `InterpFileInputs`
2) Change the if statement to `else if (use_fates)`

**Comment by ekluzek** [comment]
@slevisconsulting so Init2 and InterpFileInputs should go into clmfates_interface, rather than to the extensions of the Fire classes for FATES.

I think it will help to have the general conversation about the Fire class methods and the refactoring I'm proposing to make sure you are clear on some of the OO concepts being used here. So I think we should plan on that before you proceed forward. We can make sure we talk about what I specifically plan on doing, as well as what you need to do here.

I think most of the conversations can be resolved now. So I'll go through and make sure I do that. There are only a couple that will be outstanding. Once we go through there will only be a couple last changes to make.

File: src/utils/clmfates_interfaceMod.F90 [hunk 21]
```diff
@@ -638,6 +653,13 @@ subroutine dynamics_driv(this, nc, bounds_clump,      &
 
          c = this%f2hmap(nc)%fcolumn(s)
 
+         if (use_fates_spitfire > 1) then
+            g = col%gridcell(c)
+            this%fates(nc)%bc_in(s)%lightning24 = this%fates_fire_data_method%lnfm24(g) * 24._r8  ! #/km2/hr to #/km2/day
```
slevisconsulting, line 658: Both compilation errors refer to the following line. I'm posting here in case the solution is obvious to you @ekluzek or anyone:

`/fs/cgd/data0/slevis/git_FATES/clm5_190325/cime/../src/utils/clmfates_interfaceMod.F90(658): error #6460: This is not a field name that is defined in the encompassing structure.   [LIGHTNING24]`
`            this%fates(nc)%bc_in(s)%lightning24 = this%fates_fire_data_method%lnfm24(g) * 24._r8  ! #/km2/hr to #/km2/day
`------------------------------------^`
`/fs/cgd/data0/slevis/git_FATES/clm5_190325/cime/../src/utils/clmfates_interfaceMod.F90(658): error #6460: This is not a field name that is defined in the encompassing structure.   [LNFM24]`
`            this%fates(nc)%bc_in(s)%lightning24 = this%fates_fire_data_method%lnfm24(g) * 24._r8  ! #/km2/hr to #/km2/day`
`------------------------------------------------------------------------------^`
`compilation aborted for /fs/cgd/data0/slevis/git_FATES/clm5_190325/cime/../src/utils/clmfates_interfaceMod.F90 (code 1)
gmake: *** [clmfates_interfaceMod.o] Error 1`

slevisconsulting: Update: I corrected the first of the two errors (FATES HEAD was detached, so I reconnected to my branch).

I will pursue the second one tomorrow.

src/main/clm_driver.F90
```diff
@@ -1137,7 +1141,7 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
                temperature_inst%t_ref2m_patch, temperature_inst%t_soisno_col)
        end if
 
-       call cnfire_method_inst%UpdateAccVars(bounds_proc)
+       call fates_fire_data_inst%UpdateAccVars(bounds_proc)
```
slevisconsulting, line 1144: @ekluzek 
I'm afraid I don't see how this line will work without an `if` statement like this:

`if (use_fates_spitfire > 1) then`
`call fates_fire_data_inst%UpdateAccVars(bounds_proc)`
`else`
`call fates_fire_no_data_inst%UpdateAccVars(bounds_proc)`

The same comment goes for the corresponding lines in `clm_initializeMod` and `clm_instMod`.

slevisconsulting: @ekluzek 
Am I still missing new code that links these two options to the `fates_fire_data_method`? According to the template that I see for routines like CNFireArea, for example? Does all this infrastructure serve to avoid `if` statements and improve performance?

ekluzek: I'm not seeing what you are missing offhand. And we should be able to get this to work how we want, because we use it elsewhere. Let me checkout your code and see if I can see what's missing.

What test case are you trying? And do you have everything needed checked in at this point?

ekluzek:
I am seeing changes that are needed now that I'm looking directly at the code. There's a couple ways to proceed.  I can figure out the changes and post the diff needed, or I could try to explain inline in github, or. I could submit a PR to your branch, or lastly if you authorize me -- I can push directly to your branch. Which of those sound like the best solution?

src/main/clm_driver.F90
ekluzek: The issue here is that there needs to be a subroutine in clmfates_interface that does the UpdateAccVars for fire (since the fire data instance is inside of it). This is something I neglected to notice the other day, sorry about that.

src/main/clm_initializeMod.F90
```diff
@@ -606,6 +614,8 @@ subroutine initialize2( )
        call crop_inst%initAccVars(bounds_proc)
     end if
 
+    call fates_fire_data_inst%initAccVars(bounds_proc)
```
ekluzek, line 617: Same thing here as above...

slevisconsulting:
> I am seeing changes that are needed now that I'm looking directly at the code. There's a couple ways to proceed. I can figure out the changes and post the diff needed, or I could try to explain inline in github, or. I could submit a PR to your branch, or lastly if you authorize me -- I can push directly to your branch. Which of those sound like the best solution?

I think simplest for me would be to authorize you to push directly to my branch. How do I do that?

src/main/clm_instMod.F90
```diff
@@ -433,6 +437,8 @@ subroutine clm_instInit(bounds)
        call crop_inst%InitAccBuffer(bounds)
     end if
 
+    call fates_fire_data_inst%InitAccBuffer(bounds)
```
ekluzek, line 440: Same thing here. Although here this call can probably be inside of another clmfates_interface subroutine.

ekluzek:
Go to your fork to CTSM on github. Go to settings tab on the right top, and then "manage access" option on the left down one. Then you can add me as a collaborator. At some point it asks you to enter your password.

It'll send me an email and then I just need to accept it. I've done this with several people and it makes it a little easier to work on the same code together. Sometimes if there's something more extensive I'll still do a Pull Request, to their branch, so they can code review it beforehand. But, here we are talking about small changes. So I'll get it working for a simple case and then push my commits to your branch. I'll try to explain as well as I can in the commit notes.

Thanks.

slevisconsulting:
> It'll send me an email and then I just need to accept it.

Ok, you should get a email.

Thanks @ekluzek

ekluzek:
I realized another thing missing here is the interface layer that defines the FATES. interface to the. object. That layer is important so that clmfates_interface can use them while referencing the abstract type rather than the specific instance. I talk about this in #983, but it didn't occur to me that it's also required here.

slevisconsulting:
> I realized another thing missing here is the interface layer that defines the FATES. interface to the. object. That layer is important so that clmfates_interface can use them while referencing the abstract type rather than the specific instance. I talk about this in #983, but it didn't occur to me that it's also required here.

Should we discuss this so that I can give it a go, or will you include it in your commits to my branch? I vote for the latter :-)

ekluzek:
@slevisconsulting I've nearly got this working. I ran into a problem where I had assumed you could extend a concrete class with an abstract one -- but it looks like. that isn't possible in FORTRAN. So I'm looking into the details of that. I'm also going to work on the refactoring I proposed a bit, so I can propose the closest solution here. So I'll get back to you after I do more work on this. Let me know if it's critical to get this in ASAP. Otherwise, I'll figure on doing it this week in a few days.

slevisconsulting:
> [...] I'll figure on doing it this week in a few days.

That sounds perfect. Thanks again, @ekluzek .

ekluzek:
OK @slevisconsulting why don't you give this a whirl now. I ran two simple test cases with it. SMS_Ld5.f19_g17.I2000Clm50Fates.cheyenne_intel.clm-FatesColdDef and SMS_D_Ld2.1x1_brazil.IHistClm50BgcQianGs.cheyenne_gnu.clm-ciso_bombspike1963. You'll need to resolve conflicts because of recent updates to fates_next_api. As well as do more testing.

slevisconsulting:
@ekluzek this far I have done the following:
1) `git pull`
2) Here in github I resolved the conflicts in clmfates_interfaceMod.F90 and Externals_CLM.cfg that you pointed out in your last post. I resolved them by picking the option that looked like my branch and not like fates_next_api. I may have made the wrong choice...
3) `git pull`
4) ./case_build FAILED
5) git reset f530c993ed448b837670a077c6654f74d7318cf5
6) git checkout -- <all files modified in the last pull/merge>
7) **./case_build PASSED** but due to steps (5) and (6) my local branch is behind my branch on github. This will have to wait...
8) **./case_submit FAILS because the fates_params file has changed.** I'm updating it as we speak and finding that a large number of new variables have been added. I wonder if @pollybuotte has generated a new version of `fates_params_PPine_Cedar.nc` that I could use. If not, I will keep updating.

slevisconsulting:
I generated a new fates_params file with @pollybuotte's help.

The model now runs with use_fates_spitfire = 0 and 1.

The model fails with use_fates_spitfire = 2 in subroutine dynamics_driv in this line:
`lnfm24 = this%fates_fire_data_method%GetLight24()`
I will continue debugging tomorrow.

slevisconsulting:
Update:
The model now also runs with use_fates_spitfire = 2 (Li et al. lightning) and use_fates_spitfire = 3 (Bin's ignitions). The model simulates FIRE_AREA > 0 when I set `cg_strikes = 1` and `fire_threshold = 0` as we had found to be the case in my first implementation of this work. The default values cg_strikes = 0.2 and fire_threshold = 50 give FIRE_AREA = 0 at CZ2.

@ekluzek I would like to push to this PR the last few mods required to get it working, but I don't know how to handle the merge that I made a few lines up (https://github.com/ESCOMP/CTSM/pull/991/commits/01e2b00f99e187421e7c883d59691bcaa49101cb). In my local version I backed out of this merge to get things working. Is there a way to back out of the merge here in the PR, as well?

slevisconsulting:
**Cheyenne testing recommended by @rgknox PASSed:**
`./create_fastest_cheyenne.sh no-basegen compare`

**Izumi testing FAILs in the SHAREDLIB_BUILD step with this error:**
/fs/cgd/data0/slevis/git_FATES/clm5_190325/cime/src/share/timing/Makefile:57: recipe for target 'gptl.o' failed
gmake: Leaving directory '/scratch/cluster/slevis/ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.izumi_nag.clm-FatesHydro.20200606_090915_1r6qh3/bld/nag/mvapich2/debug/nothreads/gptl'
gmake: *** [gptl.o] Error 127
ERROR: gmake: *** [gptl.o] Error 127

I don't know how to resolve this. I didn't have this problem before merging the latest codes from fates_next_api & fates with my lightning branches. I will continue working on cheyenne for now.

@glemieux and everyone, do you know if fates_next_api was tested on izumi during the recent flurry of updates?

ekluzek:
@slevisconsulting izumi was recently upgraded and as such requires a cime update. We are getting that in place, but it's not quite ready yet. So I would ignore this problem for now, until we get a cime version that we've tested and know works with ctsm.

slevisconsulting:
As recommended by @rgknox in our meeting last Friday:
- I ran FATES testing on cheyenne:  PASS
- I added a new test to FATES testing that exercises the lighting dataset code with fates_spitfire_mode = 3 ("successful ignitions" mode). I refrained from adding two tests (one for lightning datasets and one for successful ignitions data) because the only difference between the two cases is a simple if-statement in SPITFIRE.

**NB. In the corresponding commit message I pointed out that this test has uncovered a preexisting problem that I have documented in ExpectedTestFails.xml and which was missed before due to the short lengths of other FATES tests.** Let me know if you have questions about that.

cime_config/testdefs/ExpectedTestFails.xml
```diff
@@ -12,7 +12,7 @@
   </category>
   <category name="fates">
     <entry issue="NGEET/fates/#559" >FAIL SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro RUN</entry>
-    <entry issue="667"              >FAIL ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro COMPARE_base_rest</entry>
+    <entry issue="991"              >FAIL ERS_D_Lm12.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesSuccIgnitions COMPARE_base_rest</entry>
```
slevisconsulting, line 15: The tests referencing issue #667 seem fixed. I did not remove the rest of them from this list in case I am mistaken.

However, the **COMPARE_base_rest** failure reappears now with this new test. I have confirmed that the failure also occurs with `fates_spitfire_mode = 1` which suggests to me a preexisting problem.

slevisconsulting: I further confirmed that the **COMPARE_base_rest** failure is a preexisting problem by switching to fates_next_api and master and submitting this test:
ERS_D_Lm12.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-Fates

@glemieux @ekluzek would you like me to open an issue?

slevisconsulting:
This work is ready for final review. It includes this PR #991 on the CTSM side and PR https://github.com/NGEET/fates/pull/635 on the FATES side.

Other than submitting your review(s), please let me know if there are any other final steps that must be completed before we're ready to merge with fates_next_api and master.

Thanks, everyone!

bld/unit_testers/build-namelist_test.pl
```diff
@@ -822,7 +822,7 @@ sub make_env_run {
                                      conopts=>"-phys clm4_0",
                                    },
      "usespitfireButNOTFATES"    =>{ options=>"-envxml_dir . -no-megan",
-                                     namelst=>"use_fates_spitfire=.true.",
+                                     namelst=>"fates_spitfire_mode>1",
```
slevisconsulting, line 825: I need to change this to `fates_spitfire_mode>0`

bld/CLMBuildNamelist.pm
```diff
@@ -940,7 +940,7 @@ sub setup_cmdl_fire_light_res {
        add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'fire_method');
     }
     my $fire_method = remove_leading_and_trailing_quotes( $nl->get_value('fire_method') );
-    if ( defined($fire_method) && ! &value_is_true($nl_flags->{'use_cn'}) ) {
+    if ( defined($fire_method) && ! &value_is_true($nl_flags->{'use_cn'}) && ! &value_is_true($nl_flags->{'use_fates'}) ) {
        $log->fatal_error("fire_method is being set even though bgc is NOT cn or bgc.");
```
slevisconsulting, line 944: I need to update this `fatal_error` comment.

src/biogeochem/CNFireBaseMod.F90
```diff
@@ -130,8 +131,11 @@ subroutine CNFireInit( this, bounds, NLFilename )
        allocate( this%forc_hdm(bounds%begg:bounds%endg) )
        this%forc_hdm(bounds%begg:) = nan
 
-       call this%hdm_init(bounds, NLFilename)
-       call this%hdm_interp(bounds)
+       if (use_cn) then
+          call this%hdm_init(bounds, NLFilename)
+          call this%hdm_interp(bounds)
+       end if
```
slevisconsulting, line 137: @jkshuman removing this `if (use_cn)` will give SPITFIRE access to the human pop. density dataset.

src/biogeochem/CNFireBaseMod.F90
```diff
@@ -249,13 +253,18 @@ subroutine CNFireInterp(this,bounds)
...
     ! !ARGUMENTS:
     class(cnfire_base_type) :: this
     type(bounds_type), intent(in) :: bounds  
     !-----------------------------------------------------------------------
 
     if ( this%need_lightning_and_popdens ) then
-       call this%hdm_interp(bounds)
+       if (use_cn) then
+          call this%hdm_interp(bounds)
+       end if
```
slevisconsulting, line 267: @jkshuman this is another `if (use_cn)` that would need to be removed to allow SPITFIRE access to the human pop. density dataset.

bld/unit_testers/build-namelist_test.pl
slevisconsulting: Ok, done.

bld/CLMBuildNamelist.pm
slevisconsulting: Done.

cime_config/testdefs/ExpectedTestFails.xml
glemieux: @slevisconsulting yes please open an issue.  I'm not quite familiar with the background so detailing it in an issue may help me.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -675,11 +686,30 @@ subroutine dynamics_driv(this, nc, bounds_clump,      &
...
 
          c = this%f2hmap(nc)%fcolumn(s)
 
+         if (fates_spitfire_mode > 1) then
+            g = col%gridcell(c)
+            do ifp = 1, this%fates(nc)%sites(s)%youngest_patch%patchno
+               p = ifp + col%patchi(c)
+
+               this%fates(nc)%bc_in(s)%lightning24(ifp) = lnfm24(g) * 24._r8  ! #/km2/hr to #/km2/day
+               this%fates(nc)%bc_in(s)%pop_density(ifp) = this%fates_fire_data_method%forc_hdm(g)
```
slevisconsulting, line 709: @ekluzek regarding my code mods to pass the population density to FATES:

I am comfortable with everything except line 709...
```
this%fates(nc)%bc_in(s)%pop_density(ifp) = this%fates_fire_data_method%forc_hdm(g)
```

I'm pointing it out in case you can tell that it won't work right off the bat. I plan to begin testing tomorrow, so I'll know soon enough :-)

ekluzek: It looks correct to me, from just glancing at it...

slevisconsulting: Thanks @ekluzek .

bld/CLMBuildNamelist.pm
```diff
@@ -945,18 +945,12 @@ sub setup_cmdl_fire_light_res {
     }
     if ( defined($fire_method) && $fire_method eq "nofire" ) {
        $nl_flags->{$var} = ".false.";
-    } elsif ( &value_is_true($nl->get_value('use_cn')) ) {
+#   } elsif ( &value_is_true($nl->get_value('use_cn')) || $nl_flags->{'fates_spitfire_mode'} > 1 ) {
+    } elsif ( &value_is_true($nl->get_value('use_cn')) || &value_is_true($nl->get_value('use_fates')) ) {
```
slevisconsulting, line 949: New line 949 works
```
    } elsif ( &value_is_true($nl->get_value('use_cn')) || &value_is_true($nl->get_value('use_fates')) ) {
```
However, I wish the commented-out line above it would work because it would narrow things down, but it doesn't work. The run proceeds to the next else and sets `cnfireson = .false.`.

@ekluzek do you see an obvious problem with new line 948?

slevisconsulting:
FATES testing passes except for the restart test that was already failing as reported in #667 .

This PR is now ready for review again:
@glemieux do you recommend the reviews be done before the upcoming re-basing?
Either way, I am requesting reviews from @ekluzek and @glemieux . All others are welcome to review if interested.

glemieux:
@slevisconsulting yes, ideally the reviews would be completed before the rebase so as to bring in the a consistent set of code that includes review change requests.  That said, if there are non-critical change requests that could be brought in later without much hassle and tests are all passing, I'm inclined to rebase this sooner rather that later.  Can you point me to the test directory on cheyenne for confirmation, please?

slevisconsulting:
> Can you point me to the test directory on cheyenne for confirmation, please?

You prob. already know this: While Cheyenne is down, you can access the same directory from Casper:
`/glade/scratch/slevis/clmfates_tests`

Note that this test
`ERS_D_Lm12.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-Fates_nat_and_anthro_ignitions`
came back as a FAIL because it timed out. I ran a longer variation on this test that PASSed:
`/glade/scratch/slevis/ERS_D_Lm36.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-Fates_nat_and_anthro_ignitions.20200712_122433_bhgfz5`

glemieux:
> > Can you point me to the test directory on cheyenne for confirmation, please?
> 
> You prob. already know this: While Cheyenne is down, you can access the same directory from Casper:
 
I didn't know that; thanks for the tip!

slevisconsulting:
> > > Can you point me to the test directory on cheyenne for confirmation, please?
> > 
> > 
> > You prob. already know this: While Cheyenne is down, you can access the same directory from Casper:
> 
> I didn't know that; thanks for the tip!

@glemieux I should have pointed out that today specifically Casper will also be down. Sorry for not mentioning sooner...

bld/unit_testers/build-namelist_test.pl
```diff
@@ -822,7 +822,7 @@ sub make_env_run {
                                      conopts=>"-phys clm4_0",
                                    },
      "usespitfireButNOTFATES"    =>{ options=>"-envxml_dir . -no-megan",
```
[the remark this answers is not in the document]
slevisconsulting: @jkshuman this belongs to a list of options that trigger namelist FAILs. The specific one triggers a failure if `fates_spitfire_mode > 0` and -bgc has not been set to fates.

src/main/clm_initializeMod.F90
```diff
@@ -474,6 +474,12 @@ subroutine initialize2( )
        end if
     else
        call SatellitePhenologyInit(bounds_proc)
+
+       ! fates_spitfire_mode is assigned an integer value in the namelist
+       ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
```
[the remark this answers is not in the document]
slevisconsulting: I'm leaning against this one because it introduces the risk of being missed in a future update of `fates_spitfire_mode`.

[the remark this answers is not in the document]
slevisconsulting: Though I see the benefit of including some explanation of the `> 1` and maybe the naming of these various modes (that I introduced elsewhere) needs to be used here, as well.

src/main/clm_varctl.F90
```diff
@@ -207,7 +207,7 @@ module clm_varctl
   integer, public            :: fates_parteh_mode = -9                 ! 1 => carbon only
                                                                        ! 2 => C+N+P (not enabled yet)
                                                                        ! no others enabled
-  logical, public            :: use_fates_spitfire = .false.           ! true => use spitfire model
+  integer, public            :: fates_spitfire_mode = 0                ! > 0 => use spitfire model: see bld/namelist_files/namelist_definition_clm4_5.xml for details
```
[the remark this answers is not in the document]
slevisconsulting: [Same as my most recent comment](https://github.com/ESCOMP/CTSM/pull/991#discussion_r454540753):

Name the various modes in a central location and then use them in all cases like this instead of hardwiring.

src/main/clm_initializeMod.F90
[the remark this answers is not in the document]
jkshuman: valid point, though it may be worth a quick explanation in more than 1 central location

src/main/clm_varctl.F90
jkshuman, line 210: How about this as a change?
```suggestion
  integer, public            :: fates_spitfire_mode = 0                
  ! > 0 => use spitfire model: see bld/namelist_files/namelist_definition_clm4_5.xml for details
  ! ignitions: 1=constant, >1=external data sources
```

src/main/clm_varctl.F90
```diff
@@ -207,7 +207,9 @@ module clm_varctl
   integer, public            :: fates_parteh_mode = -9                 ! 1 => carbon only
                                                                        ! 2 => C+N+P (not enabled yet)
                                                                        ! no others enabled
-  logical, public            :: use_fates_spitfire = .false.           ! true => use spitfire model
+  integer, public            :: fates_spitfire_mode = 0                
+  ! > 0 => use spitfire model: see bld/namelist_files/namelist_definition_clm4_5.xml for details
+  ! ignitions: 1=constant, >1=external data sources
```
jkshuman, line 212:
```suggestion
  ! ignitions: 1=constant, >1=external data sources (lightning and/or anthropogenic)
```

slevisconsulting:
> @slevisconsulting yes, ideally the reviews would be completed before the rebase so as to bring in the a consistent set of code that includes review change requests. That said, if there are non-critical change requests that could be brought in later without much hassle and tests are all passing, I'm inclined to rebase this sooner rather that later. Can you point me to the test directory on cheyenne for confirmation, please?

@glemieux you may have seen that I made some minor revisions in response to @jkshuman 's review. I have a few more to make once cheyenne comes up. I just want to make sure that you will let me know when I should stop further modifications to avoid conflicts with your re-basing.

glemieux:
Thanks for the heads up @slevisconsulting

src/main/clm_varctl.F90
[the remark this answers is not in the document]
slevisconsulting: I completed this in the current PR's latest commit and in https://github.com/NGEET/fates/pull/635.

The model builds, but I get a runtime system error in my cesm.log that looks like a problem leftover from cheyenne's downtime. What do you suggest @ekluzek :
```dplace failed. Verify that the numatools module is loaded.
MPT ERROR: could not run executable (case #1)
        (HPE MPT 2.19  02/23/19 05:31:12)```

src/main/clm_initializeMod.F90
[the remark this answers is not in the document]
slevisconsulting: I replaced the `> 1` with `> scalar_lightning` in the latest commit.

src/main/clm_varctl.F90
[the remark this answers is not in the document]
slevisconsulting: Good news:
Whatever this was, it's been fixed, and my test now runs :-)

slevisconsulting:
@glemieux I reran FATES testing successfully after my revisions in response to @jkshuman 's reviews.

Testing directory:
`/glade/scratch/slevis/clmfates_tests`

PR #999, opened 2020-04-28
Remove Copyright file

### LINKED ISSUE #997: We should remove the "Copyright" file
I noticed that we have an old "Copyright" file at the top level of the repository, that is both out of date and incorrect – critically, using the legally incorrect term, "public domain".

I recently updated this file in CESM: https://github.com/ESCOMP/CESM/pull/141

However, I'd suggest that we simply remove this file from CTSM. It's not clear to me that we really need to maintain this file, since other code isn't included in our repository but rather downloaded separately via manage_externals, and I think it's confusing to have separate LICENSE and COPYRIGHT files. However, we may want to have a note in the LICENSE file saying that this license only applies to the code in this repository, and that externals obtained via manage_externals have their own licenses.

billsacks:
### Description of changes

Remove Copyright file

### Specific notes

Discussion from https://github.com/ESCOMP/CTSM/issues/997:

I noticed that we have an old "Copyright" file at the top level of the
repository, that is both out of date and incorrect – critically, using
the legally incorrect term, "public domain".

I recently updated this file in CESM: ESCOMP/CESM#141

However, I'd suggest that we simply remove this file from CTSM. It's not
clear to me that we really need to maintain this file, since other code
isn't included in our repository but rather downloaded separately via
manage_externals, and I think it's confusing to have separate LICENSE
and COPYRIGHT files. However, we may want to have a note in the LICENSE
file saying that this license only applies to the code in this
repository, and that externals obtained via manage_externals have their
own licenses.

- Resolves #997


billsacks:
@ekluzek if you approve, I'll merge this directly to the release branch, but won't make a tag.

PR #1003, opened 2020-05-06
Allow for resorbtion in transition from live to dead wood N

wwieder:
### Description of changes
allows for resorbtion in transition from live to dead wood N, which also move to NPOOL for free.

### Specific notes
allows resorbtion of live_to_dead wood N flux
moves live stem and root retranslocated N to NPOOL for free

This was first noticed in single point simulations by @susancheng. @rosiealice provided input & suggestions.

#443 

All modifications increased GPP and CUE, resulting in higher LAI and VEG C pools, but the collateral changes to the C cycle seem minimized with these suggested changes here. Results seem to be behaving appropriately, with no excess N accumulating in RETRANSN and NPOOL, but is incorrectly being handled in CNNStateUpdate as presently implemented.

none

Testing performed, if any:
none

I'm sure I'll need to clean up some of this.

### DIFF
--- a/src/biogeochem/CNNStateUpdate1Mod.F90
+++ b/src/biogeochem/CNNStateUpdate1Mod.F90
@@ -184,11 +184,16 @@ contains
             ns_veg%livestemn_patch(p)    = ns_veg%livestemn_patch(p)  - nf_veg%livestemn_to_deadstemn_patch(p)*dt
             ns_veg%deadstemn_patch(p)    = ns_veg%deadstemn_patch(p)  + nf_veg%livestemn_to_deadstemn_patch(p)*dt
             ns_veg%livestemn_patch(p)    = ns_veg%livestemn_patch(p)  - nf_veg%livestemn_to_retransn_patch(p)*dt
-            ns_veg%retransn_patch(p)     = ns_veg%retransn_patch(p)   + nf_veg%livestemn_to_retransn_patch(p)*dt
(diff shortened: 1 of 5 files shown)


wwieder:
This is all work that I did a while ago and never merged back into trunk.  After cleaning things up I'll make a clean comparison to check things are working sensibly.

wwieder:
@ekluzek I'm guessing the case below is too old to rerun, it's from dev003 (maybe something to do with `p_old` vs. `p_new`).  But I'd like to look at a few simulations before cleaning up this PR.
Currently I get this error when running `preview_namelist`
`Invalid values ['10800.0']`
`ERROR: Variable 'coupling_period' has invalid value ['10800.0'].`

/glade/work/wwieder/ctsm/ctsm1.0.dev003/cime/scripts/clm5_4x5_woodCN_tests_5a_freeRetrans_50_500_spin

Do I need to update to a current tag to be able to re-run?  Can I do this easily without reverting my changes by just checking out from master and migrating changes to src?

ekluzek:
@wwieder actually that is an easy thing to fix. It just has to do with an easy python 2 to python 3 fix. And actually if you run with python 2, you wouldn't likely have to change anything.

I think that error is just in the ROF component. So you probably just need to update MOSART to the latest (which shouldn't change answers). So that would be an easy thing to do.

wwieder:
additional testing done with updated code base in #1063

PR #1007, opened 2020-05-07
fates interface refactor (fix circular deps)

rgknox:
### Description of changes

These changes fix initialization sequence problems with FATES, and remove some circular dependencies that are blocking some fates development efforts. These changes should also enable ERS tests with cohort age tracking.

This is a major API change, and synced with:

https://github.com/NGEET/fates/pull/647

### Specific notes

Contributors other than yourself, if any:  @xuchongang

Results should not change, except that cohort age test should pass now

### Testing:
TBD


glemieux:
Good catch adding in the latest bug fix.  I should be done reviewing this in the next couple hours.

glemieux:
I think this looks good.

rgknox:
In testing this fix, I ran into some problems with the testing and wanted to report the issue and solution.

Prior to this set of changes, we were calling a routine that defines the dimensions of the history file, before we had determined the number of fates chemical elements are enabled.  This dimension, fates_elem was just writing zero's do the dimension variables.  This problem was not detected yet, because not many people (none?) have really been using the element dimension (as it really only has one species in it, carbon).   Also, it appears that the testing software, cprnc, will not calculate RRMS differences between bases and test datasets if they are all zero's.  But in my tests, since the ordering fixed fates_elem, which is now not zero, it was breaking cprnc, and it was now trying to perform a divide by zero on the base dataset, since the the filter was removed now having non-zero's in the test dataset.   I verified that fates_elem is zero in the base dataset, and is 1 in the test dataset:
```
ctsm_baselines> ncdump -v fates_levelem fates-sci.1.35.2_api.11.0.0-clm5.0.30-C39a99d9e-F3bba389b.fates.cheyenne.intel/ERS_Ld5.f19_g16.I2000Clm45Fates.cheyenne_intel.clm-FatesColdDef/clm2.h0.0001-01-01-00000.nc | grep levelem
	fates_levelem = 1 ;
	int fates_levelem(fates_levelem) ;
		fates_levelem:long_name = "FATES element (C,N,P,...) identifier" ;
		fates_levelem:units = "-" ;
 fates_levelem = 1 ;
```

```
tss/ctsm_baselines> ncdump -v fates_levelem fates-sci.1.35.1_api.10.0.0-clm5.0.30-C242d6d14-F987f521a.fates.cheyenne.intel/ERS_Ld5.f19_g16.I2000Clm45Fates.cheyenne_intel.clm-FatesColdDef/clm2.h0.0001-01-01-00000.nc  | grep levelem
	fates_levelem = 1 ;
	int fates_levelem(fates_levelem) ;
		fates_levelem:long_name = "FATES element (C,N,P,...) identifier" ;
		fates_levelem:units = "-" ;
 fates_levelem = 0 ;
```

PR #1016, opened 2020-05-18
Heat storage biomass

### LINKED ISSUE #176: CNDV bug: ntree used before it's set in pftconMod.F90
**Erik Kluzek < erik > - 2017-08-25 18:31:41 -0600**
**Bugzilla Id:** [2506](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2506)
**Bugzilla CC:**  dlawren,  mvertens,  rfisher,  slevis, 

Sam Levis found this problem. The variable ntree is used before it's set in pftconMod.F90, which means the variable "tree" will be set incorrectly. It looks like the only use of the variable "tree" is when Dynamic Vegetation is used in CNDVLightMod.F90.

This goes back to the beginning of pftconMod.F90 in some refactoring that went on in clm4_5_1_r097.

swensosc:
### Description of changes
Add heat stored in biomass to surface energy balance calculation.

### Specific notes

Ronny Meier

Papers describing BHS simulations:
R. Meier, Davin, E., Swenson, S., Lawrence, D., and Schwaab, Jo. (2019). Biomass heat storage dampens diurnal temperature variations in forests. Environmental Research Letters. 14. 084026. 10.1088/1748-9326/ab2b4e. 

S.C. Swenson, Burns, S. P., and Lawrence, D. M. ( 2019). The impact of biomass heat storage on the canopy energy balance and atmospheric stability in the community land model, Journal of Advances in Modeling Earth Systems, 11, 83– 98. https://doi.org/10.1029/2018MS001476 

 Finishes fixing #342
 Fixes most of #1246
Fixes #176

Yes.  The code modifies the surface energy balance, which will change climate.

Any User Interface Changes (namelist or namelist defaults changes)? New namelist for this

Testing performed, if any: Standard testing on cheyenne and izumi now all PASS as expected
Single point tower site simulations were run for 10-15 locations using observed meteorology and assessed them against observed surface energy fluxes.  I also ran decade-long global simulations with use_biomass_heat_storage both on and off. Simulations were run on the Hobart cluster using the intel compiler.


src/main/clm_varcon.F90
```diff
@@ -120,11 +120,12 @@ module clm_varcon
 
   real(r8), public :: thk_bedrock = 3.0_r8  ! thermal conductivity of 'typical' saturated granitic rock 
                                     ! (Clauser and Huenges, 1995)(W/m/K)
-  real(r8), public :: csol_bedrock = 2.0e6_r8 ! vol. heat capacity of granite/sandstone  J/(m3 K)(Shabbir, 2000) !scs
```
ekluzek, line 123: Remove the !scs comment

ekluzek:
Also can you tell us exactly what test cases you ran with the two new things turned on? How long did you run them for and what did you do to verify they were correct? Thanks.

src/main/clm_varcon.F90
swensosc: I think the '!scs' is actually on master?  History fields that could be activated are 'T_STEM' and 'DHSDT_CANOPY'.  I would not activate the new fields in FrictionVelocityMod.  I don't think there are restrictions for the namelist variables.

dlawrenncar:
Seems like we should activate those two history fields.  Might as well be
able to track what is happening with the biomass heat storage code.

billsacks:
@olyson - can you please give this a light review for the science? (request from @dlawrenncar / @swensosc ).

olyson:
From a science perspective, I think this good to go.  The parameterization and its effects on the surface energy balance have been documented in two published papers.  I myself have run single-point simulations using this parameterization and noted improvement in the modeling of latent and sensible heat fluxes compared to observations.  Improvements in the modeling of nighttime momentum flux and canopy air temperature were also noted in the published papers.  It is a real physical process that should be represented and accounts for a missing energy balance term in the model.

I see a discussion about history field variables.  Should AGSB and AGLB be on by default in CanopyStateType.F90?

Should these parameters be on the parameter file instead of being hardcoded:

    ! biomass parameters
    real(r8), parameter :: k_vert = 0.1            !vertical distribution of stem
    real(r8), parameter :: k_cyl_vol = 1.0         !departure from cylindrical volume
    real(r8), parameter :: k_cyl_area = 1.0        !departure from cylindrical area
    real(r8), parameter :: k_internal = 0.0        !self-absorbtion of leaf/stem longwave
    real(r8), parameter :: min_stem_diameter = 0.01!minimum stem diameter for which to calculate stem interactions

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -77,7 +78,7 @@ module CanopyFluxesMod
   logical,  public :: perchroot_alt = .false.  
   !
   ! !PRIVATE DATA MEMBERS:
-  logical, private :: use_undercanopy_stability = .true.      ! use undercanopy stability term or not
```
ekluzek, line 80: Since by default this is set in the namelist, the default in the code doesn't really matter here.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -404,22 +404,49 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+    real(r8) :: lw_stem(bounds%begp:bounds%endp)   !internal longwave stem
+    real(r8) :: lw_leaf(bounds%begp:bounds%endp)   !internal longwave leaf
+    real(r8) :: sa_stem(bounds%begp:bounds%endp)   !surface area stem m2/m2_ground
+    real(r8) :: sa_leaf(bounds%begp:bounds%endp)   !surface area leaf m2/m2_ground
+    real(r8) :: sa_internal(bounds%begp:bounds%endp)   !min(sa_stem,sa_leaf)
+    real(r8) :: uuc(bounds%begp:bounds%endp)       ! undercanopy windspeed
+    real(r8) :: carea_stem                         !cross-sectional area of stem
+
+    ! biomass parameters
+    real(r8), parameter :: k_vert = 0.1            !vertical distribution of stem
```
ekluzek, line 422: Shouldn't these all go on the params file?

src/main/clm_varcon.F90
ekluzek: OK, this comment is taken care of

[x] scs comment
[x] Added T_STEM and DHSDT_CANOPY to history when bhs is turned on.
[x] New history fields are default off

src/biogeophys/CanopyFluxesMod.F90
ekluzek: Nothing needs to be done here.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: @swensosc can you comment on these new parameters in CanopyFluxesMod -- are they something that would be adjusted and hence should be on the params file?

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -629,8 +673,92 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+            ! double in spirit of full surface area for sensible heat
+            sa_leaf(p) = 2.*sa_leaf(p)
+
+            sa_stem(p) = nstem(patch%itype(p))*(htop(p)*shr_const_pi*dbh(p))
+            ! adjust for departure of cylindrical stem model
+            sa_stem(p) = k_cyl_area * sa_stem(p)
+
+            ! do not calculate separate leaf/stem heat capacity for grasses
+            ! or other pfts if dbh is below minimum value
+            if(patch%itype(p) > 11 .or. dbh(p) < min_stem_diameter) then
```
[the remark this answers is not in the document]
ekluzek: This was converted into a is_shrub and is_tree logicals. And it looks like they all work now.

[x] Change if to check if tree or shrub
[x[ Fix #176 for is_tree
[x] Fix @21 to add grass as a logical
[x] Add shrub as a logical

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -851,20 +987,17 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
 
             ri = ( grav*htop(p) * (taf(p) - t_grnd(c)) ) / (taf(p) * uaf(p) **2.00_r8)
 
-            !! modify csoilc value (0.004) if the under-canopy is in stable condition
-
-            if (use_undercanopy_stability .and. (taf(p) - t_grnd(c) ) > 0._r8) then
```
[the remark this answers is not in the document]
ekluzek: Looks like this was taken care of.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -629,8 +673,92 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+            ! internal longwave fluxes between leaf and stem
+            ! (use same area of interaction i.e. ignore leaf <-> leaf)
+            sa_internal(p) = min(sa_leaf(p),sa_stem(p))
+            sa_internal(p) = k_internal * sa_internal(p)
+
+            ! calculate specify heat capacity of vegetation
+            ! as weighted averaged of dry biomass and water
+            ! lma_dry has units of kg dry mass/m2 here (table 2 of bonan 2017) 
+            ! cdry_biomass = 1400 J/kg/K, cwater = 4188 J/kg/K
+            cp_leaf(p)  = leaf_biomass(p) * (1400._r8*(1.-fbw(patch%itype(p))) + (fbw(patch%itype(p)))*4188._r8)
```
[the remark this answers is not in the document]
ekluzek: It makes sense that cwater=4188 should go into clm_varcon. If you think cdry_biomass=1400 might need to vary, it probably should go into the params file. But, it would be easier to start with it just being on clm_varcon.

[ ] Make cwater a parameter in clm_varcon
[ ] Make cdry_biomass a parameter either in clm_varcon or on params file.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -958,7 +1105,9 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
                canopy_cond(p) = (laisun(p)/(rb(p)+rssun(p)) + laisha(p)/(rb(p)+rssha(p)))/max(elai(p), 0.01_r8)
             end if
 
-            efpot = forc_rho(c)*wtl*(qsatl(p)-qaf(p))
+! should be the same expression used in Photosynthesis/getqflx
+!scs            efpot = forc_rho(c)*elai(p)/rb(p)*(qsatl(p)-qaf(p))
```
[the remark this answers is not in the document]
ekluzek: Think this is taken care of.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1000,6 +1149,7 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
             ! Moved the original subroutine in-line...
 
             wtaq    = frac_veg_nosno(p)/raw(p,1)                        ! air
+!scs            wtlq    = frac_veg_nosno(p)*elai(p)/rb(p) * rpp   ! leaf
```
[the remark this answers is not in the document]
ekluzek: OK, taken care of.

src/biogeophys/TemperatureType.F90
```diff
@@ -388,6 +390,11 @@ subroutine InitHistory(this, bounds, is_simple_buildtemp, is_prog_buildtemp )
          avgflag='A', long_name='Urban daily maximum of average 2-m temperature', &
          ptr_patch=this%t_ref2m_max_u_patch, set_nourb=spval, default='inactive')
 
+    this%t_stem_patch(begp:endp) = spval
```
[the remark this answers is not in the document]
ekluzek: taken care of.

src/biogeophys/EnergyFluxType.F90
```diff
@@ -433,6 +440,16 @@ subroutine InitHistory(this, bounds, is_simple_buildtemp)
          avgflag='A', long_name='sensible heat from veg', &
          ptr_patch=this%eflx_sh_veg_patch, set_lake=0._r8, c2l_scale_type='urbanf')
 
+    this%eflx_sh_stem_patch(begp:endp) = spval
+    call hist_addfld1d (fname='FSH_STEM', units='W/m^2',  &
+         avgflag='A', long_name='sensible heat from stem', &
+         ptr_patch=this%eflx_sh_stem_patch, c2l_scale_type='urbanf',default = 'inactive')
+
+    this%dhsdt_canopy_patch(begp:endp) = spval
```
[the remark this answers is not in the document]
ekluzek: [ ] Move to inside the if (use_biomass_heat_storage) then

src/main/initGridCellsMod.F90
```diff
@@ -71,6 +71,7 @@ subroutine initGridcells(glc_behavior)
     integer :: nclumps              ! number of clumps on this processor
     type(bounds_type) :: bounds_proc
     type(bounds_type) :: bounds_clump
+
```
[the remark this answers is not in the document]
ekluzek: This was removed. I removed another one, so now identical to main-dev.

src/main/subgridMod.F90
```diff
@@ -73,6 +73,7 @@ subroutine subgrid_get_gcellinfo (gi, glc_behavior, &
     ! atm_topo is arbitrary for the sake of getting these counts. We don't have a true
     ! atm_topo value at the point of this call, so use 0.
     real(r8), parameter :: atm_topo = 0._r8
+
```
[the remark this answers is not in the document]
ekluzek: This was removed. I removed another one, so identical to ctsm1.0.dev113.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1269,13 +1474,13 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
          ! Downward longwave radiation below the canopy
 
          dlrad(p) = (1._r8-emv(p))*emg(c)*forc_lwrad(c) + &
-              emv(p)*emg(c)*sb*tlbef(p)**3*(tlbef(p) + 4._r8*dt_veg(p))
+              emv(p)*emg(c)*sb*(tlbef(p)**3*(tlbef(p) + 4._r8*dt_veg(p))*(1.-frac_rad_abs_by_stem(p))+ts_ini(p)**3*(ts_ini(p) + 4._r8*dt_stem(p))*frac_rad_abs_by_stem(p))
```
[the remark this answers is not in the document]
ekluzek: [ ] - Shorten long lines by breaking up with continuation lines &

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1183,25 +1355,58 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
 
-         delt    = wtal(p)*t_grnd(c)-wtl0(p)*t_veg(p)-wta0(p)*thm(p)
          taux(p) = -forc_rho(c)*forc_u(g)/ram1(p)
          tauy(p) = -forc_rho(c)*forc_v(g)/ram1(p)
          eflx_sh_grnd(p) = cpair*forc_rho(c)*wtg(p)*delt
 
          ! compute individual sensible heat fluxes
-         delt_snow = wtal(p)*t_soisno(c,snl(c)+1)-wtl0(p)*t_veg(p)-wta0(p)*thm(p)
-         eflx_sh_snow(p) = cpair*forc_rho(c)*wtg(p)*delt_snow
+         delt_snow = wtal(p)*t_soisno(c,snl(c)+1)-wtl0(p)*t_veg(p)-wta0(p)*thm(p)-wtstem0(p)*t_stem(p)
```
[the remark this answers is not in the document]
ekluzek: This was just a comment. So I'll resolve this.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -876,6 +1009,13 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
             svpts(p) = el(p)                         ! pa
             eah(p) = forc_pbot(c) * qaf(p) / 0.622_r8   ! pa
             rhaf(p) = eah(p)/svpts(p)
+! add history fields
```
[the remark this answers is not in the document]
ekluzek: Taken care of.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -629,8 +673,92 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
          btran2(p)  = btran0
+         dhsdt_canopy(p) = 0._r8
+         eflx_sh_stem(p) = 0._r8
       end do
 
+      ! calculate biomass heat capacities
+      do f = 1, fn
+         p = filterp(f)
+
+         if(use_biomass_heat_storage) then
```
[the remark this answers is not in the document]
ekluzek: The else just sets some variables to zero and one tiny sum. I suppose you are just trying to save writing a new loop. Putting the if outside the loop is better for optimization, but branch prediction will likely not cause too much loss of performance.

I'll work on this one.

[ ] - Put if statement outside of loop.

src/biogeophys/CanopyFluxesMod.F90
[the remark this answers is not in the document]
ekluzek: @swensosc can you take this one?

src/biogeophys/EnergyFluxType.F90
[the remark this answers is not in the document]
ekluzek: @swensosc can you take this one?

src/biogeophys/CanopyFluxesMod.F90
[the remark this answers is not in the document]
ekluzek: @swensosc can you take this one?

src/biogeophys/CanopyFluxesMod.F90
ekluzek: We decided to just add a comment about this. They don't need to go on the params file right now.

src/biogeophys/CanopyFluxesMod.F90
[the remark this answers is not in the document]
ekluzek: Sean put these both in clm_varcon.

src/biogeophys/CanopyFluxesMod.F90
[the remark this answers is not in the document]
ekluzek: OK, I went ahead and made this change.

ekluzek:
@swensosc I noticed that you have a bit in the code where stem biomass is multiplied by 10 when Accelerated Decomposition mode is turned on (spinup_state==2). I thought this might be an obscure linkage to the soil biogeochemistry that a user isn't likely to appreciate. so unless you really think it should be in place, I wonder if it shouldn't be removed? If you think it should be there, I could add some documentation about in the code and in regard to spinup_state.

ekluzek (approved):

dlawrenncar:
I don't think that stem biomass should be modified during accelerated
spinup.  That amplification only applies to non-living carbon pools, I
think.

wwieder:
I think ADspinup applies to wood C and N pools too.
https://github.com/ESCOMP/CTSM/blob/622a196b5e51bf4fb60591b8c02a02fe7b342e8b/src/biogeochem/CNVegStructUpdateMod.F90#L210

https://github.com/ESCOMP/CTSM/blob/e39e9f3ecdbf91b825ffd25bbd6276b65d74f10b/src/biogeochem/CNVegNitrogenStateType.F90#L722

https://github.com/ESCOMP/CTSM/blob/30e7216fe8de808fca3410169d7f78cad8b89ef8/src/biogeochem/CNVegCarbonStateType.F90#L1244

Since actual wood C pools are 10x larger than what's in AD mode, I'm guessing the original code from @swensosc is correct, @ekluzek  (as with the height calculation in `CNVegStrucUpdate`).

dlawrenncar:
Scratch that previous comment.  I forgot that deadstemc was modified during
accelerated spinup.  I believe that Sean is correctly modifying the
deadstemc during spinup because deadstemc is artificially reduced during
accelerated spinup and if you want the physics to be the same between
accelerated and non-accelerated spinup, you need to multiply by this
factor.  I think, therefore, that what is needed is a comment to make it
clear why this is being done.  Comment could be:

! During accelerated spinup (spinup_state = 2), deadstemc is decreased by a
factor of 10; to get correct stem_biomass during spinup for heat storage
calculations, stem_biomass needs
! to be converted back to its correct value.

I see that this correction is really an approximation since stem_biomass is
the sum of deadstemc and livestemc and livestemc is not modified during
accelerated spinup.  However, I think deadstemc is generally much larger
than livestemc so this minor inconsistency probably doesn't matter much,
especially since this is only an issue during accelerated spinup.

ekluzek:
@dlawrenncar and @wwieder thanks for chiming in. OK, so the one thing I still wonder about is if that hard wired constant "10." shouldn't be related to a parameter that's already in the code elsewhere? It's either related to some other hard wired constant or related to a parameter that's already saved in the code somewhere.

It looks like the 10 in CNVegNitrogenStateType.F90 is also hardwired in, so it would be really good to at least make this a constant in the code so that if it get's changed it get's changed together.

src/biogeochem/CNVegStructUpdateMod.F90
```diff
@@ -210,14 +216,28 @@ subroutine CNVegStructUpdate(num_soilp, filter_soilp, &
...
+               if (use_biomass_heat_storage) then
+                  ! Assumes fbw the same for leaves and stems
+                  leaf_biomass(p) = max(0.0025_r8,leafc(p)) &
+                       * c_to_b * 1.e-3_r8 / (1._r8 - fbw(ivt(p)))
+
+                  stem_biomass(p) = (deadstemc(p) + livestemc(p)) &
+                       * c_to_b * 1.e-3_r8 / (1._r8 - fbw(ivt(p)))
+                  
+                  if (spinup_state == 2) then
+                     stem_biomass(p) = 10._r8 * stem_biomass(p)
```
wwieder, line 237: if (spinup_state == 2) then
    stem_biomass(p) = (10._r8 * deadstemc(p) + livestemc(p)) &
                           * c_to_b * 1.e-3_r8 / (1._r8 - fbw(ivt(p)))

wwieder: @ekluzek this sounds nice to make the 10._r8 a constant if it's not too time consuming

Given @dlawrenncar 's comment, it seems like replacing the statement after the logical `spinup_state == 2` would be more accurate  (see above)

ekluzek:
@swensosc running the test suite on the updated version last night, there are a bunch of fails. And there are also a bunch of clm4_5, and clm5_0 tests that I would expect to be identical that show answer changes. Do you know offhand what changes would cause answers to change when biomass heat storage is off? And could you direct me to look at them?

I think the best way to do this would be to have a small tag that does the answer changing thing, and then bring this tag in next after that, so it'll only change answers for clm5_1 physics.

ekluzek:
With @swensosc changes I'm able to show at least one of the tests that were different than baseline is now identical. So I'll rerun the rest and make sure they are identical.

We do still have a bunch of tests that fail. I suspect there is maybe one problem that's causing them all to fail. So solving one or two issues will hopefully get it to work.

ekluzek:
I thought there was an incompatibility with running BHS with FATES. So I did some work to disable the two. But, now I've figured out the problem is really just in setting dhsdt_canopy to zero for more landunits. @swensosc noticed this as well. A ton of more tests pass if I just make sure it's zeroed out for the non BHS case.

ekluzek:
@swensosc and I talked. We think we now know the last of the remaining things to do. nstem for non-biomass heat storage needs to match the "stocking" value in CNVegStructUpdateMod.F90. I should be able to do that in pftcon. And we need to set dhsdt_canopy to zero for glacier, lake, and wetland landunit types. We could optionally set it for urban as well, but urban doesn't matter as the check in BalanceCheck doesn't use the term for urban points. I also need to push my change with the updated paramsfile.

ekluzek:
OK, we think we've solved the baseline changes. The issue I see now is that several of the new clm5_1 tests fail (with BHS on) with getting NaN's in surface temperature sent to the coupler.

Here's an example of tests that pass and fail:

================================================================================
These tests passed
================================================================================
ERP_P36x2_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs			
LILACSMOKE_Vnuopc_D_Ld2.f10_f10_musgs.I2000Ctsm50NwpSpAsRs.cheyenne_intel.clm-lilac			
SMS_D_Ld3.f10_f10_musgs.I2000Clm50BgcCru.cheyenne_intel.clm-default			
SMS_D_Ld9.f09_g17.I1850Clm50BgcNoAnthro.cheyenne_intel.clm-decStart1851_noinitial			
SMS_Ld5.f10_f10_musgs.ISSP585Clm50BgcCrop.cheyenne_intel.clm-ciso_dec2050Start			
SMS_Lm1.f19_g17.I1850Clm50BgcCropCmip6waccm.cheyenne_intel.clm-basic			
SMS_Ln9.ne30pg2_ne30pg2_mg17.I1850Clm50Sp.cheyenne_intel.clm-clm50cam6LndTuningMode			
SMS_Ln9.ne30pg2_ne30pg2_mg17.I2000Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode			
================================================================================
These tests failed
================================================================================
ERI_D_Ld9.f10_f10_musgs.I1850Clm51Bgc.cheyenne_gnu.clm-default		
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm51Bgc.cheyenne_gnu.clm-monthly		
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm51Bgc.cheyenne_intel.clm-monthly		
ERS_Ly5_P144x1.f10_f10_musgs.IHistClm51BgcCrop.cheyenne_intel.clm-cropMonthOutput		
LCISO_Lm13.f10_f10_musgs.IHistClm51BgcCrop.cheyenne_intel.clm-ciso_monthly		
LVG_Ld5_D.f10_f10_musgs.I1850Clm51Bgc.cheyenne_intel.clm-no_vector_output

I'm working on tracking down the problem with NaN's in a SMS_D_Ld9.f10_f10_musgs.I1850Clm51Bgc.cheyenne_gnu.clm-default test.

ekluzek:
I ran more of the longer tests from the test suite to ensure they all don't change answers.

SMS_Lm1.f10_f10_musgs.I1850Clm50BgcCropCmip6waccm.cheyenne_gnu.clm-basic			
SMS_Ly1_Mmpi-serial.1x1_brazil.IHistClm50BgcQianRs.cheyenne_gnu.clm-output_bgc_highfreq			
SMS_Ly3_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.cheyenne_intel.clm-clm50dynroots

They all remained identical. A longer 13-month test with clm5_1 did change answers as expected.

ekluzek:
I've got one unexpected fail now, answers are changing for a PGI test compared to baseline.

ekluzek:
I've verified that other baseline tests fail with PGI compared to ctsm5.1.dev019.

SMS.f10_f10_musgs.I2000Clm51Sp.izumi_pgi.clm-default
SMS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-tracer_consistency
SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50Sp.izumi_pgi.clm-ptsRLA
SMS_D_Ld1_P48x1.f10_f10_musgs.I2000Clm50BgcCru.izumi_pgi.clm-af_bias_v7
SMS_D_Ld5.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-irrig_alternate
SMS_D_Ld5.f10_f10_musgs.I2000Clm50Fates.izumi_pgi.clm-FatesColdDef
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm50FatesRs.izumi_pgi.clm-FatesColdDef
SMS_D_P48x1_Ld5.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-irrig_spunup
SMS_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50Sp.izumi_pgi.clm-ptsRLA
SMS_Ld5_D_P48x1.f10_f10_musgs.IHistClm50Bgc.izumi_pgi.clm-monthly
SMS_P48x1_D_Ld5.f10_f10_musgs.I2000Clm50Cn.izumi_pgi.clm-default

(Note, by the way an urban-only single point test is identical as expected since it doesn't use CanopyFluxes).

@swensosc has done some work on this and says the following...

> For the SMS.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi test, I can get it to work if I replace these three calculations with the baseline code:
> dt_veg(p) =
> dlrad(p) =
>  ulrad(p) =
> 
> Then I went variable by variable for just the dt_veg calculation to see what would make it fail.  But there is nothing I can add that doesn't make it fail.  For example, just doing this:
>  dt_veg(p) = (sabv(p) + air(p) &
>                  + bir(p)*t_veg(p)**4 + cir(p)*lw_grnd &
>                  - efsh - efe(p) &
>                  ) &
>                  / ((- 4._r8*bir(p)*t_veg(p)**3) &
>                  +dc1*wtga(p) +dc2*wtgaq*qsatldT(p) + (cp_leaf(p)/dtime))
> makes it fail.  (the only difference is the addition of the cp_leaf/dtime).  If I explicitly change (cp_leaf(p)/dtime) to 0._r8 it will pass.  But cp_leaf is 0!  I even check if (cp_leaf(p)/dtime) /= 0._r8 and it is never true

ekluzek:
OK, I have some changes that allow the PGI test SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50Sp.izumi_pgi.clm-ptsRLA. I had to include some if statements for dlrad and ulrad. But, they show that the answers are only roundoff different due to changes in order of operations of those statements. So I'm verifying the other PGI tests pass as well. If so I'll then remove the if statements and we'll go with the version that changes answers for PGI. The verification step that shows that answers are only roundoff different makes me confident that the changes are OK. Since, answers are identical for other tests for intel, nag, and gnu compilers I have confidence that this is OK.

ekluzek:
OK a few BGC PGI tests (and one FATES test) with PGI still change answers. But, most of the tests are fine now and compare identical to baseline as expected.

```
SMS_D_Ld10.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-tracer_consistency
SMS_D_Ld1_P48x1.f10_f10_musgs.I2000Clm50BgcCru.izumi_pgi.clm-af_bias_v7
SMS_D_Ld5.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-irrig_alternate
SMS_D_Ld5.f10_f10_musgs.I2000Clm50Fates.izumi_pgi.clm-FatesColdDef
SMS_D_P48x1_Ld5.f10_f10_musgs.I2000Clm50BgcCrop.izumi_pgi.clm-irrig_spunup
SMS_Ld5_D_P48x1.f10_f10_musgs.IHistClm50Bgc.izumi_pgi.clm-monthly
SMS_P48x1_D_Ld5.f10_f10_musgs.I2000Clm50Cn.izumi_pgi.clm-default
```

@billsacks @wwieder and @dlawrenncar I could do more work to track down differences in BGC for PGI. But, I feel like maybe what I've done so far is enough. PGI is a secondary compiler that we sometimes have trouble with. The other compilers are showing expected results: intel, gnu, and nag. As it is I plan on taking out some of the changes to get PGI results identical since I can show it's only a roundoff change due to change of operations. On the other hand -- I'm pretty sure the changes must be in the file CNVegStructUpdateMod.F90, so it might not be hard to track down. Although the FATES change makes no sense to me.

billsacks:
@ekluzek I'm not sure I completely understand what you're saying, but I think you're saying that you and/or @swensosc were able to put in some temporary changes that give you confidence that the differences are only roundoff-level, due to something like changing order of operations. If so, then I think it's fine to bring this in: I sometimes do things where I do things like adding some temporary code to verify that the before and after terms in some expression differ by no more than roundoff, then temporarily revert to the old term in order to get bit-for-bit answers... I think you're saying you did something vaguely similar to that.

But I don't understand what you feel the status is of the FATES difference.

ekluzek:
OK, to make sure this is correct, I'm going to create a small roundoff answer changing tag that brings in some of these changes as a precursor to this tag. This tag should then be identical to that one for non CLM5.1 physics for all compilers. I'll temporarily add in the comparison that @billsacks suggested to ensure it is roundoff. This will give me confidence in answers for both the clm5.1 version (with BHS on) and non clm5.1 version (BHS off). This will also be good, because I'm adding a couple extra tests and compsets in, and it will be good to have a baseline version that includes it.

ekluzek:
We should use the shr_const_mod variable SHR_CONST_CPFW for c_water. Looks like right now it's set to the same value.

src/biogeochem/CNVegStructUpdateMod.F90
ekluzek: OK, this has been done.

ekluzek:
The Spinup test SSP_D_Ld10.f19_g17.I1850Clm51Bgc.cheyenne_intel.clm-rtmColdSSP is failing in the first step of running spinup_state=2 (AD mode) because of an energy balance error. This is after the normal skipping the balance check for the first two steps...

For example...

> 1447: global gridcell index =         2852
> 1073: errseb         =  -4.335382911602181E-004
> 1073: clm model is stopping - error is greater than 1e-5 (W/m2)
> 1073: sabv           =   0.000000000000000E+000
> 1073: sabg           =   0.000000000000000E+000  0.000000000000000E+000
> 1073:  0.000000000000000E+000
> 1073: forc_tot      =   0.000000000000000E+000
> 1073: eflx_lwrad_net =    35.6740742249761
> 1073: eflx_sh_tot    =   -4.21088834421538
> 1073: eflx_lh_tot    =   0.824540927653914
> 1023: global landunit index =         5642
> 415: ERROR: ERROR in BalanceCheckMod.F90 at line 713

ekluzek:
Fates with BHS is currently failing in a smoke test due to balance check energy error...

SMS_D_Ld5.f45_f45_mg37.I2000Clm51Fates.cheyenne_intel.clm-FatesColdDef

Single point tests seem to work OK, but I get the balance check fail with: SSP_D_Ld10_Mmpi-serial.5x5_amazon.I2000Clm51Bgc.cheyenne_intel.clm-rtmColdSSP

lnd.log reports the following...

>  nstep =          2   TS =   296.400310574612320
>  energy balance in canopy           73 , err= -0.154427559099435     
>  WARNING: BalanceCheck: surface flux energy balance error (W/m2)
>  nstep          =            3
>  errseb         =  -0.154427559099688     
>  clm model is stopping - error is greater than 1e-5 (W/m2)
>  sabv           =   0.000000000000000E+000
>  sabg           =   0.000000000000000E+000  0.000000000000000E+000
>   0.000000000000000E+000
>  forc_tot      =   0.000000000000000E+000
>  eflx_lwrad_net =    20.3566634829095     
>  eflx_sh_tot    =   -12.5507874961278     
>  eflx_lh_tot    =   -3.19457626961661     
>  eflx_soil_grnd =   -82.9190400804914     
>  dhsdt_canopy   =    78.4621679224260     
>  fsa fsr =   0.000000000000000E+000  0.000000000000000E+000
>  fabd fabi =   0.000000000000000E+000  0.000000000000000E+000
>   0.000000000000000E+000  0.000000000000000E+000
>  albd albi =    1.00000000000000        1.00000000000000     
>    1.00000000000000        1.00000000000000     
>  ftii ftdd ftid =   0.000000000000000E+000  0.000000000000000E+000
>   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
>   0.000000000000000E+000
>  elai esai =    2.96125454231256       0.500000000000000     
>  clm model is stopping
>  local  patch      index =           73
>  global patch      index =           99
>  global column   index =           38
>  global landunit index =           29
>  global gridcell index =           12
>  gridcell longitude    =    291.500000000000     
>  gridcell latitude     =   -9.50000000000000     
>  pft      type         =            6
>  column   type         =            1
>  landunit type         =            1

Since, sabv, sabg, and forc_tot are zero. That means the sum of: eflx_lwrad_net, eflx_sh_tot, eflx_lh_tot, eflx_soil_grnd, and dhsdt_canopy should also sum to zero.

ekluzek:
@swensosc to run a standalone test you just give the test name after "./create_test ". There's a few other options I usually give to change the wallclock time and queue. I also typically tell it to do "-r ." to create it's case directory under the current directory rather than on scratch.

So I'll do this sort of thing

`qcmd -- ./create_test SMS_D_Ld5.f45_f45_mg37.I2000Clm51Fates.cheyenne_intel.clm-FatesColdDef -r . --queue premium --walltime 00:20:00 &`

I'll look into the BgcCrop issue, it would be good for you to look into the FATES cold-start issue. I'll look to see if we have other cold start tests as well.

ekluzek:
In our discussion today at noon we thought that taper should be on the parameter file. And we should have BGC use taper and nstem from the parameter file even when BHS is off. At this point it shouldn't be hard to do that for both. Since, nstem is used both for BGC and BHS it's possible that tuning for BHS on and off could have different values for it. But, we will ignore that difference for now.

ekluzek:
OK, it turns out my SSP spinup tests that are failing are NOT coldstarts. They are labeled as rtmColdSSP which is certainly a bad label because they aren't RTM anymore and they aren't coldstarts right now. So I'll file a separate issue about that. They are starting up from initial conditions that are vastly different than the resolution run, so they aren't really "well spunup", but they are NOT starting from scratch.

ekluzek:
Besides all of the Fates tests, the other normally run cold start tests are:

  ``` xml
<test name="ERP_D_Ld5" grid="f10_f10_musgs" compset="IHistClm50BgcCrop" testmods="clm/allActive">
  <test name="SMS_D_Ld9" grid="f09_g17" compset="I1850Clm50BgcNoAnthro" testmods="clm/decStart1851_noinitial">
  <test name="ERS_D_Ld7_Mmpi-serial" grid="1x1_smallvilleIA" compset="IHistClm50BgcCropRs" testmods="clm/decStart1851_noinitial">
  <test name="ERS_Lm20_Mmpi-serial" grid="1x1_smallvilleIA" compset="I2000Clm50BgcCropQianRs" testmods="clm/monthly_noinitial">
  <test name="SMS_D_Ld1_Mmpi-serial" grid="f45_f45_mg37" compset="I2000Clm50Sp" testmods="clm/ptsRLA">
  <test name="SMS_Ld1_Mmpi-serial" grid="f45_f45_mg37" compset="I2000Clm50Sp" testmods="clm/ptsRLA">
  <test name="SMS_Ld1_Mmpi-serial" grid="f45_f45_mg37" compset="I2000Clm50Sp" testmods="clm/ptsRLB">
  <test name="SMS_Ld1_Mmpi-serial" grid="f45_f45_mg37" compset="I2000Clm50Sp" testmods="clm/ptsROA">
  <test name="ERP_D_P36x2_Ld3" grid="f10_f10_musgs" compset="I2000Clm51BgcCrop" testmods="clm/coldStart">
  <test name="LWISO_Ld10" grid="f10_f10_musgs" compset="I2000Clm50BgcCrop" testmods="clm/coldStart">
```

So there's one that check BHS for a coldstart...

ERP_D_P36x2_Ld3.f10_f10_musgs.I2000Clm51BgcCrop.cheyenne_intel.clm-coldStart

and that has been passing.

File: src/biogeophys/CanopyFluxesMod.F90 [hunk 18]
```diff
@@ -634,15 +678,104 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
-      cp_leaf              = 0._r8
-      lw_stem              = 0._r8
-      sa_internal          = 0._r8
-      dt_stem              = 0._r8
-      tl_ini               = 0._r8
-      ts_ini               = 0._r8
+      !
+      ! Calculate biomass heat capacities
+      !
+      if(use_biomass_heat_storage) then
```
rgknox, line 687: My impression is that for this to work with FATES, we would split the rest of this clause for use_fates and use_cn, where each has its own subroutine that returns patch level 
sa_leaf(p)  frac_rad_abs_by_stem(p)  sa_stem(p) sa_internal(p) cp_leaf(p)  cp_stem(p) rstem(p)

And the subroutine for the fates side would probably live in clmfates_interfaceMod.F90

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1207,26 +1381,47 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
-                - lw_leaf + lw_stem - eflx_sh_veg(p) - hvap*qflx_evap_veg(p) &
-                - ((t_veg(p)-tl_ini)*cp_leaf/dtime)
+                - lw_leaf(p) + lw_stem(p) - eflx_sh_veg(p) - hvap*qflx_evap_veg(p) &
+                - ((t_veg(p)-tl_ini(p))*cp_leaf(p)/dtime)
+
+         !  Update stem temperature; adjust outgoing longwave
+         !  does not account for changes in SH or internal LW,
+         !  as that would change result for t_veg above
+         if (use_biomass_heat_storage) then
+            if (stem_biomass(p) > 0._r8) then
```
rgknox, line 1393: for compatability with FATES, we would also have to have separate code here

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -634,15 +678,104 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+            ! as weighted averaged of dry biomass and water
+            ! lma_dry has units of kg dry mass/m2 here
+            ! (Appendix B of Bonan et al., GMD, 2018) 
+
+            cp_leaf(p)  = leaf_biomass(p) * (c_dry_biomass*(1.-fbw(patch%itype(p))) + (fbw(patch%itype(p)))*c_water)
+
+            ! cp-stem will have units J/k/ground_area
+            cp_stem(p) = stem_biomass(p) * (c_dry_biomass*(1.-fbw(patch%itype(p))) + (fbw(patch%itype(p)))*c_water)
+            ! adjust for departure from cylindrical stem model
+            cp_stem(p) = k_cyl_vol * cp_stem(p)
```
rgknox, line 759: @swensosc I wasn't expecting this conversion factor for the heat capacity calculation.  It seemed more of a geometry thing (affecting fluxes) and not a mass thing (affecting heat capacity).  Is it a moot point because this term propogates to the desired calculation either way?

swensosc:
are you referring to this:

cp_stem(p) = k_cyl_vol * cp_stem(p)

src/biogeophys/CanopyFluxesMod.F90
ekluzek: No, really we just want FATES to provide stem_biomass and leaf_biomass. There's an if inside this section you outline for either BGC or SP that will possibly need a use_fates option to provide those two. 

It's actually probably better if those two were provided for them outside of this so that CanopyFluxes doesn't have to know if FATES or BGC or SP are running, it just uses the biomass provided into it.

I think this should be a simple refactoring we do later.

I've now got an issue for this here: #1247

ekluzek: I'm going to resolve this since I've moved it to the issue.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: I don't actually think any of these terms will be different in the FATES case. Most of these terms are from the atmosphere or surface or biomass heating terms. Getting the biomass for stem and leaf are the things that we need from FATES.

ekluzek: Because I opened #1247. I'm resolving this conversation.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: @rgknox @swensosc responded below. But, I'm putting his answer here so to connect to this conversation.

> are you referring to this:
> 
> cp_stem(p) = k_cyl_vol * cp_stem(p)

ekluzek: Note currently k_cyl_vol is 1, so it doesn't change it. The description is that this is departure from having a perfect cylindrical stem.

ekluzek: Resolving this conversation so this PR can be resolved.

ekluzek: Resolving this so the discussion should go in #1247.

Issue #315, opened 2018-03-06
Perform gridcell-level ch4 balance check bracketing the entire run loop


billsacks:
As for carbon and nitrogen, we should also have a gridcell-level ch4 balance check that brackets the entire run loop, including dynamic landunits area changes.

billsacks:
Tagging this high priority along with #314 to avoid further problems like #675 .

PR #1022, opened 2020-05-26
Gridcell-level error check for methane (CH4)

slevisconsulting:
### Description of changes

Bracket the model time-step loop to calculate balance checks at the gridcell level, as explained in #315

### Specific notes

Avoid duplication by using calculations from the column-level checks.

@billsacks 

CTSM Issues Fixed (include github issue #): #315 

No

No

Testing performed, if any:
The following tests PASS bit-for-bit against dev097:
ERI_D_Ld9.f10_f10_musgs.I1850Clm50Bgc.cheyenne_gnu.clm-default
ERS_Ly3_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianGs.cheyenne_gnu.clm-cropMonthOutput
ERP_D.f10_f10_musgs.IHistClm50Bgc.cheyenne_gnu.clm-decStart

### DIFF
--- a/src/biogeochem/ch4Mod.F90
+++ b/src/biogeochem/ch4Mod.F90
@@ -190,7 +190,7 @@ module ch4Mod
      ! false. This could be a scalar, but scalars cause problems with threading, so we use
      ! a column-level array (column-level for convenience, because it is referenced in
      ! column-level loops).
-     logical , pointer, private :: ch4_first_time_col         (:)   ! col whether this is the first time step that includes ch4
+     logical , pointer, private :: ch4_first_time_grc         (:)   ! grc whether this is the first time step that includes ch4
      !
      real(r8), pointer, public :: finundated_col             (:)   ! col fractional inundated area (excluding dedicated wetland cols)
      real(r8), pointer, public :: finundated_pre_snow_col    (:)   ! col fractional inundated area (excluding dedicated wetland cols) before snow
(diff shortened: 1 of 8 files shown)


src/biogeochem/ch4Mod.F90
```diff
@@ -2174,22 +2236,22 @@ subroutine ch4 (bounds, num_soilc, filter_soilc, num_lakec, filter_lakec, &
          do fc = 1, num_lakec
             c = filter_lakec(fc)
 
-            if (.not. ch4_first_time(c)) then
+            if (.not. ch4_first_time_col(c)) then
```
slevisconsulting, line 2239: Note that my four tests PASS without a similar if statement in the gridcell-level check (next do loop).

@billsacks I feel this is ready for review and then I will run the full test suite.

src/biogeochem/ch4Mod.F90
```diff
@@ -1554,13 +1560,61 @@ subroutine readParams ( ncid )
...
+    type(ch4_type)   , intent(inout) :: ch4_inst
+    !
+    ! !LOCAL VARIABLES:
+
+    character(len=*), parameter       :: subname = 'ch4_init_gridcell_balance_check'
+    !-----------------------------------------------------------------------
+
+    ! Initialize to zero for columns outside the filters because will
+    ! average up to gridcell
+    ch4_inst%totcolch4_bef_col(bounds%begc:bounds%endc) = 0._r8
```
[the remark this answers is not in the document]
slevisconsulting: Removing this works. To explain though, I preemptively inserted this and the other initialization because I was burned in #984 debugging for what ended up being partly undefined column variables entering the c2g subroutine.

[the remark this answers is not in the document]
billsacks: Makes sense. And truly, there isn't really a "right" answer here: the tradeoff between performance and being extra-safe is rarely a clear one.

src/biogeochem/ch4Mod.F90
slevisconsulting: From your wording I understand that I may not need both the col and grc level variables if I have the grc level variable. That doesn't seem right to me (though I could easily be wrong), so I made a grc level copy of the col level variable for use in the gridcell level balance check.

slevisconsulting: Oh, now I caught what you meant by "subgrid level is arbitrary."

slevisconsulting:
The test suite on cheyenne has returned only one unexpected FAIL that I need to investigate:
ERP_D_Ld3.f09_g17.I2000Clm50SpGs.cheyenne_intel.clm-prescribed BASELINE ctsm1.0.dev097: DIFF

slevisconsulting:
Good news: The test that was failing now passes, so I'm resubmitting the full test suite.

slevisconsulting:
Test suite:
cheyenne PASS
izumi PASS

This is ready for a second review and I need to prep the ChangeLog.

slevisconsulting:
Test-suites ran successfully on cheyenne and izumi. All tests passed. I have updated the ChangeLog.

@billsacks this is ready for merging when it's convenient or appropriate.

Issue #803, opened 2019-09-12
Assumption that building width (Wb) equals street width can be relaxed and Wb can be derived from morphology


olyson:
### Brief summary of bug

There is an implicit assumption in the urban building energy model that building width equals street width.  However, this assumption can/should be relaxed and building width can be derived from the morphology dataset.

### General bug information

**CTSM version you are using:** release-clm

**Does this bug cause significantly incorrect results in the model's science?** Yes for urban (greater than roundoff), No for general climate simulations (same gridcell/regional/global climate).

**Configurations affected:** All configurations with urban landunits

### Details of bug

There is an implicit assumption in the urban building energy model that building width equals street width.  However, this assumption can be relaxed and building width can be derived from the Jackson morphology dataset.  Specifically, the equations which use H/Ws (building height over street width) should use H/Wb (building height over building width).  Building width can be derived from (Ws*froof)/(1-froof), where froof is the roof (building) fraction.
A simulation with the new equations indicate that as expected, for froof=0.5, answers are identical (because H/Ws = H/Wb).  Differences in other cases depend on the degree of departure from froof=0.5 and the presence and magnitude of space heating and air conditioning.  A full assessment of differences by variable, season, density type, and region using ctsm1.0.dev098 can be found in the attached file.  This uses the default CLM5 urban datasets.

[CTSM98_BUILDENERGY_Analysis_Pub.xlsx](https://github.com/ESCOMP/CTSM/files/4820477/CTSM98_BUILDENERGY_Analysis_Pub.xlsx)

A similar assessment using release-cesm2.0.1 but with the optional new urban datasets (Oleson and Feddema 2019) can be found in this attached file:

[CLM50_BUILDENERGY_Analysis_Pub.xlsx](https://github.com/ESCOMP/CTSM/files/4820475/CLM50_BUILDENERGY_Analysis_Pub.xlsx)

I have SourceMods for this currently, but can put together a branch.  It would be nice to get this into the release branch, in particular for the CESM2 large ensemble runs, since it's not going to affect climate and would make the urban output more useful.  Along with the new urban properties data (see [PR#591](https://github.com/ESCOMP/ctsm/pull/591)).

olyson:
I created a branch for the code changes:

https://github.com/olyson/ctsm/tree/fixbuildenergy

ekluzek:
[...]

PR #1034, opened 2020-06-08
Building energy fix (issue #803).

olyson:
### Description of changes
There is an implicit assumption in the urban building energy model that building width equals street width. However, this assumption has been relaxed and building width is now derived from the morphology dataset.

### Specific notes
There is an implicit assumption in the urban building energy model that building width equals street width. However, this assumption has been relaxed and building width is now derived from the Jackson morphology dataset. Specifically, the equations which use H/Ws (building height over street width) now use H/Wb (building height over building width). Building width is derived from (Ws*froof)/(1-froof), where froof is the roof (building) fraction.
A simulation with the new equations indicate that as expected, for froof=0.5, answers are identical (because H/Ws = H/Wb). Differences in other cases depend on the degree of departure from froof=0.5 and the presence and magnitude of space heating and air conditioning. 

CTSM Issues Fixed (include github issue #): #803

Are answers expected to change (and if so in what way)?  Yes, but only for urban landunits.  The effects on grid cell averages are very small as shown by the diagnostics from this year 2000 case:

http://webext.cgd.ucar.edu/I2000/clm50_ctsm10d098_1deg_GSWP3V1_CON_FIXBUILDENERGY_2000/lnd/clm50_ctsm10d098_1deg_GSWP3V1_CON_FIXBUILDENERGY_2000.1991_2010-clm50_ctsm10d098_1deg_GSWP3V1_CON_2000.1991_2010/setsIndex.html

Testing performed, if any:
The clm_short testlist was run.  All pass except comparison to ctsm1.0.dev098 baseline was not bfb as expected.

### DIFF
--- a/src/biogeophys/UrbBuildTempOleson2015Mod.F90
+++ b/src/biogeophys/UrbBuildTempOleson2015Mod.F90
@@ -229,6 +229,7 @@ contains
     integer, parameter :: neq = 5          ! number of equation/unknowns
     integer  :: fc,fl,c,l                  ! indices
     real(r8) :: dtime                      ! land model time step (s)
+    real(r8) :: building_hwr(bounds%begl:bounds%endl)      ! building height to building width ratio (-)
     real(r8) :: t_roof_inner_bef(bounds%begl:bounds%endl)  ! roof inside surface temperature at previous time step (K)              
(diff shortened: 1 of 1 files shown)


ekluzek:
I appended this onto the luna-bug fix branch tag and ran the testing. All tests have passed as expected, and the clm4_5 answers don't change as expected.

Issue #290, opened 2018-02-13
flexibleCN and luna code cleanup


bandre-ucar:
Migration of trello card. Some of this may have already been done or no longer relevant.

# TODO list

- [ ] move flexibleCN nitrogen parameters to namelist or netcdf
- [ ] remove excessive control parameters - candidates: carbon_storage_excess_opt (remove, old code?), carbon_excess_opt (remove, old code?), dynamic_plant_alloc_opt(remove, old code?)
- [ ] update metadata (long name, description) for netcdf parameters.
- [ ] duplicate code cleanup between NutrientCompetition modules.
- [ ] luna derived type - see email below
- [ ] Proper indentation (such as in CNGapMortalityMod.F90 and CNPhenologyMod.F90 )
- [ ] Use veg indices rather than magic numbers (such as 11 and 14 used in CNGapMortalityMod.F90 should be replaced by nbrdlf_dcd_brl_shrub and nc4_grass
- [ ] Set oneatm to SHR_CONST_PSTD in clm_varcon.F90
- [ ] LUNA parameters on file? forc_pbot_ref can come from SHR_CONST_PSTD, lots of magic numbers in LunaMod.F90
- [ ] Formatting, headers, etc. in LunaMod.F90, make default private
- [ ] FlexibleCN magic numbers. Some could be parameters or on params file
- [ ] FlexibleCN formatting and headers etc. NutrientCompetitionFlexibleCNMod.F90
- [ ] r120 added over a dozen fields to default history files, do they really need to be there? (they were default inactive previously)
- [ ] need names and units for variables added to params file.
- [ ] Reconsider lnc_opt. Make more clear. Think about FUN=on,flexCN=off,Luna=on case -- does that work and valid?

# Note From Dave Lawrence
```
These are notes that Charlie took based on meeting with Bardan. Could be helpful during the code cleanup.

notes, meeting with bardan jan. 26 2016

switches:

NutrientCompetitionFlexibleCNMod.F90

carbon_resp_opt: if C:N is too high, burn off excess or not.
downreg_opt: main flexcn switch. should really always be false when flexcn is activated, so we should remove thoe blocks of code in NutrientCompetitionFlexibleCNMod where that is not the case.
CN_residual_opt: only needed when cn_partition_opt is zero; if 1 then what do you do with excess residual nitrogen
CN_partition_opt: two options
CN_partition_opt == 0 is give all the plant parts other than the leaf N, and then give leaf the remaining N
[...]

PR #1039, opened 2020-06-13
Move some hardcoded parameters to the paramsfile

### LINKED ISSUE #138: Use tau_* values on params file
**Erik Kluzek < erik > - 2016-06-10 14:30:46 -0600**
**Bugzilla Id:** [2320](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2320)
**Bugzilla CC:**  cdkoven,  dlawren,  rfisher, 

The "tau_*" values are currently using the hardcoded values in SoilBiogeochemDecompCascadeBGCMod.F90 rather than the values on the params file. The values on the params file are read in -- but discarded. The issue seems to be that answers change when you read in from the params file, rather than doing the inverse that the code has. So if the values read in are the inverse then you can do the divide in the code and preserve answers.

      ! the belowground parameters from century
      tau_l1 = 1./18.5
      tau_l2_l3 = 1./4.9
      tau_s1 = 1./7.3
      tau_s2 = 1./0.2
      tau_s3 = 1./.0045

      ! century leaves wood decomposition rates open, within range of 0 - 0.5 yr^-1
      tau_cwd  = 1./0.3

      ! Todo:  FIX(SPM,032414) - the explicit divide gives different results than when that
      ! value is placed in the parameters netcdf file.  To get bfb, keep the 
      ! divide in source.

      !tau_l1 = params_inst%tau_l1_bgc
      !tau_l2_l3 = params_inst%tau_l2_l3_bgc
      !tau_s1 = params_inst%tau_s1_bgc
      !tau_s2 = params_inst%tau_s2_bgc
      !tau_s3 = params_inst%tau_s3_bgc

      !set turnover rate of coarse woody debris
      !tau_cwd = params_inst%tau_cwd_bgc

There are similar issues in SoilBiogeochemLittVertTranspMod.F90 and in biogeochem/ch4Mod.F90. They are documented with the following type of comment in the code...

! FIX(FIX(SPM,032414),032414) can't be read off of param file.  not bfb since it is a divide

So this means there are several variables on the params file that get read in and then ignored.

### LINKED ISSUE #206: Add prh30 to params file for fire tuning
_From @ekluzek on May 23, 2017 17:57_

Fang would like a new constant for tuning the CLM50 fire model. This constant only applies to Li2016 model and not the older version. So this acts a little bit different than other constants which are shared between both. So we may need to add a namelist read to just the newer version.

Here's the email from Fang which defines what to do...

Hi Dave and Erik,

     It would be better if Erik could help change (in CNFireLi2016Mod.F90)

arh30=1._r8-max(0.7_r8, min(1._r8,rh30_col(c)/90._r8))

to

arh30=1._r8-max(prh30, min(1._r8,rh30_col(c)/90._r8)) and define prh30=0.7_r8

In this way, we could tune the parameter (prh30) by just changing user_nl_clm if necessary, like ufuel...

    Dave, low burned area in tropical deciduous tree region (casued by very high BTRAN2 and fuel load in r229) could be solved better if I set prh30 to 0._r8 or 0.4_r8, but I'm not sure it's performance in coupling case, CLM5's later version, or newer forcing data, so I didn't suggest change in prh30 value this time.

Thanks,

Fang

_Copied from original issue: NCAR/CLM#8_

### LINKED ISSUE #995: Soil turnover times on parameter file and hard coded?
Here it looks like the soil turnover times are being read from the parameter file in `DecompCascadeBGCreadNML`
https://github.com/ESCOMP/CTSM/blob/605ee9dc7c61219498f50d25bde29decc87bc4f2/src/soilbiogeochem/SoilBiogeochemDecompCascadeBGCMod.F90#L183

but then they are hard coded in `decomp_rate_constants_bgc`
https://github.com/ESCOMP/CTSM/blob/605ee9dc7c61219498f50d25bde29decc87bc4f2/src/soilbiogeochem/SoilBiogeochemDecompCascadeBGCMod.F90#L730

Values here look the same, but this may complicated the PPE @dlawrenncar wants to do?

ekluzek:
### Description of changes

Change some hardcoded parameters to go on the parameter files. This is needed in preparation of running the Perturbed Parameter Ensemble.

### Specific notes

Contributors other than yourself, if any: This is primarily work of @olyson

CTSM Issues Fixed (include github issue #): #995 #138, #206

Fixes #995
Fixes #138
Fixes #206
Some work on #290

Are answers expected to change (and if so in what way)? No, expected to be bit-for-bit

Any User Interface Changes (namelist or namelist defaults changes)? New params files
   for both clm4_5 and clm5_0

Testing performed, if any: Ran test suite on cheyenne for gnu and intel


ekluzek:
In running testing on cheyenne, these two tests show difference in answers...

ERS_Lm20_Mmpi-serial.1x1_smallvilleIA.I2000Clm50BgcCropQianGs.cheyenne_gnu.clm-monthly_noinitial
ERS_Ly5_Mmpi-serial.1x1_smallvilleIA.I2000Clm50BgcCropQianGs.cheyenne_gnu.clm-ciso_monthly

Tests of five years or more that passed...
			
ERS_Ly5_P144x1.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-cropMonthOutput			
ERS_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianGs.cheyenne_intel.clm-cropMonthOutput			
SMS_D_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm45BgcCropQianGs.cheyenne_intel.clm-cropMonthOutput

olyson:
That's interesting.  I ran the aux_clm test suite with gnu and the only failure was an expected one.  However, those two tests that failed for you don't appear to be part of the aux_clm test suite?

ekluzek:
@olyson this is after the update to ctsm1.0.dev100, and those two tests are part of the test suite there. They might not have been in the older ctsm1.0.dev089 version the branch came from.

ekluzek:
OK, I looked through to figure this out. In the older baseline version ctsm1.0.dev089 these tests don't complete, but I was able to run one for 460 days. I ran that in ctsm1.0.dev089 and in a few of the olyson branch versions and they don't change answers.

I realized the problem might be due to PFT changes in the params file due to the miscanthus,switchgrass update. It made changes to the paramsfile to convert them from C3 to C4 types. When I fix the new paramsfile for this change on the newly added PFT variable of theta_cj. --  answers go back to being identical.

olyson:
That's great Erik, thanks for solving this.

ekluzek:
OK, this has passed testing on cheyenne and izumi, and I'll update the Change files and bring it to master.

PR #1040, opened 2020-06-15
CTSM changes to enable transient FATES land use

### LINKED ISSUE #1002: allow transient land-use files when FATES is on?
Hi All,
I'd like to start working towards some implementation of linking up the land use datasets with FATES.  A first requirement is that we allow that to happen in the scripting infrastructure.  Currently its forbidden by https://github.com/ESCOMP/CTSM/blob/master/bld/CLMBuildNamelist.pm#L2205.  Does anyone object to the removal of those lines of code, at least in the fates_next_api branch?

ckoven:
### Description of changes
These changes allow the passing of harvest rates to FATES, thus enabling FATES to work with transient land use.  This represents the CTSM changes that correspond with the changes made in FATES PR https://github.com/NGEET/fates/pull/663.

Specifically, to make thus work, I added a new subroutine in src/biogeochem/dynHarvestMod.F90 that, differs from the existing code in two important ways.  (1) is that, rather than updating the variable `harvest` that is global to that module, it passes an intent-out variable with the harvest information.  (2) is that, instead of summing all of the harvest rate fields (primary forest, primary non-forest, secondary young forest, secondary old forest, secondary non-forest), it keeps them separate and lets FATES deal with them separately.  It then passes this information to the FATES interface as a fates input boundary condition.

### Specific notes

collaborated with @aldivi on this.

fixes #1002 

This should only change answers when fates is activated and harvest is turned on.

No, though the code now applies do_harvest logic for fates if that is enabled.

Testing performed, if any:
I've tested this code over California to show that the basic functionality works, but more testing and in particular, global transient historical simulations, are still needed.
I've not yet run the test suite for this.


bld/CLMBuildNamelist.pm
```diff
@@ -2682,12 +2679,13 @@ sub setup_logic_do_harvest {
       if (string_is_undef_or_empty($nl->get_value('flanduse_timeseries'))) {
          $cannot_be_true = "$var can only be set to true when running a transient case (flanduse_timeseries non-blank)";
       }
-      elsif (!&value_is_true($nl->get_value('use_cn'))) {
-         $cannot_be_true = "$var can only be set to true when running with CN (use_cn = true)";
-      }
-      elsif (&value_is_true($nl->get_value('use_fates'))) {
-         $cannot_be_true = "$var currently doesn't work with ED";
-      }
+      #cdk disabling this for now, but probably there is a better way to do it.
```
ekluzek, line 2682: This should be setup so that if use_cn AND use_fates are both false cannot_be_true would then be set.

src/biogeochem/dynHarvestMod.F90
```diff
@@ -173,6 +180,61 @@ subroutine dynHarvest_interp(bounds)
...
+    logical          , intent(out):: after_start_of_harvest_ts
+    !
+    ! !LOCAL VARIABLES:
+    integer               :: varnum       ! counter for harvest variables
+    real(r8), allocatable :: this_data(:) ! data for a single harvest variable
+
+    character(len=*), parameter :: subname = 'dynHarvest_interp'
+    !-----------------------------------------------------------------------
+
+    !!! NOTE I AM NOT USING THE BOUNDS_LEVEL_PROC ASSERTION HERE BECAUSE OF WHERE IN THE CALLING
```
ekluzek, line 214: bounds is going to be whatever type of bounds that was sent into it. The caller needs to determine what type of bounds is sent down. I would remove the language that specifies what type of bounds this is. It looks like the caller is actually a clump level bounds, but this routine doesn't need to know what type it is.

bld/CLMBuildNamelist.pm
ckoven: I think this is fixed now, @ekluzek please check to confirm, but otherwise I'll mark as resolved?

src/biogeochem/dynHarvestMod.F90
ckoven: this one too?

aldivi: This function does require specific proc-level bounds, otherwise it will not work correctly because the harvest bounds are at a different level than the clumping, site, column, etc. The caller should currently be passing these bounds correctly, the comment in the type declaration line notes the required bounds from the caller. I am not sure about the line starting with "!!!..." as I think it was there from before.

src/biogeochem/dynHarvestMod.F90
```diff
@@ -43,16 +45,18 @@ module dynHarvestMod
...
+  character(len=64), parameter, public :: harvest_varnames(num_harvest_inst) = &
        [character(len=64) :: 'HARVEST_VH1', 'HARVEST_VH2', 'HARVEST_SH1', 'HARVEST_SH2', 'HARVEST_SH3']
   
   type(dyn_var_time_uninterp_type) :: harvest_inst(num_harvest_inst)   ! value of each harvest variable
 
   real(r8) , allocatable   :: harvest(:) ! harvest rates
-  logical                  :: do_harvest ! whether we're in a period when we should do harvest
-  character(len=*), parameter :: string_not_set = "not_set"  ! string to initialize with to indicate string wasn't set
-  character(len=64)        :: harvest_units = string_not_set ! units from harvest variables 
+  logical, public, protected   :: do_harvest ! whether we're in a period when we should do harvest
```
ekluzek, line 55: This should remain private, as it's only accessed through get_do_harvest outside of here.

src/biogeochem/dynHarvestMod.F90
```diff
@@ -43,16 +45,18 @@ module dynHarvestMod
...
+  character(len=*),  parameter :: string_not_set = "not_set"  ! string to initialize with to indicate string wasn't set
```
ekluzek, line 56: This is default private, but would be good to declare it that way.

Actually it would be good to divide the public from the private data, just to make it more clear what is what, and start each section with a comment on if it's public or private.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2667,4 +2712,61 @@ subroutine hlm_bounds_to_fates_bounds(hlm, fates)
    
  end subroutine hlm_bounds_to_fates_bounds
 
+ ! ======================================================================================
+
+ subroutine GetAndSetTime()
+
+   ! CLM MODULES
```
ekluzek, line 2719: I'd like a little explanation for what this is doing. It seems to be figuring out the day of the year in order for fates to understand it. I'm a little confused as to it setting the Fates time as I would think the time-stepping determines that.  But, maybe this is just temporarily setting it and then it will be reset later?

src/biogeochem/dynHarvestMod.F90
```diff
@@ -23,13 +23,15 @@ module dynHarvestMod
...
   use PatchType               , only : patch                
+  use clm_varctl              , only : use_fates
   !
   ! !PUBLIC MEMBER FUNCTIONS:
   implicit none
   private
   !
   public :: dynHarvest_init    ! initialize data structures for harvest information
   public :: dynHarvest_interp  ! get harvest data for current time step, if needed
+  public :: dynHarvest_interp_resolve_harvesttypes  ! get harvest data for current time step, if needed, harvest-type-resolved
```
ekluzek, line 34: Add a comment that this is only called for FATES, and CNHarvest is only called for use_cn and without FATES

src/biogeochem/dynHarvestMod.F90
```diff
@@ -23,13 +23,15 @@ module dynHarvestMod
   use clm_varcon              , only : grlnd
```
ekluzek, line 23: I actually think that this should be moved out of biogeochem and placed in dyn_subgrid. As it now is both dyn_subgrid infrastructure and not just biogeochem.

src/utils/clmfates_interfaceMod.F90
glemieux: @ekluzek This is part of reinstating the uninitialized hlm_model_day fix https://github.com/ESCOMP/CTSM/pull/820 that was accidentally lost during the update to release-clm5.0.30 a while back.  The subroutine creation was an effort to modularize the calls to getting and setting time for fates.  The code was repeated three times in the `init`, `dynamic_driv` and `restart` subroutines.  

Reinstating this fix got pushed up the queue as I had to test on Lawrencium while Cheyenne was down and Charlie suggested just rolling the fix into his PR.

src/biogeochem/dynHarvestMod.F90
ckoven: @ekluzek ok I moved the module to dyn_subgrid.

src/biogeochem/dynHarvestMod.F90
ckoven: added comments on this.

ckoven:
Hi @ekluzek thanks for the comments.  I think I've addressed them, let us know if it is ok to merge now.

src/utils/clmfates_interfaceMod.F90
ekluzek: Oh, I see this was something unrelated that was lost. And it's just work at putting stuff into a subroutine, so it doesn't relate to the rest of the PR, and isn't anything specific for time for the harvesting.

I still like the idea of having a short description of what this is. But, that's the only thing I suggest then.

glemieux:
Re-ran fates suite tests.  Results PASSed as expected:
```
/glade/u/home/glemieux/scratch/clmed-tests/fates-sci.1.40.0_api.13.0.0-clm5.0.30-C2922c98b-Fffaa76e2.fates.cheyenne.intel
/glade/u/home/glemieux/scratch/clmed-tests/fates-sci.1.40.0_api.13.0.0-clm5.0.30-C2922c98b-Fffaa76e2.fates.cheyenne.gnu
```

PR #1068, opened 2020-07-02
Add LILAC!

billsacks:
### Description of changes

Add LILAC!

### Specific notes

Contributors other than yourself, if any: @mvertens @negin513 @jhamman @slevisconsulting @barlage @dlawrenncar 

CTSM Issues Fixed (include github issue #): A number of issues that have already been closed (see https://github.com/ESCOMP/CTSM/projects/23)

Any User Interface Changes (namelist or namelist defaults changes)? None (other than LILAC)

Testing performed, if any:
Full aux_clm test suite: passes


billsacks:
@ekluzek I just looked back over the diffs that aren't LILAC-specific. Many of the Fortran changes relate to changes in the time manager:

1. No longer setting the stop time in model initialization – which I'm pretty sure we discussed a few months ago, though I'm not sure if you ever saw the actual code changes... anyway, they were pretty straightforward: mainly removing some code that wasn't accomplishing anything

2. Always getting dtime from the driver rather than from the namelist. Again, we discussed this, but I'm not sure if you ever saw the actual changes.

I don't think you need to look at these changes: I'm confident enough in them.

The one commit you **should** probably look at is 294f9e706 - there I made some changes to decrease the verbosity of build-namelist a bit, as well as using bsd:glob rather than glob (as per perl's deprecation messages).

ekluzek:
Awesome. This is great. I took a quick look over the things you mention. We did talk about them beforehand. I had already approved that commit that you mention on Valentines day.

billsacks:
Thanks a lot @ekluzek . I had this nagging feeling that you had approved that commit, but couldn't find a record of it – so thanks.

billsacks:
> I had this nagging feeling that you had approved that commit, but couldn't find a record of it – so thanks.

Oh hey, it's right there in the conversation attached to the commit! I forgot to check that :-)

billsacks:
Good news!: I finally have a LILAC smoke test working. For example: `LILACSMOKE_Vnuopc_P4x1_D_Ld2.f10_f10_musgs.I2000Ctsm50NwpSpAsRsGs.cheyenne_intel.clm-lilac`. Here is the test script, if anyone is interested:

https://github.com/ESCOMP/CTSM/blob/lilac_cap/cime_config/SystemTests/lilacsmoke.py

The highest priority things I still want to do here are:
- Do some manual testing to ensure this reports a failure at all points where it should fail
- Get baseline comparison & generation working (probably requires creating some sym links so that output files appear in the expected location)

Issue #43, opened 2017-12-16
dynamic landunits: handle methane with changing lake area


billsacks:
**Bill Sacks < sacks@ucar.edu > - 2016-02-16 13:38:10 -0700**
**Bugzilla Id:** [2283](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2283)
**Bugzilla CC:**  erik@ucar.edu,  oleson@ucar.edu,  rfisher@ucar.edu, 

The methane code (ch4Mod.F90) implicitly assumes that finundated is 1 for lakes. Thus, if lake area increases, the right thing to do is probably to immediately inundate any land it took over, with code similar to what happens right now when inundated area increases.

Decreases in lake area should be handled appropriately now, but increases in lake area are not. Currently, there is no way for lakes to increase in area, but this will need to be revisited once they can.

billsacks:
When trying to run a dynamic lakes BGC case (specifically, `ERS_Lm25.1x1_smallvilleIA.IHistClm50BgcCropQianRsGs.cheyenne_gnu.clm-smallville_dynlakes_monthly` from 3ab74de5), I am getting a gridcell methane conservation issue. I haven't dug into it, but I'm thinking it may be related to this issue.

billsacks:
Now that dynamic lakes are nearly on master, I'm changing this from an enhancement to a bug.

slevisconsulting:
This issue describes a ch4 balance error with transient lakes. 
I see a ch4 balance error in #1868 but:
- I have yet to confirm that my problem is due to transient lakes and
- My balance error appears in hybrid/startup cases (e.g. 1901, 1920, 1921) though not in 1850 where the whole simulation started and not in restart/branch cases.

billsacks:
@slevisconsulting pointed out that we have this current workaround in place to bypass the gridcell-level methane checks on the first time step of each year when using dynamic lakes:

https://github.com/ESCOMP/CTSM/blob/616905bbbac55f506fccc5aef8c45bbbdbba94a2/src/biogeochem/ch4Mod.F90#L2330-L2332

together with some code that bypasses the ch4 balance checks if this flag is true.

As @slevisconsulting pointed out in https://github.com/ESCOMP/CTSM/discussions/1868#discussioncomment-3901132 it's likely that there is a problem in time step 0 of a hybrid case that may need a similar workaround to the one above.

ekluzek:
[...]

PR #1073, opened 2020-07-06
Dynamical lakes and new lake cover map 

Ivanderkelen:
Description of changes

Dynamic lakes representing reservoir construction and new lake cover input map using HydroLAKES and GRanD data.

Update (2020-08-17): this will be split into two separate PRs; this PR will just contain tools changes, and a new PR (which will come to master first) will contain all other changes.

Specific notes

New namelist handle use_dynlakes is added, off by default

Testing performed, if any:

NOTE: Be sure to check your coding style against the standard
(https://github.com/ESCOMP/CTSM/wiki/List-of-common-problems).


Ivanderkelen:
Thanks, @billsacks for your feedback. I am now in the process of reviewing the diffs of this branch relative to master, and will notify you once this is finished and I got some testing done.

Ivanderkelen:
I found issue #43, which might interfere with the dynlakes branch (not investigated in detail).

Also note, these developments add lake water and heat to the total gridcell water and heat content, and consequently, TWS now includes lake water, as it didn't do before.  

When starting my simulations @billsacks had an first review and made the following comments:

(1) You accidentally added some files. Not a big deal, but something we'll want to clean up before this is merged in eventually:

$ git diff --name-status HEAD..Ivanderkelen/dynlakes | grep '^A' | grep -v dynlakeFileMod
A tools/mksurfdata_map/landuse_timeseries_hist_16pfts_Irrig_CMIP6_simyr1850-2015.txt
A tools/mksurfdata_map/landuse_timeseries_hist_dynlake_simyr1900-2009.txt
A tools/mksurfdata_map/surfdata_0.9x1.25_hist_16pfts_Irrig_CMIP6_simyr1850_c190722.namelist
A tools/mksurfdata_map/surfdata_1.9x2.5_hist_16pfts_Irrig_CMIP6_simyr1850_c190715.namelist

_This was done intentionally, as requested by Erik to be able to update the mksurfdat tool to also take the new lake fraction map I created._

(2) It looks like the code you added in surfrdMod – to read the haslake variable – will only work if there is actually a landuse_timeseries file present. Some runs don't have this, which would cause this code to crash. Not important to fix now, but we'll want to fix it eventually.

(3) In dynSubgridControl, you prevent running with dynlakes and fates. I think these would be compatible. (We can't currently run with both transient veg and fates, but I think it would work to run with transient lakes & fates.)

_This originally came from following the harvest example. I deleted the lines preventing it but didn't test it with fates._

billsacks:
> I found issue #43, which might interfere with the dynlakes branch (not investigated in detail).

Thanks for bringing that to our attention. My sense is that #43 should be resolved at some point, but it isn't critical for the first iteration of the dynamic lake work. It is likely needed for the correct operation of the methane code with dynamic lakes, though (which is active when running with Bgc compsets, not with Sp [satellite phenology]). If you have time to look into it at some point in the next few months, that would be great, but I don't think that should hold up bringing your initial changes to master.

billsacks:
> (1) You accidentally added some files. Not a big deal, but something we'll want to clean up before this is merged in eventually:
> 
> $ git diff --name-status HEAD..Ivanderkelen/dynlakes | grep '^A' | grep -v dynlakeFileMod
> A tools/mksurfdata_map/landuse_timeseries_hist_16pfts_Irrig_CMIP6_simyr1850-2015.txt
> A tools/mksurfdata_map/landuse_timeseries_hist_dynlake_simyr1900-2009.txt
> A tools/mksurfdata_map/surfdata_0.9x1.25_hist_16pfts_Irrig_CMIP6_simyr1850_c190722.namelist
> A tools/mksurfdata_map/surfdata_1.9x2.5_hist_16pfts_Irrig_CMIP6_simyr1850_c190715.namelist
> 
> _This was done intentionally, as requested by Erik to be able to update the mksurfdat tool to also take the new lake fraction map I created._

- [ ] @ekluzek can you please comment on whether you still want these files, and if so, whether they'll need to be updated in some way?

ekluzek:
> > (1) You accidentally added some files. Not a big deal, but something we'll want to clean up before this is merged in eventually:
> > $ git diff --name-status HEAD..Ivanderkelen/dynlakes | grep '^A' | grep -v dynlakeFileMod
> > A tools/mksurfdata_map/landuse_timeseries_hist_16pfts_Irrig_CMIP6_simyr1850-2015.txt
> > A tools/mksurfdata_map/landuse_timeseries_hist_dynlake_simyr1900-2009.txt
> > A tools/mksurfdata_map/surfdata_0.9x1.25_hist_16pfts_Irrig_CMIP6_simyr1850_c190722.namelist
> > A tools/mksurfdata_map/surfdata_1.9x2.5_hist_16pfts_Irrig_CMIP6_simyr1850_c190715.namelist
> > _This was done intentionally, as requested by Erik to be able to update the mksurfdat tool to also take the new lake fraction map I created._
> 
> * [ ]  @ekluzek can you please comment on whether you still want these files, and if so, whether they'll need to be updated in some way?
> 
> Actually, I see that the updated list of additions/deletions/modifications of similar files is:
> 
> ```
> A       tools/mksurfdata_map/landuse_timeseries_hist_16pfts_Irrig_CMIP6_simyr1850-2015.txt
> D       tools/mksurfdata_map/landuse_timeseries_hist_78pfts_simyr1850-2015.txt
> A       tools/mksurfdata_map/landuse_timeseries_hist_dynlake_simyr1900-2009.txt
> M       tools/mksurfdata_map/mksurfdata_map.namelist
> ```

We will want the sample namelist update. And possibly we want the sample "dynlake" text file. I assume it must have the sample format for adding dynamic lakes for each year. We will also want to get the dynlake option into the mksurfdata.pl script. Having the sample dynamic lake text file, should help in figuring out the changes for mksurfdata.pl.

tools/mksurfdata_map/mksurfdata_map.namelist
```diff
@@ -24,7 +24,8 @@
  map_fch4         = '/glade/p/cesm/cseg/inputdata/lnd/clm2/mappingdata/maps/10x15/map_360x720_cruncep_to_10x15_nomask_aave_da_c130326.nc'
  mksrf_fsoitex    = '/glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/mksrf_soitex.10level.c010119.nc'
  mksrf_forganic   = '/glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/mksrf_organic_10level_5x5min_ISRIC-WISE-NCSCD_nlev7_c120830.nc'
- mksrf_flakwat    = '/glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/mksrf_LakePnDepth_3x3min_simyr2004_csplk_c151015.nc'
+ mksrf_flakwat    = '/glade/u/home/ivanderk/clm5.0/own_inputfile/mksurf_LakePnDepth_3x3min_simyr2009_csplk_c190603.nc
+'
```
ekluzek, line 28: There's extra whitespace at the end of the filename and onto the next line, that should be removed.

tools/mksurfdata_map/src/mksurfdat.F90
```diff
@@ -1159,16 +1165,26 @@ program mksurfdat
                 call abort()
              end if
           end if
+          
+          
+          ! IV Read input lake pct data
+          read(nfdynlak, '(A195,1x,I4)', iostat=ier) string, year
```
ekluzek, line 1171: Read this in as year2, and then compare to year from above and abort if they aren't the same. This is similar to the abort for harvest above.

tools/mksurfdata_map/src/mksurfdat.F90
```diff
@@ -1127,6 +1130,9 @@ program mksurfdat
        ! Read in each dynamic pft landuse dataset
 
        nfdyn = getavu(); call opnfil (mksrf_fdynuse, nfdyn, 'f')
+       
+       ! IV read in dynamic lake dataset
+       nfdynlak = getavu(); call opnfil (mksrf_fdynlak, nfdynlak, 'f')
```
ekluzek, line 1135: What happens if the dynamic lake file isn't provided? We need a way for it to abort. We also should decide on if lakes should always be dynamic for transient conditions, or if we you pick and choose? It would probably be best to always have it on the datasets, but then decide in CTSM whether to activate it or not. We have something similar for harvest.

tools/mksurfdata_map/src/mkfileMod.F90
```diff
@@ -553,6 +553,10 @@ subroutine mkfile(domain, fname, harvdata, dynlanduse)
                long_name=mkharvest_longname(ind2D(j)), units=mkharvest_units(ind2D(j)) )
        end do
        deallocate(ind1D, ind2D)
+       
+       ! IV: add lake field in here?
```
ekluzek, line 557: Can remove the comment. It's clear it needs to be on the landuse.timeseries file.

billsacks:
> > ```
> > A       tools/mksurfdata_map/landuse_timeseries_hist_16pfts_Irrig_CMIP6_simyr1850-2015.txt
> > D       tools/mksurfdata_map/landuse_timeseries_hist_78pfts_simyr1850-2015.txt
> > A       tools/mksurfdata_map/landuse_timeseries_hist_dynlake_simyr1900-2009.txt
> > M       tools/mksurfdata_map/mksurfdata_map.namelist
> > ```
> 
> 
> We will want the sample namelist update. And possibly we want the sample "dynlake" text file. I assume it must have the sample format for adding dynamic lakes for each year. We will also want to get the dynlake option into the mksurfdata.pl script. Having the sample dynamic lake text file, should help in figuring out the changes for mksurfdata.pl.

Thanks, @ekluzek . However:

- [ ] Can you clarify if we actually want this change from `landuse_timeseries_hist_78pfts_simyr1850-2015.txt` to `landuse_timeseries_hist_16pfts_Irrig_CMIP6_simyr1850-2015.txt`. Also note that the dynlake file has different sim years from the pft file; is that a problem for these demonstration namelists?

tools/mksurfdata_map/src/mksurfdat.F90
billsacks: - [ ] I actually wonder if we want somewhat more rework than just what @ekluzek suggested above. Specifically: the code as written assumes separate files for the year-by-year lake data vs. the year-by-year pft and harvest data. Would it make sense to put these all on the same file? (I feel like the most straightforward thing would be to change the format of this file slightly so that there is one line per year, with format: 'year pftfilename harvestfilename lakefilename'. However, that's more a comment for @ekluzek because I don't expect you (@Ivanderkelen) to do that rework unless you're actively excited about doing it. And I'm fine with sticking with the current format of having separate lines for each file, with multiple lines per year, for now.)

- [ ] We'll also need changes in mksurfdata.pl to add the names of the dynamic lake files to the year-by-year text file(s).

How exactly we handle this will depend on whether we're going to put transient lake areas on all of our landuse timeseries files moving forward. @ekluzek do you feel we should just go ahead and do that, assuming we have the data? I'm thinking yes, but do you have thoughts on it?

tools/mksurfdata_map/src/mksurfdat.F90
billsacks: As I mention in a comment below, I'm thinking we should put the lake file names on the same file as the pft & harvest file names. @ekluzek does that make sense to you, or do you see a reason to keep them on separate files? My assumption (like yours, @ekluzek ) is that we'll always put the transient lake data on the landuse timeseries files, but this will only be activated if the flag `do_transient_lakes` is true. By default I think this will be false for now, while this is still an experimental setting.

tools/mksurfdata_map/src/mkfileMod.F90
billsacks: I have removed this in a cleanup commit that I will have @Ivanderkelen merge into this branch.

billsacks (changes requested):
This is part 1 of my review. I have looked through the changes in bld and tools, which I don't think I reviewed before. Some of the questions are ones that are at least partly for @ekluzek and @dlawrenncar . I'll try to talk with them about some of this tomorrow (Thursday) morning. I haven't gotten a chance to do a final review of the changes in src, and given my schedule, this probably won't happen until next week. I am also going to open a PR with a few minor cleanup changes; please merge that into this branch if you approve.

A few big-picture comments:

- [ ]  We'll need to talk about how to get the necessary raw data for creating landuse timeseries files out of the box.

- [x] Somewhat related to the above: we'll want at least one test that exercises this option, using either an out-of-the-box landuse timeseries file or a one-off landuse timeseries file (even single point would be okay). (I can set up this test once we figure out the answer to the above question.)

billsacks:
- [ ] One more possibly big thought: Do we actually have the same years of transient data for lakes as for PFTs? And even if we do now, do we want to count on this being true in the code? If not, I'm thinking we should rework things in a few places:

- In mksurfdata_map: We should keep the lake file names on a different file (in contrast to my suggestion above), but have a separate time loop for reading the lake file names, since the years will differ.

- In mksurfdata_map and in the Fortran code: I think it may be cleanest if we put lake areas on a separate file from the other transient landcover variables. The significant downside of this is that it means more files that we need to manage in our xml database. But I'm having trouble imagining how it could work to have both sets of variables on one file if they apply over a different range of years: I think this would require either having multiple time axes (which I think is only available with NetCDF4) or having missing data along with some way of denoting which years are valid for each data field (which would require some new code and would involve wasted space on the files).

- (Other than that, I don't think many changes would be needed in the Fortran source code: I'm pretty sure I wrote the dynamic landunits code to be flexible enough to handle different sets of years for different variables as long as they are given on different files, but I'd want to double-check that.)

ekluzek:
@billsacks let's plan on talking about this more tomorrow. But, since both lake and PFT's only change yearly, you just keep a yearly time axis, but have either lake or PFT repeat it's first or last years if it's not available over the same time range.

In this case PFT's change over 1850 to 1899, but lakes start at 1900. So lake will use the value of 1900 for 1850 to 1900. then at that point they are in sync until the end of the series in 2000. Similarly for future scenarios, lake would then be constant for years after 2000. That way you don't have to have two landuse.timeseries files, although you do then have a lot of times where lake has the same value for quite a few time-slices. But, I think that's OK.

billsacks:
> since both lake and PFT's only change yearly, you just keep a yearly time axis, but have either lake or PFT repeat it's first or last years if it's not available over the same time range.
> 
> In this case PFT's change over 1850 to 1899, but lakes start at 1900. So lake will use the value of 1900 for 1850 to 1900. then at that point they are in sync until the end of the series in 2000. Similarly for future scenarios, lake would then be constant for years after 2000. That way you don't have to have two landuse.timeseries files, although you do then have a lot of times where lake has the same value for quite a few time-slices. But, I think that's OK.

Good point. As long as lakes are the ones with fewer available years, this won't be a big deal. If there were many years for lakes where we did not have PFT data, that would start to lead to wasted space, but we can deal with that later, if that ever becomes an issue.

- [ ] So then we will need logic either in mksurfdata.pl (when building the text file) or in mksurfdat.F90 (when looping over years) to replicate years of lake data when necessary. (Or I suppose we could just manually replicate the file names in the xml if both of these possible solutions end up being hard to implement.)

billsacks:
@Ivanderkelen - From discussion with Erik and Dave, we decided that we'll split this PR into two pieces:

1. Changes outside of tools (src + the small changes in bld). I'll work on getting these changes to master at high priority.

2. The changes in the tools directory (mksurfdata_map) that are needed to create out-of-the-box datasets with transient lake area. More discussion and work is needed here, so this will be done on a somewhat longer time-frame.

I will do the work to separate the changes into two branches, and thus two separate PRs. **Since the discussion in this current PR has nearly all focused on mksurfdata_map so far, I'd like to keep this PR as the one with just the tools changes, and keep discussion here focused on that piece.** Next week, I will plan to back out the changes outside of tools on this branch. I will then create a separate PR for (1).

tools/mksurfdata_map/mksurfdata_map.namelist
```diff
@@ -24,7 +24,8 @@
  map_fch4         = '/glade/p/cesm/cseg/inputdata/lnd/clm2/mappingdata/maps/10x15/map_360x720_cruncep_to_10x15_nomask_aave_da_c130326.nc'
  mksrf_fsoitex    = '/glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/mksrf_soitex.10level.c010119.nc'
  mksrf_forganic   = '/glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/mksrf_organic_10level_5x5min_ISRIC-WISE-NCSCD_nlev7_c120830.nc'
- mksrf_flakwat    = '/glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/mksrf_LakePnDepth_3x3min_simyr2004_csplk_c151015.nc'
+ mksrf_flakwat    = '/glade/u/home/ivanderk/clm5.0/own_inputfile/mksurf_LakePnDepth_3x3min_simyr2009_csplk_c190603.nc
```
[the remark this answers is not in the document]
billsacks: From discussion today, @dlawrenncar felt that we *should* update our out-of-the-box lake dataset with the new one you have been using.

- [ ] @ekluzek is going to open an issue to do that update (which will be separate from the changes here)

- [ ] then we should update `mksurfdata_map.namelist` to point to the new, official file, rather than the one in Inne's home directory

billsacks:
- [ ] One other change I realized will be needed here: adding the HASLAKE variable to the landuse timeseries file (I don't see that being written right now, but the code will require it).

ekluzek:
@billsacks instead of having a variable on the file -- why doesn't it just check for the existence of the lake variable on the file? I don't see the need for a separate variable to say that another variable exists. You should be able to check the variable for existence and then use that to set the logical flag for it in the code.

billsacks:
@ekluzek HASLAKE is a spatial variable that says whether there will be any lake area at any point in the timeseries.

ekluzek:
Ahh, now that makes sense. I should've looked at the code first. But, thanks for that clarification.

Ivanderkelen:
> > since both lake and PFT's only change yearly, you just keep a yearly time axis, but have either lake or PFT repeat it's first or last years if it's not available over the same time range.
> > In this case PFT's change over 1850 to 1899, but lakes start at 1900. So lake will use the value of 1900 for 1850 to 1900. then at that point they are in sync until the end of the series in 2000. Similarly for future scenarios, lake would then be constant for years after 2000. That way you don't have to have two landuse.timeseries files, although you do then have a lot of times where lake has the same value for quite a few time-slices. But, I think that's OK.
> 
> Good point. As long as lakes are the ones with fewer available years, this won't be a big deal. If there were many years for lakes where we did not have PFT data, that would start to lead to wasted space, but we can deal with that later, if that ever becomes an issue.
> 
> * [ ]  So then we will need logic either in mksurfdata.pl (when building the text file) or in mksurfdat.F90 (when looping over years) to replicate years of lake data when necessary. (Or I suppose we could just manually replicate the file names in the xml if both of these possible solutions end up being hard to implement.)

Thank  you @ekluzek and @billsacks for looking the review! I am now back and ready to work on this. 

- About the time periods: I have 'newer' the lake maps spanning the period 1850-2017. There is a file for every year, and when lake surface area stays the same it is a copy of the previous year. The raw files are located here: `/glade/scratch/ivanderk/inputdata/rawdata/timeseries/`. 
The data originates from the vector-based GRanD and HydroLAKES databases, and are converted to raster using the following tool: https://github.com/VUB-HYDR/polygon_to_cellareafraction 
They will have to be transferred to an official location at one point. Let me know if any actions are required from my side for this. 

- I have also and updated namelist file on this, if desired I can add this to the repo. `/glade/u/home/ivanderk/clm5.0/tools/mksurfdata_map/surfdata_0.9x1.25_hist_16pfts_Irrig_CMIP6_simyr1850_c200217.namelist`

tools/mksurfdata_map/mksurfdata_map.namelist
Ivanderkelen: This is removed in 0a38f6d7fc1ab68ce5f0845da6a1a64b8724b2bd.

tools/mksurfdata_map/src/mksurfdat.F90
```diff
@@ -1399,13 +1423,28 @@ subroutine normalizencheck_landuse(ldomain)
...
+          ! recalculate sum
+          suma = pctlak(n) + pctwet(n) + pcturb(n) + pctgla(n)
+          
+          if  (suma > (100._r8 + tol_loose)) then
+            write(6,*) subname, ' ERROR: pctwet + pcturb + pctgla must be'
+            write(6,*) '<= 100% before calling this subroutine'
+            write(6,*) 'n, pctlak, pctwet, pcturb, pctgla = ', &
+                n, pctwet(n), pcturb(n), pctgla(n)
+            call abort()
+          end if
```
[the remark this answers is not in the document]
Ivanderkelen: 1. I included the check on the sum of all land cover types not exceeding > 100 % within the time loop (55a0fb480a67059d36de25eecec80a731a021b57). However, different to the code for the surface dataset, I kept that when special land units > 100%,  pct_lak is adjusted to fill up the cell to 100%., instead of subtracting the excess from the most dominant land cover. To me this makes most sense, as the other special land units are constant in time. Do you agree? 

2. About rounding to the nearest integer: I didn't add this. I don't have strong opinions on this, and don't really know why this would be necessary. But if so, I could still add it.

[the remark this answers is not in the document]
billsacks: Thanks, @Ivanderkelen . I'm thinking that it may be important to be consistent in the logic for the surface dataset and the landuse_timeseries file – otherwise, we could get different behavior from using a landuse timeseries file even if the lake area is supposed to be the same there as in the surface dataset. Also, other landunits like urban may become dynamic in the future.

For similar reasons, I'm now leaning more towards thinking this should be consistent in terms of the rounding to nearest integer. We could change this rounding code if it's important to capture lake areas < 0.5% on the grid cell, or to capture more precision in the lake area dynamics, but if we do change this for the landuse timeseries file, I'm still thinking that we should be consistent in the surface dataset, too.

However, I'm also realizing that, for the sake of consistency and correctness – especially with the treatment of wetland % – we may actually need slightly more extensive changes here. In particular, I think it may be important in some cases that we reset pct area of the various special landunits to their pre-corrected values, and then redo the various landunit adjustments from the beginning. This way, we'll be doing the same adjustments from scratch each year, rather than (for example) letting the year 1850 adjusted wetland % impact what we calculate in 1851.

Here is my thinking in more detail:

- [ ] First, to make all of this easier and avoid code duplication: Move the whole block of code after the call to `change_landuse` up through the call to `normalizencheck_landuse` (around lines 700-something in mksurfdat.F90) into the top of `normalizencheck_landuse`: I think we're going to want to redo just about all of this for the landuse timeseries file now, and the parts that aren't needed seem like they won't hurt to redo each year.

- [ ] Right after the call to change_landuse (around line 713), but before the call to normalizencheck_landuse: Set 4 new variables:

  ```Fortran
  pctlak_orig(:) = pctlak(:)
  pctwet_orig(:) = pctwet(:)
  pcturb_orig(:) = pcturb(:)
  pctgla_orig(:) = pctgla(:)
  ```

- [ ] Remove the code I just had you add in the time loop before calling normalizencheck_landuse in that loop, since this will now be done in normalizencheck_landuse according to the above item.

- [ ] Right before calling normalizencheck_landuse in the time loop, reset landunit areas that haven't been read back to their original (pre-adjustment) values:

  ```Fortran
  ! For landunits NOT read each year: reset to their pre-adjustment values in preparation for redoing landunit area normalization
  pctwet(:) = pctwet_orig(:)
  pcturb(:) = pcturb_orig(:)
  pctgla(:) = pctgla_orig(:)
  ```

  (Then will come the existing call to normalizencheck_landuse in the time loop.)

- [ ] To make sure this was done right: generate a surface dataset at f09 or f19 resolution from master and after your code changes, and make sure the two are bit-for-bit. (A quick way to check if two files are bit-for-bit identical is to run `/glade/p/cesmdata/cseg/tools/cime/tools/cprnc/cprnc -m /path/to/file1.nc /path/to/file2.nc > cprnc_report.txt`; near the bottom it should report that the two files are identical if they are, or report how many variables differ.) Probably best to do the same with a landuse timeseries file, too; there we expect some diffs due to the inclusion of dynamic lakes now, but I want to be sure that the only differences are as expected. I can help you with this step if you'd like.

- [ ] An additional helpful test would be to compare PCT_LAKE in the first year of a landuse timeseries file with PCT_LAKE on a surface dataset corresponding to that year; hopefully the two will be identical.


Whew! I know that's a lot, but hopefully most of it is relatively straightforward to do. That said: @Ivanderkelen and/or @ekluzek - **I'd welcome your thoughts on the above, because I don't feel 100% sure that what I'm suggesting is correct, and I sometimes feel that, for every issue I fix with mksurfdata_map, I manage to introduce a new issue.** Also, @Ivanderkelen , let me know if this feels like too much for you and if you'd like help implementing the above. The good thing is that we should be able to test to ensure that we aren't messing up any existing fields.

billsacks:
@Ivanderkelen I should add: please wait to do any major work on mksurfdat.F90: I have a small set of changes coming to master today or early next week (#1119 ) and you'll get conflicts if you try to do some of the rework I suggested before then. So, once #1119 is merged, you should update your branch to the latest version of master, then can do the rework I suggested.

tools/mksurfdata_map/src/mksurfdat.F90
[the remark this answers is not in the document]
Ivanderkelen: Thanks for writing this out! At first sight, it all dounds logic and straightforward to do. I'll keep an eye on https://github.com/ESCOMP/CTSM/pull/1073 and start when it is merged to master.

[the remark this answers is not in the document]
Ivanderkelen: The prosed changes are applied in d8b85afb2d0fd11251e434c5f57a151489bfe05f. 
Maybe good to take a look at the comments in the code, they might be insufficient. 

I had to remove the deallocation of organic, pctclay, pctsand and soicol, as they are rounded now within the normalizencheck_landuse subroutine, and thus repeated in the time loop. (which I think doesn't harm, as it would be more consistent to keep them in the subroutine rather than putting them seperate from rouding the other landunits?). 

I did the testing you proposed: both surfacedatasets (before and after changes) are bit for bit the same. Also, PCT_LAKE from surfdat and first timestep of landuse.timeseries are exactly the same. The differences in the two landuse.timeseries files are due to the rounding (all between -0.5 and 0.5). 

I am not sure yet how important this is to get more detailed dynamics than 0.5%, and will have a look at this tomorrow.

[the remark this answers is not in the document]
billsacks: Thanks so much @Ivanderkelen . I took a look at your most recent changes and they look great.

tools/mksurfdata_map/src/mksurfdat.F90
```diff
@@ -1186,9 +1156,17 @@ program mksurfdat
                 call abort()
              end if
           end do
+          
 
-          call change_landuse(ldomain, dynpft=.true.)
```
billsacks, line 1190: - [ ] Comparing the changes in this PR with those done by @keerzhang1 in #1579, one difference I see is the removal of this change_landuse call in the time loop in this PR (it was left in place in #1579). It probably doesn't matter hugely, but I'm thinking that it's probably best to leave this call to change_landuse in place, as is done in #1579 .

billsacks:
The relevant changes from this PR are now incorporated into #1579 (by @keerzhang1). I have also made a few notes in that PR summarizing the relevant outstanding discussion in this PR. Therefore, I am closing this PR since it is superseded by #1579 . Thank you, @Ivanderkelen , for all of your work on this PR, which provided a very helpful basis for #1579 !

Issue #1030, opened 2020-06-02
mpi-serial nag case fails with pio2 when defining h1 history restart file


billsacks:
### Brief summary of bug

The test `ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly` fails when using pio2 when defining the h1 history restart file.

### General bug information

**CTSM version you are using:** `ctsm1.0.dev093-17-ge55f7451`; also suspected to be an issue on ctsm master if cime is switched to a tag on master (`cime5.8.23`) rather than pointing to the cime branch that we're currently using (which was created to get around this issue) (cime master uses pio2 for this test, whereas the cime branch uses pio1).

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** Only observed for this one test; unknown if it is a more general issue

### Details of bug

The above-referenced test dies when writing the restart files, with:

```
 Opened file ./ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly.20200601_170252_ohlcig.clm2.r.2003-01-01-00000.nc to write 422
 htape_create : Opening netcdf rhtape ./ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly.20200601_170252_ohlcig.clm2.rh0.2003-01-01-00000.nc
 Opened file ./ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly.20200601_170252_ohlcig.clm2.rh0.2003-01-01-00000.nc to write 423
 htape_create : Successfully defined netcdf restart history file  1
 htape_create : Opening netcdf rhtape ./ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly.20200601_170252_ohlcig.clm2.rh1.2003-01-01-00000.nc
 Opened file ./ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly.20200601_170252_ohlcig.clm2.rh1.2003-01-01-00000.nc to write 424
 htape_create : Successfully defined netcdf restart history file  2
```

The job output file indicates a segmentation fault.

[...]

Issue #1029, opened 2020-06-02
FATES tests fail with pio2


billsacks:
### Brief summary of bug

When testing with pio2, many (all?) FATES tests fail when either writing to or closing the history file. It's not clear to me whether this issue originates in pio2, ctsm or fates, but @jedwards4b strongly believes the issue is not in pio2 itself.

### General bug information

**CTSM version you are using:** Observed with `ctsm1.0.dev093-17-ge55f7451`; also suspected to be an issue on ctsm master if cime is switched to a tag on master (`cime5.8.23`) rather than pointing to the cime branch that we're currently using (which was created to get around this issue) (cime master uses pio2 for this test, whereas the cime branch uses pio1). (I have confirmed my suspicions for one of the tests listed below.)

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** FATES tests with pio2; only observed so far for mpi-serial tests, because those are the only ones that I have tested with pio2; it may or may not exist for non-mpi-serial tests.

### Details of bug

The following FATES tests fail when run with pio2:

```
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm45FatesGs.izumi_nag.clm-FatesColdDef
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesGs.izumi_nag.clm-FatesColdDef
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesGs.cheyenne_gnu.clm-FatesColdDef
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm45FatesGs.cheyenne_intel.clm-FatesColdDef
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesGs.cheyenne_intel.clm-FatesColdDef
ERS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm45FatesGs.cheyenne_intel.clm-FatesColdDef
ERS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm50FatesGs.cheyenne_intel.clm-FatesColdDef
```

I haven't looked closely at the izumi failures. The cheyenne failures break down into two different failure types; I can't tell if the underlying issue is the same or different in these two types:

(1) These tests:

```
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesGs.cheyenne_gnu.clm-FatesColdDef
[...]

PR #1085, opened 2020-07-23
Jedwards/io fixes

jedwards4b:
### Description of changes
History write changes for compatibility with pio2

### Specific notes
Only write time independent fields in the file once.   In pio2 variables of similar type are aggregated so that all are written to the file at the same time.  When an existing file is reopened and appended the information in the variable description about whether the variable includes the unlimited dimension is lost and so variables with and without this dimension are aggregated for write, this creates an error in pio at write time when it is determined that the variables have different dimension counts. 
This can be considered a pio bug and issue https://github.com/NCAR/ParallelIO/issues/1670 has been opened in that repo. 
This PR applies a workaround so that non-time dependent fields are only written to the file once.    I note that a comment in histFileMod.F90 indicates that these fields should be written each time because they may in fact change values.   This seems like a real problem to me and if this is the case then perhaps they should not be considered time independent.   I have also added pio_syncfile function which is more efficient than closing and immediately reopening a file. 

CTSM Issues Fixed (include github issue #): #1029 #1030 

Testing performed, if any:
The following tests were run with both pio1 and pio2:
ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.cheyenne_intel 
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesGs.cheyenne_gnu.clm-FatesColdDef

### DIFF
--- a/src/main/histFileMod.F90
+++ b/src/main/histFileMod.F90
@@ -13,14 +13,14 @@ module histFileMod
   use spmdMod        , only : masterproc
   use abortutils     , only : endrun
   use clm_varctl     , only : iulog, use_vertsoilc, use_fates
-  use clm_varcon     , only : spval, ispval, dzsoi_decomp 
+  use clm_varcon     , only : spval, ispval, dzsoi_decomp
   use clm_varcon     , only : grlnd, nameg, namel, namec, namep, nameCohort
   use decompMod      , only : get_proc_bounds, get_proc_global, bounds_type
   use GetGlobalValuesMod , only : GetGlobalIndexArray
-  use GridcellType   , only : grc                
-  use LandunitType   , only : lun                
-  use ColumnType     , only : col                
-  use PatchType      , only : patch                
+  use GridcellType   , only : grc
+  use LandunitType   , only : lun
+  use ColumnType     , only : col
+  use PatchType      , only : patch
   use EDTypesMod     , only : nclmax
   use EDTypesMod     , only : nlevleaf
   use FatesInterfaceMod , only : nlevsclass, nlevage
@@ -29,7 +29,7 @@ module histFileMod
   use FatesLitterMod    , only : ncwd
   use EDTypesMod        , only : num_elements_fates => num_elements
   use FatesInterfaceMod , only : maxveg_fates => numpft
-  use ncdio_pio 
+  use ncdio_pio
 
   !
   implicit none
@@ -163,7 +163,7 @@ module histFileMod
   private :: hfields_write             ! Write a variable to a history tape
   private :: hfields_1dinfo            ! Define/output 1d subgrid info if appropriate
   private :: hist_update_hbuf_field_1d ! Updates history buffer for specific field and tape
-  private :: hist_update_hbuf_field_2d ! Updates history buffer for specific field and tape 
+  private :: hist_update_hbuf_field_2d ! Updates history buffer for specific field and tape
   private :: hist_set_snow_field_2d    ! Set values in history field dimensioned by levsno
   private :: list_index                ! Find index of field in exclude list
   private :: set_hist_filename         ! Determine history dataset filenames
@@ -202,7 +202,7 @@ module histFileMod
                                                ! for 2D arrays, where the second dimension is allowed
                                                ! to be 1
      integer :: num2d                          ! size of hbuf second dimension (e.g. number of vertical levels)
-     integer :: hpindex                        ! history pointer index 
+     integer :: hpindex                        ! history pointer index
      character(len=scale_type_strlen) :: p2c_scale_type       ! scale factor when averaging patch to column
      character(len=scale_type_strlen) :: c2l_scale_type       ! scale factor when averaging column to landunit
      character(len=scale_type_strlen) :: l2g_scale_type       ! scale factor when averaging landunit to gridcell
@@ -359,7 +359,7 @@ contains
     integer :: numl         ! total number of landunits across all processors
     integer :: numc         ! total number of columns across all processors
     integer :: nump         ! total number of pfts across all processors
-    type(bounds_type) :: bounds                  
+    type(bounds_type) :: bounds
     character(len=*),parameter :: subname = 'masterlist_addfld'
     !------------------------------------------------------------------------
 
@@ -681,7 +681,7 @@ contains
     fexcl(:,8)  = hist_fexcl8(:)
     fexcl(:,9)  = hist_fexcl9(:)
     fexcl(:,10) = hist_fexcl10(:)
- 
+
 
     ! First ensure contents of fincl and fexcl are valid names
 
@@ -861,7 +861,7 @@ contains
   logical function is_mapping_upto_subgrid( type1d, type1d_out ) result ( mapping)
     !
     ! !DESCRIPTION:
-    ! 
+    !
     ! Return true if this field will be mapped into a higher subgrid level
     ! If false it will be output on it's native grid
     !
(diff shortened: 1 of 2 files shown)


ekluzek:
I was concerned about @jedwards4b comment about time constant variables changing. I found the place in the code that I think he is referring to, and it's from 6 to 8 years ago, and I would say it's incorrect. Here it is...

``` fortran
       ! Most of this is constant and only needs to be done on tape(t)%ntimes=1
       ! But, some may change for dynamic PATCH mode for example

       if (ldomain%isgrid2d) then
          call ncd_io(varname='lon', data=lon1d, ncid=nfid(t), flag='write')
          call ncd_io(varname='lat', data=lat1d, ncid=nfid(t), flag='write')
       else
          call ncd_io(varname='lon', data=ldomain%lonc, dim1name=grlnd, ncid=nfid(t), flag='write')
          call ncd_io(varname='lat', data=ldomain%latc, dim1name=grlnd, ncid=nfid(t), flag='write')
       end if
       call ncd_io(varname='area'    , data=ldomain%area, dim1name=grlnd, ncid=nfid(t), flag='write')
       call ncd_io(varname='landfrac', data=ldomain%frac, dim1name=grlnd, ncid=nfid(t), flag='write')
       call ncd_io(varname='landmask', data=ldomain%mask, dim1name=grlnd, ncid=nfid(t), flag='write')
       call ncd_io(varname='pftmask' , data=ldomain%pftm, dim1name=grlnd, ncid=nfid(t), flag='write')
       call ncd_io(varname='nbedrock' , data=grc%nbedrock, dim1name=grlnd, ncid=nfid(t), flag='write')
```

None, of those variables can change. pftmask is the mask for PFT's, which can't change in time even though PFT's themselves change. So the comment just looks incorrect to me.

jedwards4b:
Yes, that is the comment I was referring to.  Thanks @ekluzek.   I am running ctsm tests on cheyenne and izumi now.

billsacks:
@jedwards4b and @ekluzek - I have a few follow-up questions on where things stand with this PR:

1. Were you able to run all of the aux_clm tests with these changes? If so, did you run them with the defaults for PIO (so using PIO1) or with PIO2, or both? If with PIO2, did you include the  fix for https://github.com/NCAR/ParallelIO/issues/1670? And what was the status of this testing?

2. It looks like, with the latest commit, you have gone back to calling `htape_timeconst` every time because (contrary to the name of this subroutine) some of the things in here aren't time-constant. Does that work, or do we need to add some logic to only do those writes once?

3. It looks like the main behavior change in this PR is now that the fields in `hfields_1dinfo` are just written once, in the first time, rather than being rewritten each time. It looks like this will potentially change behavior in the (probably rare, but possible) case where there is a multi-time 1-d history file that spans the year boundary: on master, it looks like this would give the subgrid areas in the last time, but with this PR, it would give the subgrid areas in the first time. This may not be a big deal, but it does feel less self-consistent, since the time on the file gives the ending time yet these fields would now apply to the starting time. I'm wondering:

    a. Is this change actually still needed, given that it looks like you have fixed the underlying pio2 issue?

    b. If this change is still needed, perhaps @ekluzek or I should look into whether we can change the logic so that these fields are only written in the last time, rather than the logic in this PR where they are written in the first time. (The main potential gotcha I see with that is that it means the fields would be undefined on the file until the last time is written. In particular, would this be a problem for a file that spans a restart?)

jedwards4b:
1. All of the tests ran on izumi with pio2.   Only one failure 
0724-153311iz_nag: 35 tests
    FAIL ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly RUN time=1293
As far as I can tell this fail is not associated with the code change

On cheyenne I had the following 
 ./cs.status.fails 
0727-085945ch_gnu: 46 tests
    FAIL SMS_Ld1.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasGs.cheyenne_gnu.clm-default NLCOMP

 
0727-085945ch_int: 162 tests
    FAIL DAE_N2_D_Lh12.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-DA_multidrv RUN time=1224
    PEND DAE_N2_D_Lh12.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-DA_multidrv COMPARE_base_da
    FAIL ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic COMPARE_base_rest
    FAIL ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic BASELINE ctsm1.0.dev105: DIFF
    FAIL SMS.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.ISSP585Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode RUN time=1223

 
========================================================================
Non-PASS results for select phases:
TPUTCOMP non-passes: 31
MEMCOMP non-passes: 16

2.  I did change it to call at each write again so that the time variables are written, but I changed so that only the time dependent variables are written.

3. If these fields are time dependent then you should change so that they are written for each time, if they are not time dependent then it doesn't matter when you write them.
 a. I haven't brought the new version of pio into cime yet. 
 b. Again if you are changing the time that a time-independent field is written and expecting a different result then you have a problem and you should reconsider the nature of these variables.

billsacks:
@jedwards4b thanks a lot for your replies. Some follow-ups below:

(1) Test results:

*Remaining issues*

- [ ] `FAIL ERS_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianGs.izumi_nag.clm-monthly RUN time=1293`: this is issue #1030 . So I guess your changes here do not resolve that issue, and some other fixes – either in CTSM or in PIO2 – are needed for this case.

- [ ] `ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic`: There seems to be a problem in the writing of `time_bnds` on the cpl aux hist file (`ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic.C.0727-085945ch_int.cpl.hl2x1yr_glc.0003-01-01-00000.nc.base` and `ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic.C.0727-085945ch_int.cpl.hl2x1yr_glc.0003-01-01-00000.nc.rest`): in your runs, time_bnds are `730, 0.999999999999535` and `730, 0`, whereas in a recent run from CTSM master (with pio1), time_bnds is `730, 1095`, as expected. This should have nothing to do with the CTSM changes, but maybe is something else that isn't working right with pio2 (possibly due to issues in the mct driver code, not necessarily pio2 issues per se).

*No need to worry about these*

- `FAIL DAE_N2_D_Lh12.f10_f10_musgs.I2000Clm50BgcCropGs.cheyenne_intel.clm-DA_multidrv RUN time=1224`: looks like it just ran out of wallclock time, maybe due to system slowness (so no need to worry about this one)

- `SMS.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.ISSP585Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode`: this seems to be a new test, which looks like it failed for @ekluzek , too.

(2) `htape_timeconst`: Thanks; no more follow-ups needed here.

(3) Behavior change in `hfields_1dinfo`: I understand and appreciate your points here, but given how overwhelmed we are right now with a backlog of issues, I'd prefer to avoid this change if it isn't actually needed, since it is likely to require some additional discussion and/or code changes. My understanding is that, once we can point to a version of cime with updated pio2, the change you made here won't be necessary; is that correct?

- [ ] I think we should then back out this change to `hfields_1dinfo`, if it isn't actually needed.

jedwards4b:
Superceeded by #1095

PR #1094, opened 2020-07-31
Make most QSat output arguments optional

billsacks:
### Description of changes

Make most QSat output arguments optional

This has two benefits:

(1) It prevents callers from needing to create dummy variables to hold
    outputs that they don't actually need.

(2) If neither qsdT nor esdT are needed, then we can avoid doing some
    expensive calculations.

This is partly for the benefit of #1086 

### Specific notes

Testing performed, if any:
- `SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default` with comparison against master (bfb)
- `PFS_Ld20.f09_g17.I2000Clm50BgcCrop.cheyenne_intel` underway to look at performance changes, if any

### DIFF
--- a/src/biogeophys/QSatMod.F90
+++ b/src/biogeophys/QSatMod.F90
@@ -61,8 +61,9 @@ contains
   subroutine QSat (T, p, qs, es, qsdT, esdT)
     !
     ! !DESCRIPTION:
-    ! Computes saturation mixing ratio and (optionally) the change in saturation
-    ! mixing ratio with respect to temperature.
+    ! Computes saturation mixing ratio and (optionally) the change in saturation mixing
+    ! ratio with respect to temperature. Mixing ratio and specific humidity are
+    ! approximately equal and can be treated as the same.
     ! Reference:  Polynomial approximations from:
     !             Piotr J. Flatau, et al.,1992:  Polynomial fits to saturation
     !             vapor pressure.  Journal of Applied Meteorology, 31, 1507-1513.
@@ -86,9 +87,7 @@ contains
     real(r8) :: td,vp,vp1,vp2
     !-----------------------------------------------------------------------
 
-    td = T - SHR_CONST_TKFRZ
-    if (td > 100.0_r8) td=100.0_r8
-    if (td < -75.0_r8) td=-75.0_r8
+    td = min(100.0_r8, max(-75.0_r8, T - SHR_CONST_TKFRZ))
 
     if (td >= 0.0_r8) then
        es_local = a0 + td*(a1 + td*(a2 + td*(a3 + td*(a4 &
--- a/src/biogeophys/QSatMod.F90
+++ b/src/biogeophys/QSatMod.F90
@@ -14,49 +14,48 @@ module QSatMod
   public :: QSat
   !-----------------------------------------------------------------------
 
-    ! For water vapor (temperature range 0C-100C)
-    real(r8), parameter :: a0 =  6.11213476_r8
-    real(r8), parameter :: a1 =  0.444007856_r8
-    real(r8), parameter :: a2 =  0.143064234e-01_r8
-    real(r8), parameter :: a3 =  0.264461437e-03_r8
-    real(r8), parameter :: a4 =  0.305903558e-05_r8
-    real(r8), parameter :: a5 =  0.196237241e-07_r8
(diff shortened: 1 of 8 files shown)


billsacks:
@slevisconsulting Note in particular that I have rearranged the order of the output arguments. Now qs (qsat) is the first output argument, and is the only required output argument.

src/biogeophys/QSatMod.F90
```diff
@@ -12,58 +12,56 @@ module QSatMod
...
+contains
 
   !-----------------------------------------------------------------------
-  subroutine QSat (T, p, es, esdT, qs, qsdT)
+  subroutine QSat (T, p, qs, es, qsdT, esdT)
     !
     ! !DESCRIPTION:
-    ! Computes saturation mixing ratio and the change in saturation
+    ! Computes saturation mixing ratio and (optionally) the change in saturation
     ! mixing ratio with respect to temperature.
```
slevisconsulting, line 65: @billsacks all your mods make sense to me. Thank you for introducing this simplification and enhanced flexibility to subroutine QSat.

I have a suggested addition to the above comment. I had forgotten that mixing ratio and specific humidity are treated as equal in meteorology, and I may not be alone in forgetting:
```suggestion
    ! mixing ratio with respect to temperature. Mixing ratio and 
    specific humidity are approximately equal and can be treated 
    as the same.
```

src/biogeophys/QSatMod.F90
```diff
@@ -77,91 +75,56 @@ subroutine QSat (T, p, es, esdT, qs, qsdT)
...
+    real(r8) :: esdT_local  ! local version of esdT (in case esdT is not present)
     real(r8) :: td,vp,vp1,vp2
     !-----------------------------------------------------------------------
 
-    T_limit = T - SHR_CONST_TKFRZ
-    if (T_limit > 100.0_r8) T_limit=100.0_r8
-    if (T_limit < -75.0_r8) T_limit=-75.0_r8
+    td = T - SHR_CONST_TKFRZ
+    if (td > 100.0_r8) td=100.0_r8
+    if (td < -75.0_r8) td=-75.0_r8
```
slevisconsulting, line 91: Would this be an answer-changing modification?
```suggestion
    td = min(100.0_r8, max(-75.0_r8, T - SHR_CONST_TKFRZ))
```

src/biogeophys/QSatMod.F90
billsacks: Done - thanks

src/biogeophys/QSatMod.F90
billsacks: Done. I personally have a really hard time with mins & maxes so I don't tend to jump to them as a solution, but I do see how this more clearly shows the intent of this code - at least to someone who has less of a hard time with mins and maxes than I do :-)

billsacks: By the way, I *think* this should be a bit-for-bit change, but I'll back it out if baseline comparisons fail.

billsacks:
I checked the timing of this branch vs. master, with a focus on the timing pieces that include calls to the QSat routine. My conclusions are that these changes don't have a significant performance impact (either positive or negative).

Specifically, I checked the following, just via a single run of `PFS_Ld20.f09_g17.I2000Clm50BgcCrop.cheyenne_intel` in master and on this branch:
- BareGroundFluxes: bgflux
- CanopyFluxes: canflux
- LakeFluxes: bgplake
- CalculateSurfaceHumidity: bgp1
- UrbanFluxes: uflux
- downscale_forcings: drvinit

bgp1 and uflux took a bit more time in the new vs. old, but the differences seem well within machine variability; others were either about the same in both or slightly faster in the new.

billsacks:
I have run the full test suite on this branch, with comparisons against ctsm1.0.dev105. All tests passed and were bit-for-bit except for two that ran out of wallclock time:

```
    FAIL SMS.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.ISSP585Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode RUN time=1222
    PEND SMS.ne30pg2_ne30pg2_mg17.I2000Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode RUN
```

Issue #1041, opened 2020-06-15
Deprecate setting of specific files with CLM_USRDAT_NAME mode


ekluzek:
Setting CLM_USRDAT_NAME adds a useful ability to setup a case for an arbitrary resolution, without having to add it into config_grids in cime. When it was initially added it also provided a mechanism to use special names for files so that they could be used in that same case. With user_nl_clm, shell_commands and user-mod directories now -- this ability to name files this way isn't as useful anymore. I propose that it should be deprecated, and eventually removed.

ekluzek:
In looking at the code with @negin513 it looks to me that this is already in place. The only thing left to do is to bring in the cime update that removes it there.

ekluzek:
@jedwards4b is planning on removing the setting of a generic domain file for CLM_USRDAT_NAME in cime.

Issue #1020, opened 2020-05-21
Duplicate code in namelist_defaults_ctsm.xml


glemieux:
In bringing fates_next_api up to date with ctsm1.0.dev093 I'm seeing what looks like duplicate code in namelist_defaults_ctsm.xml:
https://github.com/ESCOMP/CTSM/blob/9db8e2ec88c231b57b9ad6a46dc2f1601d28e816/bld/namelist_files/namelist_defaults_ctsm.xml#L1122-L1142

I see that the `hgrid` file string is different in the second input; can I assume this is the most recent file or am I misinterpreting the code here?

ekluzek:
Thanks for pointing this out @glemieux. I agree this is duplicated and shouldn't be. This is using the perl parsing methods and so the variable with the most matching attributes is chosen first, and when attributes are the same it will pick the first one given. So by that it should be getting the older file, rather than the newer one, which doesn't sound right.

But, changing which file is pointed to will change answers for the prescribed test case. So we'd want to do that as it's own thing.

glemieux:
Thanks @ekluzek.  Btw I think I found another set of duplicates in namelist_definition_ctsm.xml:
https://github.com/ESCOMP/CTSM/blob/9db8e2ec88c231b57b9ad6a46dc2f1601d28e816/bld/namelist_files/namelist_definition_ctsm.xml#L1634-L1694

ekluzek:
Yep, that looks wrongly duplicated as well.

negin513:
This has been resolved in https://github.com/ESCOMP/CTSM/pull/1187 so I close this issue.

PR #961, opened 2020-04-02
Fixes for the LUNA dayl bugs

ekluzek:
### Description of changes

Fixes two of the LUNA day length bugs that @lmbirch89 found. Use the night term for the night section rather than the day version. And instead of assuming a 12 max day length, use the actual max daylength.

### Specific notes

Contributors other than yourself, if any: @lmbirch89

CTSM Issues Fixed (include github issue #): #953 #958 

 Fixes #953 
 Fixes #958

Are answers expected to change (and if so in what way)? Yes
 (when LUNA on, which is always for CLM5.0, but NOT for CLM4.5)

Any User Interface Changes (namelist or namelist defaults changes)? No

Testing performed, if any: regular
  Running regular testing now on cheyenne and izumi.
  @olyson will also run a longer simulation with it.

Issue #881, opened 2020-01-08
Remove maxpft and maxpatch_pft


ekluzek:
We left maxpft and maxpatch_pft both in the code and in the build-namelist. It looks like to me these do nothing now, and should just be removed. It actually makes the code confusing. n_dom_pfts and the reading of what's one the surface dataset replaces the need for these.

billsacks:
Slightly related?: #37

negin513:
`maxpft` is also being used in `test_irrigation.pf` code as an input to `setupEnvironment`. Should I just remove this and change wherever we are using `setupEnvironment` in the test?

https://github.com/ESCOMP/CTSM/blob/d3c31a3764c26135d58797cb2e753dbd438e5e62/src/biogeophys/test/Irrigation_test/test_irrigation.pf#L264-L270

billsacks:
It looks like maxpft is just used within that unit test module. While similar conceptually to maxpft in the main code, I think it's fine to keep maxpft in that unit test module even though you're getting rid of maxpft from the main code. Let me know if you think differently: I just took a very quick look.

ekluzek:
I agree with @billsacks that it's not critical to remove it from a unit test. However, if this is something removed from the main code it might make the unit test harder to understand. So it would be good to remove it, if doing so is reasonably easy.

billsacks:
In the unit test code, this is just used to allocate some variables to be big enough for what the unit test needs. I think it serves the same function as mxpft in pftconMod. I wouldn't spend time trying to change the unit test code right now unless the unit tests aren't working.

negin513:
This has been

PR #1102, opened 2020-08-10
Some bit-for-bit changes needed for the Perturbed Parameter Ensemble work

### LINKED ISSUE #1053: Non-standard namelist parameter: cmb_cmplt_fact
cmb_cmplt_fact is a parameter under namelist control pertaining to the fire model (combustion completion factor [unitless])

As far as I can tell it is the only real-valued, namelist-controlled parameter that is not a single value. Instead it has two elements, which are by default: 0.5 and 0.25. The first values corresponds to litter and the second to coarse woody debris. 

If possible we might split this into two parameters to avoid having a special case when it comes to editing namelist parameters.

### LINKED ISSUE #1055: Expand potential vcmax:jmax ratio
For the perturbed parameter experiment we are interested in exploring lower jmax:vcmax ratios than are currently allowed in the model. This will require commenting out a cap on wc2wj that exists in LunaMod:

https://github.com/ESCOMP/CTSM/blob/3029203daa9af4dc1ac03242462f3315746d20ca/src/biogeophys/LunaMod.F90#L1050

### LINKED ISSUE #1060: Some code cleanup of luna bug fixes
These are from some comments of the luna bug fix branch (PR #961). 

My main thought was to avoid having "yr" / "year" in the name. I suggested vcmx_last_active_day, but I'm not sure if that's the best. @ekluzek suggested something including "last_valid", which also seems reasonable. Good naming is so hard... (and yet so important).

 @billsacks
billsacks on Apr 16 Member
Another note about these names: they need _patch suffixes. Also, it's probably best if they maintain the 25_z that appears in vcmx25_z and jmx25_z variable names, to show the correspondence between these variables.

Photosynthesis:

As in LunaMod, this comment is misleading since we're not taking the running average. We should be consistent with LunaMod.

**_Initial values for the previous year settings should also be fixed (but as potentially answer changing that will be separate from this)._**

long names here should match what we're using in the description above, something about the previous year?

Minor correction, but jmx25_z_patch and jmx_prevyr long_name should be maximum rate of electron transport at 25 celsius for canopy layers, with units umol electron/m**2/s

### LINKED ISSUE #1062: Make sure some namelist parameters have a default value
There's a couple namelist parameters we need to make sure are on the lnd_in file by default for the PPE work.

freelivfix_slope_wet
freelivfix_intercept

The list of all parameters for this work are here:

https://docs.google.com/spreadsheets/d/1OtkaO_uAmafWKR9kgtRC2Ge6d6fkhymngSpben5SJ_Q

There are still some hardcoded soil parameters on the list, but it looks like they aren't going to be dealt with.

ekluzek:
### Description of changes

Some bit-for-bit changes needed for the Perturbed Parameter Ensemble branch work.

### Specific notes

Contributors other than yourself, if any: Negin Sobhani (@negin513) and Will Wieder (@wwieder )

 Fixes #1060
 Fixes #1062
 Fixes #1053
 Fixes #1055

Testing performed, if any: Ran some testing on izumi with earlier version

N.S.: Cheyenne (PASS) except for NLCOMP


ekluzek:
We are going to have @negin513 finish this off and bring this to main.

ekluzek:
These need to be done or finished at this point...

#1060
#1062 -- partially done

These could be done, but aren't required. They should be easy.

#881
#1020
#1041
#1102

ekluzek:
Looks like we should remove denitrif_nitrateconc_* from the namelist as it's already on the params file and used from there.

ekluzek:
@negin513 Chris just made a cime tag. ctsm1.0.dev111 will be pointing to a branch tag rather than a trunk tag. So you should update to cime5.8.32 in this tag after you've updated to ctsm1.0.dev111. I don't think there should be much of a difference between the branch tag that cime5.8.32, so answers shouldn't change for example.

negin513:
> @negin513 Chris just made a cime tag. ctsm1.0.dev111 will be pointing to a branch tag rather than a trunk tag. So you should update to cime5.8.32 in this tag after you've updated to ctsm1.0.dev111. I don't think there should be much of a difference between the branch tag that cime5.8.32, so answers shouldn't change for example.

After updating to cime5.8.32, should I push back to your branch?

negin513:
@wwieder and @billsacks 
Just to clarify and confirm this discussion in  https://github.com/ESCOMP/CTSM/pull/961#pullrequestreview-393876398 needed for #1060:
You are suggesting to change :
`vcmx_prevyr` to  `vcmx25_z_last_valid_patch`
`jmx_prevyr` to `jmx25_z_last_valid_patch`

negin513:
> @negin513 Chris just made a cime tag. ctsm1.0.dev111 will be pointing to a branch tag rather than a trunk tag. So you should update to cime5.8.32 in this tag after you've updated to ctsm1.0.dev111. I don't think there should be much of a difference between the branch tag that cime5.8.32, so answers shouldn't change for example.

Acknowledged and Done!

billsacks:
- [ ] @negin513 here is something important: If you want to maintain backwards compatibility with restart files that used an old variable name (e.g., `vcmx_prevyr`), you cannot simply change the varname. I'm not sure if this is important in this case, but if it is, the mechanism to do this is to use a colon-delimited string, `newname:oldname`. So in this case `vcmx25_z_last_valid_patch:vcmx_prevyr`. This will write the new name to new restart files, but will fall back on the old name if it can't find the new one. However, it's also possible that `vcmx_prevyr` is good enough for the sake of the restart variable name... I'm not sure what others think about this.

negin513:
> Looks like we should remove denitrif_nitrateconc_* from the namelist as it's already on the params file and used from there.

Acknowledged and Done!

billsacks:
Thanks @negin513 for fixing the restart var thing.

ekluzek:
@wwieder can you review these changes that relate to luna? and things like long_names and units?

ekluzek (commented):
I can't seem to review this other than to add a comment (presumably because it's on my fork and I started the PR).  But, this looks good to me.

bld/unit_testers/build-namelist_test.pl
```diff
@@ -637,6 +637,11 @@ sub make_config_cache {
                                      GLC_TWO_WAY_COUPLING=>"FALSE",
                                      phys=>"clm4_5",
                                    },
+     "freeliv wo fun"            =>{ options=>"-bgc bgc -envxml_dir .",
```
wwieder, line 640: - [ ] not sure what this title means?  is there free living nfix without fun?

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -346,8 +346,8 @@ subroutine InitAllocate(this, bounds)
       ! statements.
       allocate(this%vcmx25_z_patch  (begp:endp,1:nlevcan)) ; this%vcmx25_z_patch    (:,:) = 30._r8
       allocate(this%jmx25_z_patch   (begp:endp,1:nlevcan)) ; this%jmx25_z_patch     (:,:) = 60._r8 
-      allocate(this%vcmx_prevyr     (begp:endp,1:nlevcan)) ; this%vcmx_prevyr       (:,:) = 30._r8
-      allocate(this%jmx_prevyr      (begp:endp,1:nlevcan)) ; this%jmx_prevyr        (:,:) = 60._r8
+      allocate(this%vcmx25_z_last_valid_patch     (begp:endp,1:nlevcan)) ; this%vcmx25_z_last_valid_patch       (:,:) = 30._r8
```
wwieder, line 349: - [ ] these variable names seem descriptive, as @billsacks was asking

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -886,20 +886,20 @@ subroutine Restart(this, bounds, ncid, flag)
...
-         long_name='Maximum carboxylation rate at 25 celcius for canopy layers', units='umol CO2/m**2/s', &
+         long_name='Maximum carboxylation rate at 25 Celcius for canopy layers', units='umol CO2/m**2/s', &
          interpinic_flag='interp', readvar=readvar, data=this%vcmx25_z_patch)
       call restartvar(ncid=ncid, flag=flag, varname='jmx25_z', xtype=ncd_double,  &
          dim1name='pft', dim2name='levcan', switchdim=.true., &
-         long_name='Maximum carboxylation rate at 25 celcius for canopy layers', units='umol CO2/m**2/s', &
+         long_name='Maximum rate of electron transport at 25 Celcius for canopy layers', units='umol electrons/m**2/s', &
          interpinic_flag='interp', readvar=readvar, data=this%jmx25_z_patch)
-      call restartvar(ncid=ncid, flag=flag, varname='vcmx_prevyr', xtype=ncd_double,  &
+      call restartvar(ncid=ncid, flag=flag, varname='vcmx25_z_last_valid_patch:vcmx_prevyr', xtype=ncd_double,  &
```
wwieder, line 895: - [ ] Why do we need to keep `vcmx_prevyr` at all? Can this just be `vcmx25_z_last_valid_patch`

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -886,20 +886,20 @@ subroutine Restart(this, bounds, ncid, flag)
...
          dim1name='pft', dim2name='levcan', switchdim=.true., &
          long_name='avg carboxylation rate at 25 celsius for canopy layers', units='umol CO2/m**2/s', &
-         interpinic_flag='interp', readvar=readvar, data=this%vcmx_prevyr)
-      call restartvar(ncid=ncid, flag=flag, varname='jmx_prevyr', xtype=ncd_double,  &
+         interpinic_flag='interp', readvar=readvar, data=this%vcmx25_z_last_valid_patch)
+      call restartvar(ncid=ncid, flag=flag, varname='jmx25_z_last_valid_patch:jmx_prevyr', xtype=ncd_double,  &
```
wwieder, line 899: - [ ] As above, but for `jmx25_z_last_valid_patch`

src/biogeophys/PhotosynthesisMod.F90
billsacks: @wwieder are you asking for a change here, or approving this? Your comment makes it sound like you're approving this change, but you put a check box in....

src/biogeophys/PhotosynthesisMod.F90
billsacks: This (and for jmx) is needed for backwards compatibility with the last iteration of restart files. New restart files will just use the new name.

src/biogeophys/PhotosynthesisMod.F90
wwieder: sorry, just responding to a question @negin513 had in the PR

src/biogeophys/PhotosynthesisMod.F90
wwieder: sound good, marked as resolved.

wwieder:
BTW, @ekluzek and @negin513 sorry I missed the review request last week.  Thanks for the email reminder.

bld/unit_testers/build-namelist_test.pl
ekluzek: This actually is to make sure that you can't set free living nfix if FUN isn't on. So it's an error check for the namelist.

ekluzek: I went ahead and clicked the checkbox, but that is technically against our convention of the author making sure these are checked. So uncheck the box if my answer is insufficient.

src/biogeophys/PhotosynthesisMod.F90
negin513: > * [x]  these variable names seem descriptive, as @billsacks was asking

@wwieder Perfect! Thanks for your comment.

src/biogeophys/PhotosynthesisMod.F90
negin513: Correct! That was for backward compatibility based on @billsacks suggestion.

src/biogeophys/PhotosynthesisMod.F90
negin513: Thanks for your comment. Great!

negin513:
> BTW, @ekluzek and @negin513 sorry I missed the review request last week. Thanks for the email reminder.

@wwieder Thanks for taking the time to review this PR.

negin513:
This PR is done and it seems like there are no more issues to be addressed. I don't have the authorization to merge this PR with the master. @ekluzek and @billsacks Please merge this with the master for dev113.

doc/ChangeLog
```diff
@@ -1,4 +1,112 @@
...
+
+[ ] ctsm5_0-nwp
+
+[ ] clm4_5
+
+Notes of particular relevance for users
+---------------------------------------
+
+Caveats for users (e.g., need to interpolate initial conditions):
+cmb_cmplt_fact was a parameter under namelist control which included an array.
```
ekluzek, line 39: @negin513 one suggestion here. I would explicitly mention that if someone has cmb_cmplt_fact in a user_nl_clm file that worked before -- it won't work now since it's. split into two. It's good to try to think about it from the user's perspective and what will break for them because of the change. The thing here is that a user_nl_clm change that worked before won't anymore because of this. We want to try to explain things from the user perspective on what's different (what's new, what's removed) and what's different, and especially as in this case what would work before -- but won't work anymore because of the change.

ekluzek: I think there's three different types of user's that we want to make the ChangeLog useful for. One is for the scientists that just want to know the big picture science that is happening to the model. The next are the User's who are just using the model, and we want to tell them what's changed from their perspective. This section here is primarily targeted to that group.

The next section for developers is primarily targeting you, and Bill and I. But, also could be other developers working in the guts of the code. That section should give notes about the code changes about things like reminding us to clean this up later, that a code section should be refactored, or whatever useful bit of advice we might want to remember later. I'll use it to let you and Bill know about something, but I also use it as a reminder for me as well.

ekluzek:
> This PR is done and it seems like there are no more issues to be addressed.
> 
> I don't have the authorization to merge this PR with the master. @ekluzek and @billsacks Please merge this with the master for ctsm1.0.dev113 tag.

Hi @negin513 thanks so much for working on this and getting this done. You do have write access to escomp/master so you should be able to push the tag. And both @billsacks and I would like you to have that access to make it easier for you to make tags. So we approve and appoint you to being one that can make CTSM main development tags with all the rights, privileges and responsibilities that entails (we call this an "INTEGRATOR" in the document below). :-) If this was the military we'd pin a new button to your uniform. So just imagine we did that. :-)

The next steps are in the doc/README.CHECKLIST.master_tags and include these steps. I'm happy to help you through this if you'd like to for your first go around. Just let me know.

> ---- THE FOLLOWING CAN ONLY BE DONE BY INTEGRATORS ----
> 
> (7) Merge the PR to master when review is approved
> 
> (8) Compare master to branch show that they are identical
> 
> git diff master remote/feature-branch
> 
> This should show no diffs
> 
> (9) Make an annotated tag on master
> 
> (10) Push master and tag to ESCOMP/ctsm

PR #1119, opened 2020-08-21
Move mksurfdata_map rounding of special landunits to before wetland fill

### LINKED ISSUE #1118: mksurfdata_map rounding/truncation of special landunits should happen before wetland filling
### Brief summary of bug

I think the current ordering of the rounding/truncation of special landunits in mksurfdata_map relative to the wetland filling could cause issues in some edge cases. These should be reversed. I haven't actually observed any problems; this comes just from looking at the code.

### General bug information

**Configurations affected:** Probably unusual conditions in mksurfdata_map

### Details of bug

I think this may be important to avoid errors and/or wrong behavior in some edge cases where we have < 0.5% cover of glacier and/or lake. **I think the best thing to do would be to move the whole block of truncation code to above the wetland adjustment.** i.e., move this block:

https://github.com/ESCOMP/CTSM/blob/6c99904ea74c10a0d02367162b151580f4d7def2/tools/mksurfdata_map/src/mksurfdat.F90#L737-L747

to above this block:

https://github.com/ESCOMP/CTSM/blob/6c99904ea74c10a0d02367162b151580f4d7def2/tools/mksurfdata_map/src/mksurfdat.F90#L714-L735

In addition to the edge case where pctlak or pctgla is < 0.5%, this could also be relevant in a case like this: pctlak = 5.4%, pctgla = 94.4%. With the current code, we'd first get pctwet = 0.2%. But then, after the rounding, we'd end up with pctlak = 5%, pctgla = 94%, pctwet = 0%. Then later, in normalizencheck_landuse, I think we'd end up with 1% natural veg, which could trigger the 'vegetation found outside the pft mask' abort. (Caveat: I may be missing something here, since there are so many corrections done at different points. It's also possible that there are other problems with doing the rounding first, but I'm not seeing any off-hand.)

(This split off from #1001 )

billsacks:
### Description of changes

I think this may be important to avoid errors and/or wrong behavior in
some edge cases where we have < 0.5% cover of glacier and/or lake, and
other cases.

See https://github.com/ESCOMP/CTSM/issues/1118 for detailed thoughts

### Specific notes

Resolves #1118 

Are answers expected to change (and if so in what way)? Possible answer changes, but this should be rare

Any User Interface Changes (namelist or namelist defaults changes)? N

Testing performed, if any:

I created a subset of our standard surface datasets with `make -f Makefile.data global-present` before and after this change. The three surface datasets created in this way (f09, f19 and f10) were all bit-for-bit. There could be answer changes in rare cases, but this demonstrates that I haven't majorly broken anything with this change.
- Update (2020-08-27): as noted below, I also compared f45.

### DIFF
--- a/tools/mksurfdata_map/README.developers
+++ b/tools/mksurfdata_map/README.developers
@@ -207,7 +207,7 @@ And here are some things to check for when making new landuse.timeseries
 datasets (which often happens at the same time, and most of the above applies
 as well):
 
-- Compare one of the productive resolution datasets to a previous version.
+- Compare one of the production resolution datasets to a previous version.
 
   - If part of it should be identical (for example the historical period) make
     sure it is identical as expected (using cprnc make sure the historical period
--- a/tools/mksurfdata_map/src/mksurfdat.F90
+++ b/tools/mksurfdata_map/src/mksurfdat.F90
@@ -711,6 +711,18 @@ program mksurfdat
 
     do n = 1,ns_o
 
+       ! Truncate all percentage fields on output grid. This is needed to
+       ! insure that wt is zero (not a very small number such as
+       ! 1e-16) where it really should be zero
+       
+       do k = 1,nlevsoi
+          pctsand(n,k) = float(nint(pctsand(n,k)))
+          pctclay(n,k) = float(nint(pctclay(n,k)))
+       end do
+       pctlak(n) = float(nint(pctlak(n)))
+       pctwet(n) = float(nint(pctwet(n)))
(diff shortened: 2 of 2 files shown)


billsacks:
At @ekluzek 's request, I also compared 4x5 surface datasets with and without this change and they, too, are identical (from `./mksurfdata.pl -no-crop -vic -glc_nec 10 -y 2000 -res 4x5`)

billsacks:
Tools tests all pass, with comparison against ctsm1.0.dev110

PR #1122, opened 2020-08-25
updates to current surfdata & landuse files in tools/contrib

wwieder:
### Description of changes
Updates sources of surfacedata and land use time series in tools/contrib
 
### Specific notes
updates source of data used by singlept script in tools/contrib directory.  

Not sure if we should also point to newer:
- datm forcing /glade/p/cgd/tss/CTSM_datm_forcing_data/atm_forcing.datm7.GSWP3.0.5d.v1.1.c181207
-surface data /glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/release-clm5.0*
  
NA

Not sure how much these files with code modificatons?

NA

NA

Testing performed, if any:
created new surface dataset for single point simulation


billsacks:
@ekluzek we can just merge this to master without testing, right?

wwieder:
yep, no changes beyond the files we're pointing to.

PR #1126, opened 2020-08-27
Hydraulic redistribution

djk2120:
### Description of changes
Adding four new history fields relevant to plant hydraulic processes:
 - QHR (patch-level hydraulic redistribution)
 - VEGWPLN (local noon vegetation water potential)
 - VEGWPPD (predawn vegetation water potential)
 - VPD_CAN: canopy vapor pressure deficit (functional input to Medlyn model)

1123

No.

No, except for the availability of new history fields.

Testing performed, if any:
--compset I2000Clm50Sp --res f09_g17_gl4
Ran a five-day global simulation to confirm that the new history fields were writing out as expected. Compared local noon and predawn vegwp to half-hourly vegwp (at one point) and found agreement between the daily average and the corresponding timeslices in the half-hourly output.

### DIFF
--- a/src/biogeophys/CanopyStateType.F90
+++ b/src/biogeophys/CanopyStateType.F90
@@ -48,6 +48,8 @@ module CanopyStateType
      real(r8) , pointer :: rscanopy_patch           (:)   ! patch canopy stomatal resistance (s/m) (ED specific)
 
      real(r8) , pointer :: vegwp_patch              (:,:) ! patch vegetation water matric potential (mm)
+     real(r8) , pointer :: vegwp_ln_patch           (:,:) ! patch vegetation water matric potential at local noon (mm)
+     real(r8) , pointer :: vegwp_pd_patch           (:,:) ! patch vegetation water matric potential from 12am-2am (mm)
 
      real(r8)           :: leaf_mr_vcm = spval            ! Scalar constant of leaf respiration with Vcmax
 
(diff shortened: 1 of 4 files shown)


ekluzek:
This could be combined with other bit-for-bit work. @negin513 perhaps this should be combined with the other bit-for-bit tag? #1102

ekluzek:
@djk2120 you want these new variables on the PPE work right? Is it required for that work or just helpful?

djk2120:
I would say these are very helpful for the PPE work, but not strictly
required.

dlawrenncar:
I think we should consider this a requirement for the PPE.

ekluzek:
@negin513 since @dlawrenncar wants this for PPE work, I assigned it to you. You could combine this with the #1102 PR or make it as a separate tag, which ever you think is more efficient for you. We should make sure a few of us review this, but it is small and should be bit-for-bit so should be easy to handle.

negin513:
I merged this as part of #1187 but during the testing I've noticed FAILs in `COMPARE_base_rest` phase. 
So, I did the following:
1. I brought this up to the master and tested it again with `clm_short` test suite and I still receive FAILs in `COMPARE_base_rest` phase in ERP and ERS tests.

2. I used this branch as is and tested it with `clm_short` test suite and I still receive FAILS in `COMPARE_base_rest` phase for ERP and ERS tests. 

Therefore, I conclude the changes here are not BFB and causing FAILS in `COMPARE_base_rest`. 

I checked the outputs to figure out what are the fields that are not BFB.
 * From my preliminary investigation, `VEGWPPD` and `VEGWPLN`are the two fields that are not BFB. For example for `ERS_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default` test :

```
 RMS VEGWPPD                          1.1501E+04            NORMALIZED  3.6253E-02
 RMS VEGWPLN                          1.3917E+03            NORMALIZED  1.5362E-03
```

Please let me know your thoughts.

djk2120:
Hi Negin,
Some of this is over my head as I am not familiar with the testing suite.
What is ERP/ERS?

Both VEGWPPD and VEGWPLN are new fields introduced by this pull request,
that would not have been in previous versions of the model. Does that
explain it?

-Daniel

ekluzek:
The ERS and ERP are exact restart tests. They are showing that the new history fields are showing different results after restart then when run all in one go. Usually, the easiest way to fix that is to just add these two to the restart file. I would do that and see if it fixes the problem. It is important to get these restart tests passing.

That said, if we don't have to have these on the restart file that would be preferred. It seems like you might be able to look into it and see if it really has to be on the restart file or not.

ekluzek (approved):
@negin513 is testing this with her PR. There are a few changes she'll make there for this PR. @djk2120 is also going to make some future changes that will resolve some other questions I had.

negin513:
This is merged with #1187.

PR #1137, opened 2020-09-01
Merge ctsm1.0.dev105 tag into fates_main_api 

glemieux:
### Description of changes
This PR brings the fates_main_api branch up to date with [ctsm1.0.dev105](https://github.com/ESCOMP/CTSM/releases/tag/ctsm1.0.dev105) tag to incorporate fixes to address #1125.

### Specific notes

Contributors other than yourself, if any:

CTSM Issues Fixed (include github issue #): #1125 

Are answers expected to change (and if so in what way)? Only fates outputs expected to change relative to ctsm1.0dev.105 tag

Any User Interface Changes (namelist or namelist defaults changes)? No

Testing performed, if any:
`aux_clm` suite outputs: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_ctsmdev105-aux_clm`
`fates` suite outputs: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_ctsmdev105-merge`

**NOTE: Be sure to check your coding style against the standard
(https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines) and review
the list of common problems to watch out for
(https://github.com/ESCOMP/CTSM/wiki/List-of-common-problems).**

PR #1149, opened 2020-09-13
LILACSMOKE: Call case.load_env() before building the atm driver

billsacks:
### Description of changes

In LILACSMOKE test: Call case.load_env() before building the atm driver. This is needed for the atm driver build to use the correct module environment.

### Specific notes

Things were working for me, and apparently for @ekluzek and @negin513 , without this change. But this is probably because our default module environment was similar to cime's cheyenne_intel environment. @glemieux was running into trouble because he has intel17 loaded in his default environment (whereas the environment loaded by cime uses intel19). This should fix that issue.

Are answers expected to change (and if so in what way)? Answers *may* change for the LILACSMOKE test, if baselines were generated with a slightly different compiler version than the one loaded by cime. However, I do not expect answer changes.

Testing performed, if any:
- I first reproduced the error reported in https://github.com/ESCOMP/CTSM/pull/1137#issuecomment-690797478 by changing my environment to match @glemieux 's. Specifically, I replaced my `.profile` with his, and replaced the contents of my `.lmod.d` directory with his. So this used intel17. This led to the same issue with the LILACSMOKE test that he reported.
- I then reran the LILACSMOKE test using @glemieux 's environment but with the fix in this PR; now it passed.
- I also ran the LILACSMOKE test using my environment (which uses intel 19.0.5) with this fix; it also passed.

### DIFF
--- a/cime_config/SystemTests/lilacsmoke.py
+++ b/cime_config/SystemTests/lilacsmoke.py
@@ -111,6 +111,12 @@ class LILACSMOKE(SystemTestsCommon):
             debug=str(self._case.get_value('DEBUG')).upper(),
             ctsm_mkfile=os.path.join(caseroot, 'lilac_build', 'ctsm.mk'))
         makecmd = 'make {makevars} atm_driver'.format(makevars=makevars)
+
+        # Normally the user will source either ctsm_build_environment.sh or
+        # ctsm_build_environment.csh before building the atmosphere model. In the context
+        # of this test case, case.load_env does the equivalent.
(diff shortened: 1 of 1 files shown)


billsacks:
@ekluzek although this isn't critical like the last LILAC-related PR (it only impacts testing, not what a typical user will experience), the impact is so limited that I feel it's safe to bring it to master without any additional testing. (I have run the one test that it impacts.) Are you okay with this? If so, I'll bring it to master.

ekluzek (approved):

PR #1151, opened 2020-09-15
Have a copy of btran2 that's just inside CNFire

### LINKED ISSUE #1144: Move btran2 to just inside of Fire model
btran2 is a bit of data in EnergyFluxes type that's just used as an alternative to btran for the Li Fire model. A scientific question would be could the fire model be tuned to use the regular value as well as the new PHS formulation? But, that would be an involved effort.

In the meantime, since this is just for use by the Li Fire model, it would be good to isolate it to just inside the fire model. Right now it's data in EnergyFluxes, it's initialized in CanopyFluxes, and then calculated in SoilMoistStress%calc_root_moist_stress_clm45default. And then it's only use is in CNFireLi2014Mod and CNFireLi2016Mod. clmfates_interfaceMod.F90 also uses it, but actually just sets it to a missing value, so it must not be used within FATES.

The calculation just involves this bit of code which is fairly well contained.

``` fortran
            s_node = max(h2osoi_vol(c,j)/watsat(c,j), 0.01_r8)

            call soil_water_retention_curve%soil_suction(c, j, s_node, soilstate_inst, smp_node_lf)

            smp_node_lf = max(smpsc(patch%itype(p)), smp_node_lf)
            btran2(p)   = btran2(p) +rootfr(p,j)*max(0._r8,min((smp_node_lf - smpsc(patch%itype(p))) / &
                    (smpso(patch%itype(p)) - smpsc(patch%itype(p))), 1._r8))
```

So, s_node, smp_node_lf would be local variables. h2osoi_vol, watsat, soil_water_retention_curve, soilstate_inst, smpsc, rootfr, smpso are then passed into the subroutine for calculating it. So it needs the types: waterstatebulk_inst, soilstate_inst, soil_water_retention_curve, and pftcon sent in. btran0 is a parameter in CanopyFluxes that is used for initialization, but it's just set to 0.0_r8, so I don't know that a big effort needs to be made to use it for btran2.

Note, #889 has some adjustments to the above, but if it's made to be inside LiFire model it'll be more straightforward to handle those changes without adding an extra switch for it outside of the existing fire_method switch.

ekluzek:
### Description of changes

Move btran2 so it's just inside of CNFire

### Specific notes

Contributors other than yourself, if any: self

  Fixes #1144

Testing performed, if any:
 Currently just running the test: SMS_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default
 Test builds and runs

### DIFF
--- a/src/biogeochem/CNFireBaseMod.F90
+++ b/src/biogeochem/CNFireBaseMod.F90
@@ -74,6 +74,7 @@ module CNFireBaseMod
   type, abstract, extends(fire_base_type) :: cnfire_base_type
     private
       ! !PRIVATE MEMBER DATA:
+      ! !PUBLIC MEMBER DATA (used by extensions of the base class):
       real(r8), public, pointer :: btran2_patch   (:)   ! patch root zone soil wetness factor (0 to 1)
 
     contains
@@ -141,7 +142,7 @@ contains
     allocate(this%btran2_patch             (begp:endp))             ; this%btran2_patch            (:)   = nan
     ! History file
     this%btran2_patch(begp:endp) = spval
-    call hist_addfld1d(fname='BTRAN2FIRE', units='unitless',  &
+    call hist_addfld1d(fname='BTRAN2', units='unitless',  &
          avgflag='A', long_name='root zone soil wetness factor', &
          ptr_patch=this%btran2_patch, l2g_scale_type='veg')
   end subroutine CNFireInit
@@ -160,7 +161,7 @@ contains
 
     logical :: readvar
 
-    call restartvar(ncid=ncid, flag=flag, varname='btran2fire', xtype=ncd_double,  &
+    call restartvar(ncid=ncid, flag=flag, varname='btran2', xtype=ncd_double,  &
          dim1name='pft', &
          long_name='', units='', &
          interpinic_flag='interp', readvar=readvar, data=this%btran2_patch)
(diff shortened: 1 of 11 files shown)


billsacks:
It looks like you haven't removed the existing btran2 calculation; is there a reason for that?

ekluzek:
I'm still getting this to work. The original btran2 and the new btran2 are different, and I'm trying to track down why and get them to match. Obviously in the end I'll remove the original one and only have the new one. and I'll need to make sure btran2 is only calculated when CN is on.

ekluzek:
Also @billsacks there will be some specific questions I have that I'll want your opinion on. But, I figure I'll wait until this is working correctly. I'll want both you and @negin513 to give me some feedback on how this is done. Mostly in terms of the OO perspective.

billsacks:
Okay, sounds good. Let me know when you want my feedback.

ekluzek:
OK, so @billsacks helped me to see the bit-for-bit issue I was having. So I have two design questions I would like to run by @billsacks and @negin513. But, I think what I should do at this point is to get the feedback -- but not change it -- until later. I could do a small refactor later to make this better. I'll just make an issue out of it, and add it to another later bit-for-bit tag.

ekluzek:
I think this is a good thing, but a consequence of this is that BTRAN2 is dropped off of history files for Sp cases now.

ekluzek:
The following tests are showing answers identical to the baseline:

SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm45FatesRsGs.cheyenne_intel.clm-FatesColdDef
ERI_D_Ld9.f10_f10_musgs.I1850Clm45Bgc.cheyenne_gnu.clm-default
ERI_D_Ld9.f10_f10_musgs.I1850Clm45Bgc.cheyenne_gnu.clm-default
ERI_D_Ld9.f10_f10_musgs.I1850Clm45Bgc.cheyenne_gnu.clm-default
ERS_Ly20_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRsGs.cheyenne_intel.clm-cropMonthOutput

billsacks:
See #1155 for my proposed changes to this PR

ekluzek:
We talked. The plan is that I've pulled this PR into my ctsm5_1 PR and that will come in as it is as the first tag.

Issue #1153, opened 2020-09-18
Fire btran2 is only computed for exposed veg patches, but is used over all veg patches


billsacks:
### Brief summary of bug

From examining the fire code, it looks like btran2 is used over all vegetated patches, but it is only computed each time step over the exposed veg patches. I think this means that, for non-exposed patches, it is using the btran2 value from whenever the patch was last exposed. This seems scientifically questionable.

### General bug information

**CTSM version you are using:** master

**Does this bug cause significantly incorrect results in the model's science?** Yes (I think... but impact is probably relatively small)

**Configurations affected:** All configurations with fire active

### Details of bug

btran2 is accessed for all vegetated patches:

https://github.com/ESCOMP/CTSM/blob/5cd28a0db17c22f10447c819edfad38d5239dab3/src/biogeochem/CNFireLi2016Mod.F90#L346-L360

yet I'm pretty sure its calculation is just done over the exposed veg filter (in calc_root_moist_stress_clm45default).

#1151 refactors this, but I don't think it changes the points over which btran2 is calculated. In fact, I noticed this when reviewing #1151, because that PR makes the filter over which btran2 is calculated more explicit.

I think this means that the btran2 values over non-exposed vegetated patches are taken from some previous time step – whenever the given patch was last exposed. I'm not sure how much of an impact this has on the results, but it feels wrong.

Probably what should be done is that some assumed btran2 value is used over non-exposed vegetated patches – probably 1. In principle this could require retuning of the fire model, but I hope the impact would be relatively small because there hopefully aren't many fires in these non-exposed (which I think implies snow-covered) points.

I'm hopeful that this change would also allow us to remove btran2 from the restart file.

ekluzek:
This is the advantage of having the fire-specific setting of btran2 in the fire code, it makes this more obvious. When they are strewn around it's harder to tell if things like this match up.

We should bring @lifang0209 into the conversation to ask about if this was intentional or not, and the impacts of changing it.

billsacks:
From @dlawrenncar and myself: if there's no exposed veg, let's set btran2 to 1. I'll fold this into the upcoming btran2 tag.

billsacks:
[...]

PR #1155, opened 2020-09-18
Calculate btran2 inside fire - and call the new routine from within the fire code

### LINKED ISSUE #1139: Decrease expense of ne0ARCTICGRISne30x8 test and C96 tests
The test `SMS_Ln9.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.ISSP585Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode` takes 470 core-hours to run, which is about 3x as expensive as any other test in the aux_clm test suite. To put this in perspective: If the aux_clm test suite is run 100 times per year (which is not at all unreasonable), this one test would use up nearly 1% of the SEWG annual allocation.

Nearly all of the time in this test is spent in model initialization – probably (though I haven't confirmed) in init_interp.

We should think about what is really important about this test, and trim it down so that it is just testing the aspects that are truly important to test here. My guess is that we're not getting any additional code coverage (of the Fortran code) from this test, so it is mostly about testing compatibility of datasets. If so, can we cover this well enough with a cold start test? Or even by just relying on build-namelist testing for this resolution?

I know that it would be ideal to have a somewhat realistic test of every supported resolution, but our aux_clm test suite is unmanageably large, so I feel we need to be a little more thoughtful here, and cut some corners even if it means slightly increasing the risk that issues will be discovered in production. I feel it's important to ensure that as much as possible of our Fortran code is covered in the test suite, but if a user occasionally runs into a problem where a run at a particular (fairly obscure) resolution doesn't get off the ground, I feel that's an acceptable risk to take.

### LINKED ISSUE #1144: Move btran2 to just inside of Fire model

billsacks:
### Description of changes

This is an extension of #1151 that additionally moves the call to the new routine so that the fire code itself calls this routine.

This has a few advantages in my mind:

- It makes the logic of CNFireArea more clear (rather than depending on
  a btran2 variable that is calculated from some other part of the code)

- This avoids having the biogeophysics depend on the biogeochemistry

- This lets us avoid doing this btran calc if using no-fire – or other,
  future, fire methods that don't need it

The placement in the driver loop differs from before; in order to get
bit-for-bit results, we need to save h2osoi_vol and used that saved
version from earlier in the driver loop.

@ekluzek if you're happy with this, then I'd propose that, before bringing this to master, we change this so that it uses the real h2osoi_vol, and get rid of the temporary saved variable that I added to get bit-for-bit. This will be answer changing, but only in that we'll be using h2osoi_vol from later in the time step when computing btran2. I'm thinking that would be an acceptable level of answer change, unless we're trying to avoid any answer changes as we finalize the PPE branch (in which case we could save this for a cleanup tag in a few weeks).

To be honest, I went into this hoping it would be clean and easy. It turned out to be a bit messier because of the need to pass some extra arguments down a few levels of the calling chain. I personally still prefer this solution (though in hindsight I'm not sure if it was worth the time I spent on it), but I'm not completely tied to it if you feel it's problematic for some reason.

I realize that this might cause conflicts when merging with your PR to bring in the clm5.1 fire code. If we want to go with this version, I'm happy to help with that merge.

### Specific notes

Resolves #1144 
Resolves #1139 

Are answers expected to change (and if so in what way)? Currently no. But if I make the final change suggested above, then answers will change for all BGC/CN tests. I expect these answer changes to be greater than roundoff but non-climate-changing.

Testing performed, if any:
- aux_clm on izumi: all pass and bfb (FIELDLIST diffs for Sp cases)
- ERP_Ly3_P72x2.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-cropMonthOutput: passes and bfb
- aux_clm on cheyenne: in progress


billsacks:
@ekluzek - in looking at this, you could either look at the full changes (which include all of your changes plus mine on top) or you could just look at my two added commits:

- e345c390332e8178a0fb814305abc3a656273f4e makes the code changes
  - Note: a somewhat-related change I made in this commit was to remove CNFireArea from CNFireBaseMod: It seemed unnecessary to have it in there, and having it there just means extra maintenance when we change the interface (like now). In addition, removing it will enforce at compile time (rather than runtime) that any fire implementations implement this method.

- c7723f9178e77f8a253281a90f64c364879ff0e3 removes the expensive tests that we talked about removing (I wanted to do this before running the test suite on this branch)

billsacks:
With the cheyenne-intel test suite partly done: I'm seeing answer changes in two tests:

- `ERI_Ld9.f45_g37.I2000Clm50BgcCruGs.cheyenne_intel.clm-nofire` changes answers in just the BTRAN2 field. This is expected because this variable is no longer calculated for the nofire case.

- `SMS_Lm1.f10_f10_musgs.I1850Clm50BgcCropCmip6waccm.cheyenne_gnu.clm-basic`: this changes answers in the BTRAN2 field but nothing else. I'll look into this a bit, but if I can't figure it out relatively quickly, I may say we should just let this go, since this doesn't appear to impact anything other than BTRAN2.

billsacks:
With all but three tests done now (the three ne0 tests), I'm seeing the following answer changes:

- nofire tests (`ERI_Ld9.f45_g37.I2000Clm50BgcCruGs.cheyenne_intel.clm-nofire`): diffs just in BTRAN2, as noted above

- dynroots tests (`ERS_D_Ld3.f19_g17_gl4.I1850Clm50BgcCrop.cheyenne_intel.clm-clm50dynroots`, `SMS_Lm1.f19_g17_gl4.I1850Clm50Bgc.cheyenne_intel.clm-clm50dynroots`, `SMS_Ly3_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRsGs.cheyenne_intel.clm-clm50dynroots`): Diffs in many fields. I expected these diffs but forgot to mention it in my initial comment: now btran2 is calculated after dynroot updates in each time step, rather than before.

- Tests where the compared clm2 h0 file includes the 0th time step (`SMS_Lm1.f10_f10_musgs.I1850Clm50BgcCropCmip6waccm.cheyenne_gnu.clm-basic`, `SMS_Lm1_D.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-output_crop_highfreq`, `SMS_Lm1_D.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-snowlayers_3_monthly`, `SMS_Lm1.f19_g17.I1850Clm50BgcCropCmip6waccm.cheyenne_intel.clm-basic`, `SMS_Ln9.ne30pg2_ne30pg2_mg17.I2000Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode`): Diffs only in BTRAN2 field. So it seems like my changes have led to a difference in BTRAN2 in the 0th time step, which doesn't appear to feed back on anything else. I can't see why this would happen, but it feels unimportant to me.

dlawrenncar:
These all seem fine.  Certainly the dynroots doesn't matter much since we
have never really gotten that feature to work and it needs to be revisited
before it can be utilized.

ekluzek:
@billsacks will pull this in as a separate tag after the ctsm5_1 tag.

The tests that change will be moved to the new test list "ctsm_sci". You should consider if #1153 should also be handled here as well.

ekluzek (changes requested):
We went over these changes together. These are some nice cleanup changes to the change's I'm making. I especially like having the calculation of the fire code btran2 in the fire model itself -- instead of outside it in biogeophys. This also fixes a problem I had there where I had to add some explicit if statements as well as doing a double "%" reaching two levels down into a class which is a bad practice.

The testing changes are going to change a bit as we discussed to move them to ctsm_sci test list.

src/biogeophys/HydrologyNoDrainageMod.F90
```diff
@@ -272,6 +273,13 @@ subroutine HydrologyNoDrainage(bounds, &
        call water_inst%TracerConsistencyCheck(bounds, 'before main snow code in HydrologyNoDrainage')
     end if
 
+    do j = 1, nlevgrnd
```
ekluzek, line 276: So is the idea that we would remove this calculation of h2osoi_vol_prehydrology as an answer changing tag when we are ready to do a larger answer change? Should we start planning for such an answering change tag?

billsacks: @ekluzek @dlawrenncar - Sorry, I wasn't very clear in describing this. My plan, if it's okay with both of you, is to make this additional answer change as part of the tag where I bring these changes in. So my intention was: after @ekluzek makes the ctsm5.1 fire tag, I would:
- Merge that into this branch
- Remove the temporary `h2osoi_vol_prehydrology`, so that the btran2 calculation depends on the updated version of h2osoi_vol in the time step
- Run test suite

This will change answers for all CN/BGC cases because btran2 will now use the later-in-timestep version of h2osoi_vol. My sense is that this shouldn't have a significant impact, but I can also hold off on making that answer change if you want to be conservative about answer changes right now.

billsacks:
> You should consider if #1153 should also be handled here as well.

It felt to me like that should go in a different tag, since it would probably be a more significant change and would likely require some scientific sign-off.

src/biogeophys/HydrologyNoDrainageMod.F90
ekluzek: OK, makes sense.

billsacks:
@ekluzek I have reworked the testing of ne0 and C96, I think according to our discussion. You can let me know if you have any feedback.

In case it helps:

```
$ ./query_testlists --xml-category ctsm_sci
ctsm_sci: SMS_Ln9.C96_t061.I2000Clm50SpRsGs.cheyenne_intel.clm-default                                                          # We have one C96 test in aux_clm; this is another that uses a different compset. No need to run this additional C96 test with every tag, but include it in the less frequent ctsm_sci testing.
ctsm_sci: SMS_Ln9.C96_C96_mg17.IHistClm50Sp.cheyenne_intel.clm-decStart                                                         # We have one C96 test in aux_clm; this is another that uses a different compset. No need to run this additional C96 test with every tag, but include it in the less frequent ctsm_sci testing. decStart doesn't accomplish a lot in this very short test, but we're using it anyway to be consistent with other Hist tests.
ctsm_sci: ERS_Ln9.ne0ARCTICne30x4_ne0ARCTICne30x4_mt12.IHistClm50SpGs.cheyenne_intel.clm-clm50cam6LndTuningMode_1979Start       # Run ARCTIC for transient case starting in 1979 as for AMIP CAM cases, be sure to run with stub GLC
for ERS test as otherwise it won't work for a sub-day test (no need to run this high core count test with every tag, but include it in the less frequent ctsm_sci testing)"
ctsm_sci: SMS_Ln9.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.IHistClm50Sp.cheyenne_intel.clm-clm50cam6LndTuningMode_1979Start # Run ARCTICGRIS for transient case starting in 1979 as for AMIP CAM cases (no need to run this high core count test with every tag, but include it in the less frequent ctsm_sci testing)"
ctsm_sci: SMS_Ln9.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.ISSP585Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode    # Run ARCTICGRIS for future transient case (do not run this expensive test with every tag, but include it in the less frequent ctsm_sci testing)"
ctsm_sci: SMS_Ln9.ne0CONUSne30x8_ne0CONUSne30x8_mt12.IHistClm50Sp.cheyenne_intel.clm-clm50cam6LndTuningMode_2013Start           # Run CONUS for transient case starting in 2013 as for CAM case (no need to run this high core count test with every tag, but include it in the less frequent ctsm_sci testing)"
```

```
$ ./query_testlists --xml-category aux_clm | grep -e ne0 -e C96
aux_clm: SMS_Ln9.C96_C96_mg17.IHistClm50BgcCrop.cheyenne_intel.clm-default                                               # Want one C96 test in the aux_clm test suite; just a short smoke test to make sure it can get off the ground
```

ekluzek:
This looks like what I expected from our discussion. Something I just thought of though is that of these are going to be nine step tests you should use compsets with stub glacier and stub river.

billsacks:
> Something I just thought of though is that of these are going to be nine step tests you should use compsets with stub glacier and stub river.

I know that's needed for restart tests, but I didn't think it was needed for smoke tests; is it?

ekluzek:
Ahh, yes you are right! It's just the restart tests that can't do the 9 step tests.

billsacks:
@ekluzek you have this marked as changes requested, but I believe I addressed all of your earlier concerns (which I think just related to the test suite); can you please mark this as approved if you indeed approve now?

src/biogeophys/HydrologyNoDrainageMod.F90
ekluzek: This was just a question. So resolving this.

ekluzek (approved):

ekluzek:
Yes, looks good.

billsacks:
Testing on 0a970625, where I have merged up to the latest master and done the same refactoring for the 2021 fire version: I ran all Clm51 cases in the aux_clm test list, plus a single long Clm45 and Clm50 case:

```
ERI_D_Ld9.f10_f10_musgs.I1850Clm51BgcGs.cheyenne_gnu.clm-default
ERI_N2_Ld9.f19_g17.I2000Clm51BgcCropGs.cheyenne_intel.clm-default
ERP_Ld9.f45_g37.I2000Clm51BgcGs.cheyenne_intel.clm-default
ERP_D.f10_f10_musgs.IHistClm51BgcGs.cheyenne_gnu.clm-decStart
ERP_D.f10_f10_musgs.IHistClm51BgcGs.cheyenne_intel.clm-decStart
ERP_D_Ld10_P36x2.f10_f10_musgs.IHistClm51BgcCropGs.cheyenne_intel.clm-ciso_decStart
ERP_D_P36x2_Ld3.f10_f10_musgs.I2000Clm51BgcCropGs.cheyenne_intel.clm-coldStart
ERP_D_P36x2_Ld5.f10_f10_musgs.I2000Clm51BgcCropGs.cheyenne_intel.clm-irrig_spunup
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm51BgcGs.cheyenne_intel.clm-monthly
ERP_P36x2_Lm13.f10_f10_musgs.IHistClm51BgcGs.cheyenne_gnu.clm-monthly
ERP_P72x2_Lm25.f10_f10_musgs.I2000Clm51BgcCropGs.cheyenne_intel.clm-monthly
ERS_Ly5_P144x1.f10_f10_musgs.IHistClm51BgcCropGs.cheyenne_intel.clm-cropMonthOutput
LVG_Ld5_D.f10_f10_musgs.I1850Clm51BgcGs.cheyenne_intel.clm-no_vector_output
LCISO_Lm13.f10_f10_musgs.IHistClm51BgcCropGs.cheyenne_intel.clm-ciso_monthly
SMS_Lm13.f19_g17.I2000Clm51BgcCropGs.cheyenne_intel.clm-cropMonthOutput
SSP_D_Ld10.f19_g17.I1850Clm51BgcGs.cheyenne_intel.clm-rtmColdSSP

ERP_D_Ld5_P48x1.f10_f10_musgs.I1850Clm51BgcGs.izumi_nag.clm-ciso
ERP_D_P48x1.f10_f10_musgs.IHistClm51BgcGs.izumi_nag.clm-decStart
PEM_Ld1.f10_f10_musgs.I2000Clm51BgcCropGs.izumi_intel.clm-crop
SMS_D.f10_f10_musgs.I2000Clm51BgcCropGs.izumi_intel.clm-crop
SMS_D.f10_f10_musgs.I2000Clm51BgcCropGs.izumi_pgi.clm-crop
SMS_D.f10_f10_musgs.I2000Clm51BgcCropGs.izumi_gnu.clm-crop
SMS_Ld5_D_P48x1.f10_f10_musgs.IHistClm51BgcGs.izumi_nag.clm-decStart

```

All of these tests pass and are bit-for-bit. This gives me confidence that the merge and the application to the 2021 fire version went smoothly. I'm now moving on to some answer- changing pieces of this tag.

billsacks:
I reran all of the testing noted in https://github.com/ESCOMP/CTSM/pull/1155#issuecomment-702250123 from f7a98bf10; all of these are still bit-for-bit with ctsm5.1.dev006.

billsacks:
@ekluzek My preliminary testing looks good. I'm running final testing on this tonight. While developing, I separated these changes into batches of answer-changing and bit-for-bit (or possibly changing the BTRAN2 history field and nothing else) changes. The answer-changing commits are small (ec8d075d5, b1cf8b9f9, 4291f064f). I generated baselines with those answer changes, and then am testing the final changes against those baselines.

I don't think you need to review these final changes (beyond what you have already reviewed): the answer changes are small and I'm fairly confident that I have accurately done those as we have discussed; as for the remaining larger changes, most/all are things we already discussed, and the test suite should pick up issues if there were problems in my refactor. That said, if you do look them over, I recommend turning on the setting to ignore whitespace changes in the diffs, since about 2/3 of the changes are just whitespace changes from refactoring some loops as we discussed this afternoon.

Issue #139, opened 2017-12-16
Remove or reduce duplication in Photosynthesis code for Plant Hydraulic Stress


ekluzek:
**Erik Kluzek < erik > - 2016-06-24 14:09:30 -0600**
**Bugzilla Id:** [2323](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2323)
**Bugzilla CC:**  andre,  dlawren,  oleson,  rfisher,  sacks, 

clm4_5_8_r182 brings in plant hydraulic stress (PHS) as an option to the code. The older version needs to stay in place as well to support clm4_5 answers. The new option is brought in largely by duplicating the code in PhotosynthesisMod.F90 for PHS. Almost all subroutines have a PHS version and a non-PHS version. The effects of at least six namelist control items is also reproduced: lnc_opt, reduce_dayl_factor, vcmax_opt, leafresp_method, leafacc, and light_inhibit. Thus any testing for these control items must be duplicated for both PHS on and off. And the size of the file more than doubled from 2k to 2.5k.

This duplication leads to double maintenance of code where changes in the PHS section of the code has to be duplicated in the non-PHS section of the code. It also leads to bugs where a change is made on one side and not the other.

Because, there is so much code duplication this can be managed better by breaking it up into smaller segments that can be called by both PHS and non-PHS versions. The top section of Photosynthesis is almost identically duplicated in PhotosyntesisHydraulicStress until the loop through canopy layers above snow. The sections after that are also duplicated with the PHS setting both sun and shade. I think the duplication can be removed if it's broken up into three parts by three loops -- calculate_respiration loop, leaf-level-photosynthesis-and-stomotal-conductance, and canopy-photosynthesis-and-stomotal-conductance. The second section includes the call to hybrid or hybrid_PHS.

Other ways of handling the duplication are to set magic numbers in one place as parameters in the code, and to break up the namelist control items into small subroutines/functions that can be called in either branch. Thus these control items are tested whether called from the PHS branch or the non-PHS branch. At least some of these methods should be used to reduce the code duplication.

ekluzek:
**Erik Kluzek < erik > - 2016-06-24 14:34:37 -0600**

Also the statement functions are reproduced. The statement functions need to be removed and the module versions used. We'll need to check that this doesn't change answers for PHS on or off.

djk2120:
I shared some thoughts on this today at the CLM meeting. 

The CLM4.5 code structure was not amenable to PHS:
[...]

Issue #440, opened 2018-07-11
leafcn_max < target leafcn?


wwieder:
It looks like the parameter leafcn_max is less than the target leafcn for some pfts? I don't yet understand how this effects simulations, but it doesn't look like the stoichiometry of other vegetation pools (livewdcn) has similar issues. 

### General bug information

**CTSM version you are using:** [output of `git describe`]
clm4518/n04/cime5.4.0-alpha.03

**Does this bug cause significantly incorrect results in the model's science?** [Yes / No]
Yes? But not yet tested.

**Configurations affected:** [Fill this in if known.]
default clm5 or any simulations with active bgc and flexCN

### Details of bug
ncdump -v leafcn,leafcn_min,leafcn_max /glade/p/cesmdata/cseg/inputdata/lnd/clm2/paramdata/clm5_params.c171117.nc

### Important details of your setup / configuration so we can reproduce the bug
to correct this, leafcn_min and leafcn_max should be +/- 10 from the target leafcn (as with livewdcn).
We'll need to do testing to see how this potentially changes answers.

susancheng:
I ran a point simulation for NET with the original parameter file (where leafcn > max leaf cn) and another where I set leaf cn < max leaf cn. There wasn't much of a difference in the leaf C:N. But it's confusing why max leaf cn in the parameter file seems to be ignored.  

Related to #443, deadstem C:N was ~290 and livestem C:N was ~28, even though the parameter file sets the target deadwdcn is 500 and livewdcn is 50.

ekluzek:
See this related note in #290.

> the +- 10 or +-15s used in the leafcn_min leafcn_max calc in NutrientCompetitionFlexibleCNMod.F90 ought to all be based on an uptake number and a burn-off number
> e.g. change "frac_resp = (leafcn_actual - leafcn_max) / 10.0_r8" to "frac_resp = (leafcn_actual - leafcn_max) / 15.0_r8" and then make the 15 an input parameter (organ_CN_offoptimal_range_uptake and organ_CN_offoptimal_range_resp)

ekluzek:
I thought that this might be something that was being fixed on the hardcodep branch, but in looking at the code -- it's not. It's still hardwired. But, from above there is still different opinions on how this should be resolved.

wwieder:
This is confusing

I'd suggest removing the min and max values from the parameter file and using the range defined in the code. **This is may be easiest**.

[...]

PR #1164, opened 2020-09-27
Some more hardcoded parameters to the params file

### LINKED ISSUE #808: Remove camdenNJ test from prebeta test list, replace it with working tests
From #64 

Decision from today's clm-cmt meeting: get rid of this test, and get rid of support for camdenNJ grid entirely.

In its place we will either add:

(1) A single test in the prebeta test list that tests single point (with global forcing data) using an ERI test; then also add the same test to aux_clm

or

(2) Two tests in the prebeta test list: one that tests single point (with global forcing data), and another that is an ERI test of a global configuration; both of those should be tests that we already have in aux_clm.

### LINKED ISSUE #886: Warning if FUN is on, but FlexibleCN is not
According to @wwieder , you should only be able to use `use_fun` if you are also using `use_flexibleCN`. i.e., it should be an error if `use_fun` is true but `use_flexibleCN` is not. Currently, CLMBuildNamelist allows you to set `use_fun` without `use_flexibleCN` (by modifying `user_nl_clm` in a CLM4.5 compset). I haven't checked whether this causes a runtime failure, but I don't see any runtime error check in controlMod, which is where I would expect it. We should add checks in both CLMBuildNamelist and in controlMod that prevent this combination of `use_fun = .true.` but `use_flexibleCN = .false.`.

@wwieder should we also prevent the reverse? i.e., is it okay to set `use_flexibleCN = .true.` with `use_fun = .false.`?

### LINKED ISSUE #1168: Remove duplication between phys versions in CLM_BLDNML_OPTS settings (and fix some others) in config_component.xml
When reviewing the changes for ctsm5.1.dev001, I noticed that we have duplication here:

https://github.com/ESCOMP/CTSM/blob/2b4802d568900df5be8f320fa61b81eb58bd8a8d/cime_config/config_component.xml#L171-L198

Given that this section already has felt error-prone to me, I feel that we shouldn't make it worse by duplicating settings for different physics versions. So, for example, rather than these three lines:

```xml
<value compset="_CLM45%[^_]*SP"		 >-bgc sp</value>
<value compset="_CLM50%[^_]*SP"		 >-bgc sp</value>
<value compset="_CLM51%[^_]*SP"		 >-bgc sp</value>
```

(and we'll need more in the future)

I feel we should have a single line like this:

```
<value compset="_CLM[^_]*%SP"		 >-bgc sp</value>
```

I think the same applies to:
- CLM_NML_USE_CASE: the match for 1850_noanthro_control needs to be changed to work for other phys versions
- CLM_USER_MODS

@ekluzek it looks like you intentionally removed some options (like CNDV) for CLM51. I think this removal can still be accomplished by leaving these out of the description section, as you have done.

### LINKED ISSUE #1185: Remove wrtdia
I noticed that our standard testing (clm-default testmod) sets `wrtdia = .true.`. Is this actually important, either for testing or in general? It looks like it outputs a global average temperature every time step. I have never personally looked at this. It seems like this adds an unnecessary communication synchronization point. I noticed that, at least in the first test I checked (`ERS_Ly5_P72x1.f10_f10_musgs.IHistClm45BgcCrop.cheyenne_intel.clm-cropMonthOutput`) this subroutine takes a whopping 1.4% of the land run time, in terms of max time. My guess is that this is just from the synchronization time rather than it actually taking a long time, but even so, I'm wondering if we can remove this from our testing – and possibly remove this option entirely.

### LINKED ISSUE #1245: intrinsic water use efficiency diagnostic
We would like to add intrinsic water use efficiency (iWUE) as a history field.

You can compute iWUE from half hourly data with standard output:
iWUE = FPSN / (1e-6*(GSSUN * LAISUN+GSSHA * LAISHA))

However, the monthly mean of iWUE, is not equivalent to calculating iWUE using monthly mean FPSN,GSSUN,LAISUN, etc. Therefore we would like to add iWUE as a diagnostic variable, such that daily and monthly means will accurately capture iWUE.

### LINKED ISSUE #1250: Error in HumanIndexMod (QSat_2 subroutine)
### Brief summary of bug

Calculation of derivative in QSat_2 is missing a term (reported by J. Buzan).  This only affects diagnostic quantities that use the advanced wet bulb temperature calculation (when calc_human_stress_indices = 'ALL').   The corrected web bulb temperature and associated indices differ by +-0.00001 K in a 1 month simulation.

### General bug information

**CTSM version you are using:** ctsm5.1.dev021

**Configurations affected:** Any that use calc_human_stress_indices = 'ALL' (not the default in CLM/CTSM).

### Details of bug

This line in QSat_2:

fdT = -lambd_a*(1._r8/T_k + vkp*de_mbdT/pminuse + gdT)

Should be:

fdT = -lambd_a*(1._r8/T_k + vkp*de_mbdT/pminuse + gdT) * foftk

Corrected in my branch:

https://github.com/olyson/ctsm/tree/fixhumanindexmod

### LINKED ISSUE #1284: change where stem_biomass is set
These are changes that @olyson is running with for the PPE work. Because stem_biomass should be set to zero in the InitCold method, I think it's equivalent to what's currently being done. But I need to check for sure. This moves the setting of stem_biomass as well as the if woody statement up to where leaf_biomass is set. It also clearly sets stem_biomass to zero there. So it looks like this...

```
            ! calculate vegetation physiological parameters used in biomass heat storage
            if (use_biomass_heat_storage) then
               ! Assumes fbw (fraction of biomass that is water) is the same for leaves and stems
               leaf_biomass(p) = max(0.0025_r8,leafc(p)) &
                    * c_to_b * 1.e-3_r8 / (1._r8 - fbw(ivt(p)))

               if (woody(ivt(p)) == 1._r8) then
                  stem_biomass(p) = (spinup_factor_deadwood*deadstemc(p) + livestemc(p)) &
                       * c_to_b * 1.e-3_r8 / (1._r8 - fbw(ivt(p)))
               else
                  stem_biomass(p) = 0._r8
               end if
            else
               leaf_biomass(p) = 0._r8
               stem_biomass(p) = 0._r8
            end if
```

### LINKED ISSUE #1296: Make a checklist of issues to check for when running a simulation.
We need to create a README file that gives some suggestions on things to check for your simulation to make sure it's correct.

@olyson

Here's things I can think of right now:

-  Make sure CLM_ env settings are correct (./xmlquery -p CLM)
- Make sure you are using the correct CLM_PHYSICS_VERSION
- Make sure you are running the appropriate overall CLM vegetation model. The "-bgc" option of either Satellite Phenology (sp), or Full BioGeoChemistry (bgc), or FATES (fates) (./xmlquery CLM_BLDNML_OPTS)
-  Also if you are running the bgc model, check to see if you should be running the prognostic crop model (option -crop in CLM_BLDNML_OPTS)
-  Make sure the LND_TUNING_MODE is correct (./xmlquery LND_TUNING_MODE)
-  For an "I compset" make sure you are running over the right forcing years (usually ./xmlqueary -p ATM_CLMNCEP_YR)
- First and align year for streams should be the start year of a historical simulation (grep stream_year_first; grep model_year_align)
- Last year for streams should be the last year you are going to run to (or beyond it)  (grep stream_year_last)
-  Make sure you are starting from appropriate spunup initial conditions (check finidat or nrevsn)

Should we also give instructions on how to do a spinup sequence? Maybe that should wait until CN-Matrix comes to CTSM main-dev.

### LINKED ISSUE #1379: Invalid xml for namelist_definition file: missing end tag
There is a missing `</entry>` tag here:

https://github.com/ESCOMP/CTSM/blob/c9775200b5e3ad0c1cbbffb94c396bc47b7e3d4c/bld/namelist_files/namelist_definition_ctsm.xml#L334-L336

I'm surprised this doesn't cause any problems in parsing the xml, but in any case, we should fix it.

### LINKED ISSUE #1382: write out friction velocity to history files by default
The NEON simulations (and other tower simulations need this).

Also update the long-name to remove 'dust model'

### LINKED ISSUE #1399: Some cleanup for arctic luna changes
There were some comments from #947 that didn't make it into the final PR #990.

Make onset_thresh a logical rather than a real. This should cleanup the logic around this...

https://github.com/ESCOMP/CTSM/pull/947#discussion_r397487334

Move some constants to parameter file. Improve some of the logic.

https://github.com/ESCOMP/CTSM/pull/947#discussion_r397502359

Change soila10 from patch to column

https://github.com/ESCOMP/CTSM/pull/947#discussion_r398080425

Another parameter

https://github.com/ESCOMP/CTSM/pull/947#discussion_r400453574

### LINKED ISSUE #1401: 1-Degree methane finundated stream data is labeled as if it's 2-degree
### Brief summary of bug
In namelist_defaults_ctsm.xml - the specification of the finundated stream seems incorrect

### General bug information
In namelist_defaults_ctsm.xml - the specification of the finundated stream seems incorrect

**CTSM version you are using:** ctsm5.1.dev038

**Configurations affected:** Any configuration where methane is used (e.g. %Bgc)

### Details of bug
The following is the logic in namelist_defaults_ctsm.xml
```
<finundation_res finundation_method="TWS_inversion">1.9x2.5</finundation_res>
<finundation_res finundation_method="ZWT_inversion">1.9x2.5</finundation_res>

<stream_fldfilename_ch4finundated finundation_method="TWS_inversion" hgrid="1.9x2.5"
>lnd/clm2/paramdata/finundated_inversiondata_0.9x1.25_c170706.nc</stream_fldfilename_ch4finundated>
<stream_fldfilename_ch4finundated finundation_method="ZWT_inversion" hgrid="1.9x2.5"
>lnd/clm2/paramdata/finundated_inversiondata_0.9x1.25_c170706.nc</stream_fldfilename_ch4finundated>
```
In CLMBuildNamelist.pm the following logic is used to parse this:
```
my $finundation_method = remove_leading_and_trailing_quotes($nl->get_value('finundation_method' ));
add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'finundation_res',
          'finundation_method'=>$finundation_method );
add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_fldfilename_ch4finundated',
         'finundation_method'=>$finundation_method,
         'finundation_res'=>$nl->get_value('finundation_res') );
add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'use_aereoxid_prog',
```
So it looks like

1. The only file is for 0.9x1.25 - so the finundation_res_finundation_methods seems wrong (although this variable does not even appear in the namelist_defaults_ctsm.xml file).
2. hgrid="1.9x2.5" is not used

So unless I am not seeing something it seems like all you want in the namelist_defaults_ctsm.xml is
```
<stream_fldfilename_ch4finundated finundation_method="TWS_inversion"
>lnd/clm2/paramdata/finundated_inversiondata_0.9x1.25_c170706.nc</stream_fldfilename_ch4finundated>
<stream_fldfilename_ch4finundated finundation_method="ZWT_inversion"
>lnd/clm2/paramdata/finundated_inversiondata_0.9x1.25_c170706.nc</stream_fldfilename_ch4finundated>
```

[Specify anything relevant: the compset, resolution, machine, compiler, any xml or namelist changes, etc. You don't have to repeat anything that you have already noted above.]

[Fill this in with anything relevant that you haven't already noted; if there is nothing to add, delete this section.]

### LINKED ISSUE #1403: Invalid test running single-point forcing data before or after it's valid data series
### Brief summary of bug

@mvertens noticed this running all of the tests with NUOPC. This actually points out a legit problem as using extend to extrapolate the hourly forcing data before or after it is invalid.

**SMS_Ly1_Mmpi-serial.1x1_vancouverCAN.I1PtClm50SpRs.cheyenne_gnu.clm-output_sp_highfreq**

### General bug information

**CTSM version you are using:** ctsm5.1.dev043

Running before or after meteorological forcing data is invalid and causes invalid scientific results.

**Configurations affected:** Running I1Pt compsets before or after it's valid data period

### Details of bug

Running this with NUOPC properly detects this as a problem since it has taxmode as limit, while mct has taxmode as cycle, so it improperly allows this to happen.

The case is running too long for data that only lasts for 13 days. It should only run for 12 days to make sure it's within the time-span of the data.

This is what NUOPC test run shows for this case. MCT isn't failing (even though it really should).
>  ERROR: (shr_stream_findBounds) ERROR: rDateIn gt rDategvd limit true

### LINKED ISSUE #1427: Division of critical day length by 2 is incorrect
This comes from this comment in a previous PR...

https://github.com/ESCOMP/CTSM/pull/947#discussion_r399663433

It refers to this section of code...

```
        ! Note: The divide by 2 of crit_dayl is just to make sure
        ! onset doesn't happen with less than 6 hours of daylight
        else if (season_decid_temperate == 0 .and.  onset_gddflag == 1.0_r8 .and. &
                soila10 > SHR_CONST_TKFRZ .and. &
                t_a5min > SHR_CONST_TKFRZ .and. ws_flag==1.0_r8 .and. &
                dayl>(crit_dayl/2.0_r8) .and.  snow_5day<snow5d_thresh_for_onset) then
```

The problem is that the description for the reason for the divide by 2 is incorrect, as it allows onset to happen at 5 hours of daylight rather than 10. The description makes it sound like it makes it harder when it's actual function is to make it easier for onset to happen.

### LINKED ISSUE #1468: Remove vestiges of ch4 0.9x1.25 grid in mksurfdata_map/mkmapdata
The methane input was moved from the surface dataset to it's own stream file, but there are still vestiges of it in mkmapdata and mksurfdata_map. All of this is now uneeded so can be removed. This would also remove the need of creating mapping files at 0.9x1.25 resolution which would save some time and complexity. Currently this unneeded input is added to the surface datasets, so it also causes confusion.

### LINKED ISSUE #1469: Remove unused 0.1x0.1 grid
Remove the unused 0.1x0.1 grid, which was partially removed with work on mksurfdata_map, but not fully removed.

### LINKED ISSUE #1472: hist_master_list_file dies if master_list_file.rst already exists
### Brief summary of bug

When hist_master_file is active if you've already written the file it dies on open.

### General bug information

**CTSM version you are using:** ctsm5.1.dev053
**Configurations affected:** master_list_file.rst=.true.
### Details of bug

The model aborts on trying to open the master_list_file.rst if it already exists. We really want to have it just overwrite the file.

Run ERP_P36x2_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs twice and it'll abort the second time.

```
At line 382 of file /glade/work/erik/ctsm_hardcodep/src/main/histFileMod.F90 (unit = 96)
0:Fortran runtime error: Cannot open file 'master_list_file.rst': File exists
0:
0:Error termination. Backtrace:
0:#0  0x2b0e2c4c786c in already_open
0:	at ../../../libgfortran/io/open.c:720
0:#1  0x61db40 in __histfilemod_MOD_hist_printflds
0:	at /glade/work/erik/ctsm_hardcodep/src/main/histFileMod.F90:382
0:#2  0x553046 in __clm_initializemod_MOD_initialize2
0:	at /glade/work/erik/ctsm_hardcodep/src/main/clm_initializeMod.F90:369
0:#3  0x5146d1 in __lnd_comp_mct_MOD_lnd_init_mct
0:	at /glade/work/erik/ctsm_hardcodep/src/cpl/mct/lnd_comp_mct.F90:241
0:#4  0x430cfe in __component_mod_MOD_component_init_cc
0:	at /glade/work/erik/ctsm_hardcodep/components/cpl7/driver/main/component_mod.F90:248
0:#5  0x41df9e in __cime_comp_mod_MOD_cime_init
0:	at /glade/work/erik/ctsm_hardcodep/components/cpl7/driver/main/cime_comp_mod.F90:1421
0:#6  0x42cd25 in cime_driver
0:	at /glade/work/erik/ctsm_hardcodep/components/cpl7/driver/main/cime_driver.F90:122
0:#7  0x42cf17 in main
0:	at /glade/work/erik/ctsm_hardcodep/components/cpl7/driver/main/cime_driver.F90:23
-1:MPT ERROR: MPI_COMM_WORLD rank 0 has terminated without calling MPI_Finalize()
-1:	aborting job
t
```

ekluzek:
### Description of changes

Brings more hardcoded parameters to the paramsfile. Some bit-for-bit cleanup. A fix for HumanStressIndex. Adding IWUELN and VPD_2M diagnostic to history output. Add in ability to run Medlyn photosynthesis for non-PHS cases (not turned on yet though).

I have added two new diagnostic fields available for history outuput.
IWUELN: the intrinsic water use efficiency at local noon (umolCO2/molH2O)
VPD_2M: vapor pressure deficit at 2m reference height (pascals)

The Medlyn stomatal conductance functionality was deployed for CLM5. It had not been implemented for use_hydrstress=.false (i.e. SMS).
Previously if we opted for:
use_hydrstress=.false and stomatalcond_method = 'Medlyn2011', the model would actually use the Ball-Berry stomatal conductance model.

### Specific notes

Contributors other than yourself, if any: @olyson @djk2120 J Buzan

Both are patch-level variables. I have added VPD_2M for users who want to calculate ecosystem scale inherent water use efficiency relevant for flux tower point simulations.

  Fixes #1472 Done
  Fixes #1469 Done
  Fixes #1468 Done
  Fixes #1427 Done
  Fixes #1403 Done
  Fixes #1401 Done
  Fixes #1399 Done
  Fixes #1296 Done
  Fixes #1245 Done
  Fixes #1250 Done
  Fixes #1284 Done
  Fixes #1382 Done
  Fixes #1379 Done
  Fixes #1185 Done
  Fixes #1168 Done
 Partial work on #1356 (but use_lch4, use_nitrif_dentrif remain)
  Fixes #886 Done
  Fixes #808 Done

  Medlyn change relates to #139

Are answers expected to change (and if so in what way)? No (a few diagnostic fields change for one test)

extra-outputs test mod changes answers for diagnostic variables because of fix in human-stress module
Many answers will change when use_hydrstress=.false., stomatalcond_method = 'Medlyn2011' including FPSN, QVEGT, etc.
use_vertsoilc=FALSE and BGC=CN are no longer options
There are some new options for handling arctic seasonal deciduous plants which also changes answers if activated

There are two new diagnostic variables for Medlyn-PHS

Two new history fields: IWUELN and VPD_2M

New params file. Remove some namelist items.

Testing performed, if any: regular OK


ekluzek:
As part of this I'd like to try to make sure fields that are on the paramsfile, but aren't read in the code are removed. See #440.

ekluzek:
OK, I've updated and brought in the hardcodep3 changes and I see the following 22 tests that unexpectedly show answer changes. Most are Clm51Bgc, but there are two Sp, one Fates, and one Clm50 (no Clm45). All are at least several days, and some are longer. Both gnu and intel showup. I've only tested on cheyenne so far...

FAIL ERI_D_Ld9.f10_f10_mg37.I1850Clm51Bgc.cheyenne_gnu.clm-default BASELINE ctsm5.1.dev042: DIFF
FAIL ERI_N2_Ld9.f19_g17.I2000Clm51BgcCrop.cheyenne_intel.clm-default BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D.f10_f10_mg37.IHistClm51Bgc.cheyenne_gnu.clm-decStart BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D.f10_f10_mg37.IHistClm51Bgc.cheyenne_intel.clm-decStart BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D_Ld10_P36x2.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-ciso_decStart BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D_Ld10_P36x2_Vnuopc.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-ciso_decStart BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D_Ld5.f10_f10_mg37.I2000Clm51Sp.cheyenne_intel.clm-decStart BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D_Ld5.f10_f10_mg37.IHistClm51Sp.cheyenne_intel.clm-default BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D_P36x2_Ld3.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-coldStart BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_D_P36x2_Ld5.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-irrig_spunup BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_Ld9.f45_g37.I2000Clm51Bgc.cheyenne_intel.clm-default BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_P36x2_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_P36x2_Lm13.f10_f10_mg37.IHistClm51Bgc.cheyenne_gnu.clm-monthly BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_P36x2_Lm13.f10_f10_mg37.IHistClm51Bgc.cheyenne_intel.clm-monthly BASELINE ctsm5.1.dev042: DIFF
FAIL ERP_P72x2_Lm25.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-monthly BASELINE ctsm5.1.dev042: DIFF
FAIL ERS_Ly5_P144x1.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-cropMonthOutput BASELINE ctsm5.1.dev042: DIFF
FAIL LCISO_Lm13.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-ciso_monthly BASELINE ctsm5.1.dev042: DIFF
FAIL LVG_Ld5_D.f10_f10_mg37.I1850Clm51Bgc.cheyenne_intel.clm-no_vector_output BASELINE ctsm5.1.dev042: DIFF
FAIL SMS_D_Ld5.f45_f45_mg37.I2000Clm51Fates.cheyenne_intel.clm-FatesColdDef BASELINE ctsm5.1.dev042: DIFF
FAIL SMS_D_Vnuopc_Mmpi-serial.CLM_USRDAT.I1PtClm51Bgc.cheyenne_intel.clm-NEON_NIWO BASELINE ctsm5.1.dev042: DIFF
FAIL SMS_Lm13.f19_g17.I2000Clm51BgcCrop.cheyenne_intel.clm-cropMonthOutput BASELINE ctsm5.1.dev042: DIFF
FAIL SSP_D_Ld10.f19_g17.I1850Clm51Bgc.cheyenne_intel.clm-rtmColdSSP BASELINE ctsm5.1.dev042: DIFF

ekluzek:
@wwieder @olyson and @lmbirch89 could you review CNPhenology in this PR in light of #1409? I have comments complete now and would like you to look it over.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -125,11 +155,12 @@ subroutine CNPhenologyReadNML( NLFilename )
     integer :: ierr                 ! error code
     integer :: unitn                ! unit for namelist file
 
+    character(len=25) :: min_crtical_dayl_method   ! Method to determine critical day length for onset
```
olyson, line 158: Should these be "critical" instead of "crtical" for consistency?

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -166,6 +209,45 @@ subroutine CNPhenologyReadNML( NLFilename )
...
+       call endrun(msg="ERROR critical_daylight_method "//errmsg(sourcefile, __LINE__))
+    end if
+  end subroutine CNPhenologySetNML
+  
+  !-----------------------------------------------------------------------
+  subroutine CNPhenologySetParams( )
+    !
+    ! !DESCRIPTION:
+    ! Set the parameters for unit-testing
+    !
```
olyson, line 234: Should the following have unit definitions, e.g., seconds, etc?

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -372,6 +457,31 @@ subroutine CNPhenologyInit(bounds)
...
+            call endrun(msg="ERROR crit_dayl can NOT be higher than seconds in a day"//errmsg(sourcefile, __LINE__))
+        end if
+    end if
+    if ( (critical_daylight_method == critical_daylight_depends_on_lat) .or. &
+         (critical_daylight_method == critical_daylight_depends_on_latnveg) )then
+        if ( params_inst%crit_dayl_lat_slope <= 0.0_r8 )then
+            call endrun(msg="ERROR crit_dayl_lat_slope can not be negative or zero"//errmsg(sourcefile, __LINE__))
+        end if
+        if ( params_inst%crit_dayl_lat_slope >= (secspday - params_inst%crit_dayl_at_high_lat)/ &
+                                                 (90._r8 - critical_offset_high_lat) )then
```
olyson, line 480: I don't understand "can not higher than allowing a day at the poles"

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+    !
+    ! !LOCAL VARIABLES:
+    !-----------------------------------------------------------------------
+
+    select case( critical_daylight_method )
+    ! Critical day length depends on both vegetation type and latitude
+    case(critical_daylight_depends_on_latnveg)
+        ! Critical day length for offset is fixed for temperate type vegetation
+        if ( pftcon%season_decid_temperate(patch%itype(p)) == 1 )then
+           crit_daylat = crit_dayl
```
olyson, line 1058: longer and shorter?

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+  end function SeasonalCriticalDaylength
+
+  !-----------------------------------------------------------------------
+  function SeasonalDecidOnset( onset_gdd, onset_gddflag, soilt, soila10, t_a5min, dayl, &
+                               snow_5day, ws_flag, crit_onset_gdd, season_decid_temperate ) &
+                       result( do_onset )
+
+    !
+    ! !DESCRIPTION:
+    ! Function to determine if seasonal decicious leaf onset should happen.
```
olyson, line 1100: Change decicious to deciduous.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+    !
+    ! !USES:
+    use shr_const_mod   , only: SHR_CONST_TKFRZ
+    ! !ARGUMENTS:
+    real(r8), intent(INOUT) :: onset_gdd      ! onset growing degree days 
+    real(r8), intent(INOUT) :: onset_gddflag  ! Onset freeze flag
```
olyson, line 1106: Change temperture to temperature

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+       ! shrub, C3 arctic grass)
+       if (onset_gdd > crit_onset_gdd .and.  season_decid_temperate == 1) then
+           do_onset = .true.
+        ! Note: The min_critial_daylength_onset is because for some coastal
+        ! points the other triggers could allow onset in January/February
+        ! which isn't sustainable and is a degenerate case. To prevent this
+        ! condition was added, but now the other conditions aren't triggered
+        ! until much later so it's value just needs to be high enough to prevent
+        ! the degenerate case of happening too early, and low enough that it
+        ! doesn't restrict onset. As such the value of this parameter shouldn't
```
olyson, line 1163: Change degnerate to degenerate

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+    if ( onset_thresh_depends_on_veg) then
+       ! separate into non-arctic seasonally deciduous pfts
+       ! (temperate broadleaf deciduous
+       ! tree) and arctic/boreal seasonally deciduous pfts (boreal
+       ! needleleaf deciduous tree,
+       ! boreal broadleaf deciduous tree, boreal broadleaf deciduous
+       ! shrub, C3 arctic grass)
+       if (onset_gdd > crit_onset_gdd .and.  season_decid_temperate == 1) then
+           do_onset = .true.
+        ! Note: The min_critial_daylength_onset is because for some coastal
```
olyson, line 1157: min_critical_daylength_onset

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+        ! which isn't sustainable and is a degenerate case. To prevent this
+        ! condition was added, but now the other conditions aren't triggered
+        ! until much later so it's value just needs to be high enough to prevent
+        ! the degenerate case of happening too early, and low enough that it
+        ! doesn't restrict onset. As such the value of this parameter shouldn't
+        ! matter for reasonable values between the two degnerate cases.
+        else if (season_decid_temperate == 0 .and.  onset_gddflag == 1.0_r8 .and. &
+                soila10 > SHR_CONST_TKFRZ .and. &
+                t_a5min > SHR_CONST_TKFRZ .and. ws_flag==1.0_r8 .and. &
+                dayl>min_critial_daylength_onset .and.  snow_5day<snow5d_thresh_for_onset) then
```
olyson, line 1168: min_critical_daylength_onset

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+    end if
+    if ( onset_thresh_depends_on_veg) then
+       ! separate into non-arctic seasonally deciduous pfts
+       ! (temperate broadleaf deciduous
+       ! tree) and arctic/boreal seasonally deciduous pfts (boreal
+       ! needleleaf deciduous tree,
+       ! boreal broadleaf deciduous tree, boreal broadleaf deciduous
+       ! shrub, C3 arctic grass)
+       if (onset_gdd > crit_onset_gdd .and.  season_decid_temperate == 1) then
+           do_onset = .true.
```
olyson, line 1156: The check "dayl>min_critical_daylength_onset" in the if statement was added because...."

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+        ! doesn't restrict onset. As such the value of this parameter shouldn't
+        ! matter for reasonable values between the two degnerate cases.
+        else if (season_decid_temperate == 0 .and.  onset_gddflag == 1.0_r8 .and. &
+                soila10 > SHR_CONST_TKFRZ .and. &
+                t_a5min > SHR_CONST_TKFRZ .and. ws_flag==1.0_r8 .and. &
+                dayl>min_critial_daylength_onset .and.  snow_5day<snow5d_thresh_for_onset) then
+           do_onset = .true.
+        end if
+    else
+       ! set onset_flag if critical growing degree-day sum is exceeded
```
olyson, line 1172: We are setting do_onset here, not onset_flag

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+    real(r8), intent(IN)    :: t_a5min        ! 5-day running mean of min 2-m temperature
+    real(r8), intent(IN)    :: dayl           ! Day length
+    real(r8), intent(IN)    :: snow_5day      ! 5-day average of snow
+    real(r8), intent(IN)    :: ws_flag        ! winter-summer solstice flag (0 or 1)
+    real(r8), intent(IN)    :: crit_onset_gdd ! critical onset growing degree-day sum
+    real(r8), intent(IN)    :: season_decid_temperate  ! If this is a temperate seasonal decidious type 
+    logical :: do_onset                       ! Flag if onset should happen (return value)
+    !
+    ! !LOCAL VARIABLES:
+    real(r8), parameter :: snow5d_thresh_for_onset        = 0.1_r8          ! 5-day snow depth threshold for onset
```
olyson, line 1118: Is 39300/2 correct?  This is about 5.4583 hours....

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -944,16 +1015,7 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
                   if (days_active(p) > 355._r8) pftmayexist(p) = .false.
                end if
 
-               if ( min_crtical_dayl_depends_on_lat )then
```
wwieder, line 947: it seems like some of these comments may be helpful either in the code or when the parameters are set

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -972,6 +1034,147 @@ subroutine CNSeasonDecidPhenology (num_soilp, filter_soilp       , &
...
+
+    select case( critical_daylight_method )
+    ! Critical day length depends on both vegetation type and latitude
+    case(critical_daylight_depends_on_latnveg)
+        ! Critical day length for offset is fixed for temperate type vegetation
+        if ( pftcon%season_decid_temperate(patch%itype(p)) == 1 )then
+           crit_daylat = crit_dayl
+        ! For Arctic vegetation -- critical daylength is higher at high latitudes and shorter
+        ! at midlatitudes
+        else
```
wwieder, line 1061: wait, what happen to this formula when you're north of 65 degrees?  It makes crit_daylat > crit_dayl_at_high_lat.  Is that what's @lmbirch89 intended? Should `crit_daylat` be more than 15 hr (54000 seconds)?

src/biogeochem/CNPhenologyMod.F90
wwieder: or just add a comment about where this value is coming from

wwieder (commented):
looks good, Erik.  let me know if my comments aren't clear

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -96,8 +119,15 @@ module CNPhenologyMod
...
-  logical,  private :: min_crtical_dayl_depends_on_lat = .false. ! If critical day-length for onset depends on latitude
+
+  ! Constants for seasonal decidious leaf onset and offset
   logical,  private :: onset_thresh_depends_on_veg     = .false. ! If onset threshold depends on vegetation type
+  integer,  public, parameter :: critical_daylight_constant           = 1
+  integer,  public, parameter :: critical_daylight_depends_on_lat     = critical_daylight_constant + 1
+  integer,  public, parameter :: critical_daylight_depends_on_veg     = critical_daylight_depends_on_lat + 1
+  integer,  public, parameter :: critical_daylight_depends_on_latnveg = critical_daylight_depends_on_veg + 1
+  integer,  private :: critical_daylight_method = critical_daylight_constant
+  real(r8), parameter :: critical_offset_high_lat         = 65._r8     ! Start of what's considered high latitude (degrees)
```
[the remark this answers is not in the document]
wwieder: Oh, this is part of the function that determines the how latitude influences critical_dayl_lat maybe some of Leah's comments from the old code or a more descriptive comment here would be more helpful?  I can brainstorm ideas if it's helpful.

src/biogeochem/CNPhenologyMod.F90
ekluzek: OK, got that one.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Adding units. Because these are only used for unit-testing they actually should be pretty static.

src/biogeochem/CNPhenologyMod.F90
ekluzek: OK got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it

src/biogeochem/CNPhenologyMod.F90
ekluzek: Got it.

src/biogeochem/CNPhenologyMod.F90
ekluzek: Yes, 39300/2 is what we are currently using and what we are testing with. You are correct that this is a bit screwy, but changing this could possibly change answers, so I want to leave it alone, unless we do more testing with it changed. So I suppose I can add a comment regarding that.

src/biogeochem/CNPhenologyMod.F90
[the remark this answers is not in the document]
ekluzek: This is what Leah used, but she didn't have any comments about it.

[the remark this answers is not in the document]
ekluzek: I added a comment line, please look it over.

src/biogeochem/CNPhenologyMod.F90
ekluzek: That is what's currently being done. In talking with Leah we thought we should stick with that.

ekluzek:
In doing more testing I'm seeing the following tests change answers:

FAIL ERP_P36x2_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs BASELINE ctsm5.1.dev053: DIFF
FAIL ERS_Ld3.f10_f10_mg37.I2000Clm51Bgc.cheyenne_intel.clm-ciso_cwd_hr BASELINE ctsm5.1.dev053: DIFF
FAIL ERS_Ld9.f10_f10_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-FatesColdDefCH4 BASELINE ctsm5.1.dev053: DIFF

The first and last is because use_nitrif_denitrif was allowed to be false. I'm not sure about the middle one.

src/biogeochem/CNPhenologyMod.F90
ekluzek: I'm adding these comments to the top of the code where these parameter are defined.

ekluzek:
@olyson and @wwieder I'm bringing this to main-dev today, but go ahead and look at CNPhenology and let me know if you have further suggestions and I'll bring them in a future tag. Also go ahead and go through each of the conversations you started and make sure you agree that I dealt with your comment, and then you can resolve the conversation.

ekluzek:
SMS_D_Lm1_Mmpi-serial_Vnuopc.CLM_USRDAT.I1PtClm50SpRs.cheyenne_intel.clm-USUMB_nuopc is failing because it's trying to find the UNSET fatmlndfrc file which it doesn't even need. I'll rework this so it's not needed and rerun the nuopc tests.

olyson:
I think I resolved the conversations I started.

src/biogeochem/CNPhenologyMod.F90
ekluzek: I did add some comments regarding this. So I think this is good.

wwieder: is 39300 used elsewhere in the code as a parameter for a day length trigger for offset?  If so, can we refer to the name of that parameter in the comment so that people could track down where this number came from.

Also, is this the parameter that Leah added to prevent unrealistically early onset in some grids / climates?  If so, let's include that detail on this comment too.

wwieder: Yes, let's include a comment (as I did above) that states: "this is equal to crit_dayl/2, and is used to prevent unrealistically early onset" 

I see you have more on line 1157, but I think a bit is warranted here where the parameter is defined.

ekluzek: OK I added more. I'm a little reluctant to refer to the origin of the number, since it likely doesn't mater what the specific value is. I think you could change it within some range and answers will be identical. As such the specific value doesn't matter, but I'm also reluctant to change the value because maybe there is sensitivity to it for a particular point. We'd have to do more testing to see if that's the case, and it's not really worth it to do it.

ekluzek:
A few tests are failing because of this in check_input_data

`Model ctsm no file specified for finidat`

This is because finidat for a cold start is blank, but it's still in the ctsm.input_data_list file, but apparantely new behavior for cime is that this will now trigger an error.

So I just need to remove adding the files to ctsm.input_data_list file if they are blank.

Issue #1171, opened 2020-10-01
btran2 not initialized to zero in CNFireLi2021Mod.F90


ekluzek:
### Brief summary of bug

When I updated CNFire_calc_fire_root_wetness for CNFireLi2021, I left off the initialization of btran2 to zero like it's done in the base class. It was in originally, but I left it off when I refactored for Li2021 to have it's own subroutine.

### General bug information

**CTSM version you are using:** ctsm5.1.dev001

**Does this bug cause significantly incorrect results in the model's science?** Yes (for clm5_1)

**Configurations affected:** clm5_1 with fire turned on (for Clm51Bgc compsets)

### Details of bug

The problematic commit is 7f0025bf58b49296fda63a2ec957e62e5c93ca13.

ekluzek:
One reason for this bug is that I didn't check answers for clm5_1, since they were already changing. But, I needed to separate out the answer changes from the changes that should be bit-for-bit, and neglected to do that. This would have required separating steps and creating separate baselines (or at least for a single sample case). I usually do this, but neglected to do this in this instance.

ekluzek:
I'm going to run more tests on this. But, I have shown that the following test is identical with or without the zeroing of btran2. I'd still say the zeroing is correct though.

SMS_Lm13.f19_g17.I2000Clm51BgcCropGs.cheyenne_intel.clm-cropMonthOutput

billsacks:
> I have shown that the following test is identical with or without the zeroing of btran2.

Thanks for doing that testing. That was a red flag for me, which led me to find #1173 . With the #1173 fix in place off of ctsm5.1.dev004 (before the fix here), I verified that btran2 quickly blows up into huge values.

ekluzek:
With the fix in #1173 in place my testing is now showing that answers change as they should to the ctsm5.1.dev001 tag. I'm now going to test using the methodology before btran was inside of cnfire.

ekluzek:
OK, I've now done the testing for the clm5_1 refactor that I should have done and shown that with #1173 in place the clm5_1 tests preserve answers versus the version where btran2 was calculated in SoilMoistStress. This tells me there weren't any other issues that I introduced in do that refactor of moving btran2 for clm5_1 into CNFire.

So the following tests give identical answers between those two as expected...

ERI_D_Ld9.f10_f10_musgs.I1850Clm51BgcGs.cheyenne_gnu.clm-default			
[...]

Issue #1169, opened 2020-09-30
CNFireLi2021Mod reintroduces a number of bugs that were fixed in earlier versions


billsacks:
### Brief summary of bug

From comparing CNFireLi2016Mod.F90 with CNFireLi2021Mod.F90 I see that the latter reintroduces a number of bugs that were fixed in earlier versions.

### General bug information

**CTSM version you are using:** ctsm5.1.dev004

**Does this bug cause significantly incorrect results in the model's science?** Yes

**Configurations affected:** CLM51 compsets

### Details of bug

Issues that I see are:

#### Significant issues

** @ekluzek can you please work on the following, some of which are answer-changing: **

- [x] The accumulations of trotr1_col and trotr2_col should not have `*col%wtgcell(c)` (fixed in `clm4_5_12_r195` for the older fire modules)
  - (2020-10-01) Fixed in 7a7ab79dc

- [x] The accumulation of dtrotr_col should not have `*col%wtgcell(c)` (fixed in `clm4_5_14_r223` for the older fire modules)
  - (2020-10-01) Fixed in 7a7ab79dc

- [x] nfire(c) needs to be initialized to 0 (fixed in 1b6c2bae3 for the older fire modules)
  - (2020-10-01) Fixed in b45f492f4

- [ ] Two hard-coded parameters that were recently pulled out to the parameter file have returned to being hard-coded in this version of the fire code: prh30 and ignition_efficiency

- [ ] Given these regressions, the other diffs between the 2016 and 2021 modules in the loop starting with the comment "begin column loop to calculate fractional area affected by fire" should be carefully reviewed to confirm that all of these differences are intentional (and not other regressions of bugs we have previously fixed).

#### Minor issues

**I'll fix these myself in an upcoming tag**

- [x] CNFireLi2021Mod uses `SHR_ASSERT_ALL` rather than the newer & preferred `SHR_ASSERT_ALL_FL`

- [x] dt should be set from `get_step_size_real()`

Edit: I have fixed both of these in #1155

ekluzek:
@olyson this first issue in this list would be a problem in your simulations you are doing with the arctic branch. We don't think it should matter much, but it would be an issue.

olyson:
I had restarted the spinup yesterday with that branch due to the organic_frac_squared issue, and then saw the the fire issue, so I've stopped the spinup until we decide how/if we want to resolve the fire issue.  I assume we'll discuss at the PPE meeting.

[...]

Issue #1166, opened 2020-09-30
Possible problems with namelist defaults starting in ctsm5.0.dev001


billsacks:
### Brief summary of bug

I am reviewing the changes in ctsm5.0.dev001. I noticed a few possible problems with the namelist defaults in that tag. I'm not sure if all of these are actually bugs, but I wanted to bring this to the attention of @ekluzek and others in case some should be fixed. **One of these issues, organic_frac_squared, appears to lead to incorrect results in out-of-the-box clm51 configurations (unless this change was intentional).** The other issues would just affect runs in non-standard configurations, or are cleanup issues.

### General bug information

**CTSM version you are using:** ctsm5.1.dev001

**Does this bug cause significantly incorrect results in the model's science?** Yes, I think so, for clm51 cases, because of incorrect value for organic_frac_squared.

**Configurations affected:** clm51 cases, and non-standard configurations

### Details of bug

These are really a few somewhat unrelated issues. I'm grouping them together here for ease of review:

- [ ] `organic_frac_squared` for `clm5_1` uses the same setting as `clm4_5`; I'm guessing this is wrong:

https://github.com/ESCOMP/CTSM/blob/08bc0ade130b23a67a9366120f8033897505c74b/bld/namelist_files/namelist_defaults_ctsm.xml#L163-L165

- [ ] Some fire-related namelist variables depend on the physics version directly, whereas my intuition is that they should depend on the fire method. As a thought experiment: if someone sets up a clm51 case but reverts to the li2016crufrc fire method, or conversely sets up a clm50 case but changes to li2021gswpfrc, what defaults should be used for these? Specifically, this is the case for the lightning and popdens streams: the `light_res` variable and the three `popdens` variables that depend on phys version – though it looks like the `popdens` variables don't actually differ with different phys versions, so the duplication in xml is irrelevant and could probably be removed:

https://github.com/ESCOMP/CTSM/blob/08bc0ade130b23a67a9366120f8033897505c74b/bld/namelist_files/namelist_defaults_ctsm.xml#L1569-L1573

https://github.com/ESCOMP/CTSM/blob/08bc0ade130b23a67a9366120f8033897505c74b/bld/namelist_files/namelist_defaults_ctsm.xml#L1592-L1602

[...]

PR #1172, opened 2020-10-01
Clm51 fixes

billsacks:
### Description of changes

A few fixes for CLM51 configurations: three for fire and one for the organic_frac_squared namelist default

### Specific notes

- Resolves ESCOMP/CTSM#1171
- Partially addresses ESCOMP/CTSM#1169
- Partially addresses ESCOMP/CTSM#1166

Are answers expected to change (and if so in what way)? YES: significant answer changes for CLM51

Any User Interface Changes (namelist or namelist defaults changes)? fix namelist default for organic_frac_squared

Testing performed, if any: none yet

### DIFF
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -29,6 +29,8 @@ Does this tag change answers significantly for any of the following physics conf
 
     [Put an [X] in the box for any configuration with significant answer changes.]
 
+[ ] clm5_1
+
 [ ] clm5_0
 
 [ ] ctsm5_0-nwp
--- a/bld/namelist_files/namelist_defaults_ctsm.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm.xml
@@ -160,9 +160,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <urban_traffic>.false.</urban_traffic>
 
 <!-- Soil state settings -->
-<organic_frac_squared phys="clm5_1" >.true.</organic_frac_squared>
 <organic_frac_squared phys="clm4_5" >.true.</organic_frac_squared>
 <organic_frac_squared phys="clm5_0" >.false.</organic_frac_squared>
+<organic_frac_squared phys="clm5_1" >.false.</organic_frac_squared>
 
 <soil_layerstruct_predefined structure="fast">4SL_2m</soil_layerstruct_predefined>
 <soil_layerstruct_predefined structure="standard" phys="clm5_1" >20SL_8.5m</soil_layerstruct_predefined>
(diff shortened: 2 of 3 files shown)


doc/.ChangeLog_template
```diff
@@ -29,6 +29,8 @@ Does this tag change answers significantly for any of the following physics conf
 
     [Put an [X] in the box for any configuration with significant answer changes.]
 
+[ ] clm5_1
```
billsacks, line 32: @ekluzek does it make sense to you that we should have this in here now, or did you deliberately exclude it? At some point soon we should also add a line for ctsm5_1-nwp, but I don't think we have any out of the box configurations like that yet.

ekluzek: I actually purposefully excluded it -- because I figure clm5_1 is under development and will change in many tags, until we start a ctsm5_2 physics version. At that point I thought it would make sense to include it. But, we could go either way on this. I probably should have brought this up for discussion.

ekluzek (approved):

doc/.ChangeLog_template
billsacks: Yeah, I also don't have strong opinions on this. For now I'll leave it, but I don't object to its removal. I think it comes down to whether it's helpful to the scientists to have this noted in the ChangeLog. Maybe we can discuss further at a ctsm-software meeting.

billsacks:
Testing all looks good. These expensive tests still haven't run:

```
    PEND ERS_D_Ld10.C96_C96_mg17.IHistClm50Sp.cheyenne_intel.clm-decStart RUN
    PEND ERS_Ln9.ne0ARCTICne30x4_ne0ARCTICne30x4_mt12.IHistClm50SpGs.cheyenne_intel.clm-clm50cam6LndTuningMode_1979Start RUN
    PEND SMS_Ln9.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.IHistClm50Sp.cheyenne_intel.clm-clm50cam6LndTuningMode_1979Start RUN
    PEND SMS_Ln9.ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12.ISSP585Clm50BgcCrop.cheyenne_intel.clm-clm50cam6LndTuningMode RUN
    PEND SMS_Ln9.ne0CONUSne30x8_ne0CONUSne30x8_mt12.IHistClm50Sp.cheyenne_intel.clm-clm50cam6LndTuningMode_2013Start RUN
```

These are all tests that we are about to remove from the test list, probably in my next tag. So I'm planning to go ahead with this tag even though those didn't run; I'll just note it in the ChangeLog.

Issue #29, opened 2017-12-16
add namelist option to turn on all history fields, add a CLM test that uses this


billsacks:
**Bill Sacks < sacks@ucar.edu > - 2014-07-31 14:44:46 -0600**
**Bugzilla Id:** [2022](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2022)
**Bugzilla CC:**  bandre@lbl.gov,  erik@ucar.edu,  muszala@ucar.edu,  rfisher@ucar.edu, 

You can turn on all history fields by ignoring a few lines of code in histFileMod:

Index: src/clm4_5/main/histFileMod.F90
===================================================================
--- src/clm4_5/main/histFileMod.F90    (revision 61424)
+++ src/clm4_5/main/histFileMod.F90    (working copy)
@@ -4139,9 +4139,6 @@
          p2c_scale_type=scale_type_p2c, c2l_scale_type=scale_type_c2l, l2g_scale_type=scale_type_l2g)

     l_default = 'active'
-    if (present(default)) then
-       l_default = default
-    end if
     if (trim(l_default) == 'inactive') then
        return
     else
@@ -4429,9 +4426,6 @@
          no_snow_behavior=no_snow_behavior)

     l_default = 'active'
-    if (present(default)) then
-       l_default = default
-    end if
     if (trim(l_default) == 'inactive') then
        return
     else

It would be useful to have a namelist option that does this, for testing.

Then we should add a test that uses this.

billsacks:
**Bill Sacks < sacks@ucar.edu > - 2014-07-31 14:47:52 -0600**

We probably want to use this option for one test for each configuration... this could replace the addition of select fields, which we do for some tests.

johnpaulalex:
I took a closer look at this.  It looks like the logic is spread over two places: 'global' settings are applied from namelist variables [here](https://github.com/ESCOMP/CTSM/blob/394f9fd473731cc7d87e268a385cc45fd3f3daf6/src/main/histFileMod.F90#L869) and then if the global settings have no opinion, the 'local' settings are applied.  The local logic is in Bill's code snippet above; basically all history fields are written by default, but the caller can opt out of that by passing in 'inactive' to an optional parameter called 'default' ([example](https://github.com/ESCOMP/CTSM/blob/394f9fd473731cc7d87e268a385cc45fd3f3daf6/src/biogeophys/LakeStateType.F90#L152)).

[...]

PR #1209, opened 2020-11-10
Write history fields master list to separate optional text file

slevisconsulting:
User can choose to write the history fields master list in a separate text file by setting `hist_master_list_file = .true.` in the CTSM namelist.

File contents follow recommended reStructuredText "simple table" formatting (see `https://docutils.sourceforge.io/docs/user/rst/quickref.html#tables`) for use in CTSM documentation in the sphinx environment.

### Description of changes
- New section in `subroutine hist_printflds` (called by subr. initialize2); the new section runs when `hist_master_list_file = .true.`
- New namelist option `hist_master_list_file`
- Test added to existing test in cheyenne's test suite (see testing below)

### Specific notes

@billsacks @danicalombardozzi @dlawrenncar @katiedagon @wwieder 

#34 #941 

No

One new namelist option for now, with default value
`hist_master_list_file = .false.`
A new output file (master_list_file.txt) when `hist_master_list_file = .true.`

Testing performed, if any:
- Before introducing the new namelist option, I ran existing test `ERP_D_Ld5.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-allActive` to generate the new output file.
- Once I introduced the new namelist option, I ran new test `SMS_Ld1_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCrop.cheyenne_intel.clm-write_hist_master_list`. The new test is intended to ensure that writing the history field master list to a separate file does not cause a failure.
- After consolidating the new test with an existing one, I started using `ERP_P36x2_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs`.


src/main/histFileMod.F90
```diff
@@ -301,10 +301,22 @@ subroutine hist_printflds()
...
+    integer nf, i  ! do-loop counters
+    integer width_col_1  ! widths of table columns 1...
+    integer width_col_2  ! ...2
+    integer width_col_3  ! ...3
+    integer width_col_4  ! ...and 4
+    integer width_col_sum  ! widths of columns summed, including spaces
+    character(len=1) str_w_col_1  ! string versions of width_col_1 through 4
+    character(len=2) str_w_col_2, str_w_col_3, str_w_col_4
+    character(len=3) str_w_col_sum  ! string version of width_col_sum
+    character(len=25) fmt_txt  ! format statement
```
billsacks, line 319: - [ ] [optional] I'd suggest making these string lengths longer so that this doesn't break under modification. (If need be, you can then `trim` them where they are used, but I have a feeling that isn't necessary.)

src/main/histFileMod.F90
```diff
@@ -318,6 +330,78 @@ subroutine hist_printflds()
...
+       width_col_sum = width_col_1 + width_col_2 +  &  ! sum of widths plus
+                       width_col_3 + width_col_4 + 3  ! ...3 blank spaces
+
+       ! Convert integer widths to strings for use in format statements
+       ! These write statements are not outputting to files
+       write(str_w_col_1,'(i1)') width_col_1
+       write(str_w_col_2,'(i2)') width_col_2
+       write(str_w_col_3,'(i2)') width_col_3
+       write(str_w_col_4,'(i2)') width_col_4
+       write(str_w_col_sum,'(i3)') width_col_sum
```
billsacks, line 350: - [ ] You can avoid duplicating assumptions about the number of digits by using '(i0)' formatting here

src/main/histFileMod.F90
```diff
@@ -318,6 +330,78 @@ subroutine hist_printflds()
...
     end if
 
+    ! Print master field list in separate file if namelist variable
+    ! requests it
+    if (masterproc .and. hist_master_list_file) then
+       ! hardwired table column widths
+       width_col_1 = 4  ! column that shows the variable number, nf
+       width_col_2 = 39  ! variable name column
+       width_col_3 = 94  ! long description column
+       width_col_4 = 65  ! units column
```
billsacks, line 340: - [ ] [optional] should these be determined more dynamically, for example, based on the actual maximum length of variable names, etc.?

billsacks (changes requested):
Thanks a lot for your quick work on this @slevisconsulting ! I have a number of inline comments, some of which are optional. In addition:

- [no] Have you confirmed that the resulting file appears correctly when building the documentation? As a test, you could copy and paste it into an existing page and rebuild the documentation, to ensure that it shows up correctly.
  - (2020-11-16) deferred to below

cime_config/testdefs/testlist_clm.xml
```diff
@@ -1800,6 +1800,15 @@ for ERS test as otherwise it won't work for a sub-day test (no need to run this
...
   </test>
+  <test name="SMS_Ld1_Mmpi-serial" grid="1x1_numaIA" compset="I2000Clm50BgcCrop" testmods="clm/write_hist_master_list">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="aux_clm"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:20:00</option>
+      <option name="comment"  >Ensure that writing the history field master list to a separate file does not cause failure"</option>
+    </options>
+  </test>
```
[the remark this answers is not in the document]
slevisconsulting: I came across a few candidates, but the best match I found was `testmods_dirs/clm/extra_outputs`. The name is generic enough and the outcome of this work is an extra output.

Pls LMK if I should consolidate with this one or just leave it.

src/main/clm_varctl.F90
```diff
@@ -79,6 +79,7 @@ module clm_varctl
   !----------------------------------------------------------
   !
   integer, public :: iulog = 6        ! "stdout" log file unit number, default is 6
+  integer, public :: master_list_file = 9  ! unit number for optional output file containing master history list in tabular form
```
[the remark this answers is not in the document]
slevisconsulting: @billsacks thank you for pointing out getavu. I hadn't used it before.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -729,6 +729,11 @@ SNICAR (SNow, ICe, and Aerosol Radiative model) optical data file name
 SNICAR (SNow, ICe, and Aerosol Radiative model) snow aging data file name
 </entry>
 
+<entry id="hist_master_list_file" type="logical" category="history"
+       group="clm_inparm" valid_values="" >
+If TRUE, write master field list to separate file for documentation purposes
+</entry>
+
```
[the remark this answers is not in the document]
slevisconsulting: @billsacks thank you for this reminder.

cime_config/testdefs/testlist_clm.xml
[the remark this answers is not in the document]
billsacks: Yes, that seems like a good one. Thanks for taking the time to identify that.

src/main/histFileMod.F90
slevisconsulting: Changing to lengths of 3,3,3,99. For the first three, I limit the lengths to 3 characters because I want to discourage columns wider than 999.

I tested without adding `trim` and got identical results.

billsacks: Makes sense - thanks.

src/main/histFileMod.F90
slevisconsulting: @billsacks your suggestion makes sense on one hand. On the other, I was thinking of controlling the size of this table so that each row would fit on a computer screen. For long names that get cut off, users could still search for the complete description in the code.

I'm open to going with your suggestion if that's preferred.

slevisconsulting: Unless... long names will end up wrapping in the available space. I didn't think of that earlier.

billsacks: Your rationale makes sense. Thanks for explaining.

slevisconsulting: I have built the new .rst file's html version. Columns with long strings do wrap into the next line(s), so that's good news. The issue of fitting on the screen still applies when looking at the .rst file itself.

So I would still like final feedback on the subject of including the complete length of long_name. It's less of an issue with the other columns.

slevisconsulting: To take a look and offer comments, you will find the relevant html files (master_list_file.html and customizing-the-clm-namelist.html) here:
`/glade/work/slevis/git/ctsm-docs/versions/master/html/users_guide/setting-up-and-running-a-case`

billsacks: I'll defer to others (@danicalombardozzi @katiedagon @wwieder @dlawrenncar @olyson etc.)

danicalombardozzi: I think this makes sense. It doesn't have a huge impact since it only cuts off long name descriptions for a few variables, and in all cases the primary relevant information is still included.

slevisconsulting:
@billsacks remaining tasks in this PR:
- Address #29 beyond what this PR accomplishes? As of now this PR allows users to output their case's master list (i.e. the active and inactive history fields) to a table.
- I need to run the test suites on cheyenne and izumi
- Am I forgetting anything?

billsacks:
@slevisconsulting thanks a lot! I'd say skip #29 for now. This PR doesn't address that issue (because it turns out not to have been needed), but we can save that for later.

I tried building the documentation with the changes you made, but it looks like the rst syntax you used didn't work right for including the new table in the documentation. So can you please:

- [ ] Fix the rst syntax so that the table gets included

- [ ] Build the documentation and look at the table to make sure the formatting looks right. For example, when I was looking at the diffs, I noticed that there are some rst formatting characters in some units and/or variable names (for example, `**`, as is used sometimes for exponents, means bold in rst). I'm wondering if we need to either wrap the whole table in an rst construct that prevents this from being interpreted (e.g., can you have a table within a code block?), or, if the problem only occurs in a few variables, we could change the units (or whatever is causing the problem) to avoid the problem – e.g., replacing `m**2` with `m^2`. The latest directions for building the documentation are here https://github.com/ESCOMP/CTSM/wiki/Directions-for-editing-CLM-documentation-on-github-and-sphinx

- [ ] Add a comment/note somewhere about how to recreate this table (i.e., copy it from the test in the test suite). Probably the best place for this to go would be in a README file in the directory where the rst file lives.

I have testing underway for a tag today (I hope), then you can coordinate with Negin as to whether this tag or her big upcoming tag should come next.

slevisconsulting:
> * [ ]  Fix the rst syntax so that the table gets included

I wanted to include a link to the table rather than include the table itself, because of its immense size. It took me a bit of trial and error to get a link to work, and the table now appears as a new section, which may be a problem. Should I try to link it differently?

> * [ ]  Build the documentation and look at the table to make sure the formatting looks right. For example, when I was looking at the diffs, I noticed that there are some rst formatting characters in some units and/or variable names (for example, `**`, as is used sometimes for exponents, means bold in rst). I'm wondering if we need to either wrap the whole table in an rst construct that prevents this from being interpreted (e.g., can you have a table within a code block?), or, if the problem only occurs in a few variables, we could change the units (or whatever is causing the problem) to avoid the problem – e.g., replacing `m**2` with `m^2`. The latest directions for building the documentation are here https://github.com/ESCOMP/CTSM/wiki/Directions-for-editing-CLM-documentation-on-github-and-sphinx

I noticed this in the diffs, too, but the table itself seems formatted correctly and consistently throughout, after building. In any case, the m**2 occurs only once as far as I can tell, so I will go change it in the code.

> * [ ]  Add a comment/note somewhere about how to recreate this table (i.e., copy it from the test in the test suite). Probably the best place for this to go would be in a README file in the directory where the rst file lives.

Added README.

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -586,7 +586,7 @@ subroutine InitHistory(this, bounds)
 
        endif
        this%fpsn24_patch = spval 
-       call hist_addfld1d (fname='FPSN24', units='umol CO2/m**2 ground/day',&
+       call hist_addfld1d (fname='FPSN24', units='umol CO2/m^2 ground/day',&
```
slevisconsulting, line 589: I committed this change without recreating the .rst file and then rebuilding the documentation... I will do that next.

slevisconsulting: Also noticing: I still need to add the code that inserts the new .rst file's title...

billsacks:
Thanks @slevisconsulting ! Testing looks good on my tag, so I plan to bring it to master later this morning, after confirming with Mariana that she's okay with this. Then you can go next with this unless Negin is planning to go imminently with her next tag.

src/main/histFileMod.F90
```diff
@@ -318,6 +332,93 @@ subroutine hist_printflds()
...
+       write(str_w_col_4,'(i0)') width_col_4
+       write(str_w_col_sum,'(i0)') width_col_sum
+
+       ! Open master_list_file
+       master_list_file = getavu()  ! get next available file unit number
+       open(unit = master_list_file, file = 'master_list_file.rst',  &
+            status = 'new', action = 'write', form = 'formatted')
+
+       ! File title
+       fmt_txt = '(19a)'
```
billsacks, line 366: I'm pretty sure you can just use '(a)' to avoid needing to count characters. (Not critical to fix.)

slevisconsulting:
Cheyenne test-suite: PASS
Izumi test-suite: In progress

slevisconsulting:
@billsacks I believe this is ready for you to merge.

billsacks (approved):
I realized I never changed this to approved... doing so now.

@slevisconsulting @ekluzek - I'm making one other addition to this in the merge to master: I'm removing the now-defunct findHistFields.pl.

PR #1222, opened 2020-11-23
Add ACTIVE (T or F) column to the master hist fields table and consider alphabetizing

### LINKED ISSUE #941: History fields table missing from CLM5 Users Guide
The table of history fields is missing from the current users guide (section 1.2.3.1.3, Table 1.3)
https://escomp.github.io/ctsm-docs/doc/build/html/users_guide/setting-up-and-running-a-case/customizing-the-clm-namelist.html#table-1-3-clm-history-fields

Several of the links on that page are also broken (e.g., section 1.2.3.1 and section 1.2.3.1.1)

slevisconsulting:
Following up on #1209 
- @dlawrenncar suggested adding an ACTIVE (T or F) column to the master fields table
- @danicalombardozzi suggested alphabetizing the variable list

### Description of changes
- While adding the new column, I made the code less dependent on the number of columns in the table
- I will look into alphabetizing the variable list. See @billsacks comment [here](https://github.com/ESCOMP/CTSM/issues/941#issuecomment-732490641)

### Specific notes

@billsacks @danicalombardozzi @dlawrenncar 

Resolves #941 

No.

Updated master_list_file.rst with a new column.

Testing performed, if any:
I have confirmed that
1) `./create_test ERP_P36x2_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev017`
has not changed answers 
2) the new master_list_file.rst builds and looks comparable to the previous version that had one less column


slevisconsulting:
@billsacks this code feels ready for review. I tested as before with `ERP_P36x2_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs` against the baseline and it PASSed.

I'm open to alternate approaches if you think I should have subroutinized sort_hist_list differently.

billsacks:
@slevisconsulting - It may just be that I'm resistant to change, but I don't love what your solution does to the data structures. The biggest change is that it unnecessarily duplicates the master field list N times (where N = number of history tapes); I think this will be confusing under future maintenance. In addition, it introduces actflag into history_entry, despite the fact that this is only needed (I think) for master entries, not hlist entries. In general, I find it confusing when there is information that is used in some contexts but not others like this.

This type of situation is where object orientation really shines. So I spent some time writing up the solution I was envisioning. Admittedly, this ended up being slightly more complex than I originally imagined, due to the need to define a 'copy' subroutine (I realized through testing that it doesn't work to just use intrinsic assignment on polymorphic variables).

Here are the changes you need:

```Fortran
  type, abstract :: entry_base
     type(field_info) :: field
   contains
     procedure(copy_entry_interface), deferred :: copy
  end type entry_base

  abstract interface
     subroutine copy_entry_interface(this, other)
       ! set this = other
       import :: entry_base
       class(entry_base), intent(out) :: this
       class(entry_base), intent(in) :: other
     end subroutine copy_entry_interface
  end interface

  type, extends(entry_base) :: master_entry
     logical            :: actflag(max_tapes)  ! active/inactive flag
     character(len=avgflag_strlen) :: avgflag(max_tapes)  ! time averaging flag ("X","A","M","I","SUM")
   contains
     procedure :: copy => copy_master_entry
  end type master_entry

  type, extends(entry_base) :: history_entry
     character(len=avgflag_strlen) :: avgflag  ! time averaging flag
     real(r8), pointer :: hbuf(:,:)            ! history buffer (dimensions: dim1d x num2d)
     integer , pointer :: nacs(:,:)            ! accumulation counter (dimensions: dim1d x num2d)
   contains
     procedure :: copy => copy_history_entry
  end type history_entry

  ! ...

contains

  ! ...

  subroutine copy_master_entry(this, other)
    ! set this = other
    class(master_entry), intent(out) :: this
    class(entry_base), intent(in) :: other

    select type(this)
    type is (master_entry)
       select type(other)
       type is (master_entry)
          this = other
       class default
          call endrun('Unexpected type of "other" in copy_master_entry')
       end select
    class default
       call endrun('Unexpected type of "this" in copy_master_entry')
    end select
  end subroutine copy_master_entry

  subroutine copy_history_entry(this, other)
    ! set this = other
    class(history_entry), intent(out) :: this
    class(entry_base), intent(in) :: other

    select type(this)
    type is (history_entry)
       select type(other)
       type is (history_entry)
          this = other
       class default
          call endrun('Unexpected type of "other" in copy_history_entry')
       end select
    class default
       call endrun('Unexpected type of "this" in copy_history_entry')
    end select
  end subroutine copy_history_entry

  subroutine sort_hist_list(t, n_fields, hist_list)

    ! !DESCRIPTION:

    ! !ARGUMENTS:
    integer, intent(in) :: t  ! tape index
    integer, intent(in) :: n_fields  ! number of fields
    class(entry_base), intent(inout) :: hist_list(:)

    ! !LOCAL VARIABLES:
    integer :: f, ff  ! field indices
    class(entry_base), allocatable :: tmp

    character(len=*), parameter :: subname = 'sort_hist_list'

    ! add assertion here: size of hist_list >= n_fields

    if (n_fields < 2) then
       return
    end if

    allocate(tmp, source = hist_list(1))

    do f = n_fields-1, 1, -1
       do ff = 1, f
          if (hist_list(ff)%field%name > hist_list(ff+1)%field%name) then

             call tmp%copy(hist_list(ff))
             call hist_list(ff  )%copy(hist_list(ff+1))
             call hist_list(ff+1)%copy(tmp)

          else if (hist_list(ff)%field%name == hist_list(ff+1)%field%name) then

             write(iulog,*) trim(subname),' ERROR: Duplicate field ', &
                hist_list(ff)%field%name, &
                't,ff,name=',t,ff,hist_list(ff+1)%field%name
          end if
       end do
    end do

  end subroutine sort_hist_list
```

I have not tested this in the actual CTSM code base, but I extracted it and modified it slightly in order to test it in a standalone file - see attached.

[testing.f90.txt](https://github.com/ESCOMP/CTSM/files/5599490/testing.f90.txt)

If you think this is a reasonable solution, I'm happy to help if you run into trouble when you go to actually implement and test it in CTSM.

**That said, if you saw other advantages to the rework you did in your last commit, please feel free to describe why you think that is better.**

slevisconsulting:
@billsacks your proposed code passed the test that I've been comparing to the baseline:
`ERP_P36x2_D_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs`
Thank you for sending such clear instructions.

I can begin the test suites when you think we're ready.

billsacks (approved):
Sorry for the delay. This looks good to me now - thank you!

slevisconsulting:
Both test suites passed, so this should be ready @billsacks

PR #1227, opened 2020-12-04
Update ChangeLog template

billsacks:
Main change is to instruct removing lines that don't apply in most sections (as we agreed upon at yesterday's CTSM software meeting). I have also tweaked some clarifying text in some places, and moved the
"significant changes" section higher up.

Thanks to @negin513 for suggesting this a while ago... though we didn't go with the suggestion at the time, I have eventually come around :-)

I am attaching examples of fairly minimal ChangeLog entries with the old and new conventions.
[ChangeLog_demo_old.txt](https://github.com/ESCOMP/CTSM/files/5643363/ChangeLog_demo_old.txt)
[ChangeLog_demo_new.txt](https://github.com/ESCOMP/CTSM/files/5643364/ChangeLog_demo_new.txt)

@ekluzek @negin513 @slevisconsulting @danicalombardozzi @olyson - If you get a chance, I'd like sign-off from all of you that you're happy with the new version. @danicalombardozzi and @olyson you can just look at the attached examples (no need to look at the diffs to the ChangeLog_template): is there anything that you'd prefer IS still called out explicitly rather than deleted? @ekluzek @negin513 @slevisconsulting I'd welcome feedback both based on the examples and on the clarity of explanatory text in the ChangeLog template.

### DIFF
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -70,26 +70,24 @@ Changes to tests or testing:
 
 Testing summary:
 ----------------
-[... Remove before making master tag.  Available test levels:
+[... Remove before making master tag.
 
-Model test levels one of the next five (one MUST be given):
+Nearly all CTSM tags should undergo 'regular' (aux_clm) testing.
+However, it occasionally makes sense to do more or less system testing;
+here is guidance on different available levels of system testing:
     a) doc (no source testing required)
     b) minimal (for use in rare cases where only a small change with 
                 known behavior is added ... eg. a minor bug fix. This
 		might be to just run the "short" test list, or to run
 		a single test. Whatever makes sense for the particular case.)
-    c) fates (Just run the fates test list on normal machines if CTSM source
-                only relating to fates is modified)
-    d) regular (regular tests on normal machines if CTSM source is modified)
-    e) release (regular tests plus the fates, ctsm_sci, aux_glc, mosart and rtm test lists
+    c) regular (regular tests on normal machines if CTSM source is modified)
+    d) release (regular tests plus the fates, ctsm_sci, mosart and rtm test lists
                 and normally all of the ancillary tests (build-namelist, python, ptclm, etc.) 
 		would be run as well)
-Optional Tools test level: add this if standard tools were changed:
-    f) tools (if tools are modified)
-    
-i.e. for regular testing with tools the test level would be: regular tools
-if JUST tools were changed it would be: doc tools
-which says testing of the source was not required, but the tools tests were required
+
+In addition, various other tests of the tools, python and perl
+infrastructure should be run when appropriate, as described below.
+
 ...]
 
 [Remove any lines that don't apply.]
@@ -113,10 +111,12 @@ which says testing of the source was not required, but the tools tests were requ
 
     (any machine) - 
 
-  regular tests (aux_clm):
+  regular tests (aux_clm: https://github.com/ESCOMP/CTSM/wiki/System-Testing-Guide#pre-merge-system-testing):
 
     cheyenne ---- 
     izumi ------- 
+
+  any other testing (give details below):
 
 If the tag used for baseline comparisons was NOT the previous tag, note that here:
 
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -75,7 +75,11 @@ Testing summary:
 Nearly all CTSM tags should undergo 'regular' (aux_clm) testing.
 However, it occasionally makes sense to do more or less system testing;
 here is guidance on different available levels of system testing:
-    a) doc (no source testing required)
+    a) no system testing (for use when the only changes are ones that
+                          have absolutely no impact on system runs; this
+                          includes documentation-only tags, tags that
+                          just change the tools or some python code that
+                          does not impact system runs, etc.)
     b) minimal (for use in rare cases where only a small change with 
                 known behavior is added ... eg. a minor bug fix. This
 		might be to just run the "short" test list, or to run
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -10,17 +10,6 @@ Purpose of changes
 [Fill this in with details]
 
 
-Bugs fixed or introduced
-------------------------
-
-Issues fixed (include CTSM Issue #): [If none, remove this line]
-CIME Issues fixed (include issue #): [If none, remove this line]
-
-Known bugs introduced in this tag (include github issue ID): [If none, remove this line]
-
-Known bugs found since the previous tag (include github issue ID): [If none, remove this line]
-
-
 Significant changes to scientifically-supported configurations
 --------------------------------------------------------------
 
@@ -37,8 +26,23 @@ Does this tag change answers significantly for any of the following physics conf
 
 [ ] clm4_5
 
+
+Bugs fixed or introduced
+------------------------
+[Remove any lines that don't apply. Remove entire section if nothing applies.]
+
+Issues fixed (include CTSM Issue #):
+
+CIME Issues fixed (include issue #):
+
+Known bugs introduced in this tag (include issue #):
+
+Known bugs found since the previous tag (include issue #):
+
+
 Notes of particular relevance for users
 ---------------------------------------
+[Remove any lines that don't apply. Remove entire section if nothing applies.]
 
 Caveats for users (e.g., need to interpolate initial conditions):
 
@@ -48,37 +52,34 @@ Changes made to namelist defaults (e.g., changed parameter values):
 
 Changes to the datasets (e.g., parameter, surface or initial files):
 
-Substantial timing or memory changes: [For timing changes, can check PFS test(s) in the test suite]
+Substantial timing or memory changes:
+[For timing changes, you should at least check the PFS test in the test
+ suite, whether or not you expect timing changes: ensure that its run
+ time has not increased significantly relative to the baseline.]
 
-Notes of particular relevance for developers: (including Code reviews and testing)
+
+Notes of particular relevance for developers:
 ---------------------------------------------
 NOTE: Be sure to review the steps in README.CHECKLIST.master_tags as well as the coding style in the Developers Guide
+[Remove any lines that don't apply. Remove entire section if nothing applies.]
 
 Caveats for developers (e.g., code that is duplicated that requires double maintenance):
 
 Changes to tests or testing:
 
-CTSM testing:
 
-[... Remove before making master tag.  Available test levels:
+Testing summary
+---------------
+[Remove any lines that don't apply.]
 
-    a) regular (must be run before handing off a tag to SEs and must be run
-     before committing a tag)
-    b) build_namelist (if namelists and/or build_system changed))
-    c) tools (only if tools are modified and no CTSM source is modified)
-    d) short (for use during development and in rare cases where only a small
-          change with known behavior is added ... eg. a minor bug fix)
-    e) doc (no source testing required)
+ [PASS means all tests PASS; OK means tests PASS other than
+  expected fails, including expected baseline failures.]
 
-... ]
-
- [PASS means all tests PASS and OK means tests PASS other than expected fails.]
-
-  build-namelist tests:
+  build-namelist tests (if CLMBuildNamelist.pm has changed):
 
     cheyenne - 
 
-  tools-tests (test/tools):
+  tools-tests (test/tools) (if tools have been changed):
 
     cheyenne - 
 
@@ -86,7 +87,7 @@ CTSM testing:
 
     cheyenne - 
 
-  python testing (see instructions in python/README.md; document testing done):
+  python testing (if python code has changed; see instructions in python/README.md; document testing done):
 
     (any machine) - 
 
@@ -124,6 +125,7 @@ Changes answers relative to baseline:
 
 Detailed list of changes
 ------------------------
+[Remove any lines that don't apply. Remove entire section if nothing applies.]
 
 List any externals directories updated (cime, rtm, mosart, cism, fates, etc.):
 
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -89,7 +89,7 @@ Optional Tools test level: add this if standard tools were changed:
     
 i.e. for regular testing with tools the test level would be: regular tools
 if JUST tools were changed it would be: doc tools
-which says no testing of the source was done, but the tools tests were also done
+which says testing of the source was not required, but the tools tests were required
 ...]
 
 [Remove any lines that don't apply.]
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -4,8 +4,8 @@ Originator(s):
 Date: 
 One-line Summary:
 
-Purpose of changes
-------------------
+Purpose and description of changes
+----------------------------------
 
 [Fill this in with details]
 
@@ -73,7 +73,7 @@ Testing summary
 [Remove any lines that don't apply.]
 
  [PASS means all tests PASS; OK means tests PASS other than
-  expected fails, including expected baseline failures.]
+  expected baseline failures.]
 
   build-namelist tests (if CLMBuildNamelist.pm has changed):
 
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -72,8 +72,7 @@ Testing summary
 ---------------
 [Remove any lines that don't apply.]
 
- [PASS means all tests PASS; OK means tests PASS other than
-  expected baseline failures.]
+ [PASS means all tests PASS; OK means tests PASS other than expected fails.]
 
   build-namelist tests (if CLMBuildNamelist.pm has changed):
 
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -68,8 +68,19 @@ Caveats for developers (e.g., code that is duplicated that requires double maint
 Changes to tests or testing:
 
 
-Testing summary
----------------
+Testing summary:
+----------------
+[... Remove before making master tag.  Available test levels:
+
+    a) regular (regular tests on normal machines if CTSM source is modified)
+    c) tools (if tools are modified)
+    d) minimal (for use in rare cases where only a small change with 
+                known behavior is added ... eg. a minor bug fix. This
+		might be to just run the "short" test list, or to run
+		a single test. Whatever makes sense for the particular case.)
+    e) doc (no source testing required)
+...]
+
 [Remove any lines that don't apply.]
 
  [PASS means all tests PASS; OK means tests PASS other than expected fails.]
@@ -82,7 +93,8 @@ Testing summary
 
     cheyenne - 
 
-  PTCLM testing (tools/shared/PTCLM/test):
+  PTCLM testing (tools/shared/PTCLM/test): (if cime or cime_config are changed)
+  (PTCLM is being deprecated, so we only expect this to be done on occasion)
 
     cheyenne - 
 
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -83,7 +83,8 @@ Optional Tools test level: add this if standard tools were changed:
     d) tools (if tools are modified)
     
 i.e. for regular testing with tools the test level would be: regular tools
-if JUST tools were changed it would be: tools
+if JUST tools were changed it would be: doc tools
+which says no testing of the source was done, but the tools tests were also done
 ...]
 
 [Remove any lines that don't apply.]
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -72,15 +72,20 @@ Testing summary:
 ----------------
 [... Remove before making master tag.  Available test levels:
 
-Model test levels one of the next three (one MUST be given):
-    a) regular (regular tests on normal machines if CTSM source is modified)
+Model test levels one of the next five (one MUST be given):
+    a) doc (no source testing required)
     b) minimal (for use in rare cases where only a small change with 
                 known behavior is added ... eg. a minor bug fix. This
 		might be to just run the "short" test list, or to run
 		a single test. Whatever makes sense for the particular case.)
-    c) doc (no source testing required)
+    c) fates (Just run the fates test list on normal machines if CTSM source
+                only relating to fates is modified)
+    d) regular (regular tests on normal machines if CTSM source is modified)
+    e) release (regular tests plus the fates, ctsm_sci, aux_glc, mosart and rtm test lists
+                and normally all of the ancillary tests (build-namelist, python, ptclm, etc.) 
+		would be run as well)
 Optional Tools test level: add this if standard tools were changed:
-    d) tools (if tools are modified)
+    f) tools (if tools are modified)
     
 i.e. for regular testing with tools the test level would be: regular tools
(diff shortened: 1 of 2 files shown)


doc/.ChangeLog_template
```diff
@@ -48,45 +52,42 @@ Changes made to namelist defaults (e.g., changed parameter values):
 
 Changes to the datasets (e.g., parameter, surface or initial files):
 
-Substantial timing or memory changes: [For timing changes, can check PFS test(s) in the test suite]
+Substantial timing or memory changes:
+[For timing changes, you should at least check the PFS test in the test
+ suite, whether or not you expect timing changes: ensure that its run
+ time has not increased significantly relative to the baseline.]
```
billsacks, line 58: @ekluzek @slevisconsulting @negin513 - I've gotten in the habit of checking the run time of the PFS test compared with baselines for nearly every tag (other than those where the changes are very minor). I think that's a good thing for us to do in general; do you agree? If so, is this a good enough way to call this out? I thought about adding this to the testing section, but it doesn't seem to completely fit there, since everything else in the testing section is more an objective pass/fail, whereas this is more subjective. I'm open to suggestions here.

By the way, once we update to a cime version that pulls in https://github.com/ESMCI/cime/pull/3786, this will become easier to check: you'll be able to grep for TPUTCOMP in the TestStatus.log file for the PFS test. At that point, I'll probably update the text in brackets to point that out.

olyson:
It sounds like you've already discussed the content, but one thing I find useful is the "Issues Fixed" section, which I use to update this wiki on bugs: https://github.com/ESCOMP/CTSM/wiki/Major-bugs-and-issues-clm5
I thought this was in the ChangeLog but I don't see it in ChangeLog_demo_old.txt, so maybe it already was removed...
Another thing is that answer changes used to be defined as roundoff, more than roundoff but same climate, and different climate.  Maybe this was removed because it was too subjective?  It was useful for me.

billsacks:
@olyson sorry, I should have explained things better. The examples I included were the bare minimum, which gives an example of what the ChangeLog entry would look like if there were no bugs fixed, no answer changes, etc.

> one thing I find useful is the "Issues Fixed" section

This will still be included just like it was before if there were actually any issues fixed. There are no changes in this respect (we have always deleted these lines if there were no issues fixed), except that I'm proposing moving it to a little lower down to make the "significant answer changes" section more prominent.

> Another thing is that answer changes used to be defined as roundoff, more than roundoff but same climate, and different climate. Maybe this was removed because it was too subjective? It was useful for me.

Again, this will still be included just like it was before if there were any answer changes, but will be excluded if there were no answer changes. This is the same as what we did before.

Please let me know if you have any other questions or concerns.

doc/.ChangeLog_template
ekluzek: This sounds good. I don't consistently check partially because in the past it wasn't reliable. But, it's good to know you think it's reliable now. And if so checking every time significant changes are made makes sense. The test is subjective, but it does result in a PASS/FAIL based on some arbitrary threshold. So I think reporting the result when it fails is what we should do. Machine variability makes me suspect of results that might show a small timing change that might not really be accurate.

Is there just a single PFS test? Machine variability means it might be better to have several unless the PFS test itself does multiple runs, which I don't think it does.

doc/.ChangeLog_template
```diff
@@ -48,45 +52,42 @@ Changes made to namelist defaults (e.g., changed parameter values):
...
+Notes of particular relevance for developers:
 ---------------------------------------------
 NOTE: Be sure to review the steps in README.CHECKLIST.master_tags as well as the coding style in the Developers Guide
+[Remove any lines that don't apply. Remove entire section if nothing applies.]
 
 Caveats for developers (e.g., code that is duplicated that requires double maintenance):
 
 Changes to tests or testing:
 
-CTSM testing:
```
ekluzek, line 61: I'd actually like to keep this section. I find it a useful reminder on what to test, and I always remove it before making a tag.

doc/.ChangeLog_template
```diff
@@ -48,45 +52,42 @@ Changes made to namelist defaults (e.g., changed parameter values):
...
+  tools-tests (test/tools) (if tools have been changed):
 
     cheyenne - 
 
   PTCLM testing (tools/shared/PTCLM/test):
 
     cheyenne - 
 
-  python testing (see instructions in python/README.md; document testing done):
+  python testing (if python code has changed; see instructions in python/README.md; document testing done):
```
ekluzek, line 90: Just a general note. Since these unit tests are quick, I always run them. It doesn't really hurt. And technically changes on the system could show up between tags that cause the test to fail.

Maybe this would be a good candidate to add automated testing for when a PR is submitted?

ekluzek (changes requested):
This is good! Thanks for working on this. I think this will improve the ChangeLog's.

I have a couple requests for changes. And I have a few questions that I ask.

doc/.ChangeLog_template
billsacks: > Maybe this would be a good candidate to add automated testing for when a PR is submitted?

Yeah, this would be an easier thing to add automated testing for; on the other hand, it wouldn't have a ton of value right now because the python code changes in so few PRs currently... but maybe that will change moving forward as we move more of our tooling (including build-namelist) to python.

doc/.ChangeLog_template
billsacks: This section has never made sense to me and has felt redundant with the list of possible tests below. Note that I did try to add some of the relevant information from here to the list of tests below. Can you explain what value this section has for you beyond what is already available in the list of tests below?

doc/.ChangeLog_template
ekluzek: Right part of what I was thinking was just that it would likely be easy to add in. And yes I expect that with the tool chain updates it will become more useful. Doing simple automated testing when we can, enables us to eventually figure out something that would be more complex, but more useful.

slevisconsulting:
> [...] clarity of explanatory text in the ChangeLog template.

Two comments that might be considered nit-picky:
- I would change "Purpose of changes" to "Purpose and description of changes".
- In the Testing summary I find it redundant to say "OK means tests PASS other than expected fails, including expected baseline failures." I interpret "expected fails" to include all expected fails.

doc/.ChangeLog_template
billsacks: > This sounds good. I don't consistently check partially because in the past it wasn't reliable. But, it's good to know you think it's reliable now. And if so checking every time significant changes are made makes sense. The test is subjective, but it does result in a PASS/FAIL based on some arbitrary threshold. So I think reporting the result when it fails is what we should do. Machine variability makes me suspect of results that might show a small timing change that might not really be accurate.
> 
> Is there just a single PFS test? Machine variability means it might be better to have several unless the PFS test itself does multiple runs, which I don't think it does.

I'll be honest: I've been struggling with this one. On the one hand, it would be good for us to notice when changes unintentionally slow the model down significantly. But on the other hand, I want to keep the check of this quick (e.g., no more than a minute or two of our time per tag) and relatively low in computational expense (because our running of the aux_clm test suite already takes so much of the SEWG computing allocation).

To be honest, I'm not sure how reliable it is. Up until now, I've just been checking the time of the test RUN phase, which tends to be reasonably consistent, but still sometimes has a big blip (which occasionally leads me to rerun the test). It does seem like the performance of the PFS test is more stable than many of our other tests, because there is relatively more computation time rather than i/o time, and I think there is significantly less machine variability in computation time compared with i/o time. (The PFS test doesn't do any output, but of course still does all the normal file input.)

I'm hesitant to add multiple PFS tests for the following reasons:

1. They are relatively expensive tests (20-day 1-degree tests currently)

2. Interpreting the results could be time consuming if we have multiple tests that we need to look at

3. I'm not sure how much value this would add if the multiple tests run in succession or all at once: my hunch is that the biggest variability comes from glade slowness, and my hunch (unconfirmed) is that, if one test is experiencing issues, then other tests run at that moment or a close time would be more likely to experience the same issues.

So my feeling has been that a single, high-resolution, long-ish test (as we currently have) does as good a job at averaging out the machine variability as we'd get from multiple tests run all at once.

Regarding the idea of using the PASS/FAIL: The default tolerance is set at 25%, which seems too high. I just saw that there's a way to specify a per-test throughput tolerance, so we could lower the tolerance for the PFS test. But with the way I've set up our standard cs.status script to ignore throughput tolerance failures in general (because there is a lot of variability for many other tests), this wouldn't show up in the cs.status results. So you would need to grep for this line in the PFS's TestStatus file; I'm not sure that has any advantage over grepping for it in TestStatus.log, which is where I put it in ESMCI/CIME#3786? (In principle, I guess we could suggest a new test phase in cime just for the PFS test that does the same thing as TPUTCOMP but is reported with a different phase name so that it doesn't get excluded when you are ignoring most TPUTCOMP results.)

As you can probably tell, I'm struggling here and don't feel like I have a good answer. Further thoughts are welcome.

doc/.ChangeLog_template
billsacks: Agreed.

doc/.ChangeLog_template
ekluzek: Obviously we need it to make sense to everyone making tags. This really came out of a CLM software discussion from years ago when we worked out that it was sometimes acceptable to NOT always run all of the testing. At that time we worked out some sample use cases when you wouldn't run all of the testing. The following is basically the table of those use cases. What I find useful with the table is to identify the testing level for a particular PR and then do that level of testing and report it that way. So it helps me to identify what tests I should be doing as well as what I should report on. And it also lets me feel like it's legit to not run all the testing for each tag. Otherwise, I tend to feel like I should always run all the testing. I'd expect it to be useful to others making tags in the same way.

Now, technically you are right that it's redundant in that you then report on what testing you actually run in the parts below. So if we agree that it's acceptable to not always run all the testing every tag, maybe it's not really needed. But, again I find it a useful table for me to figure out what level of testing I should do for a tag.

billsacks:
Thanks for your feedback @slevisconsulting 

> * I would change "Purpose of changes" to "Purpose and description of changes".

Good suggestion - done.

> * In the Testing summary I find it redundant to say "OK means tests PASS other than expected fails, including expected baseline failures." I interpret "expected fails" to include all expected fails.

I struggled a bit with this one. I'm not sure what was meant by the original, "OK means tests PASS other than expected fails". This was confusing to me because we have some tests in an expected fails list, and I (and I think others) have always reported tests as PASS if the only failures were ones in the expected fails list. So I've just used "OK" to mean tests pass but some baseline comparisons fail as expected. I just pushed a change that updates the wording with that meaning.

@slevisconsulting @ekluzek @negin513 are there other uses of "ok" besides "baseline comparisons fail as expected" that I'm not thinking of?

doc/.ChangeLog_template
billsacks: I remember having that discussion, but I've felt that the use cases listed here don't really apply anymore: Have we ever had a tag in the last few years for which we've just run the 'short' testing, for example; and is there ever a case where you would run the build-namelist testing but not the system tests? And on the other hand, this makes no mention of when to run ptclm or python testing. It also hasn't made sense to me to have distinct "levels"; for example, what would the "test level" be if I changed the tools and some obscure part of the python but nothing else?

I do still agree that there are (rare) cases when it makes sense not to run all of the testing – for example, if you just update something in the python directory that has no impact on most / any system runs; or if you just change documentation.

My feeling is that the best way to do this is to try to give rules for when each of the (currently five) categories of testing should be run. I have tried to do that for build-namelist,tools-tests and python testing. I'm not sure what the rule should be for PTCLM. We could also add a rule for regular tests, though I think it would be more subjective, like "should be run in nearly all cases, unless changes have no impact on system runs".

All that said: In the end, it's not a big deal to me if we keep this in if it's providing some real value for you. I can continue to routinely ignore it and just delete it when I make each tag.

slevisconsulting:
I am one of those who reports "OK" when I see any failure that appears in the expected list (not just baseline failures). That's why I prefer the more general wording "other than expected fails" without the "including..." part.

slevisconsulting:
> I am one of those who reports "OK" when I see any failure that appears in the expected list (not just baseline failures). That's why I prefer the more general wording "other than expected fails" without the "including..." part.

Of course I am happy to change how I report this if I had misinterpreted.

billsacks:
> > I am one of those who reports "OK" when I see any failure that appears in the expected list (not just baseline failures). That's why I prefer the more general wording "other than expected fails" without the "including..." part.
> 
> Of course I am happy to change how I report this if I had misinterpreted.

Oh, okay, thanks for the explanation. I'm fine either way; we should just be consistent with each other.

- [ ] @negin513 @ekluzek how do you prefer to handle this?

ekluzek:
@billsacks I also report OK when tests fail -- but it's OK because they are listed in the expected fails. And I do the same when baseline tests fail. Some of the tests just all PASS and it is good to say that. When some fail you have to check and make sure that the ones that fail are OK. So I also find this way of distinguishing the two cases helpful.

doc/.ChangeLog_template
ekluzek: Let's hear from @negin513 and @slevisconsulting to see if the table is useful to them.

I have done dead simple tags where I just run the short test or maybe one specific test. Very rare though as you point out. I find it hard to imagine a case where you would run the build-namelist testing -- but NOT the system tests. But, that could be the case if you were just fixing the build-namelist testing and didn't do anything that would affect any of the system tests. I think I've had that happen before.

PTCLM is being deprecated, but how about I made a stab at editing your PR to add a note around it? I've always thought of that test as something that I'm the only one that actually does it, but it's useful for me to see when I do. It's easy enough to edit the file directly in the PR if you are OK with my doing that.

doc/.ChangeLog_template
ekluzek: OK, having this one test sounds acceptable to me. 25% FAIL is too high, maybe 10 or 15%? We could set the threshold for this one test. I suspect machine variability might be around a few percent normally. The problem is that it can also just randomly be higher, any specific time. Maybe we say that if it fails -- you rerun the test a few times to make sure it's consistent? I think in general we are OK with changes in a few percent, but 10% is enough that we start to care.

billsacks:
> @billsacks I also report OK when tests fail -- but it's OK because they are listed in the expected fails. And I do the same when baseline tests fail. Some of the tests just all PASS and it is good to say that. When some fail you have to check and make sure that the ones that fail are OK. So I also find this way of distinguishing the two cases helpful.

Okay, thanks for the explanation. I have reverted this line to the earlier wording.

doc/.ChangeLog_template
billsacks: Yup, please go ahead and make any changes you'd like.

negin513:
Similar to @slevisconsulting and @ekluzek: I think using "OK" for both the expected fails and expected baseline failures is fine.

doc/.ChangeLog_template
negin513: I think this table has been useful for me to find the level of testing (especially at the beginning); however, I do agree with @billsacks that there are not any tags in the past year for which we've just run the `short` testing or `build-namelist` testing.

Overall, I think this includes useful information and we should keep it. Everyone should remove this paragraph before making a tag, so it does not cause any clutters.  I agree with @ekluzek to keep this in the template as a guide.

doc/.ChangeLog_template
billsacks: Okay, I have changed the throughput tolerance for the PFS test to 10% (just pushed a commit). Keep in mind that the failure won't show up when you run the standard 'cs.status' script that run_sys_tests creates, as noted above, so you'll need to explicitly check it. @ekluzek if there's any other wording you want to add here, feel free to do so. Once we update cime, we'll be able to say, "grep for TPUTCOMP in the TestStatus.log file in the PFS test" (which will show both the actual tolerance and whether it exceeds the given threshold).

doc/.ChangeLog_template
billsacks: @negin513 - to be honest, your statement that this was useful to you in the beginning is to me an argument that this should be removed or at least reworked: I feel like this is more misleading than helpful to a newcomer, since it gives the impression that there are different allowable levels of testing, whereas in fact nearly all of our tags need to be tested at the "standard" level.

That said, if it's useful to you and @ekluzek I'm still okay keeping it. But can one of you please rewrite it so that it reflects the current reality. Feel free to edit the text I've added in the list below describing when each category should be run in order to keep these two sets of instructions consistent.

slevisconsulting: > My feeling is that the best way to do this is to try to give rules for when each of the (currently five) categories of testing should be run.

I support this idea.

doc/.ChangeLog_template
```diff
@@ -4,23 +4,12 @@ Originator(s):
 Date: 
 One-line Summary:
 
-Purpose of changes
-------------------
+Purpose and description of changes
```
slevisconsulting, line 7: @billsacks thank you for making the change to this line. Now this raises a question (that I already wondered about before) about the line that says "Detailed list of changes" toward the bottom:

I take it to mean that we summarize the changes here and we describe them in detail near the bottom. I recommend more clarity about that if that's ok.

slevisconsulting (commented):
I am offering one more suggestion. I am happy to approve the PR if you need me to push the appropriate button :-)

doc/.ChangeLog_template
billsacks: Good point - that was confusing wording. Do you like my new wording better (that I just pushed)?

If you're happy with this, officially approving will help me track who is happy with the changes at this point :-)

doc/.ChangeLog_template
ekluzek: OK, so I reworked it -- feel free to comment.

This is a section that I've used and I've labeled the test level for my tags. As such I could go back and see that about 10% of those tags were something other than "regular". There was also about that amount where tools testing was done along with regular testing. I did have a few tags with short, but I changed the level description to "minimal" so you could then describe what you did. There were also a few with None or N/A that would probably now be done labeled as "doc" level.

I also removed the "build-namelist" level as that could really just be described by a special case of the "minimal" level (which in itself is really a special case category). Personally I run the build-namelist tests for every tag, because they are short and it's easy for me to fix them. So I never labeled any tags that I just ran that level.

ekluzek: I also added a ":" to the header for the section, as I plan to continue to add the level to my tags after the header.  It's mostly useful if you are doing something besides "regular", so I don't think we should require it. But, maybe recommend it when you are doing something besides "regular".

ekluzek: Can everyone comment if my changes are acceptable? If so I'll resolve this conversation.

slevisconsulting (approved):
I am approving and will let you all finalize the "Testing summary"

doc/.ChangeLog_template
```diff
@@ -48,45 +52,53 @@ Changes made to namelist defaults (e.g., changed parameter values):
...
 [... Remove before making master tag.  Available test levels:
 
-    a) regular (must be run before handing off a tag to SEs and must be run
-     before committing a tag)
-    b) build_namelist (if namelists and/or build_system changed))
-    c) tools (only if tools are modified and no CTSM source is modified)
-    d) short (for use during development and in rare cases where only a small
-          change with known behavior is added ... eg. a minor bug fix)
+    a) regular (regular tests on normal machines if CTSM source is modified)
+    c) tools (if tools are modified)
```
billsacks, line 76: - [ ] This should be changed to b, c, d rather than c, d, e

- [ ] Does the description under tools really mean, "if *only* tools are modified"? This is where I personally find it clearer to have rules for the individual test types rather than trying to define a test level: I would assume that a test level of tools means that you only ran the tools tests and not the regular tests, but I'm not sure if that's what you're thinking. (And does a test level of "tools" imply anything about whether you ran build-namelist tests, python tests, ptclm tests? This is where the whole notion of testing levels falls down for me: We have 2**5 = 32 possible combinations of test types that were / were not run with each tag, not counting the rare possibility of "minimal", so I have trouble describing that with just a couple of levels. Maybe I'm just being pedantic, but I wanted to try to explain why test levels don't make sense to my way of thinking. Anyway, it's not worth spending more time on this. If these categories make sense to you, I'm fine with this, but to be honest, I think I'll continue to ignore this because it doesn't make a lot of sense to me.)

danicalombardozzi:
All the changes and updates seem good to me. I typically look at the one-line summary and purpose [and description] of the changes, and I appreciate that they are front and center. I'm happy for the change log to include other things as outlined by folks above if they are useful

doc/.ChangeLog_template
ekluzek: OK, so I worked on it more. Fixing the numbering. And to try to explain, this is a summary of type of testing that you do, not a complete description of everything. The reason for it is to help pick what testing should be done, and also give a short summary of what was done. You should still fill out the details of what tests you actually ran. For example, I often run the build-namelist tests even if I mark it as "tools" test level. There the critical testing that needed to be run was the tools tests. But, if I ran other tests that's not bad. But, if I say it's a "regular tools" test level (so needs to run regular system tests and the tools tests) I should have run both the regular tests and the tools tests. And I should see it documented that way. I view the information on  if you ran the build-namelist tests, python tests, or ptclm tests, to be details of the testing that aren't really specified in the test level. That's why they are details that you list in turn. One of the reasons for giving the details about which tests were run -- is that I then expect those tests to work for that tag. But, I can get the general idea of what was done by looking at the test level. If I see the test level is "doc" I won't expect that you did much if any testing. If I see it was "minimal" I know I have to look at the details to see what you actually did. Since, "regular" is the usual case, I'll just expect that you at least ran the standard system tests. Having a test level for the non regular case is when this might be the most useful.

ekluzek (approved):

doc/.ChangeLog_template
ekluzek: @billsacks I'm OK if you don't use the test level and I do. Your comment on the number of factorials of testing is why I find the test level requirement summary helpful. It helps me to lay out what testing I should require for a given tag. So it's useful beforehand. And it's also a useful general summary afterwards as well so I can see if the testing done was just pretty standard or something special. I still want to also describe in detail what actual tests were run, as the level only gives the general requirements, not the specific details. Anyway, I worked on it more, and think it's described better now. And I actually added two other levels "fates" and "release" which are two levels I think will be useful to have. So hopefully this is better for everyone now.

billsacks (commented):
@ekluzek I hate to say it, but I found some of your latest changes made things more confusing to me. I have made another iteration on this, but I actually want to make one more minor change, which I'll push in a minute.

billsacks:
@ekluzek with my latest set of changes, I am finally happy – personally – with the description of testing levels. The key change for me was changing these to only talk about *system* testing levels, and presenting this as: "we usually do X, but occasionally do more or less, as given below". How is this to you?

doc/.ChangeLog_template
```diff
@@ -70,26 +70,24 @@ Changes to tests or testing:
 
 Testing summary:
 ----------------
-[... Remove before making master tag.  Available test levels:
+[... Remove before making master tag.
 
-Model test levels one of the next five (one MUST be given):
+Nearly all CTSM tags should undergo 'regular' (aux_clm) testing.
+However, it occasionally makes sense to do more or less system testing;
+here is guidance on different available levels of system testing:
```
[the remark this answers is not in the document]
ekluzek: Sounds good.

ekluzek (approved):

PR #1236, opened 2020-12-21
Mvertens/performance

### LINKED ISSUE #1083: NUOPC cap: use named constants rather than literal strings in state_getimport and state_setexport
I was looking quickly at the NUOPC cap. It looks like `state_getimport` and `state_setexport` return silently if the given field doesn't exist. I feel that this behavior, combined with the use of strings (rather than defined constants that can be checked at compile time) is an opening for bugs due to typos.

I feel it would be more robust to make these routines abort if the given field isn't found. I realize that there may then need to be some logic around the calls to `state_getimport` and `state_getexport` so that they aren't called if a given field is known not to exist.

For example, this line would probably then result in an error, because methane currently isn't added to the field list:

```Fortran
call state_getimport(importState, 'Sa_methane', bounds, output=atm2lnd_inst%forc_pch4_grc, rc=rc )
```

Alternatively, the use of strings throughout the cap could be replaced with constants. For example, the use of `Faxa_snowc` in these lines:

```Fortran
    call fldlist_add(fldsToLnd_num, fldsToLnd, 'Faxa_snowc'   )
```
```Fortran
    call state_getimport(importState, 'Faxa_snowc', bounds, output=forc_snowc, rc=rc )
```

would be replaced by reference to a named constant (whose value could still be `Faxa_snowc`). This would ensure that you don't try to reference a non-existent name in the `state_getimport` call.

@mvertens I'm tentatively assigning you. Feel free to comment if you feel my suggestions aren't feasible.

### LINKED ISSUE #1235: In nuopc cap: don't do math on garbage ndep fields when not receiving ndep from coupler
I noticed that the nuopc cap doesn't check the logical variable `ndep_from_cpl` when importing the ndep fields:

https://github.com/ESCOMP/CTSM/blob/c4cf1205bdd835b2cccbffe184e5d71858d58252/src/cpl/nuopc/lnd_import_export.F90#L540-L546

I don't understand how this works in the common case where we are *not* getting ndep from the atmosphere. For example, would the missing fields trigger a premature return (via the ChkErr return value)? Or would forc_ndep_grc get filled with garbage? Either way seems potentially problematic.

@mvertens can you provide input here?

mvertens:
### Description of changes
Performance updates for nuopc cap

### Specific notes
The cap has been rewritten to implement significant performance updates.

CTSM Issues Fixed:
Resolves #1235 
Resolves #1083 

Are answers expected to change (and if so in what way)? No (testing showed bfb results with noupc tests)

Testing performed, if any: 
ran the cmeps testlist_drv.xml and showed that the results for the following tests did not change:
```
ERS_Vnuopc_Ld5.f09_g17.I1850Clm50Sp.cheyenne_intel.clm-default
ERS_Vnuopc_Ld5.T31_g37_gl20.I1850Clm50SpG.cheyenne_intel.cism-test_coupling
ERS_Vnuopc_Ld5.f19_g17.I2000Clm51Bgc.cheyenne_intel.clm-default
```

### DIFF
--- a/src/cpl/nuopc/lnd_comp_nuopc.F90
+++ b/src/cpl/nuopc/lnd_comp_nuopc.F90
@@ -70,6 +70,7 @@ module lnd_comp_nuopc
 
   logical                :: glc_present
   logical                :: rof_prognostic
+  logical                :: atm_prognostic
   integer, parameter     :: dbug = 0
   character(*),parameter :: modName =  "(lnd_comp_nuopc)"
   character(*),parameter :: u_FILE_u = &
@@ -290,6 +291,14 @@ contains
        rof_prognostic = .true.
     end if
 
+    call NUOPC_CompAttributeGet(gcomp, name='ATM_model', value=cvalue, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    if (trim(cvalue) == 'satm' .or. trim(cvalue) == 'datm') then
+       atm_prognostic = .false.
+    else
+       atm_prognostic = .true.
+    end if
+
     call NUOPC_CompAttributeGet(gcomp, name='GLC_model', value=cvalue, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
     if (trim(cvalue) == 'sglc') then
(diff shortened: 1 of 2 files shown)


billsacks (approved):
With the changes I have pushed, I now approve this and will bring it to master soon.

Thank you @mvertens for your work on this! I appreciate that it improves performance without compromising (and maybe improving) readability!

billsacks:
I ran the one nuopc test in the aux_clm test suite, `SMS_D_Ld5_Vnuopc.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-default`; this passes and is bfb with baselines.

billsacks:
This is being superseded by a branch that has this performance work along with @mvertens 's changes for the domain file.

PR #1241, opened 2020-12-29
Small answer changes in preparation for adding option for bio-mass heat storage

ekluzek:
### Description of changes

Some small changes that can change answers in preparation of bringing in the option for bio-mass heat storage. This only changes answers for some compilers. This brings in some new terms (currently set to zero) that will be in use for clm5_1 when bio-mass heat storage is turned on. The terms that change are identical to the BHS version with the exception that some variables have a "p" index in the full BHS version as they are array terms rather than the scalars they are in here.

The full BHS version is PR #1016

### Specific notes

Are answers expected to change (and if so in what way)? Roundoff changes for some compilers

  Some new compsets for clm5.1
  Change one clm5.0 test to use the new clm5.1 compset, and add another test

Testing performed, if any: Limited testing on izumi

  I also added an explicit testing inside the code in an earlier version to ensure the changes were on the order of e-11/e-12 for the terms that change.

### DIFF
--- a/src/biogeophys/CanopyFluxesMod.F90
+++ b/src/biogeophys/CanopyFluxesMod.F90
@@ -1074,8 +1074,12 @@ contains
                   / ((1._r8-frac_rad_abs_by_stem)*(- 4._r8*bir(p)*t_veg(p)**3 &
                   + 4._r8*sa_internal*emv(p)*sb*t_veg(p)**3 &
                   +dc1*wtga+dc2*wtgaq*qsatldT(p))+ cp_leaf/dtime)
-            if ( abs(temp - dt_veg(p) ) > 1.e-12 )then
+            if( temp /= 0._r8 )then
+            if ( abs(temp - dt_veg(p) )/abs(temp) > 1.d-14 )then
+               write(iulog,*) 'Difference = ', temp - dt_veg(p)
+               write(iulog,*) 'R Difference = ', (temp - dt_veg(p))/temp
                call endrun( "Difference greater than roundoff" )
+            end if
             end if
             t_veg(p) = tlbef(p) + dt_veg(p)
             dels = dt_veg(p)
@@ -1097,8 +1101,10 @@ contains
                     - (efsh + dc1*wtga*dt_veg(p)) - (efe(p) + &
                     dc2*wtgaq*qsatldT(p)*dt_veg(p)) &
                     - (cp_leaf/dtime)*(t_veg(p) - tl_ini)
-               if ( abs(temp - err(p) ) > 1.e-12 )then
+               if( temp /= 0._r8 )then
+               if ( abs(temp - err(p) )/abs(temp) > 1.d-14 )then
                   call endrun( "Difference greater than roundoff" )
+               end if
                end if
             end if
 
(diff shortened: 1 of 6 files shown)


src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1049,6 +1066,17 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+            dt_veg(p) = ((1._r8-frac_rad_abs_by_stem)*(sabv(p) + air(p) &
+                  + bir(p)*t_veg(p)**4 + cir(p)*lw_grnd) &
+                  - efsh - efe(p) - lw_leaf + lw_stem &
+                  - (cp_leaf/dtime)*(t_veg(p) - tl_ini)) &
+                  / ((1._r8-frac_rad_abs_by_stem)*(- 4._r8*bir(p)*t_veg(p)**3 &
+                  + 4._r8*sa_internal*emv(p)*sb*t_veg(p)**3 &
+                  +dc1*wtga+dc2*wtgaq*qsatldT(p))+ cp_leaf/dtime)
+            if ( abs(temp - dt_veg(p) ) > 1.e-12 )then
+               call endrun( "Difference greater than roundoff" )
+            end if
```
billsacks, line 1079: @ekluzek - Thanks for all of your work to demonstrate only roundoff-level diffs. I know from my own experience how painful that can sometimes be.

One question: are the terms you're checking here all order 1 (in magnitude)? If not, or if you're not sure, I generally use relative diff checks for things like this:

```Fortran
if (abs(temp - dt_veg(p)) > 1.e-13_r8 * temp) then
```

ekluzek: That's a good point. Actually I'm sure they aren't order 1. dt_veg and err will be order 1, but could also be very small. ulrad and dlrad should be order 100. So doing a relative diff would be a better comparison. If I do that I'm likely to be able to show a relative difference that's a bit smaller as well.

billsacks: No need to redo things, though, if you have already done it this way and you have a general sense of the magnitude. My main worry with abs diffs is if the terms are very small to begin with - so an abs diff of 1e-12 could be a relative diff of, say, 1e-6. If you're pretty confident that isn't generally the case here, then I think this is good enough.

ekluzek: I went ahead and did a little testing with this to show it's fine. Relative diff's potentially a bit more troublesome if temp is identically zero, or negative. But with a relative diff I was able to show a tolerance of e-14 was fine for two tests.

ekluzek:
I think this is complete now. I'm running final testing and will tag once that's complete.

src/biogeophys/CanopyFluxesMod.F90
billsacks: > Relative diff's potentially a bit more troublesome if temp is identically zero, or negative.

Just a thought for the future: If `temp` is the old (presumed correct) version, then I usually have success with something like `if (abs(temp - dt_veg(p)) > 1.e-13_r8 * temp)` – or change the last `temp` to `abs(temp)` if negative values are acceptable. This avoids divide by 0. If temp (the old version) was exactly 0 at a point, then I typically expect the new version to be exactly zero, too, and I generally want to ensure that's true (which is done by the above) – since exactly 0 vs. slightly different from 0 can sometimes lead to large downstream differences.

Issue #229, opened 2018-01-29
Limit where dynamic crops are allowed


billsacks:
(From NCAR Teamwork site, 2015-11-11)

I found one area with a big potential for performance gains: Limiting where we allow dynamic crops. I did some performance runs with (a) CLM5BGC, (b) CLM5BGCCROP, and (c) CLM5BGCCROP, but only allocating memory for non-zero-weight crop columns (according to either year-1850 or year-2000 crop distributions... the difference between those two, in terms of performance, was much less than I expected).

**When we include crops, the impact of adding 0-weight columns is much greater than the ~ 10% I found when I was previously investigating the impact of dynamic landunits on performance: When running with crops, the impact of these 0-weight columns is a 1.59x increase in cost.** I think that part of the reason why I didn't consider this before was that these 0-weight columns had always been added, even before I started my dynamic landunits work.

**Currently, the impact of including crops is a 2.31x cost increase. This could be reduced to as low as ~ 1.5x if we only allocated memory where needed, according to the crop distribution in year-2000.** In practice, we would want to do something more than that, to allow for other changes with future scenarios. And things get harder if we want to allow coupling to an integrated assessment model, which could change flanduse_timeseries in unpredictable ways. But we could still achieve a significant cost savings if we restrict where dynamic crops are possible.

Note that this big cost savings comes from restricting the types of crops that can grow in each grid cell, in addition to the presence of any crops at all. If, instead, we say: "anywhere crops can grow, allocate memory for all possible crop types", then the savings are less (at best we reduce the 2.31x to 2.0x).

So how could we do this?

(1) The easiest thing from a software perspective would be to have a field on the surface dataset (or elsewhere) saying which crops are allowed to exist in each grid cell, ever. If someone was able to come up with such a dataset, the code mods needed to make use of this new field would be easy – I've pretty much already done it for the sake of these timing runs. The main downside of this approach is needing to come up with this field, particularly if we want to allow coupling to an integrated assessment model that may change flanduse_timeseries in unpredictable ways.

(2) An alternative would be to give more consideration to the idea of resizing arrays in memory as needed. It's possible that we could achieve this through something like init_interp; however, it feels like that would still take significant code rework, to allow reinitialization of the model at runtime. So I would not consider this a viable solution on the CESM2 timeframe.

billsacks:
(From NCAR Teamwork site 2015-11-11)

[...]

PR #1249, opened 2021-01-13
add namelist option to enable multiple columns having a single pft

swensosc:
### Description of changes
Enable ability to run patches on separate columns, i.e. to allow vegetation patches to evolve without competition from other pfts.
### Specific notes

Only when nameslist option is turned on; otherwise, no.
new namelist option: use_individual_pft_soil_column
Testing performed, if any:

### DIFF
--- a/src/main/lnd2glcMod.F90
+++ b/src/main/lnd2glcMod.F90
@@ -204,7 +204,9 @@ contains
       ! Make sure we haven't already assigned the coupling fields for this point
       ! (this could happen, for example, if there were multiple columns in the
       ! istsoil landunit, which we aren't prepared to handle)
-      if (fields_assigned(g,n)) then
+      !FIXTHIS!!!      if (fields_assigned(g,n)) then
+      ! This is commented out so that multiple soil columns can be enabled
+      if (1==2) then
          write(iulog,*) subname//' ERROR: attempt to assign coupling fields twice for the same index.'
          write(iulog,*) 'One possible cause is having multiple columns in the istsoil landunit,'
          write(iulog,*) 'which this routine cannot handle.'
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -1499,6 +1499,7 @@ sub process_namelist_inline_logic {
   setup_logic_subgrid($opts,  $nl_flags, $definition, $defaults, $nl);
   setup_logic_fertilizer($opts,  $nl_flags, $definition, $defaults, $nl);
   setup_logic_grainproduct($opts,  $nl_flags, $definition, $defaults, $nl, $physv);
+  setup_logic_pftsoilcolumn($opts,  $nl_flags, $definition, $defaults, $nl);
   setup_logic_soilstate($opts,  $nl_flags, $definition, $defaults, $nl);
   setup_logic_demand($opts, $nl_flags, $definition, $defaults, $nl);
   setup_logic_surface_dataset($opts,  $nl_flags, $definition, $defaults, $nl);
@@ -2088,7 +2089,12 @@ sub error_if_set {
    }
 }
 
-
+#-------------------------------------------------------------------------------
+sub setup_logic_pftsoilcolumn {                                                  
+  my ($opts, $nl_flags, $definition, $defaults, $nl, $physv) = @_;
+                                                                               
+  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'use_individual_pft_soil_column');
+}
 #-------------------------------------------------------------------------------
 
 sub setup_logic_soilstate {
--- a/bld/namelist_files/namelist_defaults_ctsm.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm.xml
@@ -155,6 +155,8 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <use_bedrock phys="clm5_0">.true.</use_bedrock>
 <use_bedrock phys="clm4_5">.false.</use_bedrock>
 
+<!-- Individual pft soil columns -->
+<use_individual_pft_soil_column>.false.</use_individual_pft_soil_column>
 
 <!-- Rooting profile namelist defaults -->
 <rooting_profile_method_water  phys="clm5_0">1</rooting_profile_method_water>
--- a/bld/namelist_files/namelist_definition_ctsm.xml
+++ b/bld/namelist_files/namelist_definition_ctsm.xml
@@ -144,6 +144,11 @@ specify spatially variable soil thickness. If not present, use bottom
 of soil column (nlevsoi).
 </entry>
 
+<entry id="use_individual_pft_soil_column" type="logical" category="clm_physics"
+       group="clm_inparm" valid_values="" >
+If TRUE, each pft exists on a separate soil column.
+</entry>
+
 <entry id="rooting_profile_method_water" type="integer" category="clm_physics"
        group="rooting_profile_inparm" valid_values="0,1,2" >
 Index of rooting profile for water
--- a/src/main/clm_varctl.F90
+++ b/src/main/clm_varctl.F90
@@ -273,6 +273,12 @@ module clm_varctl
   logical, public :: use_lai_streams = .false. ! true => use lai streams in SatellitePhenologyMod.F90
 
   !----------------------------------------------------------
+  ! each pft has individual soil column switch
+  !----------------------------------------------------------
+  
+  logical, public :: use_individual_pft_soil_column = .false. ! true => each pft exists on its own soil column
+
+  !----------------------------------------------------------
   ! bedrock / soil depth switch
   !----------------------------------------------------------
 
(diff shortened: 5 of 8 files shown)


billsacks:
I'm tentatively assigning this to @slevisconsulting . If he has time, he's going to work on this. @ekluzek I'm unassigning you, but I'll keep you as a reviewer in case you want to look at it and provide a review.

billsacks:
To first order, this looks reasonable. However, I have a number of concerns regarding both known and not-yet-known needs for this:

- [ ] How does this work with transient landuse?
    - I know that dynpft_interp needs to be updated
    - How should we determine the "template column" when a new column comes into existence? Currently, there's some code that treats the naturally vegetatated column as the template column when a new column comes into existence. The template column is used to set the initial biogeophysical state of the new column. We need some logic for how to choose the template column if there are multiple natural vegetated columns.
    - I think that a growing/shrinking vegetated landunit will currently lead to an equal growth/shrinkage of all columns (so that col%wtlunit will remain unchanged); if you would want something different from that, some code changes would be needed
    - Could use some testing in general

- [ ] For transient runs: Which columns have memory allocated for them?
    - I think this will work okay as is, but it's going to have big memory use implications until we do https://github.com/ESCOMP/CTSM/issues/229

- [ ] Which zero-weight (on the gridcell) columns are always active? Currently, zero-weight natural veg columns are always active, in part so that they can provide a spun-up state for themselves and other columns (this ties in with the above question about determining a "template" column). But this could have serious performance implications in a transient run, where a given pft - and thus, the column it sits on - may not be needed for decades. This ties in with the next question.

- [ ] For transient runs: Does any more thought need to be given to how the column weights are set in a grid cell that currently has 0 natural veg, but may grow in the future? I think what happens now is that you need to specify the PFT fractions adding to 100% in all grid cells, even those that currently have no natural veg. So maybe nothing more is needed here, because these PFT fractions will now end up giving the *column* areas. But this now carries somewhat more importance, in that the columns with non-zero area on the landunit will be active by default.

- [ ] Glaciers: handle smb forcing and maybe others

    - In lnd2glcMod:

    ```Fortran
      ! Make sure we haven't already assigned the coupling fields for this point
      ! (this could happen, for example, if there were multiple columns in the
      ! istsoil landunit, which we aren't prepared to handle)
    ```

    - See lnd2glcMod: bareland_normalization:

    ```Fortran
    ! Note: We currently aren't careful about how we would handle things if there are
    ! multiple columns within the vegetated landunit. If that possibility were introduced,
    ! this code - as well as the code in update_clm_s2x - may need to be reworked somewhat.
    ```

    - Anything else for glaciers?

- [ ] Check for any assumptions that there is just one vegetated column in init_interp
    - At the very least, we'll need to give each column its own index for init_interp. But there may be other changes needed as well.
    - See also this todo:
  https://github.com/ESCOMP/CTSM/blob/0f6985da747a9e0538df63bf6e3f0ff6b4341229/src/init_interp/initInterpMindist.F90#L637-L641

- [ ] Is this compatible with FATES? (If not, we should prevent running this combination)

- [ ] Is this compatible with CNDV? (If not, we should prevent running this combination)

- [ ] [optional] How confident are we that there aren't loops that assume only one natural vegetated column?
    This originally started as a vague memory of something Zack Subin told me 10 years ago, about needing to make sure we avoid loops that assume we only have one column in the natural veg landunit. But looking back at mods he sent me from 2011 (where I *think* he was addressing this issue for the sake of vegetated wetlands), he didn't actually have much to change, so I wonder if this concern is unfounded.

    Here's an idea for a test that would give me a fair amount of confidence: Set up a case where every vegetated landunit has two columns. Do two versions of this case: one where the columns (and the pfts on those columns) appear in reverse order of the other in memory. Gridcell-level averages from these two cases should only be roundoff-level different from each other. I would want this case to have the full complement of PFTs on each column (or close to it) to ensure that there is no code that assumes that there are at most maxsoil_patches or max_patch_per_col patches on the vegetated landunit.

    Before coming up with that test, here were some ideas of code searches we could do to try to find problem code:
    - search for maxsoil_patches and max_patch_per_col for possible candidate issues
    - could look at loops that operate over istsoil but not istcrop for possible issues (point being: anything that operates over crops would be more likely to handle this correctly... though it's still possible that those loops would assume that there is no more than maxsoil_patches patches on a landunit)
    - if we trust that we left ourselves good comments, we could search for 'assum' (capturing 'assume'/'assumption'/'assuming') in the code; I have looked through relevant comments in dyn_subgrid/, but not in other directories (there are about 300 instances of 'assum' throughout the source)
    - etc.

    **My feelings at this point on this particular todo are: if it's not too much work to set up the test idea given above (as a manual, one-time test: I don't envision this being an automated test), then that could be worthwhile to give us confidence that things are working right. If not, then it may NOT be worth the effort to look through the code for possible issues.**

dlawrenncar:
Thanks for the detailed consideration of this pull request.  I think it
would be useful to have a scientific discussion of some of these points,
perhaps at the next Software or Science meeting.  As I noted before, it
does seem that one option is to only allow this for non-transient
simulations if it is too complicated to figure out.  We don't currently
anticipate that this would be a commonly run configuration, so that should
be taken into consideration.  Understanding the implications for FATES
seems important.  Is this going to impede or help with FATES transient land
use plans, for example?

kjetilaas:
> Thanks for the detailed consideration of this pull request. I think it would be useful to have a scientific discussion of some of these points, perhaps at the next Software or Science meeting. As I noted before, it does seem that one option is to only allow this for non-transient simulations if it is too complicated to figure out. We don't currently anticipate that this would be a commonly run configuration, so that should be taken into consideration. Understanding the implications for FATES seems important. Is this going to impede or help with FATES transient land use plans, for example?
> […](#)

I just want to express interest in this PR, in case there has been some more discussions, or there are plans for that in another forum. This development would be a good starting point for some ideas I have on explicitly representing sub-grid snow heterogeneity. Also not something that would (at least in the near term) be commonly used, or something that would need transient land-use.

dlawrenncar:
Thanks for that input.  We will try to discuss at this week's software
meeting.

billsacks:
Based on discussion at today's ctsm-software meeting, we will defer trying to combine multiple vegetated columns with transient landcover change, since that is a larger task and not high priority. I have opened #1285 to track that.

- [ ] We should then add some error checking to abort the run if you try to enable this new option in a transient case. You can check `get_do_transient_pfts`, `get_do_transient_crops` and `get_do_transient_lakes`. I think it's harder to determine if transient glaciers are enabled, and it may be that we don't check that combination for now, and just rely on documenting that this option should not be used with evolving glaciers or any of those other options.

The other things in https://github.com/ESCOMP/CTSM/pull/1249#issuecomment-765725667 probably do need to be done. It's possible that we could defer the item related to glacier surface mass balance, but I'm not sure off-hand how to detect whether we're in a run where SMB is unimportant – because it can be important to scientists even in a run without CISM at all.

ekluzek:
I'm just pinging this again as it came up in #1596. It looks like there are some changes needed for this one, and @slevisconsulting is currently assigned to it. I don't know if it's active or not though. I've added "next" to it so that we'll talk about it at our next CTSM software meeting.

src/main/lnd2glcMod.F90
```diff
@@ -204,7 +204,9 @@ subroutine update_lnd2glc(this, bounds, num_do_smb_c, filter_do_smb_c, &
       ! Make sure we haven't already assigned the coupling fields for this point
       ! (this could happen, for example, if there were multiple columns in the
       ! istsoil landunit, which we aren't prepared to handle)
-      if (fields_assigned(g,n)) then
+      !FIXTHIS!!!      if (fields_assigned(g,n)) then
+      ! This is commented out so that multiple soil columns can be enabled
+      if (1==2) then
```
ekluzek, line 209: This obviously needs to be resolved.

ekluzek (changes requested):
@billsacks has several issues pointed out that need to be addressed. I concur with those. Outside of that I think this looks good other than the if statement in lnd2glcMod, but I'm pretty sure that's covered in Bill's comments.

slevisconsulting:
> I'm just pinging this again as it came up in #1596. It looks like there are some changes needed for this one, and @slevisconsulting is currently assigned to it. I don't know if it's active or not though. I've added "next" to it so that we'll talk about it at our next CTSM software meeting.

According to my notes from while I worked on the LILAC project:
I had started on #229 (see [status](https://github.com/ESCOMP/CTSM/issues/229#issuecomment-870769561) and planned to look at #1249 (this PR) after.

ekluzek:
It looks like this is an important development, but it looks to have a lot of work that needs to happen with it before it comes in. So I'm moving it to draft status.

@swensosc can you comment on how important this is to come in? Especially on CESM3 release timelines?

wwieder:
Enabling this with transient LULCC seems like a heavy lift, but maybe there's a simple way to enable this capability with static PFT distributions more easily?  @swensosc do you think this is something we'd want to enable sooner / later?

swensosc:
This is similar but separate to the hillslope use of multiple columns.  I
think this capability was meant for those who want to turn off competition
between pfts in the vegetated landunit.  So it may be a question more for
those who want to do that.

PR #1267, opened 2021-01-29
Adding IWUELN and VPD_2M diagnostic

### LINKED ISSUE #1245: intrinsic water use efficiency diagnostic

djk2120:
### Description of changes
I have added two new diagnostic fields available for history outuput.
IWUELN: the intrinsic water use efficiency at local noon (umolCO2/molH2O)
VPD_2M: vapor pressure deficit at 2m reference height (pascals)

### Specific notes
Both are patch-level variables. I have added VPD_2M for users who want to calculate ecosystem scale inherent water use efficiency relevant for flux tower point simulations.

  Fixes #1245 

 - only that there are two new diagnostic variables

 - Two new history fields: IWUELN and VPD_2M

Testing performed, if any:
 - 1-year global I2000SP simulation looked as expected. 
 - Also run_sys_tests for the clm_short test-list, which passed but for the new fields


src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1548,6 +1557,26 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+                  gs        = 1.e-6_r8*(laisun(p)*gs_mol_sun(p,iv)+laisha(p)*gs_mol_sha(p,iv))    ! 1e-6  converts umolH2O->molH2O
+                  if ( gs>0._r8 ) then
+                     iwue_ln(p)   = fpsn(p)/gs
+                  else
+                     iwue_ln(p)   = spval
+                  end if
+               else
+                  iwue_ln(p)   = spval
+               end if
+            end do
```
ekluzek, line 1578: We should add an "else" statement with an endrun so that it aborts if levcan is ever set to 1. We won't remember this detail when that happens, so need a reminder. Since, the endrun would be outside the loop that should be fine for efficiency.

- [ ] Add endrun if levcan /= 1

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1548,6 +1557,26 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
          
+         ! Calculate water use efficiency
+         !   does not support multi-layer canopy
+         if (nlevcan == 1) then 
+            do f = 1, fn
+               p = filterp(f)
+               c = patch%column(p)
+               g = patch%gridcell(p)
+               
+               if ( is_near_local_noon( grc%londeg(g), deltasec=3600 ) .and. fpsn(p)>0._r8 )then
```
ekluzek, line 1568: You could use isecsphr from clm_varcon here in place of the hardcoded 3600. But, that's not too big of a deal either way.

ekluzek (changes requested):
Just one change to make and one that you could do.

djk2120:
Just pushed a commit with the endrun added

ekluzek (approved):

ekluzek:
I merged this to the PPE branch and did full testing and many restart tests are failing mostly due to VPD2M, but some IWUELN. So I'm looking into this.

ekluzek:
I'm going to bring this one in with #1164.

ekluzek:
Closing this as there are conflicts and this branch was incorporated into #1164 which is coming in the next tag for ctsm5.1.dev054.

Issue #1270, opened 2021-02-04
Hexadecimal constants use non-standard Fortran


billsacks:
### Brief summary of bug

While trying to build with gfortran10 on my mac, I ran into an issue with what is apparently non-standard Fortran in the definition of two hexadecimal constants. The fix is easy, and I will open a PR for it shortly. This isn't high priority, but should go into the next simple bfb tag.

### General bug information

**CTSM version you are using:** master

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** all with gfortran10

### Details of bug

The issue appears in two places:

https://github.com/ESCOMP/ctsm/blob/0f6985da747a9e0538df63bf6e3f0ff6b4341229/src/biogeochem/FireEmisFactorsMod.F90#L230

https://github.com/ESCOMP/ctsm/blob/0f6985da747a9e0538df63bf6e3f0ff6b4341229/src/biogeochem/MEGANFactorsMod.F90#L277

This generates a build error like this:

```
  277 |     integer, parameter :: gen_hash_key_offset = z'000053db'
      |                                               1
Error: BOZ literal constant at (1) is neither a data-stmt-constant nor an actual argument to INT, REAL, DBLE, or CMPLX intrinsic function [see '-fno-allow-invalid-boz']
```

The fix is simply to wrap this in an 'int()' call. Alternatively, the compilation flag `-fallow-invalid-boz` can be given to gfortran to bypass the error.

PR #1271, opened 2021-02-04
Fix non-standard hexadecimal constant

billsacks:
### Description of changes

See #1270 for details

This should be folded into an upcoming tag.

### Specific notes

- Resolves ESCOMP/CTSM#1270

Testing performed, if any:
Tested on my mac with test `SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRs.bishorn_gnu.clm-o3`. With gfortran10, this passes and is bit-for-bit with a test without these changes but instead with FFLAG `-fallow-invalid-boz` added.

### DIFF
--- a/src/biogeochem/FireEmisFactorsMod.F90
+++ b/src/biogeochem/FireEmisFactorsMod.F90
@@ -227,7 +227,7 @@ contains
     integer :: i
     integer :: strlen
     integer, parameter :: tbl_max_idx = 15  ! 2**N - 1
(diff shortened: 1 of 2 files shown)


ekluzek:
I did some testing of this on the PPE branch, and didn't see any answer changes for the intel compiler.

billsacks:
I should have mentioned: I folded this into my upcoming bfb tag.

Issue #1273, opened 2021-02-05
Add testmod for FATES land use


glemieux:
Add fates testmod to provide coverage for FATES landuse introduced in #1040.  See [software engineering meeting notes](https://github.com/ESCOMP/CTSM/wiki/Meeting-Notes-2020-Software#jul-7th-2020) for relevant discussion.

Issue #1266, opened 2021-01-29
Add test for successful_ignitions mode for FATES-SPITFIRE


glemieux:
Per @ekluzek [comment](https://github.com/ESCOMP/CTSM/pull/1264#discussion_r566406680), we should create a testmod option for enabling `fates_spitfire_mode = 3`.  Currently this is only referenced in `Fates_nat_and_anthro_ignitions` testmod (which itself enables `fates_spitfire_mode = 4`):

https://github.com/ESCOMP/CTSM/blob/5f348cac1919e70df19a6edbbf7b970c3877ce22/cime_config/testdefs/testmods_dirs/clm/Fates_nat_and_anthro_ignitions/README#L14-L26

glemieux:
@ckoven do we need permission to move this data set into the CESM inputdata file structure?

ckoven:
possibly? I'll ask.

wwieder:
@ekluzek and @billsacks what do we need to do to bring this into  CESM inputdata?

ekluzek:
> @ekluzek and @billsacks what do we need to do to bring this into CESM inputdata?

It's an easy task to do. We just need to copy it over into $DIN_LOC_ROOT/lnd/clm2/firedata and then import it into svn inputdata. We also need to setup a test configuration for this. This dataset is only for California, so we should have a surface dataset that's regional and just covers California or a part of California. The hard part of this is Charlie's ask for permission  to save the dataset that he mentions doing above.

glemieux:
Just a heads up that I've got a branch in the works to address this issue and some of the other fates-related test issues (#1265,
#1273).

lmkueppers:
I'll check with Yufang and Bin about the dataset. I presume that including means that it would be publicly available to anyone, right?

ekluzek:
> I'll check with Yufang and Bin about the dataset. I presume that including means that it would be publicly available to anyone, right?

Yes, exactly. It would be publicly available for anyone. It wouldn't be obvious to find it, as it would be "hidden" in all the myriad of CESM input files. 

As an alternative to publicly available to anyone, we could reference to its location on cheyenne, and only use it there. I'd prefer if it could just be part of the normal CESM system though so would be downloaded if anyone happened to do this test. It wouldn't normally come up for anything standard it would require someone to have to do this specific test.

ekluzek:
[...]

PR #1264, opened 2021-01-27
Merge fates_main_api into master

glemieux:
### Description of changes

This PR merges fates_main_api into master.  The fates_main_api branch has integrated updates from master up to tag [ctsm5.1.dev020](https://github.com/ESCOMP/CTSM/releases/tag/ctsm5.1.dev020) per PR #1257.  The purpose of this merge is to bring the most recent fates updates into master and for fates future development to continue from ctsm master instead of on fates_main_api.

### Specific notes

The `aux_clm` suite tests were compared against `ctsm5.1.dev021` baseline.  The `fates` suite tests were compared against `fates-sci.1.43.2_api.14.2.0-ctsm5.1.dev020-C5f348cac-F5534a940` baseline.

Contributors other than yourself, if any: @ekluzek 

CTSM Issues Fixed (include github issue #): none

Are answers expected to change (and if so in what way)?  Only for `aux_clm` fates testmods only.  All else b4b.

Any User Interface Changes (namelist or namelist defaults changes)? none

Testing performed, if any:
- `aux_clm` on cheyenne: `/glade/u/home/glemieux/scratch/tests_0126-161304ch`
- `aux_clm` on izumi: `/scratch/cluster/glemieux/tests_0204-142628iz`
- `fates` on cheyenne: `/glade/u/home/glemieux/scratch/clmed-tests/basegen.fates-sci.1.43.2_api.14.2.0-ctsm5.1.dev021-C4c9a4c15-F5534a940`

**NOTE: Be sure to check your coding style against the standard
(https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines) and review
the list of common problems to watch out for
(https://github.com/ESCOMP/CTSM/wiki/List-of-common-problems).**

Issue #514, opened 2018-09-18
Add a multi-instance FATES test to the FATES test suite


rgknox:
Several FATES researchers are using multi-instance with FATES.  
I think this needs a test, at least added to category="fates"

I was thinking:

NCK_Ld9.f10_f10.I2000Clm50FatesGs  

(NCK) is the multi-instance header right?

@billsacks , @ekluzek , any thoughts, and/or do you want this test also added to aux_clm?

billsacks:
Sounds good. Yes, NCK is a multi-instance test. Note that this test is fairly weak: it ensures that you can run with 2 identical instances, and compares one of them (or maybe both?) to an identical single-instance run. There is recognition that a stronger test is needed, but nobody has taken the time to write one (e.g., see the discussion in https://github.com/ESMCI/cime/issues/445).

Another option is to do something like ERP_N2, which does a two-instance ERP test (confirming that both instances start up from restart properly). It might be worth having both an NCK and an ERP_N2 FATES test in the FATES test list. I don't think we necessarily need them in the aux_clm test list: generally, if multi-instance breaks, it's because of something at a very high level in CTSM (e.g., in the coupling layer), so it generally doesn't seem necessary to test multi-instance with different options on/off - but I can see how you'd want this in the fates test list to ensure yourselves that this is working with fates.

rgknox:
>I don't think we necessarily need them in the aux_clm test list: generally, if multi-instance breaks, it's because of something at a very high level in CTSM (e.g., in the coupling layer), so it generally doesn't seem necessary to test multi-instance with different options on/off - but I can see how you'd want this in the fates test list to ensure yourselves that this is working with fates.

Makes sense.

But I am starting to wonder now if this is really a test needed for FATES run specifically.  It seems that the multi-instance functionality is fairly orthogonal to anything FATES relevant.  I brought up the idea for creating this test, because @ckoven was having multi-instance failing on edison.  But that seems to really be more of a machine compatibility issue, and not a fates vis-a-vis multi-instance issue...  Although, I suppose that is why we test, because we don't know what will be a problem, until it fails...

billsacks:
I don't have strong feelings on this. Again, my experience is that, like you say, multi-instance is pretty orthogonal to anything related to specific options. However, there could be aspects that I'm not thinking of, such as if FATES handles any of its own creation of input / output file names.

glemieux:
[...]

PR #1275, opened 2021-02-05
FATES testing update and cleanup

### LINKED ISSUE #1265: Diversifying fates testmods included in `aux_clm` test suite
Per our [discussion in the software meeting today](https://github.com/ESCOMP/CTSM/wiki/Meeting-Notes-2021-Software#fates-merge-into-ctsm-master), this issue is meant to provide us a place to discuss the diversification of fates testmods in the `aux_clm` test suite.  With the upcoming merging of fates_main_api into ctsm master (#1264) and all future development happening directly off of master, I'm thinking that we want to have more coverage of a variety of fates testmods.  My concern is that with `aux_clm` only using `FatesColdDef` for the fates testmod, we might miss an answer changing result due to a ctsm-side changes (i.e. not initiated by fates api updates) for different fates configurations.  I'm not sure how likely this is to be though.  

Here's the questions that I have:

1. Based on @billsacks suggestion, what subset of `aux_clm` fates tests should we _replace_ with tests from `fates` suite.  The idea being that we want to minimize the number of new tests added to `aux_clm`.
2. What compset coverage is necessary?  The `fates` suite predominantly tests using `I2000Clm50FatesCru` where as `aux_clm` fates test uses `I2000ClmXXFates`, where `XX` are the different clm versions.  Similarly, should we diversify the compset coverage in the `fates` suite?
3.  From a process standpoint, if a ctsm-side change does create answer changes to fates, how do we want to handle assessing the such an update?  My assumption is that given fates team attendance to the ctsm software meeting, we will not likely miss out on fates answer-changing updates and will discuss stuff there and here on github.

glemieux:
### Description of changes

This PR is being created in response to findings during review of #1264.  At that time it was decide to file a number of issues and resolve them separately from that PR so as reduce the complexity that renaming tests can cause during a pull request.  

### Specific notes

This PR resolves #1265, #1266, and #1273.  Discussion on addressing #514 are in progress.

Contributors other than yourself, if any: @ekluzek @rgknox 

CTSM Issues Fixed (include github issue #): #1265  #1266 #1273

Are answers expected to change (and if so in what way)? Yes due to removal of erroneous cold start `xmlchange` in baseline `Fates` testmod.

Testing performed, if any:
TBD


glemieux:
Task list:
- [x] Remove erroneous `./xmlchange CLM_FORCE_COLDSTART="on"` in `Fates` base testmod
- [ ] Remove clm45 tests from cheyenne and izumi `aux_clm` category
- [ ] Update cheyenne `aux_clm` with fates testmods
- [x] Add fates land us testmod
- [x] Rename `Fates_nat_and_anthro_ignitions`
- [ ] Add `ColdDef` suffix to all fates testmods that use cold start
- [x] Remove redundant `fireFates` testmod

glemieux:
With regard to the [discussion on which testmods not to add](https://github.com/ESCOMP/CTSM/issues/1265#issuecomment-770095796) to `aux_clm`, I've also elected not to add the new FateColdDefLandUse and the `successful_ignitions` testmods since both will be new.  Additionally, I've decided not to include FateFireLightningPopDens since we only have one version of it on fates and we're still figuring out the [ERS error related to it](https://github.com/ESCOMP/CTSM/issues/667#issuecomment-640921719).

glemieux:
Adding a stretch goal to the task list:
_Expected failures_:
- [ ] Fix `fates` suite gnu-only floating point error #1338
- [ ] Fix `fates` suite exact restart errors on anthro and hydro

cime_config/testdefs/testmods_dirs/clm/Fates/shell_commands
```diff
@@ -1,3 +1,2 @@
 ./xmlchange CLM_BLDNML_OPTS="-no-megan" --append
 ./xmlchange BFBFLAG="TRUE"
-./xmlchange CLM_FORCE_COLDSTART="on"
```
rgknox, line 3: Is this variable deprecated? Unnecessary?

glemieux: I had initially thought this was reverting a mistake that I made during the transition from `fates_main_api` to `master`, but it looks like it was introduced a while back in https://github.com/ESCOMP/ctsm/commit/34d8a33b80528b49cd94cd76645eefac26acd11b, so I guess I was wrong.

That said, I think the idea was that since we have a cold start testmod we should remove the coldstart from the generic `Fates` testmod and just make everything use `FatesColdDef` (with a possible look towards the future where we use `Fates` testmod with initialization files I think).  Spurred by your question I went to double check and it looks like I overlooked `FatesPRT2` testmod.  I'll fix it to use `FatesColdDef`.

- [ ] Update `FatesPRT2` testmod to use `FatesColdDef`

ekluzek: Yes, I suggested to @glemieux that the base Fates test-mod NOT have the cold-start embedded in it, so that tests that do a cold start could inherit the cold-start as a separate test mod to include (which allows the way the cold start is done to change). This also allows some tests to inherit from a test mod that is NOT a cold-start. If you embed the FATES base test-mod with a cold-start you only allow FATES tests to be cold-starts.

glemieux:
Updated single site compsets with https://github.com/ESCOMP/CTSM/pull/1275/commits/62fb4478e2047ff0439818e85186845a68a3a6e5 to avoid using MOSART partially in response to https://github.com/ESCOMP/MOSART/issues/45

glemieux:
While going through PR #1182, I noticed that there appears to be a bunch of defunct or out-of-date expected failures and related issues.  I should go through these and clean them up as appropriate

[ ] Review ExpectedTestFails.xml, remove out-of-date entries, and close related issues

glemieux:
The fates core team discussed changing the default set of history output variables for the `Fates` and `FatesColdDef` testmods.  The thought was that the list was too short and not entirely representative of the available output.  Initially we thought about identifying a minimum set of necessary variables that would capture the greatest extent of all calculations possible, but for times sake we decided to simply use the active, non-multiplexed variables.

glemieux:
I've decided to break out the `successful_ignitions` spitfire mode regression test to its own PR: #1507.  I'm having trouble replicating a successful case run for this mode and need more time to discuss how to set it up properly.

rgknox:
@glemieux, I saw the updated test list in the last commit. It looks like we are no longer using empty tapes for our base list, and the added variables seem to include the extra dimension, just not multiplexed.  How did you design this list? Does it include all non default variables that aren't multi-plexed?

glemieux:
The added variables are all the active, multiplexed variables, but they are set to the `hist_fexcl1` variable that Charlie told us about the other day (so the list of variables should be excluded from the end result).  The list of active, multiplexed variables was shorter than the list of active, site-only variables (61 versus 86) so I elected to exclude the shorter list rather than clear the tapes and include the longer list.  I'm just about to test this out to see if it works as advertised.

rgknox:
Thanks for the explanation.  Looks good!
The work @adrifoster is doing in fixing our naming conventions will help greatly in making things clearer.  I was looking at some of those variables, like RECRUITMENT, and thinking "why is that excluded?"  Then I go to the registry and see it is dimensioned by PFT.

glemieux:
> Thanks for the explanation. Looks good! The work @adrifoster is doing in fixing our naming conventions will help greatly in making things clearer. I was looking at some of those variables, like RECRUITMENT, and thinking "why is that excluded?" Then I go to the registry and see it is dimensioned by PFT.

Totally; that was exactly my same response too.

ckoven:
I guess one question with using `hist_fexcl` is how robust it is with respect to changing variable names.  I.e., does the run crash if you try to exclude a variable that is no longer there because its name has been changed (as is the case with `hist_fincl`)? If so, that would seem to require a coordinated change between the testing infrastructure and the fates code if any of those variable names get changed.  Possibly not a big deal if we are just doing the one big change that @adrifoster is working on, but just wanted to check.

adrifoster:
@ckoven yes I had been thinking about this - it seems like we should just skip over variables in the exclude list that don't exist - and perhaps print a warning message so future cases can be modified?

I suppose the downside is that your run may be longer than you want if you don't end up actually excluding the variable you meant to exclude.

cime_config/testdefs/testmods_dirs/clm/Fates/shell_commands
rgknox: @ekluzek, if there is no restart, or finitdat file involved, what happens differently when CLM_FORCE_COLDSTART="on" ?   Does the force coldstart setting tell the scripting to ignore the file, or to use a subset of its information or something?

glemieux:
I ran preliminary tests last Friday to check the `hist_fexcl` output; I should have thought about the fact that using this, but not clearing the tapes mean that all the `active` clm multiplexed variables are getting included.  I'm going to revert to clearing the tapes and using `hist_fincl`.

cime_config/testdefs/testmods_dirs/clm/Fates/shell_commands
ekluzek: @rgknox CLM_FORCE_COLDSTART tells build-namelist to explicitly set finidat to blank so that no restart file will be used. And it won't use any of the default finidat files that are in namelist_defaults even if they match.  It will abort with a warning if you do try to set an finidat file and have CLM_FORCE_COLDSTART=="on". This only applies to a "startup" case, for a hybrid or branch case it will use the finidat or nrevsn file as provided there.

For FATES we don't have finidat files that can be used, so it doesn't really apply much. We don't have a mechanism to ignore part of the finidat file. Although there are specific variables that are allowed to be missing for a startup case, whereas a restart or branch requires all restart variables to be provided. But, that's a variable by variable choice and there aren't that many that are allowed. Does that help?

glemieux:
Ran `aux_clm` on Cheyenne against `ctsm5.1.dev057`.  All expected tests pass, although one test hit the wallclock limit: `SSP_D_Ld4.f09_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-ciso_rtmColdSSP`.  @ekluzek I'm manually bumping up the wallclock limit and retesting this one.  I see from another issue that this is an expensive test; can I safely ignore this if it passes with an increased wallclock limit?

ekluzek:
Yes, if it's just a wallclock limit that's a pass if it runs with an increase. I would suggest increasing the wallclock limit in the testlist if it looks like hitting the wallclock limit is likely to happen again.

glemieux:
Huh, resubmitting it, it ran well within the limit this time.  It looks like the previous run might have gotten hung up while reading in a stream file.  Path to the whole test suite is here: `/glade/u/home/glemieux/ctsm/tests_fates-testmods-update-aux_clm`

cime_config/testdefs/testmods_dirs/clm/Fates/shell_commands
rgknox: yes, thanks!

rgknox (approved):

glemieux:
Interestingly, running`ERP_D_P32x2_Ld3.f19_g17.I2000Clm50FatesCru.cheyenne_intel.clm-FatesColdDef` (i.e. with MOSART), passes the RUN.  I reran the river stub compset for this test from a new test build and it failed for a 4th time in a row.  I think this strongly suggests that running with MOSART in this case somehow is more stable.
Passing test directory: `/glade/scratch/glemieux/ctsm-tests/tests_fates-testmodupdate-32x2-mosarton`
Failing test directory: `/glade/scratch/glemieux/ctsm-tests/tests_fates-testmodupdate-32x2-mosartoff`

glemieux:
> Hmmm. This is a bit disconcerting, because whether there is active ROF shouldn't matter. And I can think of more reasons why having MOSART on would cause trouble than having it off. So I think this is a concerning issue. I recommend filing it as an issue that we should look at some point, because it might indicate something really wrong in memory for example.

Will do @ekluzek.  One of the other things I noticed: I think this is the only `P32x2` test that doesn't use a `Bgc` or `Sp` compset in the test list.

jkshuman:
@ekluzek looks like this is all merged. Can this be closed?

PR #1276, opened 2021-02-06
Restructure ozone code

billsacks:
### Description of changes

Restructure ozone code to better accommodate the new ozone parameterization in #1232 . The changes here were heavily inspired by @ziu1986 's changes in #1232. My main goal was to demonstrate how we can bring in a new ozone stress formulation using a select case statement rather than object-oriented polymorphism, because this will keep the code simpler. A significant side benefit of this restructure was to allow me to remove the ozone stress terms from the restart file.

### Specific notes

Contributors other than yourself, if any: Heavily motivated by changes from @ziu1986 as well as conversations with @sunnivin

Testing performed, if any:
```
ERP_D_Ld5_P48x1.f10_f10_musgs.I2000Clm50Sp.izumi_nag.clm-o3
ERP_D_P36x2_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default
ERP_P72x2_Ly3.f10_f10_musgs.I2000Clm50BgcCrop.cheyenne_intel.clm-irrig_o3_reduceOutput
```

All pass and are bit-for-bit with baselines.

### DIFF
--- a/src/biogeophys/OzoneMod.F90
+++ b/src/biogeophys/OzoneMod.F90
@@ -18,6 +18,8 @@ module OzoneMod
   use clm_varcon  , only : spval
   use OzoneBaseMod, only : ozone_base_type
   use abortutils  , only : endrun
+  use PatchType   , only : patch
+  use pftconMod   , only : pftcon
 
   implicit none
   save
@@ -52,12 +54,16 @@ module OzoneMod
      ! Public routines
      procedure, public :: Init
      procedure, public :: Restart
+     procedure, public :: CalcOzoneUptake
      procedure, public :: CalcOzoneStress
 
      ! Private routines
      procedure, private :: InitAllocate
      procedure, private :: InitHistory
      procedure, private :: InitCold
+
+     ! Calculate ozone uptake for a single point, for just sunlit or shaded leaves
+     procedure, private, nopass :: CalcOzoneUptakeOnePoint
 
      ! Calculate ozone stress for a single point, for just sunlit or shaded leaves
      procedure, private, nopass :: CalcOzoneStressOnePoint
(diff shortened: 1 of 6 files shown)


src/biogeophys/OzoneMod.F90
```diff
@@ -472,6 +425,108 @@ subroutine CalcOzoneStressOnePoint( &
...
+    integer  , intent(in) :: filter_exposedvegp(:)     ! patch filter for non-snow-covered veg
+    !
+    ! !LOCAL VARIABLES:
+
+    character(len=*), parameter :: subname = 'CalcOzoneStress'
+    !-----------------------------------------------------------------------
+
+    select case (this%stress_method)
+    case (stress_method_lombardozzi)
+       call this%CalcOzoneStressLombardozzi(bounds, num_exposedvegp, filter_exposedvegp)
```
billsacks, line 449: Here is where this will be extended to allow the Luna-specific method from @ziu1986. I envision that CalcOzoneStress will be called from the same place in the code in either case (currently clm_driver.F90) - i.e., the caller won't need to know which ozone stress method is being used.

It looks like the Luna-based method only needs to calculate the stress coefficient once a day, at the end of the day. So, to give a small performance improvement, we can put an end-of-day check either in the case block (as a conditional around the call to the new routine) or at the top of the new routine (simply returning if it isn't the end of the current day).

In terms of the calculation of the different stress coefficients (of which there will be 6): My thinking is that we can rely on the fact that they are all initialized to 1 in InitCold. So the new o3coefjmax will *not* be set explicitly in CalcOzoneStressLombardozzi, and similarly, the original o3coefv and o3coefg variables will *not* be set explicitly in your new method.

ziu1986: Am I right to assume that instead of storing the O3 coefficients we are going to store the O3 uptake in the history/restart files? Not sure if I have seen that in the proposed changes.

billsacks: The o3 uptake was already in the restart files. This is needed because it is an evolving state variable. Previously, we also stored the o3 coefficients in the restart file. That was only needed because of the driver loop ordering: the coefficients were calculated after photosynthesis in time step N and then applied in photosynthesis in time step N+1. I have reordered the code here so that the coefficients are calculated before photosynthesis in time step N, and then applied later that same time step, so they no longer need to be on the restart file.

Issue #1280, opened 2021-02-16
Snow density unrealistically large for very thin snowpacks


swensosc:
### Brief summary of bug
Snow densities derived from ctsm history output of H2OSNO and SNOWDP can be unrealistically large (>1000kg/m3) 
[Give a one or two sentence summary. This could be the same as the issue title if you feel that is a sufficient summary.]

### General bug information
Calculating snow density as rho=h2osno/snowdp from time step frequency history files shows very large values at times when the snowpack is very thin.  This typically occurs when the snowpack is unresolved, i.e. snl is 0 but h2osno_no_layers > 0.  The issue appears to happen because frac_sno (snow cover fraction) is held fixed throughout the timestep, while h2osno and snow_depth are updated due to phase change and frost deposition.  I corrected the behavior by making changes to how snow_depth is calculated during phase change, and using an updated value of frac_sno in the calculation of snowdp.  
**CTSM version you are using:** [output of `git describe`]

**Does this bug cause significantly incorrect results in the model's science?** [Yes / No]

**Configurations affected:** [Fill this in if known.]

### Details of bug

[Fill in details here.]

### Important details of your setup / configuration so we can reproduce the bug

[Specify anything relevant: the compset, resolution, machine, compiler, any xml or namelist changes, etc. You don't have to repeat anything that you have already noted above.]

### Important output or errors that show the problem

[Fill this in with anything relevant that you haven't already noted; if there is nothing to add, delete this section.]

wwieder:
looks like we're close to having a fix for the in CTSM5.1.  What needs to happen to bring #1283 to main?

billsacks:
Reading back through #1283 it looks like I had a number of concerns with how that was done, some of which were pretty fundamental – at least if we still want to support the ability to select between multiple snow cover fraction methods. So I think @swensosc was going to rethink the approach.

I wonder if we should just drop support for the old snow cover fraction method: This isn't the first time that changes / fixes have been difficult because of these two very different snow cover fraction methods. If I remember correctly, the big issue is that the different snow cover fraction methods operate very differently in terms of the data flow of how they fit in with the rest of the model (e.g., the order in which different terms are calculated that then need to feed into other calculations: what depends on what). Maybe that's not an acceptable solution, but I just wanted to pose that question....

wwieder:
how old is the 'old' snow cover fraction.  I'm OK to drop it, especially if it's a CLM4 era approach.

billsacks:
[...]

PR #1283, opened 2021-02-18
snow density fix 

swensosc:
### Description of changes
Modify updates to snow_depth and snowdp to provide better values for snow density.
### Specific notes
Simulations previously created for SnowMIP showed unrealistically large density values when the snowpack was very thin.  
This was caused by updates to snow depth that used the value of frac_sno from the beginning of the time step, rather than a value that represented the updates due to changes in snowpack due to incorporation of frozen surface water and frost.

Resolves ESCOMP/CTSM#1280

yes.  snow_depth is calculated differently in some situations, which would impact surface energy budget and albedo.

Testing performed, if any:

### DIFF
--- a/src/biogeophys/HydrologyNoDrainageMod.F90
+++ b/src/biogeophys/HydrologyNoDrainageMod.F90
@@ -208,6 +208,7 @@ contains
     real(r8) :: fracl                         ! fraction of soil layer contributing to 10cm total soil water
     real(r8) :: s_node                        ! soil wetness (-)
     real(r8) :: icefrac(bounds%begc:bounds%endc,1:nlevsoi)
+    real(r8) :: frac_sno_update               ! update to snow cover fraction for h2osfc phase change and frost deposition
     !-----------------------------------------------------------------------
 
     associate( &
@@ -239,6 +240,10 @@ contains
          snowdp             => b_waterdiagnostic_inst%snowdp_col             , & ! Input:  [real(r8) (:)   ]  area-averaged snow height (m)       
          frac_sno_eff       => b_waterdiagnostic_inst%frac_sno_eff_col       , & ! Input:  [real(r8) (:)   ]  eff.  snow cover fraction (col) [frc]    
          frac_h2osfc        => b_waterdiagnostic_inst%frac_h2osfc_col        , & ! Input:  [real(r8) (:)   ]  fraction of ground covered by surface water (0 to 1)
+
+         qflx_h2osfc_to_ice         => b_waterflux_inst%qflx_h2osfc_to_ice_col        , & ! Input: [real(r8) (:)   ]  conversion of h2osfc to ice (mm H2O /s)
+         qflx_soliddew_to_top_layer => b_waterflux_inst%qflx_soliddew_to_top_layer_col, & ! Input:  [real(r8) (:)  ]  rate of solid water deposited on top soil or snow layer (frost) (mm H2O /s) [+]
+
          snw_rds            => b_waterdiagnostic_inst%snw_rds_col            , & ! Output: [real(r8) (:,:) ]  effective snow grain radius (col,lyr) [microns, m^-6] 
          snw_rds_top        => b_waterdiagnostic_inst%snw_rds_top_col        , & ! Output: [real(r8) (:)   ]  effective snow grain size, top layer(col) [microns] 
          sno_liq_top        => b_waterdiagnostic_inst%sno_liq_top_col        , & ! Output: [real(r8) (:)   ]  liquid water fraction in top snow layer (col) [frc] 
@@ -453,8 +458,16 @@ contains
       end do
 
       ! Calculate column average snow depth
+      ! Use updated snow cover fraction to account for
+      ! surface water to ice conversion and frost deposition
+      ! that may have occurred since initial frac_sno calculation
       do c = bounds%begc,bounds%endc
-         snowdp(c) = snow_depth(c) * frac_sno_eff(c)
+         frac_sno_update = frac_sno_eff(c) &
+              + tanh(0.1 * (qflx_h2osfc_to_ice(c)+qflx_soliddew_to_top_layer(c))*dtime) &
+              * (1._r8 - frac_sno_eff(c))
+
+         snowdp(c) = snow_depth(c) * frac_sno_update
+
       end do
 
       ! Calculate snow internal temperature
--- a/src/biogeophys/SnowHydrologyMod.F90
+++ b/src/biogeophys/SnowHydrologyMod.F90
@@ -364,6 +364,7 @@ contains
          h2osoi_liq          = b_waterstate_inst%h2osoi_liq_col(begc:endc,:), &
          qflx_snow_grnd      = b_waterflux_inst%qflx_snow_grnd_col(begc:endc), &
          qflx_snow_drain     = b_waterflux_inst%qflx_snow_drain_col(begc:endc), &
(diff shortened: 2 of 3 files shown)


src/biogeophys/HydrologyNoDrainageMod.F90
```diff
@@ -453,8 +458,16 @@ subroutine HydrologyNoDrainage(bounds, &
       end do
 
       ! Calculate column average snow depth
+      ! Use updated snow cover fraction to account for
+      ! surface water to ice conversion and frost deposition
+      ! that may have occurred since initial frac_sno calculation
       do c = bounds%begc,bounds%endc
-         snowdp(c) = snow_depth(c) * frac_sno_eff(c)
+         frac_sno_update = frac_sno_eff(c) &
+              + tanh(0.1 * (qflx_h2osfc_to_ice(c)+qflx_soliddew_to_top_layer(c))*dtime) &
```
swensosc, line 466: the 0.1 should be the accum_factor from SnowCoverFractionSwensonLawrence2012Mod, but currently it is a private parameter.  What's the best way to use that parameter here?

src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -1232,7 +1238,9 @@ subroutine UpdateState_TopLayerFluxes(bounds, num_snowc, filter_snowc, &
           call endrun("In UpdateState_TopLayerFluxes, h2osoi_ice has gone significantly negative")
        end if
 
-       if (h2osoi_liq(c,lev_top(c)) < 0._r8) then
+       !       if (h2osoi_liq(c,lev_top(c)) < 0._r8) then
+       ! in some case roundoff level negative values trigger this 
+       if (h2osoi_liq(c,lev_top(c)) < -1.e-12_r8) then
```
swensosc, line 1243: the 0 was changed to -1e-12 b/c I found some cases where <0 was catching values of h2osoi_liq that were negative roundoff values of zero due to the removal of the constrained evaporation from the available moisture being roundoff level different values.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -912,14 +913,18 @@ subroutine PhaseChangeH2osfc (bounds, num_nolakec, filter_nolakec, &
 
             z_avg=frac_sno(c)*snow_depth(c)
             if (z_avg > 0._r8) then 
-               rho_avg=min(800._r8,h2osno_total(c)/z_avg)
+               rho_avg=min(denice,h2osno_total(c)/z_avg)
             else
-               rho_avg=200._r8
+               rho_avg = denice
```
swensosc, line 918: I don't know why the old values weren't denice; they should have been.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -912,14 +913,18 @@ subroutine PhaseChangeH2osfc (bounds, num_nolakec, filter_nolakec, &
...
             endif
 
             !=====================  xm < h2osfc  ====================================
             if(temp1 >= 0._r8) then ! add some frozen water to snow column
 
+               ! update snow depth
+               frac_sno_update = frac_sno(c) + tanh(0.1 * (-xm(c)))*(1._r8 - frac_sno(c))
```
swensosc, line 925: again, 0.1 should be accum_factor

billsacks (changes requested):
I have made a few comments inline. However, my biggest comment relates to the question you raised, about how to reference a constant from `SnowCoverFractionSwensonLawrence2012Mod`. As is often the case, the need to access a private module variable actually indicates a bigger issue, I think. In this case, the problem I see is that it looks like your changes might assume that you're using that snow cover fraction formulation, rather than the one in SnowCoverFractionNiuYang2007Mod or anything that might come along later.

- [ ] I think this needs to be reworked somehow so that it works consistently with any snow cover fraction method. This might require introducing one or more new subroutines in the snow cover fraction modules that do what you need.

If that's difficult, I wonder if there is a completely different way to solve the underlying problem (of inconsistency between two history variables) – such as making the H2OSNO history variable based on some new variable that is saved at the same time that `snow_depth` and `snowdp` are calculated – i.e., in the middle of the time step, rather than using the H2OSNO value from the end of the time step. I recognize, though, that that might lead to other issues, like H2OSNO now being inconsistent with some other values. Maybe a solution could be to compute the snow density diagnostic in the code at the appropriate time, ensuring that it is using self-consistent values? I'm not sure... just thinking out loud about some ideas if it's too hard to make this work generally across any snow cover fraction method.

src/biogeophys/HydrologyNoDrainageMod.F90
```diff
@@ -239,6 +240,10 @@ subroutine HydrologyNoDrainage(bounds, &
          snowdp             => b_waterdiagnostic_inst%snowdp_col             , & ! Input:  [real(r8) (:)   ]  area-averaged snow height (m)       
          frac_sno_eff       => b_waterdiagnostic_inst%frac_sno_eff_col       , & ! Input:  [real(r8) (:)   ]  eff.  snow cover fraction (col) [frc]    
          frac_h2osfc        => b_waterdiagnostic_inst%frac_h2osfc_col        , & ! Input:  [real(r8) (:)   ]  fraction of ground covered by surface water (0 to 1)
+
+         qflx_h2osfc_to_ice         => b_waterflux_inst%qflx_h2osfc_to_ice_col        , & ! Input: [real(r8) (:)   ]  conversion of h2osfc to ice (mm H2O /s)
+         qflx_soliddew_to_top_layer => b_waterflux_inst%qflx_soliddew_to_top_layer_col, & ! Input:  [real(r8) (:)  ]  rate of solid water deposited on top soil or snow layer (frost) (mm H2O /s) [+]
+
```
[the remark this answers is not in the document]
swensosc: I see two issues: 1) is the diagnostic SNOWDP and snow density correct and 2) is the actual snow depth being correctly calculated.  I think that I found a couple of issues with the latter that this pr corrects (i.e the change in snow depth due to phase change and incorporation of frozen surface water into the snowpack).  If it was just 1), then I think changing the location of the snowdp assignment could work, but I think 2) needs to include a recalculation of the scf.

src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -480,7 +482,11 @@ subroutine BulkDiag_NewSnowDiagnostics(bounds, num_c, filter_c, &
        int_snow(c) = max(int_snow(c),h2osno_total(c)) !h2osno_total could be larger due to frost
 
        ! snowmelt from previous time step * dtime
-       snowmelt(c) = qflx_snow_drain(c) * dtime
+       ! add h2osfc to ice conversion flux to snowmelt to trigger
+       ! re-calculation of scf when h2osfc freezes (snowmelt
+       ! amount is not used in sfc_method, just presence/absence)
+       snowmelt(c) = (qflx_h2osfc_to_ice(c) + qflx_snow_drain(c)) * dtime
```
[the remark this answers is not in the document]
swensosc: This was just used as a trigger to get the SCF to be updated.  Normally, it is updated when newsnow > 0 or snowmelt > 0.  If neither occurs, then SCF is not updated.  I'm not sure, but I think the latter behavior was intentional, and due to undesirable behavior caused by frost deposition or surface water freezing.  But now it seems like it would be good to update SCF for changes to due to surface water freezing.

Another way to do this would be to have a second call to scf_method%UpdateSnowDepthAndFrac where the newsnow argument is set to qflx_h2osfc_to_ice and bifall is set to denice.  I think that is conceptually what is happening.  I can test that.

swensosc:
Needs more work...

PR #1310, opened 2021-03-21
run_sys_tests: add --retry option on izumi

billsacks:
### Description of changes

Adds a `--retry` option to `run_sys_tests`, adds a mechanism to set a machine-dependent default value of this option, and sets the default to 1 for izumi. So, if you don't specify the `--retry` option to `run_sys_tests`, the `--retry` option to `create_test` will be set to 1 for izumi, 0 for cheyenne or other machines. You can override this with the `--retry` option to `run_sys_tests`. This value of this option is printed both to stdout and to the (increasingly poorly named) `SRCROOT_GIT_STATUS` file in your testroot directory.

The motivation for this change was: Recently, nearly all of my test runs on izumi have a few tests fail with system errors. It started getting annoying to need to rerun them (and wait for the reruns to complete, then check them) before making a tag. With this change, they will be retried once. We could increase the retry to 2 or 3 if 1 doesn't seem sufficient, but I wanted to start out with just 1 retry because it could start to get confusing or wasteful of machine resources if true failures are rerun multiple times, failing in the same way each time – so I figured we could see if 1 is sufficient before increasing this further.

If I understand correctly how this is implemented in create_test, it will mean that the create_test job (which, with run_sys_tests, is executed on a batch node) will remain active until all tests that it "owns" have completed. There's a chance that this will have negative repercussions in terms of job scheduling (if the scheduler sees that you already have these active jobs so gives priority to someone else), but since izumi tends to be pretty lightly loaded, I'm hopeful it will be okay.

I also made a commit to this PR that fixes some pylint warnings with recent versions of pylint.

### Specific notes

Testing performed, if any:
- python tests
- Ran full test suite on izumi via the new run_sys_tests to verify that it worked; this test suite had 3 tests that experienced system failures in their first runs but (as desired) were automatically rerun and passed the second time around
- Ran a 2-test run via run_sys_tests on cheyenne to verify that I haven't broken that

### DIFF
--- a/python/ctsm/lilac_build_ctsm.py
+++ b/python/ctsm/lilac_build_ctsm.py
@@ -473,8 +473,8 @@ def _check_and_transform_os(os_type):
                   'cnl': 'CNL'}
     try:
         os_type_transformed = transforms[os_type]
-    except KeyError:
-        raise ValueError("Unknown OS: {}".format(os_type))
+    except KeyError as exc:
+        raise ValueError("Unknown OS: {}".format(os_type)) from exc
     return os_type_transformed
 
 def _get_case_dir(build_dir):
--- a/python/ctsm/lilac_download_input_data.py
+++ b/python/ctsm/lilac_download_input_data.py
@@ -5,9 +5,9 @@ import logging
 import os
 import re
 
-from ctsm.ctsm_logging import setup_logging_pre_config, add_logging_args, process_logging_args
-
 from CIME.case import Case # pylint: disable=import-error
+
+from ctsm.ctsm_logging import setup_logging_pre_config, add_logging_args, process_logging_args
 
 logger = logging.getLogger(__name__)
 
--- a/python/ctsm/lilac_make_runtime_inputs.py
+++ b/python/ctsm/lilac_make_runtime_inputs.py
@@ -8,11 +8,11 @@ import logging
 from configparser import ConfigParser
 from configparser import NoSectionError, NoOptionError
 
+from CIME.buildnml import create_namelist_infile # pylint: disable=import-error
+
 from ctsm.ctsm_logging import setup_logging_pre_config, add_logging_args, process_logging_args
 from ctsm.path_utils import path_to_ctsm_root
 from ctsm.utils import abort
-
-from CIME.buildnml import create_namelist_infile # pylint: disable=import-error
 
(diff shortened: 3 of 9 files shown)


billsacks:
@ekluzek @negin513 @slevisconsulting I'm assigning you all as reviewers, not so much because I feel like the code changes themselves need a review (though feel free to look through them if you want) but more because I'd like your sign-off on the basic idea. Specifically: are you happy with this idea that, on izumi, we'll set a flag to automatically retry each failed test once, to help deal with system failures? I've been reluctant to use this flag in the past, because I've been concerned that it would cause more confusion than good, but (1) izumi tests are failing pretty frequently for me recently due to system issues, and (2) the test system seems pretty robust now in terms of the ability to rerun tests cleanly (rather than needing to create new tests when a test fails, which was sometimes the case until a few years ago).

Let me know what you think.

slevisconsulting (approved):
@billsacks this sounds good to me and thank you for doing it.

ekluzek (approved):
I do like this idea as I think that izumi tests should likely run without having to manually resubmit the few that typically fail. I usually have about six or so that fail each time I run on izumi. It does take some effort to check the problem and resubmit. So this should remove having to do that.

I've skimmed the code changes and overall they make sense to me. There's a few points where you are making some small changes that aren't strictly necessary here, but seem fine to do. Changing get_possibly_overridden_baseline_dir to get_possibly_overridden_mach_value helps with clarity in using it for both cases and also should be a more general name that can be used in the future as well. That change is probably responsible for the majority of the lines of code that had to change here.

billsacks:
Thanks @ekluzek and @slevisconsulting . @ekluzek - yes, this change ended up involving more code changes than I originally imagined. The way I have set up the machine information, machine defaults and related unit tests ends up requiring a fair number of changes to add a new machine-specific default that can be overridden from the command-line like this. This probably means that there's a way this could be done better, but off-hand I haven't seen one yet myself, particularly since different machine-specific options have different logic associated with them, so I'm not sure it would work easily to (for example) store them all in some generic structure.

@ekluzek are you okay with my merging it now? You have the next few tags, so I'll let you make the call as to whether you're comfortable with this being merged now.

ekluzek:
Yes you can merge it. Or I'll merge it into the next tag. I'd like to have it in place for upcoming izumi testing.

PR #1321, opened 2021-04-05
Medlyn-SMS

djk2120:
### Description of changes
The Medlyn stomatal conductance functionality was deployed for CLM5. It had not been implemented for use_hydrstress=.false (i.e. SMS). 
Previously if we opted for:
use_hydrstress=.false  and  stomatalcond_method = 'Medlyn2011', the model would actually use the Ball-Berry stomatal conductance model.

In the implementation process, I also noticed that the history fields GSSUN and GSSHA were not being populated when use_hydrstress=.false, which I also updated with this PR.

### Specific notes
No CTSM 'issue', but this resolves a known bug.

Yes. Many answers will change when use_hydrstress=.false., stomatalcond_method = 'Medlyn2011' including FPSN, QVEGT, etc.
Answers should match for use_hydrstress=.true.
Answers should mostly match for use_hydrstress=.false., stomatalcond_method = 'Ball-Berry1987'
apart from GSSUN and GSSHA, which were previously not being populated, resulting in NaNs.
 
No.

Testing performed, if any:
Ran a 5-day I2000 SP global simulation to confirm that FPSN differs between:
SMS-Medlyn and SMS-BallBerry, which were previously identical.

### DIFF
--- a/src/biogeophys/PhotosynthesisMod.F90
+++ b/src/biogeophys/PhotosynthesisMod.F90
@@ -812,6 +812,7 @@ contains
        if (      trim(stomatalcond_method) == 'Ball-Berry1987' ) then
           this%stomatalcond_mtd = stomatalcond_mtd_bb1987
        else if ( trim(stomatalcond_method) == 'Medlyn2011'     ) then
+          !zqz
           this%stomatalcond_mtd = stomatalcond_mtd_medlyn2011
        else
           call endrun(msg="ERROR bad value for stomtalcond_method in "//nmlname//"namelist"//errmsg(sourcefile, __LINE__))
--- a/src/biogeophys/PhotosynthesisMod.F90
+++ b/src/biogeophys/PhotosynthesisMod.F90
@@ -1229,6 +1229,8 @@ contains
          gs_mol     => photosyns_inst%gs_mol_patch           , & ! Output: [real(r8) (:,:) ]  leaf stomatal conductance (umol H2O/m**2/s)
          gs_mol_sun_ln => photosyns_inst%gs_mol_sun_ln_patch , & ! Output: [real(r8) (:,:) ]  sunlit leaf stomatal conductance averaged over 1 hour before to 1 hour after local noon (umol H2O/m**2/s)
          gs_mol_sha_ln => photosyns_inst%gs_mol_sha_ln_patch , & ! Output: [real(r8) (:,:) ]  shaded leaf stomatal conductance averaged over 1 hour before to 1 hour after local noon (umol H2O/m**2/s)
+         gs_mol_sun => photosyns_inst%gs_mol_sun_patch       , & ! Output: [real(r8) (:,:) ]  patch sunlit leaf stomatal conductance (umol H2O/m**2/s)
+         gs_mol_sha => photosyns_inst%gs_mol_sha_patch       , & ! Output: [real(r8) (:,:) ]  patch shaded leaf stomatal conductance (umol H2O/m**2/s)
          vcmax_z    => photosyns_inst%vcmax_z_patch          , & ! Output: [real(r8) (:,:) ]  maximum rate of carboxylation (umol co2/m**2/s)
          cp         => photosyns_inst%cp_patch               , & ! Output: [real(r8) (:)   ]  CO2 compensation point (Pa)
          kc         => photosyns_inst%kc_patch               , & ! Output: [real(r8) (:)   ]  Michaelis-Menten constant for CO2 (Pa)
@@ -1243,6 +1245,7 @@ contains
          lnc        => photosyns_inst%lnca_patch             , & ! Output: [real(r8) (:)   ]  top leaf layer leaf N concentration (gN leaf/m^2)
          light_inhibit=> photosyns_inst%light_inhibit        , & ! Input:  [logical        ]  flag if light should inhibit respiration
          leafresp_method=> photosyns_inst%leafresp_method    , & ! Input:  [integer        ]  method type to use for leaf-maint.-respiration at 25C canopy top
+         medlynintercept  => pftcon%medlynintercept          , & ! Input:  [real(r8) (:)   ]  Intercept for Medlyn stomatal conductance model method
          stomatalcond_mtd=> photosyns_inst%stomatalcond_mtd  , & ! Input:  [integer        ]  method type to use for stomatal conductance.GC.fnlprmsn15_r22845
          leaf_mr_vcm => canopystate_inst%leaf_mr_vcm           & ! Input:  [real(r8)       ]  scalar constant of leaf respiration with Vcmax
          )
@@ -1700,7 +1703,20 @@ contains
 
                ! End of ci iteration.  Check for an < 0, in which case gs_mol = bbb
 
-               if (an(p,iv) < 0._r8) gs_mol(p,iv) = bbb(p)
+               if (an(p,iv) < 0._r8) then
+                  if (stomatalcond_mtd == stomatalcond_mtd_bb1987) then
+                     gs_mol(p,iv) = bbb(p)
+                  else if ( stomatalcond_mtd == stomatalcond_mtd_medlyn2011 )then
+                     gs_mol(p,iv) = medlynintercept(patch%itype(p))
+                  end if
+               end if
+
+               ! Write stomatal conductance to the appropriate phase
+               if (phase=='sun') then
+                  gs_mol_sun(p,iv) = gs_mol(p,iv)
+               else if (phase=='sha') then
+                  gs_mol_sha(p,iv) = gs_mol(p,iv)
+               end if
 
                ! Use time period 1 hour before and 1 hour after local noon inclusive (11AM-1PM)
                if ( is_near_local_noon( grc%londeg(g), deltasec=3600 ) )then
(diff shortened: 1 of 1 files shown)


billsacks:
From CTSM-software meeting: either along with this or in a follow-up tag, we want to change this:

https://github.com/ESCOMP/CTSM/blob/cb6057edbe495a8e58d59c86880598ca413a4647/bld/namelist_files/namelist_defaults_ctsm.xml#L99-L104

so that we always use Medlyn by default with clm5_0 and clm5_1.

Probably best to do that in a follow-up tag to separate answer changes.

ekluzek:
The only tests where PHS is off that we do are with the clm50dynroots test-mod. So these tests...

ERS_D_Ld3.f19_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-clm50dynroots
SMS_Ly3_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.cheyenne_intel.clm-clm50dynroots

I'm also running this test to show that's it's good for a 5-year stint...

SMS_Ly5_P144x1.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-clm50dynroots

heroldn (commented):
Hi Daniel, thanks for doing this and apologies if I'm not using Github features properly.

I noticed on line 1335 that minimum gs is set as **bbb(p) = max (bbbopt(p)*btran(p), 1._r8)**. This looks like a ball-berry specific line and bbb is used further down on line 1655: **rs_z(p,iv) = min(rsmax0, 1._r8/bbb(p) * cf)**.

Given the Photosynthesis routine now accommodates BB and Medlyn should line 1335 be changed to something like the following, so that any calculations with bbb are using the correct minimum gs? There's probably a more elegant way to do this so bbb doesn't become a medlyn quantity, but hopefully you see what I'm getting at.
**if ( stomatalcond_mtd == stomatalcond_mtd_bb1987 )then
   bbb(p) = max (bbbopt(p)*btran(p), 1._r8)
else if ( stomatalcond_mtd == stomatalcond_mtd_medlyn2011 )then
   bbb(p) = medlynintercept(patch%itype(p))
end if**

djk2120:
Hi Nicholas,

Thanks, nice catch. You're right. I hadn't thought about the nighttime
conductances. I'll use that logic, but around L1655 (
https://github.com/djk2120/CTSM/blob/medlyn_sms/src/biogeophys/PhotosynthesisMod.F90#L1655),
and maintain bbb as a ball-berry variable. I think we could certainly clean
this up a bunch, but I'm going to save that for a future pull request where
we try to unify the PHS/SMS code

ekluzek:
@djk2120 question on this set of code. It's technically outside of this specific PR, but it's in this section of code. And comes up in my examining the code to make sure these changes make sense compared to the same changes to enable Medlyn in PHS.

For PHS you have this for modification to bbb(p) that's needed for Ball-Berry "max(bsha*bbb(p),1._r8))". As such I'd expect something similar for these two lines in PHS (in ci_func_PHS)...

``` fortran
          bquad = -(2.0 * (medlynintercept(patch%itype(p))*1.e-06_r8 + term) + (medlynslope(patch%itype(p)) * term)**2 / &
               (gb_mol*1.e-06_r8 * rh_can))
          cquad = medlynintercept(patch%itype(p))*medlynintercept(patch%itype(p))*1.e-12_r8 + &
               (2.0*medlynintercept(patch%itype(p))*1.e-06_r8 + term * (1.0 - medlynslope(patch%itype(p))* &
               medlynslope(patch%itype(p)) / rh_can)) * term
```

Shouldn't there be some modification to the medlyn parameters by bsha (and also bsun)? I totally could be wrong here, but wanted to ask since it didn't' make sense to me that Medlyn wouldn't have a similar modification as Ball-Berry.

Thanks for taking a look and letting us know.

djk2120:
Hi Erik,
This is an open science question, whether the stomatal conductance
intercept should be modified for water stress. I think generally the field
is tending towards no attenuation, which I'm guessing is why when Medlyn
was introduced (Rosie?) chose to have a constant intercept.

So even though BB and Medlyn have different treatments of the intercept, I
do think that the Medlyn implementation was an active choice to leave out
the btran effects on the intercept. I can say that I, and at least a few
others, had noticed this in the model, and consider it a positive change.

-Daniel

ekluzek:
@djk2120 OK good to know. I added a comment to that effect in the code. See commit 72b7dc0. Let me know if you have any changes to suggest for it.

djk2120:
Looks good, but can you add intercept to the first line of the comment, for
clarity.
(We do modify vcmax for water stress within the Medlyn code)

! Note for Medlyn we are NOT modifying the intercept by water stress as we
do for

ekluzek:
OK, just pushed that clarification. Thanks for looking it over.

olyson:
A couple of items that should probably be addressed, if not as part of this PR, then as a separate PR.

1. The Medlyn error check discussed in issue #1398 

2. gs_mol_sun (GSSUN in the history output) and gs_mol_sha (GSSHA) don't seem be calculated/prescribed at night, so their values at night remain at the last value calculated during daytime.  Seems like they should be set to the Medlyn intercept value (minimum stomatal conductance) at nighttime?

heroldn:
Regarding Keith's second point, putting the following lines in the IF condition in the PhotosynthesisMod module at [line 3250](https://github.com/ESCOMP/CTSM/blob/114c4950a1191b1b8b88a151df5c665a8aedefa0/src/biogeophys/PhotosynthesisMod.F90#L3250) seems to do the trick. Output at the time step frequency shows GSSUN and GSSHA now being minimum values at night. Importantly, I see no change in TV, QVEGE or QVEGT, indicating this bug thankfully doesn't have a climate impact (WHY that is is beyond me, obviously the gs_mol* variables don't get used to determine any of these variables, interested in anyone's thoughts on this).
```
               gs_mol_sun(p,iv) = gsminsun
               gs_mol_sha(p,iv) = gsminsha
```

Trying to confirm the above for the Photosynthesis module (for when PHS=off) revealed another bug, where GSSUN and GSSHA aren't recorded. That module is missing code such as below. Currently GSSUN and GSSHA are static over time.
```
               ! Write stomatal conductance to the appropriate phase
               if (phase=='sun') then
                  gs_mol_sun(p,iv) = gs_mol(p,iv)
               else if (phase=='sha') then
                  gs_mol_sha(p,iv) = gs_mol(p,iv)
               end if
```

olyson:
Regarding the second comment above (static GSSUN/GSSHA), I think Daniel has already addressed this based on his comment above: "In the implementation process, I also noticed that the history fields GSSUN and GSSHA were not being populated when use_hydrstress=.false, which I also updated with this PR."

olyson:
Regarding the first comment, the stomatal resistance is actually what is used in CanopyFluxesMod to calculate, e.g., transpiration.  in the Photosynthesis code. Changing gs_mol_sun and gs_mol_sha at night, even using the current suggestion of setting them to the minimum stomatal conductance, doesn't affect stomatal resistance because it is still set by (in release-cesm2.1.3):

rs_z_sun(p,iv) = min(rsmax0, 1._r8/(max( bsun(p)*gsminsun, 1._r8 )) * cf)
rs_z_sha(p,iv) = min(rsmax0, 1._r8/(max( bsha(p)*gsminsha, 1._r8 )) * cf)

So it might be better to set gs_mol_sun and gs_mol_sha to the inverse of the stomatal resistance at night, after making the appropriate unit conversions, to make them consistent.  Just a suggestion.

ekluzek:
> A couple of items that should probably be addressed, if not as part of this PR, then as a separate PR.
> 
> 1. The Medlyn error check discussed in issue [Medlyn stomatal conductance error check fails #1398](https://github.com/ESCOMP/CTSM/issues/1398)
> 2. gs_mol_sun (GSSUN in the history output) and gs_mol_sha (GSSHA) don't seem be calculated/prescribed at night, so their values at night remain at the last value calculated during daytime.  Seems like they should be set to the Medlyn intercept value (minimum stomatal conductance) at nighttime?

@djk2120 can you take care of these two things? I could do these if needed if "1" has clear code, but it would be best if you can do it. You are also the one that needs to Vet these changes, so it would be best to have you do it. And you are also the one that should evaluate if the error check should be done. Let us know if you can work on this. Thanks.

djk2120:
Sure- that all sounds good.
I'll be available once I'm done with the TRENDY simulations, i.e. Sept-1

heroldn:
@olyson ah you're right Daniel fixed this (and I might have got those lines from his code). Thanks for the info re conductance. Good to know we can still use the climate from our runs.

ekluzek:
I've included this early version as part of #1164 and will bring it in as is with it. The option isn't actually invoked there anyway, so it just allows the new functionality. But, I'll leave this PR open so that the two new additions can be added on (nighttime change and the error check).

ekluzek:
Note, I have a compare to baseline case that is showing differences (it's marking everything as missing value, so it's calling the fillvalue mask as different). This is without the option invoked (which is the default) and I expect all cases where the new option isn't turned on to give identical results. So this needs to be resolved.

SMS_Ld1.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasRs.cheyenne_gnu.clm-default

ekluzek:
The first set of changes for this will be coming in with ctsm5.1.dev054. But, I'm going to leave this open so that more work can go into this as has been planned.

ekluzek:
It looks like the current set of changes got merged to master with ctsm5.1.dev054, but it didn't close the PR (which is good). In order to close this further changes need to be done, and then it should allow us to merge those changes to main-dev and close the PR.

olyson:
A suggestion I'm passing on from a CESM Forum question is to add endrun calls to these types of Ball-Berry error checks:

if (abs(gs_mol(p,iv)-gs_mol_err) > 1.e-01_r8) then
   write (iulog,*) 'Ball-Berry error check - stomatal conductance error:'
   write (iulog,*) gs_mol(p,iv), gs_mol_err
end if

and any Medlyn error checks that are implemented.

Issue #510, opened 2018-09-13
total_plant_stored_h2o_col should be set upon restart


billsacks:
### Brief summary of bug

`total_plant_stored_h2o_col` is referenced early in the driver loop, when computing total water content for the sake of dynamic landunits adjustments as well as water balance checks. However, as far as I can tell, it isn't set until CanopyFluxes (call clm_fates%wrap_hydraulics_drive). I believe this means that things will be done wrong in the first time step after restart – and possibly the first time step after cold start – in a run with FATES.

### General bug information

**CTSM version you are using:** ctsm1.0.dev011

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** Runs with FATES

billsacks:
@rgknox could you take a look at this?

billsacks:
(Not urgent, but should be looked at at some point.)

rgknox:
Sure thing @billsacks

rgknox:
tagging @xuchonggang and @bchristo

jbeisman:
Brad and I will check on this early next week.

Yours
Chonggang

xuchongang:
Based on the discussion with Brad, this bug echos well with the current behaviors of mass balance error for the first time step. Hope to resolve it this week.

billsacks:
Just making sure this is hasn't slipped through the cracks.

xuchongang:
@billsacks and @rgknox , for the ELM, it is set as zero for cold start, which should be ok. However, for the restart, it need to be fed into the restart file. We have not yet finished the restart component for hydraulics, but it should be a task to be done.

Issue #1314, opened 2021-03-25
FATES hydro mass-balance issues


rgknox:
### Brief summary of bug

FATES-hydro is a special mode where vegetation can uptake and store water via the transpiration stream. This requires an accounting of how much water is stored in FATES vegetation, which must be passed back to CLM for its water mass balance checking.  The water mass balance check (BalanceCheckMod.F90 at line 760) is failing. Details below.

### General bug information

**CTSM version you are using:** [output of `git describe`]

hash: 3ac4305    
tag: ctsm5.1.dev029-3-g3ac43057

**Does this bug cause significantly incorrect results in the model's science?** [Yes / No]

Not sure, but I don't think so. 

**Configurations affected:** [Fill this in if known.]

These namelist options must be set:

use_fates = .true.
use_fates_planthydro = .true.

### Details of bug

I generate this error using a specialized site setup for Barro Colorado Island Panama, where I also initialize a forest stand structure from inventory data.  The simulation runs 26+ months and then fails with the following error:

```
 clm: calling FATES model        38449
FATES Dynamics: 1902-03-13
 dbh counter:           13  is woody:  T
 dbh counter:           13  is woody:  T
 dbh counter:           17  is woody:  T
 dbh counter:           17  is woody:  T
 clm: leaving fates model           1           1
 WARNING:  grid cell-level water balance error  nstep=        38449  local indexg=            1  errh2o_grc=   -2.2061354929889532E-003
 CTSM is stopping because errh2o >    1.0000000000000001E-005  mm
 nstep                     =        38449
 errh2o_grc                =   -2.2061354929889532E-003
 forc_rain                 =    0.0000000000000000     
 forc_snow                 =    0.0000000000000000     
 endwb_grc                 =    2989.3396629148806     
 begwb_grc                 =    2989.3595745851790     
 qflx_evap_tot             =    1.9854765757965773E-002
 qflx_sfc_irrig            =    0.0000000000000000     
 qflx_surf                 =    0.0000000000000000     
 qflx_qrgwl                =   -2.1492309525845731E-003
 qflx_drain                =    0.0000000000000000     
 qflx_ice_runoff           =    0.0000000000000000     
 qflx_snwcp_discarded_ice  =    0.0000000000000000     
 qflx_snwcp_discarded_liq  =    0.0000000000000000     
 deltawb                   =   -1.9911670298370154E-002
 deltawb/dtime             =   -1.1062039054650086E-005
[...]

PR #1334, opened 2021-04-13
Bypass grid-level water mass check when fates hydro is active

rgknox:
### Description of changes

With changes forthcoming in FATES PR: https://github.com/NGEET/fates/pull/727 

Note: **This need not be synchronized with FATES 727**.

Column-scale water balance checks with FATES-HYDRO are passing with consistency, including scenarios where plants are allowed to by dynamic (grow, recruit, die and non-mortal turnover).  However, we are still failing a grid-level mass balance check.

Fixes: #510 

See issue: #1314 

This PR is a temporary fix, that bypasses grid-scale water mass checks only for fates-hydro runs.  Since column level checks pass, the expectation is that there is a bug on the upscaling from column to grid that is presently unknown. However, this PR will allow hydro tests to pass which will prevent further regressions.

Note that column level plant stored water was also added to the restart file (CLM side) to enable tests to pass.

### Specific notes

CTSM Issues Fixed (include github issue #):  Does not fix, but addresses #1314 partially.

Are answers expected to change (and if so in what way)?  Answers should go from no-answers, to has-answers for fates-hydro. Should not impact other answers.

Any User Interface Changes (namelist or namelist defaults changes)?  No

Testing performed, if any:

TBD

### DIFF
--- a/src/biogeophys/WaterDiagnosticType.F90
+++ b/src/biogeophys/WaterDiagnosticType.F90
@@ -296,7 +296,7 @@ contains
     ! !USES:
     use clm_varcon       , only : nameg, namec
     use ncdio_pio        , only : file_desc_t, ncd_double
-    use clm_varctl         , only : use_fates_planthydro
+    use clm_varctl       , only : use_fates_planthydro
     use restUtilMod
     !
     ! !ARGUMENTS:
--- a/src/biogeophys/WaterDiagnosticType.F90
+++ b/src/biogeophys/WaterDiagnosticType.F90
@@ -296,6 +296,7 @@ contains
     ! !USES:
     use clm_varcon       , only : nameg, namec
     use ncdio_pio        , only : file_desc_t, ncd_double
+    use clm_varctl         , only : use_fates_planthydro
     use restUtilMod
     !
     ! !ARGUMENTS:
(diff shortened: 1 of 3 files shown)


billsacks (changes requested):
This is fine. Can you just please add a comment like this above the line of code that you changed:

```Fortran
  ! BUG(wjs, 2021-04-13, ESCOMP/CTSM#1314) Temporarily bypassing gridcell-level check with use_fates_planthydro until issue 1314 is resolved
```

(So we have some explanation that we can refer back to in case this temporary code stays in place for a while.)

ekluzek (approved):
I agree with @billsacks just make sure you have some comments in the code that document about this. I'd even suggest adding a note to the Fates Hydro test that notes that grid-level mass check is turned off. It's really easy to forget these temporary work arounds are put in place, and they often last much longer than you'd like.

rgknox:
will do, great suggestions!

billsacks (approved):
Great, thanks!

rgknox:
fyi, I'm testing out some changes in this PR to hopefully address:
https://github.com/ESCOMP/CTSM/issues/510

src/biogeophys/WaterDiagnosticType.F90
```diff
@@ -329,6 +329,15 @@ subroutine Restart(this, bounds, ncid, flag)
          units='kg/kg', &
          interpinic_flag='interp', readvar=readvar, data=this%qaf_lun)
 
+
+    call restartvar(ncid=ncid, flag=flag, &
```
billsacks, line 333: - [ ] Please put this in a fates_hydro conditional

rgknox:
@billsacks , updated with conditional on the restart

billsacks:
Thanks @rgknox . I'll bring this into my upcoming tag.

Just FYI for the future: we prefer if the person who creates a check box be the one to check it off (see https://github.com/ESCOMP/CTSM/wiki/Pull-request-review-workflow). Not at all a big deal here where I just had one, but for big reviews, it helps us control-freak reviewers be able to create check lists of things we want done and then be able to verify that what we've asked was actually done.

rgknox:
I agree to not push your buttons @billsacks

PR #1374, opened 2021-05-17
Local solar time option for average flag

RonnyMeier:
### Description of changes

This allows to have instantaneous output at a specific local solar time (e.g., always at noon). Besides the options 'A','I','X','M', and 'SUM', it is now possible to choose 'LXXXXX'  in hist_avgflag_pertape, where XXXXX is a five digit number of the desired time into the day in seconds (e.g., L03600 --> output a 1 am, L43200 --> output at noon). Note that the implementation interpolates linearly in time. 

### Specific notes

The current implementation can be problematic, if the desired time is closer than dtime/2 to midnight. In this case, the interpolation in time will interpolate between shortly after midnight of the day of interest and shortly before midnight of the same day (i.e., 24h - dtime later). This is problematic for daily output. The error becomes small for monthly output or larger.

In hist_avgflag_pertape it is possible to choose 'LXXXXX'.

Testing performed, if any:

Tested output with hist_dov2xy = .true. and .false. for two- and three-dimensional variables. Results appear to be correct/reasonable. @olyson confirmed that results look good.


ekluzek:
@RonnyMeier I think this is a nice increase in functionality that sounds useful and flexible to me. I think we already have some history terms that that are hardcoded to appear at local noon, that could be changed to use this mechanism. This is general flexible code, so it can be easily used for anything a researcher wants to see. The previous mechanism was just hardwired for a few variables and required it to be hardwired for anything else desired.

I took a quick review of the code and nothing jumps out at me. From your own caveat the one thing that I suggest is that we  add at least a warning for near midnight, if not having the model die for that case. Most likely there isn't enough of a need for exact midnight (and 11pm or 1am would suffice). The ambiguity of which time period it would pick for midnight sounds cautionary enough that maybe it shouldn't be allowed.

Also I'll have to look closer, but your initial description says it's instantaneous which implies to me it would output the last such instance for whatever time interval your outputting in. But, your comment about monthly output or larger sounds like it's doing an average over the time's within the averaging period which is different. If it is doing an average I'm not sure the problem with midnight is as big of a deal. If it is doing it as instantaneous that just means it's picking the last available time for the period. So it's a bit odd, but you should be able to figure it out...

RonnyMeier:
@ekluzek Thank you for having a look. Indeed the local noon variables could use this functionality instead of the hardcoding. Note that the output would change slightly since the current implementation is choosing the time step closest to local noon, while mine interpolates linearly in time.

I agree that users should be made aware of the caveat. Since the problem is marginal for e.g. monthly output, I would settle for a warning. The implementation is averaging over all outputs at a local time in the given period. In that sense, instantaneous is a bit misleading.  Might be better to just call it "at specified local time" similar to the local noon variables. 

How is the general procedure? Should I make another pull request with the added warning or will you clean up the code before merging?

ekluzek:
Hi @RonnyMeier we talked about this one a bit this morning. We don't need you to submit a new PR, all you need to do is push changes to your branch and they'll show up here.

So yes the one change we'd like you to make is to add a warning about the issue near midnight. Just commit and push that to this branch.

ekluzek:
@RonnyMeier the other thing we wanted to hear from you about is the choice you made to do the time interpretation to get an exact time estimate rather than just using the nearest model time. Do you think that's important? And important enough for the cost of complexity in the code? Did that make a difference in your analysis for example? Let us know what you think. I'm wondering if it's worth the extra complexity. Thanks....

ekluzek:
Final thing to mention as a seperate step after this is in we'll convert current local noon variables to this mechanism. That will include an apparent change in answers. So we can bring this in as a simple bit for bit and then do that change as an answer changing thing.

dlawrenncar:
Interesting.  It probably doesn't matter too much in terms of the
analysis.  Seems like benefits and limitations with either approach.
Personally happy to go ahead and utilize what you've already implemented.

ekluzek:
We'll. bring it in as it is in that case. I mostly wanted to hear your reasoning behind it.

@RonnyMeier could you show the plot of the case with the interpolation so it's easier to assess the difference? I do see what you mean by the lines, showing both would just clarify it.

RonnyMeier:
Below you can see the same example first with linear interpolation and second with the closest point in time (note that the simulations are different).

I have put now the weight calculation for the linear interpolation into a function to reduce redundancy in the code. I did a test with this updated implementation in CESM2.1.2, which yielded the same result as previously. Both the warning and the version with the new function are now in my branch "localtime".

**Linear interpolation in time**
TSKIN_13_30_Feb_linear

**Nearest neighbor in time**
TSKIN_13_30_Feb_closest

billsacks:
Thank you @RonnyMeier for your explanations – and for all of your work on this. I had also been wondering about the value of interpolating rather than simply picking the closest time, but your rationale makes sense. Thanks for explaining.

ekluzek (approved):

ekluzek:
@RonnyMeier I'm bringing this in under #1411, note there was a small fix I needed to put in place to get it to work...

168fea1e9b3835f5fa89c480cbec09aa6fd3869c

Just FYI.

ekluzek:
@RonnyMeier @dlawrenncar @wwieder @olyson the syntax for this right now is number of seconds in the day, so goes from L00000 to L86400 taking up five digits. I'm wondering if using four digits for hour and minute would be better, so L1230 would be 12:30 pm for example. One downside is that it would have to be military time so go from L0000 to L2400. Timesteps are typically many minutes rather than seconds as well, so you don't really have accuracy to within a second (although the time interpolation mitigates that).

RonnyMeier:
@ekluzek My apologies for the bug. We just had a system update of our supercomputer before my push and I could not test the newly implemented parts.

About the time information I agree that "military style" would be more user friendly. I chose seconds into day, because it was easiest to implement and I was not expecting that this would go into the official code eventually. One thing to consider is that the ATM implementation of the local time uses seconds into day as well in the namelist input. So it might be good to have some consistency there. But then again their implementation differs on other aspects anyways (e.g. averaging over a chosen time window. time information as a separate namelist input). 

Anyways, feel free to adapt to what you consider is best. I can also have a closer look next week when I'm back from my vacations.

wwieder:
FWIW I like the military time suggestion, it's more intuitive to read for users.

billsacks:
@ekluzek in an effort to keep our test list from growing too large, what would you think about combining this new test with the existing `extra_outputs` testmod so that it gets triggered in that existing test? (I haven't looked at whether there would be issues with doing that.) Not a huge deal... though it's probably something I'll at least want to do when I (sometime, eventually) get around to overhauling the test suite.

ekluzek:
Yes, that should be fine I'll do that. I didn't initially see any appropriate test mod to include it with at first. But, that one makes sense.

I'm also going to leave it in seconds per day for now. I looked into the change and it would require changing several blocks of repeated code (which should likely be condensed to small functions anyway). But, that can happen later.

ekluzek:
We are going to leave it as seconds in the day as that's more consistent with other things in CESM (tod) and is consistent with how this is done in CAM. We also think that if the averaging window is smaller than a day, that some time samples will just be missing, which is fine. If we find later that it blows up in that case, we'll make sure they are missing.

PR #1435, opened 2021-07-23
Point to CTSM-specific forums

billsacks:
### Description of changes

@mnlevy1981 pointed out that many of the CESM-General forum posts relate to CTSM. So he wondered if the problem may be that we are pointing to the overall CESM forums from various points in the CTSM repository, and some users may think these forums are all for CTSM. So, at his suggestion, I am adding links directly to the CTSM forums. This has the downside that the URL is more likely to change in the future and we'll have to update these links, but I'm thinking it could be worth that risk.

In the README and github.io documentation, I have kept links to the overall CESM forums in addition to the CTSM forums, because the former could be more appropriate for some questions (e.g., porting) – though if someone doesn't read carefully, they may still end up in the overall forums when they should really go to the CTSM forums, so I'm open to getting rid of the overall forum link if people think that would be better.

In the issue chooser, I made the link just go to the CTSM forums; I didn't feel there was a clean way to provide links to both the CTSM forums and overall CESM forums.

### Specific notes

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,5 @@
 blank_issues_enabled: true
 contact_links:
   - name: CESM forums
-    url: https://xenforo.cgd.ucar.edu/cesm/
+    url: https://bb.cgd.ucar.edu/cesm/forums/ctsm-clm-mosart-rtm.134/
     about: For support with model use, troubleshooting, etc., please use the CTSM forum
--- a/README.md
+++ b/README.md
@@ -25,10 +25,9 @@ and
 
 https://github.com/ESCOMP/ctsm/wiki/Recommended-git-setup
 
-For support with model use, troubleshooting, etc., please use the CTSM forum (or other
-appropriate forum) here:
-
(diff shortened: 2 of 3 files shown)


mnlevy1981:
@olyson do you think this would be useful? You responded to all the posts I moved to the CTSM forum from CESM General, so I assume you have a feel for whether the last four weeks were typical or an abnormal amount of misplaced posts. I made this suggestion to @billsacks under the assumption that I was seeing the usual number of people posting CTSM questions in CESM General, but if that's not the case then this PR may be unnecessary.

olyson:
@mnlevy1981, I think this would be useful.  
The downside may be that the CTSM forum may get more porting-related questions than before, so I wonder if we should additionally mention specifically in the documentation that the CESM infrastructure forum is an appropriate place for a porting question, regardless of whether the error/problem appears to be related to CTSM.   I don't really have the software expertise to answer those types of questions.  Happy to hear people's thoughts on that.

olyson:
Ok, looks like Bill has specifically mentioned infrastructure/porting in the documentation, so I think this is fine.

olyson (approved):
Looks good.

PR #1447, opened 2021-07-29
Improve and document method for getting diagnostics on a problem point

### LINKED ISSUE #164: BalanceCheck: Calls get_proc_bounds from threaded region
**Gautam Bisht < gbisht > - 2017-04-05 14:01:12 -0600**
**Bugzilla Id:** [2439](http://bugz.cgd.ucar.edu/show_bug.cgi?id=2439)
**Bugzilla Depends:**  2027,  2310,  2384, 
**Bugzilla CC:**  rfisher, 

The bug was discovered by Azamat Mametjanov (azamat@mcs.anl.gov) with ACME, but it is also applicable to CLM. BalanceCheck() is called by each thread. When a balance check is violated, the thread attempts to determine the global index and if CLM had more than 1 thread, the code will crash. Below is the call sequence:

BalanceCheck()
 - GetGlobalIndex()
   - call get_proc_bounds()

### LINKED ISSUE #1424: Change all the endrun calls to write out the point context information
Most of the calls to endrun, only print a message. There's another interface where you give information on point context. We don't use that version nearly as often as the version that just gives the message. It would be good to use the version that writes the information on the subgrid context: gridcell, land-unit, column, and patch point number. This is more important because of #1420 where each processor has those start at index 1, so it becomes important to identify the processor ID. Using this form would give this extra information out.

On cheyenne at least this is output by default in output, but this feature may not be available on all machines.

billsacks:
### Description of changes

Following from the changes @mvertens made to change the lower bounds to 1 on all processors: The one downside of that change was that it made it a bit harder to write diagnostics on a given point. This PR cleans up the method for writing these diagnostics and adds relevant documentation.

### Specific notes

Contributors other than yourself, if any: others helped with the design / naming for the rework

- Resolves #164 
- Resolves #1424 

Testing performed, if any: so far, aux_clm on izumi


billsacks:
I don't think it's worth anyone's time to review this carefully, but you may want to skim through it to get a sense of what the new endrun calls look like.

Also, I would welcome feedback on the documentation from @olyson @ekluzek or others. See the changes to trouble-shooting.rst. I have also pushed the built documentation to https://escomp.github.io/ctsm-docs/versions/master/html/users_guide/trouble-shooting/trouble-shooting.html#general-advice-on-debugging-run-time-problems

olyson:
I read through the updated documentation, very detailed and clear. I'll try it out on my next troubleshooting exercise and let you know if I have any suggestions/comments.  Thanks!

ekluzek:
@billsacks there's a couple of things I'm wondering about since you have gone through and changed all the endrun statements, about going ahead with a couple further things that hopefully wouldn't add much time, but make things more consistent.

One is if the endrun statement should have the "ERROR:: " prefix already embedded into it, so it doesn't have to be added to each call. Another is if errMsg(sourcefile, __LINE__) shouldn't be added to each call, because it's makes it so much easier to find the call. A lot of them have it, but not all, and it would help to have that consistency. Lastly, I wonder about removing any uses of "subname" in endrun calls that have the errMsg(sourcefile, __LINE__) line as it isn't really needed if the first is there, and the subname has to be maintained if the subroutine name changes, so there's less maintenance for using the errMsg(sourcefile, __LINE__) form.

billsacks:
Thanks for your thoughts @ekluzek, but I feel these should be deferred to a separate tag. I did see these ideas before doing final testing, but since I had already made all of the changes to endrun calls, and had done preliminary testing, it wasn't going to be any more efficient to do these now than to do them in a separate tag – and I felt I had already spent too much time on this cleanup tag. And actually, the reason I didn't spend more time discussing this when you brought up these ideas vaguely last week was because, even if we had come up with these ideas before I made the bulk of my endrun changes, I don't think it would have been much more efficient for me to do these all at once than to do different ones separately – since each one involves different changes and touches different sets of endrun calls. (So if I were personally doing these all in one tag, I probably would have done multiple passes anyway, each time making one type of change.)

That said, I do feel like these are good ideas long-term. See specific replies inline:

> One is if the endrun statement should have the "ERROR:: " prefix already embedded into it, so it doesn't have to be added to each call.

I believe this is already added by `shr_abort_abort`. So we currently have multiple occurrences of "ERROR" in some calls, e.g.,:

```
ERROR: ERROR in /Users/sacks/ctsm/ctsm2/src/biogeochem/CNBalanceCheckMod.F90 at line 311
```

So this is a matter of finding all of the places where we add "ERROR" to the endrun msg and deleting that.

> Another is if errMsg(sourcefile, **LINE**) shouldn't be added to each call, because it's makes it so much easier to find the call. A lot of them have it, but not all, and it would help to have that consistency.

That probably makes sense.

> Lastly, I wonder about removing any uses of "subname" in endrun calls that have the errMsg(sourcefile, **LINE**) line as it isn't really needed if the first is there, and the subname has to be maintained if the subroutine name changes, so there's less maintenance for using the errMsg(sourcefile, **LINE**) form.

Yeah, that probably makes sense, too. The subname can be a bit helpful, but maybe not enough so to warrant keeping it if we're consistent about printing the file & line numbers.

ekluzek:
Putting it in a separate tag is fine, whatever would be most efficient for your workflow.

billsacks:
And just so we're on the same page: I do *not* plan to make those other changes in the near future: I'm under too much pressure to make progress on other projects.

PR #1448, opened 2021-07-31
Python script for modifying fsurdat files (after they have been generated with the tool-chain)

### LINKED ISSUE #1558: Enable tracking netcdf files with git lfs
See discussion in #1548 . Needed for #1448 .

slevisconsulting:
### Description of changes
Introducing new Python script  `fsurdat_modifier` using `subset_data` as my initial template. Currently this script just opens and reads an fsurdat file and writes a copy of the same file. In this PR, I will introduce functions that modify the fsurdat before outputting it.

### Specific notes
For now fsurdat_modifier.py runs as follows:
`./fsurdat_modifier.py --fsurdat_in </path/input_filename.nc> --fsurdat_out </path/output_filename.nc>`
or
`./fsurdat_modifier.py` for default entries of `fsurdat_in fsurdat_out`.
or
`./fsurdat_modifier.py --help` to see options (though this still lists options I found in the template code that I started from)

Contributors other than yourself, if any: @ekluzek @billsacks 

Resolves #1558 

Testing performed, if any:
I have executed the script in the ways explained a few lines up.


slevisconsulting:
[Slides](https://docs.google.com/presentation/d/1F9tlo-cIrkeF7GOCm_C-Je37aXQm3hIZNSJpEeNbxrc/edit?usp=sharing) that I presented to the CSSI group to discuss our options. The group decided to go with fsurdat_modifier.py, rather than the paleoToolkit or other method that would modify raw datasets.

While I'm sharing links, I take additional notes [here.](https://docs.google.com/document/d/16mhubUD03gtK_uj0aUFhmoK8RosNpQK9M90zOKVfxwI/edit?usp=sharing)

slevisconsulting:
@billsacks @ekluzek 
I requested a review already because I want to ensure that I start with a good template.

@billsacks I saw your emailed comments to @negin513 about her work, and I believe that they apply to mine since I used one of her scripts as my starting point. Feel free to just refer me back to that email if you feel I should start from there before you and @ekluzek spend much time reviewing this.

Also I didn't know whether to add others to this PR, yet. Feel free to add (or tell me to add) anyone you think necessary.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
+#! /usr/bin/env python
```
billsacks, line 1: - [ ] this should be python3

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+    ./fsurdat_modifier.py
+ 
+To remove NPL (ncar_pylib) from your environment on Cheyenne/Casper:
+    deactivate
+-------------------------------------------------------------------
+
+"""
+
+#  Import libraries
+from __future__ import print_function
```
billsacks, line 32: - [ ] If you change the shebang line to python3 (as you should), then this isn't necessary

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+import subprocess
+import argparse
+
+import xarray as xr
+
+from datetime import date
+from getpass import getuser
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
+
+myname = getuser()
```
billsacks, line 47: - [ ] It seems non-standard to call a function to set a variable at the top level like this – and it won't work smoothly once this is broken up into pieces in the python directory. So this should just be inlined where it's needed.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+
+def get_parser():                                                                                                                                                                   
+        """
+        Get parser object for this script.
+
+        Command-line inputs
+        - fsurdat_in: input file (str)
+        - fsurdat_out: output file (str)
+        - variable: variable to modify (str)
+        - value: value assigned to the variable to be modified (float)
```
billsacks, line 57: Can delete this part of the comment, which is already out of date and will get more so.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+                               formatter_class=argparse.RawDescriptionHelpFormatter)
+
+        parser.print_usage = parser.print_help
+
+        parser.add_argument('--dom_pft', 
+                    help='Dominant PFT if overwrite_single_pft = .true. [default: %(default)s] ', 
+                    action="store", 
+                    dest="dom_pft",
+                    type =int,
+                    default=7)
```
billsacks, line 69: - [ ] For settings like this where there isn't an obvious default, I would leave out the default so it needs to be explicitly specified.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+        parser.add_argument('--uni_snow', 
+                    help='Turn on the flag to create uniform snowpack. [default: %(default)s]', 
+                    action="store", 
+                    dest="uni_snow",
+                    default=False)
```
billsacks, line 74: - [ ] I feel like the typical thing to do for logicals like this (and others below) is to use action="store_true" (then you don't need default=False)

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+                    action="store", 
+                    dest="no_saturation_excess",
+                    type =bool,
+                    default=False)
+        parser.add_argument('--fsurdat_in', 
+                    help = 'Input surface dataset. [default: %(default)s]', 
+                    action = "store", 
+                    dest = "fsurdat_in",
+                    type = str,
+                    default = "/glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/release-clm5.0.18/surfdata_0.9x1.25_hist_78pfts_CMIP6_simyr2000_c190214.nc")
```
billsacks, line 97: - [ ] Again, for this and fsurdat_out, get rid of the default so it needs to be explicitly set

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+
+    def __init__(self, overwrite_single_pft, dominant_pft,
+                 zero_nonveg_landunits, uniform_snowpack, no_saturation_excess):
+        super().__init__()
+        self.overwrite_single_pft = overwrite_single_pft
+        self.dominant_pft = dominant_pft
+        self.zero_nonveg_landunits = zero_nonveg_landunits
+        self.uniform_snowpack = uniform_snowpack
+        self.no_saturation_excess = no_saturation_excess
```
billsacks, line 143: - [ ] The member variables of the class should probably be "private" (leading underscore) to note that they shouldn't be accessed directly from outside the class. e.g., `self._overwrite_single_pft`.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+    @staticmethod
+    def add_tag_to_filename(filename, tag):
```
billsacks, line 147: - [ ] This should be moved to a more general utility module so it can be shared between different scripts that need it.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+           os.abort()
+
+        today = date.today()
+        today_string = today.strftime("%y%m%d")
+
+        return(basename[:cend] + "_" + tag + "_c" + today_string + '.nc')
+
+
+    @staticmethod
+    def update_metadata(nc, filename):
```
billsacks, line 171: - [ ] This function would be another good candidate for a general utility module that can be shared between scripts.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+        today = date.today()
+        today_string = today.strftime("%Y-%m-%d")
+
+        #get git hash
+        sha = get_git_sha()
+
+        nc.attrs['Created_on'] = today_string
+        nc.attrs['Created_by'] = myname
+        nc.attrs['Created_with'] = os.path.abspath(__file__) + " -- " + sha
+        nc.attrs['Created_from'] = filename
```
billsacks, line 183: - [ ] To be announced soon: The new required metadata for inputdata files will be the following; "N/A" can be put in place for anything not applicable (e.g., contact in this case?):
  - title
  - summary (Plain English descriptive short one sentence, maybe two)
  - creator (Name only, no email)
  - contact (e.g., CAM Subboard at https://bb.cgd.ucar.edu)
  - creation_date (e.g., 2019-11-21)
  - data_script (script or instructions used to generate the dataset)
  - description (anything else relevant)
    - Maybe capturing the command-line here would be good (sys.argv)

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+
+        print ("Creating surface dataset file")
+        filename = self.fsurf_in
+        print("Open file: " + filename)
+        f1 = xr.open_dataset(filename)
+
+        # modify surface data properties
+        if self.overwrite_single_pft:
+            f1['PCT_NAT_PFT'][:,:,:] = 0
+            f1['PCT_NAT_PFT'][:,:,self.dominant_pft] = 100
```
billsacks, line 206: - [ ] [OPTIONAL] I would probably put the body of each of these conditionals in its own little function. That way, as they (inevitably) grow more complex over time, each operation will stay self contained.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+            f1['FMAX'][:,:] = 0.
+
+        # specify dimension order 
+        f1= f1.transpose(u'time', u'cft', u'lsmpft', u'natpft', u'nglcec', u'nglcecp1', u'nlevsoi', u'nlevurb', u'numrad', u'numurbl', 'lsmlat', 'lsmlon')
+        
+        #update attributes
+        self.update_metadata(f1, filename)
+
+        # mode 'w' overwrites file if it exists
+        f1.to_netcdf(path=self.fsurf_out, mode='w')
```
billsacks, line 226: - [ ] Should have a check if the given file already exists, aborting if so.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+        print('Successfully created file (fsurf_out) :' + self.fsurf_out)
+        f1.close()
+
+
+def setup_logging(log_file, log_level):
```
billsacks, line 231: - [ ] This would be another good candidate for general functionality to be shared between CTSM scripts. Actually, I have already started such a module: https://github.com/escomp/CTSM/blob/master/python/ctsm/ctsm_logging.py . There is extra stuff done in this setup_logging function; we should discuss what here we really want.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+    stream_handler = logging.StreamHandler(sys.stdout)
+    stream_handler.setFormatter(fmt)
+    root_logger.addHandler(stream_handler)
+
+    # redirect stdout/err to log file
+    StreamToLogger.setup_stdout()
+    StreamToLogger.setup_stderr()
+
+
+class StreamToLogger(object):
```
billsacks, line 260: - [ ] Again, this should be moved to a general place. However, I'd like to have some discussion on what here we really want. At a glance, it looks like this is doing something like redirecting print statements to the logger. My feeling is that we shouldn't do that, but instead should explicitly make logger calls at the appropriate level (logger.info, logger.debug, logger.warning, etc.) because that lets the user set the desired logging level for the script.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+def main ():
+
+    # Parse arguments from the command line
+    args = get_parser().parse_args()
+
+    # Set up logging
+    today = date.today()
+    today_string = today.strftime("%Y%m%d")
+    pwd = os.getcwd()
+    log_file = os.path.join(pwd, today_string + '.log')
```
billsacks, line 316: - [ ] I'm generally not a fan of logging to a file: it clutters up the directory, you need to manage possible file name collisions, and if a user wants it, they can redirect output to the desired file when running the command in the shell.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+    overwrite_single_pft = args.overwrite_single_pft
+    dominant_pft         = args.dom_pft
+    zero_nonveg_landunits= args.zero_nonveg
+    uniform_snowpack     = args.uni_snow
+    no_saturation_excess = args.no_saturation_excess
+
+    #--  Create ModifyFsurdat Object
+    modify_fsurdat = ModifyFsurdat(overwrite_single_pft, dominant_pft,
+                                   zero_nonveg_landunits, uniform_snowpack,
+                                   no_saturation_excess)
```
billsacks, line 334: - [ ] Given that this argument list is likely to become long, I'd suggest:
  - Not bothering making local copies of the args, but instead passing things like args.uni_snow directly
  - Using argument keywords in the function call to avoid accidentally associating the wrong arguments

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+    print(modify_fsurdat)
+
+    #--  Set input and output filenames
+    fsurf_in = args.fsurdat_in
+    fsurf_out = args.fsurdat_out
+
+    modify_fsurdat.fsurf_in = fsurf_in
+    modify_fsurdat.fsurf_out = fsurf_out
```
billsacks, line 342: - [ ] You shouldn't set these class members directly; instead, pass them as arguments to the modify function or (if they make sense as class members) set them via the constructor

billsacks (changes requested):
Okay, I have a lot of specific line comments in addition to some bigger picture things. I realize that you were following the example of some of @negin513 's scripts, so a lot of this really pertains to some things there that we have discussed but I guess she hasn't had a chance to incorporate yet.

- [x] The biggest thing is that almost all of this code should be moved to the python/ctsm directory, with code shared between scripts where possible. Then there should just be a little skeleton wrapper that loads the appropriate modules and calls the appropriate main function. Feelings of @ekluzek and me are that the executable wrapper should *not* have a ".py" extension. (Side note from last week's ctsm-software meeting: I looked briefly for a reference for this feeling and couldn't find one. However, it follows the general practice of unix utilities and (newer) CIME scripts: the scripts run by a user shouldn't have a language extension.)

- [ ] We should have a discussion about the use of logging. I made some line comments here. It looks like the way this is set up may be to redirect print statements to the logger. But I feel we should be more explicit about logging, in part because it gives you more control over different logging levels. For example, see https://docs.python.org/3/howto/logging.html#when-to-use-logging.

- [ ] [OPTIONAL] We may also want to discuss whether it's better to use a config file here rather than command-line arguments. I see pros & cons of each, and don't have any strong feelings on the matter, so we could just proceed with the command line argument-based approach for now, but it may be worth a bit of discussion up-front.

tools/contrib/fsurdat_modifier.py
```diff
@@ -0,0 +1,356 @@
...
+    ncar_pylib
+
+-------------------------------------------------------------------
+To see available options:
+    ./fsurdat_modifier.py --help
+-------------------------------------------------------------------
+
+This script reads a surface dataset (fsurdat file) and outputs a
+modified copy of the same file.
+
```
wwieder, line 20: Maybe these notes can clarify how this is different than the `modify_singlept_site_neon.py` script that @negin513 created and what it should be used for? I have two other high level questions:

1. Is there code that should be shared between these scripts (to @billsacks' point about putting this in the python directory.
2. Should we consider renaming the wrapper scripts we're calling so they are descriptive and consistent?

slevisconsulting:
@billsacks I think I have brought the code in this PR closer to your recommendations. Pls let me know if I took a wrong turn anywhere and whether there are additional improvements and corrections that you envision.

@billsacks @ekluzek @negin513 thank you for the recent meetings outlining what needed to happen.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,100 @@
...
+def get_parser():
+
+    """
+    Get parser object for this script.
+
+    Command-line inputs
+    - fsurdat_in: input file (str)
+    - fsurdat_out: output file (str)
+    - variable: variable to modify (str)
+    - value: value assigned to the variable to be modified (float)
```
billsacks, line 25: - [ ] Can delete this part of the comment, which is already out of date and will get more so

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,100 @@
...
+    args = get_parser().parse_args()
+
+    #--  Create ModifyFsurdat Object
+    modify_fsurdat = ModifyFsurdat(args.fsurdat_in,
+                                   args.fsurdat_out,
+                                   args.overwrite_single_pft,
+                                   args.dom_pft,
+                                   args.zero_nonveg_lu,
+                                   args.uni_snowpack,
+                                   args.no_saturation_excess)
```
billsacks, line 93: - [ ] For an argument list this long (and likely to grow longer), using keyword arguments helps prevent bugs (`fsurdat_in = args.fsurdat_in, fsurdat_out = args.fsurdat_out`, etc.).

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,100 @@
...
+                type =int)
+    parser.add_argument('--uni_snow',
+                help='Turn on the flag to create uniform snowpack. [default: %(default)s]',
+                action="store_true",
+                dest="uni_snowpack")
+    parser.add_argument('--overwrite_single_pft',
+                help='Turn on flag to make whole grid 100% single PFT. [default: %(default)s]',
+                action="store",
+                dest="overwrite_single_pft",
+                default=False)
```
billsacks, line 46: - [ ] For this and a few other booleans, you could follow the pattern of setting `action="store_true"` and getting rid of the `default=False`. (You did that for a few but not all.)

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,100 @@
...
+                action="store",
+                dest="no_saturation_excess",
+                type =bool,
+                default=False)
+    parser.add_argument('--fsurdat_in',
+                help = 'Input surface dataset. [default: %(default)s]',
+                action = "store",
+                dest = "fsurdat_in",
+                type = str,
+                default = "/glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/release-clm5.0.18/surfdata_0.9x1.25_hist_78pfts_CMIP6_simyr2000_c190214.nc")
```
billsacks, line 64: - [ ] I think you should remove this default and make this a required argument.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,100 @@
...
+    parser.add_argument('--fsurdat_out',
+                help = 'Output surface dataset. [default: %(default)s]',
+                action = "store",
+                dest = "fsurdat_out",
+                type = str,
+                default = "/glade/scratch/" + getuser() +
+                          "/surfdata_0.9x1.25_hist_78pfts_CMIP6_simyr2000_c190214_modified.nc")
```
billsacks, line 71: - [ ] I think you should remove this default and make this a required argument.

billsacks (changes requested):
Thanks a lot for your work on this, @slevisconsulting ! You have addressed most of my initial comments, and this is moving in the right direction; thank you.

I was thinking a little bit about the object-oriented design you use here, though. At least as of now, it looks like the public interface to the ModifyFsurdat class consists of an `__init__` method (which sets flags based on the command-line arguments) and a single main method that doesn't take any arguments (`modify`). In most cases, I feel like the design is simpler & easier to understand if situations like this are changed to be non-object-oriented – so you would replace this class with a non-object-oriented `modify_fsurdat` function whose arguments match the current arguments to the `__init__` method. Along with this, you could combine your current two files into one, keeping everything together, which feels a little better to me.

However, here I can also see what feels to me like a more intuitive object-oriented design – and one that would likely lend itself better to unit testing:
- The `__init__` method would accept just one argument, the input file, and would read that file into memory, storing it in a variable in the object
- There would then be separate methods for each modification you could do – e.g., `overwrite_single_pft`, `zero_nonveg_lu`, etc. The main program would then have logic like: `if args.overwrite_single_pft: modify_fsurdat.overwrite_single_pft(...)` and similarly for the other modifications. So the main program would end up calling multiple methods, and the in-memory data would get incrementally modified.
- There would also be a final method to write out the result to a file, which would be called at the end of the main program

What I like about this proposed design is that it feels easier and more intuitive for testing, either via unit tests or interactively: I can imagine initializing the object with some data (in a unit test context, I would imagine using a different constructor where I set the data directly rather than reading it from file), then calling one or more modifier methods, then checking the resulting data against expectations. It's possible to do the same thing with the current design, but it feels more awkward, because you have to specify up-front what changes you want to make, then the modify method does everything at once.

A possible downside of this proposed design is if certain modifications can only be done in a particular order: it feels easier to ensure that order dependence with your current design.

As with most design things, I don't feel like there is a clearly "right" design, but I thought I'd suggest this alternative to see what you think about it. I don't think it would take too much rework at this stage... and most of the rework is work I'd suggest doing anyway: breaking down the big `modify` method into separate functions for the separate operations (it's just a question of whether the single big `modify` method calls a bunch of individual, private functions, or if those functions are all public and called by the main program).

tools/contrib/fsurdat_modifier.py
slevisconsulting: @wwieder thanks. I have added clarifications that will appear in my next commit.

1. Ideally, yes, and we will continue to keep that in mind during development.
2. I'm not attached to existing names.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
slevisconsulting: Will eventually get to this (sorry for postponing).

python/ctsm/modify_fsurdat/fsurdat_modifier.py
slevisconsulting: Postponing because during development I find it convenient to not have to enter fsurdat_in and fsurdat_out every time I run.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
slevisconsulting: Postponing because during development I find it convenient to not have to enter fsurdat_in and fsurdat_out every time I run.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
slevisconsulting: Not relevant to this arg. list anymore, but implementing this suggestion when calling update_metadata.

billsacks:
+1 to @ekluzek 's comment. But I'll also add, regarding this:

> For your idealized case above the surface dataset should set land quantities for the square and set everything outside the square to wetland.

I imagined that you might keep things even simpler in this case: If the user wants to set something like constant values or a latitudinal gradient of values, you would simply fill the whole world with those values. As @ekluzek says, it doesn't matter what's on the surface dataset in grid cells that the overall run classifies as ocean, so I feel like the whole process will be simpler in many cases if you don't try to distinguish between land and non-land points in creating idealized surface datasets in examples like this (with a strip of land, for example).

python/ctsm/utils.py
```diff
@@ -33,3 +46,138 @@ def fill_template_file(path_to_template, path_to_final, substitutions):
...
+        abort("Error: {} needs to be specified in config file {}".format(item, file_path))
+
+    if allowed_values is not None:
+        if val not in allowed_values:
+            abort("Error: {} is not an allowed value for {} in config file {}\n"
+                  "Allowed values: {}".format(val, item, file_path, allowed_values))
+
+    return val
+
+def select_value(var, default, type_of_var):
```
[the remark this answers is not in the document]
slevisconsulting: Two pending items here:
- Moving `allowed_values` into `_handle_config_value`. I think we should discuss this one before I proceed.
- Add unit tests.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,115 @@
...
+    fsurdat_in = get_config_value(config, 'modify_input', 'fsurdat_in',
+                                  cfg_path)
+    fsurdat_out = get_config_value(config, 'modify_input', 'fsurdat_out',
+                                   cfg_path)
+
+    # not required: user may set these in ./modify.cfg
+    # TODO Error checking not complete for these. Should it be done here
+    # while reading or later when using them?
+    temp = get_config_value(config, 'modify_input', 'idealized', cfg_path,
+                            allowed_values=['True','False','UNSET'])
```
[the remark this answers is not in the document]
slevisconsulting: @billsacks this may need follow-up. I addressed your...
- first checkbox by using the renamed constant in utils.py `CONFIG_UNSET` and
- second checkbox by replacing `get_config_value` with `getboolean`.

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,347 @@
...
+    write_output:
+        Write out the modified fsurdat (surface dataset)
+
+    TODO: Complete list of methods here?
+    """
+
+    def __init__(self, fsurdat_in):
+
+        print("Open file: " + fsurdat_in)
+        self.file = xr.open_dataset(fsurdat_in)
```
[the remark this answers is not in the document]
slevisconsulting: Renamed `self.file` to `self._file` and back to `self.file` because in the refactor from set_in_rectangle to set_idealized `self.file` is not private.

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -21,7 +24,8 @@ class ModifyFsurdat:
 
     def __init__(self, fsurdat_in, lon_1, lon_2, lat_1, lat_2, landmask_file):
 
-        print("Opening fsurdat_in file to be modified: " + fsurdat_in)
+        logger.info(
+            'Opening fsurdat_in file to be modified: ' + fsurdat_in)
```
slevisconsulting, line 28: @billsacks I realize that replacing the two `print` statements w `logger.info` statements may be overkill, but I wanted to make sure I had the basic logging infrastructure in this tool.

slevisconsulting:
@billsacks I started working on `test_sys_fsurdat_modifier.py` using the test_sys_lilac file that you suggested as template. I intended to replicate my manual testing, as you recommended. For now I get these errors:
- No such file or directory: 'rm ../tools/modify_fsurdat/test_modify.nc'
- No such file or directory: 'rm ../tools/modify_fsurdat/test_modify_minimal.nc'
Despite the errors, I pushed to the PR in case you have feedback on what I've done with `test_sys_fsurdat_modifier.py` so far (also because I plan to take the next two days off).

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,292 @@
...
+
+        if landmask_file is not None:
+            # overwrite self.not_rectangle with data from
+            # user-specified .nc file in the .cfg file
+            self._landmask_file = xr.open_dataset(landmask_file)
+            rectangle = self._landmask_file.landmask
+            self.not_rectangle = np.logical_not(rectangle)
+
+
+    def _get_not_rectangle(self, lon_1, lon_2, lat_1, lat_2, longxy, latixy):
```
slevisconsulting, line 43: self not used in this function
put `@staticmethod` in the line before def

billsacks:
@slevisconsulting - just thinking a bit about how I might set up a function to aid system testing, ensuring that the output matches expectations: I could imagine doing one of these options (or both) – though I'm certainly not tied to this if you have other ideas – I just thought I'd share what came into my head:

(1) Have a function, `confirm_files_equal(expected, actual, fields_expected_to_differ=None)`: `expected` and `actual` could either be file paths or xarray datasets; `fields_expected_to_differ`, if provided, should be a list of strings. Loop over all variables (skipping those in `fields_expected_to_differ`) and see if they are identical between `expected` and `actual`; for any that differ, add them to a list of fields that differ, then return that list at the end. The caller can then assert that the returned list is empty; if the assertion fails, the message can be the list of fields that differ. Alternatively, put the assertions directly in the loop within this function, which could allow for more information about the unexpected differences in the individual field, the downside being that as soon as there is a failure for one field the test will abort, so you won't see the full list of fields that differ (so, pros & cons either way).

Then, for a test where you expect some fields to differ in the output compared with the input, but expect some fields to stay the same, you can call the above function with a list of fields that you expect to differ, then follow that with individual `@assertEquals` calls for the fields that you have changed, comparing against expectations.

(2) Or you could do all of the assertions at once: instead of having `fields_expected_to_differ` be simply a list of field names, it could be a dictionary of expected values for the fields expected to differ, mapping the field name to the expected values for that field. Then, in the loop through variables, if the current variable is in the keys of this dictionary, it would be compared against the corresponding value, instead of being compared against the value from the `expected` file.

python/ctsm/test/test_sys_fsurdat_modifier.py
```diff
@@ -0,0 +1,108 @@
...
+    def tearDown(self):
+        shutil.rmtree(self._tempdir, ignore_errors=True)
+
+    def test_minimalInfo(self):
+        """
+        This test specifies a minimal amount of information
+        """
+
+        # get fsurdat_in and fsurdat_out from the .cfg (config) file
+        cfg_path = 'ctsm/test/testinputs/modify_minimal.cfg'
```
slevisconsulting, line 36: @billsacks I created directory /ctsm/test/testinputs, as I understood you to recommend. As of now it contains two configure files and an fsurdat_in file:
- modify_minimal.cfg
- modify_all.cfg
- surfdata_5x5_amazon_16pfts_Irrig_CMIP6_simyr2000_c171214.nc

Would you have me add/commit/push these to the PR or would I eventually rimport them?

python/ctsm/test/test_sys_fsurdat_modifier.py
```diff
@@ -0,0 +1,108 @@
...
+        # move the generated file to _tempdir
+        fsurdat_out_final = os.path.join(self._tempdir, fsurdat_out)
+        shutil.move(fsurdat_out, fsurdat_out_final)
+
+        fsurdat_in_data = xr.open_dataset(fsurdat_in)
+        fsurdat_out_data = xr.open_dataset(fsurdat_out_final)
+
+        self.assertFalse(fsurdat_out_data.equals(fsurdat_in_data))
+        # TODO read template file and self.assertTrue is
+        # easier than needing to specify every variable...
```
slevisconsulting, line 103: @billsacks for now I just assert here the opposite of what I asserted in `def test_minimalInfo`

Now that I've come across this super-simple way of comparing the contents of .nc files, I'm tempted to place a template file in /testinputs that `def test_allInfo` compares to with `self.assertTrue`. I won't do it if you disagree :-)

python/ctsm/test/test_sys_fsurdat_modifier.py
```diff
@@ -0,0 +1,108 @@
...
+        # doesn't generate errors; however, we also do some assertions below
+
+        # move the generated file to _tempdir
+        fsurdat_out_final = os.path.join(self._tempdir, fsurdat_out)
+        shutil.move(fsurdat_out, fsurdat_out_final)
+
+        fsurdat_in_data = xr.open_dataset(fsurdat_in)
+        fsurdat_out_data = xr.open_dataset(fsurdat_out_final)
+
+        self.assertTrue(fsurdat_out_data.equals(fsurdat_in_data))
```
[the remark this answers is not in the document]
billsacks: That's fantastic!

billsacks:
I'm feeling stuck regarding what to do about the netCDF inputs and outputs for these tests. I'd like to get input from  @ekluzek and others at an upcoming CTSM-software meeting. (In the meantime, can you just keep this file where it is but not commit it?) See https://github.com/ESCOMP/CTSM/issues/1548 for my initial thoughts.

python/ctsm/utils.py
```diff
@@ -33,3 +46,149 @@ def fill_template_file(path_to_template, path_to_final, substitutions):
...
+            err.args = (str(err) + errmsg,)
+            raise
+
+    if var is not None and not is_list and convert_to_type is not None:
+        try:
+            var = convert_to_type(var)
+        except Exception as err:
+            errmsg = '--> Check .cfg file variable: ' + item
+            err.args = (str(err) + errmsg,)
+            raise
```
billsacks, line 192: - [ ] The `is_list` block assumes that `convert_to_type` is not None. I think you could consolidate this code as follows (untested); this also implements a number of my other suggestions:

``` python
def _handle_config_value(var, default, item, is_list, convert_to_type, can_be_unset, allowed_values):
    """
    Description
    -----------
    Assign the default value or the user-specified one to var.
    Convert from default type (str) to reqested type (int or float).

    If is_list is True, then default should be a list
    """
    if var == CONFIG_UNSET:
        if can_be_unset:
            return default  # default may be None
        abort('Must set a value for .cfg file variable: {}'.format(item))

    # convert string to list of strings; if there is just one element, we will get a list
    # of size one, which we will convert back to a scalar later if needed
    var = var.split()

    if convert_to_type is bool:
        try:
            var = [_convert_to_bool(v) for v in var]
        except ValueError:
            abort('Non-boolean value found for .cfg file variable: {}'.format(item))
    elif convert_to_type is not None:
        try:
            var = [convert_to_type(v) for v in var]
        except ValueError:
            abort('Wrong type for .cfg file variable: {}'.format(item))

    if allowed_values is not None:
        for v in var:
            if v not in allowed_values:
                abort("{} is not an allowed value for {} in .cfg file\n"
                      "Allowed values: {}".format(v, item, allowed_values))

    if not is_list:
        if len(var) > 1:
            abort('More than 1 element found for .cfg file variable: {}'.format(item))
        var = var[0]

    return var

def _convert_to_bool(val):
    """Convert the given value to boolean

    Conversion is as in config files 'getboolean'
    """
    if val.lower() in ['1', 'yes', 'true', 'on']:
        return True
    if val.lower() in ['0', 'no', 'false', 'off']:
        return False
    raise ValueError("{} cannot be converted to boolean".format(val))
```

- [ ] Note that I have added the `can_be_unset` handling that I suggested in an earlier comment. You should therefore remove 'CONFIG_UNSET' as an allowed value for some variables, setting `can_be_unset` to `True` for those.

- [ ] Note that I have moved the allowed value checking into here; it can then be removed from `get_config_value`

A few other notes about this:
- In the initial check for `CONFIG_UNSET`, I return immediately. I think that should work: I don't think we should need any of the later processing in this case (and some of the later processing may not work right if we try to run it after setting a default value). That also  removes the need for checking for `None` values later in the code (I assume those checks were just to handle the possibility that we had set the value to the default, which was `None`).
- I have changed `list(var.split())` to simply `var.split()`: I think this already returns a list so the conversion to list is unnecessary (unless you think otherwise)
- I have changed `list(map(convert_to_type, var))` to the equivalent list comprehension, `[convert_to_type(v) for v in var]`. They are equivalent, but I personally find the list comprehension easier to read. I'm really fine with either, though: my main point was to illustrate how you could do this with a list comprehension in case you aren't as familiar with that form; if you prefer the `map`, then by all means stick with it.
- I have changed the exception handling to call abort rather than raising an exception, because the latter can be ugly for users to have to read through; you should try this and make sure the error message looks good
- Note that I have implemented my suggested conversion-to-boolean function here (see an earlier comment in this review related to this)

- [ ] Finally, there should be some unit tests of both of these functions (`_convert_to_bool` and `_handle_config_value`)

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,292 @@
...
+                self.set_lai_sai_hgts(pft=pft, dom_nat_pft=dom_nat_pft,
+                                      var_name=var_name, var=sai)
+            var_name='MONTHLY_HEIGHT_TOP'
+            if hgt_top is not None:
+                self.set_lai_sai_hgts(pft=pft, dom_nat_pft=dom_nat_pft,
+                                      var_name=var_name, var=hgt_top)
+            var_name='MONTHLY_HEIGHT_BOT'
+            if hgt_bot is not None:
+                self.set_lai_sai_hgts(pft=pft, dom_nat_pft=dom_nat_pft,
+                                      var_name=var_name, var=hgt_bot)
```
billsacks, line 166: - [ ] Looking at the code here and in `set_lai_sai_hgts`, it looks unnecessary to have these calls to `set_lai_sai_hgts` in the `for pft` loop: you can take these out of the loop and just pass in `dom_nat_pft`, then getting rid of the `if pft == dom_nat_pft` conditional in `set_lai_sai_hgts`.

tools/modify_fsurdat/modify_template.cfg
```diff
@@ -0,0 +1,88 @@
+[modify_input]
+
+# ------------------------------------------------------------------------
+# .cfg file with inputs for fsurdat_modifier.
```
billsacks, line 4: Thanks for all of your detailed & helpful comments in this config file!

billsacks (changes requested):
This is looking really great! I do have some additional comments / requests on some of the new code; some of the bigger things are clarifications from my last round of comments.

I feel like it might be most efficient if we address some of the remaining needs - both my outstanding comments and the addition of unit tests - by working together for a couple of hours (pair programming): that way we can quickly get through these comments synchronously, which might be more efficient (or at least more fun) than continuing to go back and forth in this PR. If you agree, let me know when would be a good time to do that.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,134 @@
...
+        modify_fsurdat.file['FMAX'] = \
+         modify_fsurdat.file['FMAX'].where(modify_fsurdat.not_rectangle, other=max_sat_area)
+
+    if std_elev is not None:  # overwrite "idealized" value
+        modify_fsurdat.file['STD_ELEV'] = \
+         modify_fsurdat.file['STD_ELEV'].where(modify_fsurdat.not_rectangle, other=std_elev)
+
+    if soil_color is not None:  # overwrite "idealized" value
+        modify_fsurdat.file['SOIL_COLOR'] = \
+         modify_fsurdat.file['SOIL_COLOR'].where(modify_fsurdat.not_rectangle, other=soil_color)
```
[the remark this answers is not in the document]
billsacks: (A good general rule in object-oriented design is that all member variables of a class should be private, only accessed within functions defined in that class. This helps long-term maintainability and understandability of the code. As with any rules, there are times when it's worth breaking the rule; for example, we break this rule all over the place in CTSM, for performance reasons and because the overall design of CTSM doesn't lend itself to following this rule. But this rule is generally a good starting point for object-oriented design.)

python/ctsm/utils.py
```diff
@@ -33,3 +46,149 @@ def fill_template_file(path_to_template, path_to_final, substitutions):
...
+
+def _handle_config_value(var, default, item, is_list, convert_to_type):
+    """
+    Description
+    -----------
+    Assign the default value or the user-specified one to var.
+    Convert from default type (str) to reqested type (int or float).
+    """
+    if var == CONFIG_UNSET:
+        var = default  # default may be None
```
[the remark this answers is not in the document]
billsacks: - [ ] And those variables will also need to set the new `can_be_unset = True`

billsacks:
@slevisconsulting I have checked off the completed requests from my earlier review - which is all of them except for the addition of tests. Thank you very much for addressing all of those points and for your comments: as always, I appreciate your care in going through these points.

Again, please let me know if you'd like to go through the remaining points together. I think we can get through most of them pretty quickly together, and this might be more efficient than continuing to do this asynchronously, at least for anything that isn't super-clear.

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,292 @@
...
+        # Default values
+        zbedrock = 10
+        max_sat_area = 0  # max saturated area
+        max_inundated = 0  # max inundated fraction
+        std_elev = 0  # standard deviation of elevation
+        slope = 0  # mean topographic slope
+        pftdata_mask = 1
+        landfrac_pft = 1
+        pct_nat_veg = 100
+        pct_not_nat_veg = 0
```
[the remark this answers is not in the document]
slevisconsulting: Adding comments that these two lines must not be changed and that these values work together.

python/ctsm/utils.py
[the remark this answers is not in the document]
slevisconsulting: This far I didn't find it necessary to create `can_be_unset = True`. Let's discuss this to make sure I haven't misunderstood.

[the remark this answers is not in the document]
slevisconsulting: I see that you discuss `can_be_unset` in your next comment, so my previous comment may be irrelevant...

python/ctsm/modify_fsurdat/modify_fsurdat.py
slevisconsulting: Good catch!

python/ctsm/utils.py
slevisconsulting: As of now I have addressed this **except for the unit tests.**
(I hope I didn't miss any details, but that's what the unit tests will help with.)

slevisconsulting: @billsacks as always thank you for the well-written code and clear instructions by the way!

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -0,0 +1,134 @@
...
+    # required but fallback values available for variables omitted
+    # entirely from the .cfg file
+    idealized = config.getboolean(section, 'idealized', fallback=False)
+    zero_nonveg = config.getboolean(section, 'zero_nonveg',
+                                    fallback=False)
+
+    lnd_lat_1 = config.getfloat(section, 'lnd_lat_1', fallback=-90)
+    lnd_lat_2 = config.getfloat(section, 'lnd_lat_2', fallback=90)
+    lnd_lon_1 = config.getfloat(section, 'lnd_lon_1', fallback=0)
+    lnd_lon_2 = config.getfloat(section, 'lnd_lon_2', fallback=360)
```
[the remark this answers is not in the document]
slevisconsulting: I think this is addressed now.

billsacks:
@slevisconsulting here is the start of the system test that we worked on yesterday (`cime_config/SystemTests/fsurdatmodifyctsm.py`):

```python
"""
CTSM-specific test to ...
"""

from CIME.SystemTests.system_tests_common import SystemTestsCommon
from CIME.XML.standard_module_setup import *
from CIME.SystemTests.test_utils.user_nl_utils import append_to_user_nl_files

logger = logging.getLogger(__name__)

class FSURDATMODIFYCTSM(SystemTestsCommon):

    def __init__(self, case):
        """
        initialize an object interface to the SMS system test
        """
        SystemTestsCommon.__init__(self, case)

        self._fsurdat = ""  # this will get set in _create_config_file

        # if ".done_setup" doesn't exist, do the following:
        # call preview_namelists
        # ideally, call check_input_data with the download (but that's icing)
        # then search CaseDocs/lnd_in for the fsurdat line, get the value, and use that as the input fsurdat
        # then:
        self._create_config_file()
        self._run_modify_fsurdat()
        self._modify_user_nl()

        # Create an empty file named ".done_setup" in self._caseroot()

    def _create_config_file(self):
        cfg_file_path = os.path.join(self._get_caseroot(),
                                     'modify_fsurdat.cfg')
        ctsm_root = self._case.get_value('COMP_ROOT_DIR_LND')
        self._fsurdat = os.path.join(self._get_caseroot(), 'fsurdat.nc')
        # Do something like what you did for the python system test


    def _run_modify_fsurdat(self):
        # Run the tool similarly to what you did for the python system test

    def _modify_user_nl(self):
        append_to_user_nl_files(caseroot = self._get_caseroot(),
                                component = "clm",
                                contents = "fsurdat = {}".format(self._fsurdat))
```

python/ctsm/modify_fsurdat/fsurdat_modifier.py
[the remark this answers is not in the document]
slevisconsulting: I have addressed this now.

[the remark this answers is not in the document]
billsacks: Thanks, this is moving towards what I imagined but still could be improved for a more standard object-oriented design: Instead of code like:

```python
modify_fsurdat.file['FMAX'] = modify_fsurdat.setvar(
            modify_fsurdat.file['FMAX'], max_sat_area)
```

replace this with

```python
modify_fsurdat.setvar('FMAX', max_sat_area)
```

so that `file` becomes a private variable of this class that isn't modified outside of it.

You can let me know if you see a problem with this approach.

python/ctsm/lilac_make_runtime_inputs.py
```diff
@@ -223,11 +187,11 @@ def buildnml(cime_path, rundir):
 
     # determine if fsurdat and/or finidat should appear in the -namelist option
     extra_namelist_opts = ''
-    if fsurdat != _UNSET:
-        # NOTE(wjs, 2020-06-30) With the current logic, fsurdat should never be _UNSET,
-        # but it's possible that this will change in the future.
+    if fsurdat != None:
```
billsacks, line 190: - [ ] Thanks for this change, but the pythonic way to check for None is: `if fsurdat is not None`. Can you please change this? (See https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html)

python/ctsm/lilac_make_runtime_inputs.py
```diff
@@ -223,11 +187,11 @@ def buildnml(cime_path, rundir):
...
+        # NOTE(wjs, 2020-06-30) With the current logic, fsurdat should never be UNSET
+        # (ie None here) but it's possible that this will change in the future.
         extra_namelist_opts = extra_namelist_opts + " fsurdat = '{}' ".format(fsurdat)
-    if finidat != _UNSET:
+    if finidat != None:
```
billsacks, line 194: - [ ] As above, this should be: `if finidat is not None`

python/ctsm/utils.py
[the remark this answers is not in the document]
billsacks: You have partially addressed my above comment, starting with "If I understand this right" (thank you), but part one is still needed:

- [ ] Please document the behavior in get_config_value where anything that had the value "UNSET" will end up being None (in the docstring for `get_config_value`)

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,288 @@
...
+            (float) User's entry of MONTHLY_SAI for their dom_nat_pft
+        hgt_top:
+            (float) User's entry of MONTHLY_HEIGHT_TOP for their dom_nat_pft
+        hgt_bot:
+            (float) User's entry of MONTHLY_HEIGHT_BOT for their dom_nat_pft
+        """
+
+        # initialize 3D variable
+        self.file['PCT_NAT_PFT'][:,:,:] = self.setvar(
+            self.file['PCT_NAT_PFT'][:,:,:], 0)
```
billsacks, line 148: - [ ] Are you sure this works correctly to set this 3-d variable? The mask (the not_rectangle) is defined in terms of a single 2-d slice, so I would think that it wouldn't work to use that mask directly to subset a variable with an extra dimension. But maybe there are some smarts in xarray to do this correctly? I just want to double-check: if you're not sure, it would be a good thing to play around with to make sure it's working right. (We can also do this together if you want.)

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def setvar(self, var, val):
+        """
+        Set variable var to value val in the *rectangle* domain, which
+        is denoted as *other* in the *not_rectangle* domain.
+        """
+
+        var = var.where(self.not_rectangle, other=val)
+
+        return var
```
billsacks, line 210: - [ ] As mentioned above, there would be better [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)) in this class if this function did the setting of the file variable – changing its interface to `def setvar(self, varname, val)` and then actually doing the `self.file[varname] = ...` itself. Sorry that I didn't explain my thinking clearly before.

``` python
def setvar(self, varname, val):
"""
Sets the value for a single-level field
"""
    self.file[varname] = self.file[varname].where(self.not_rectangle, other=val)
```

- [ ] Also, if I'm right in my comment above that this doesn't work right for setting a 3-d variable, then you could add a check in this function that `self.file[varname]` has the same size/shape as `self.not_rectangle`

- [ ] I see that you may need to then have a separate version of this function for multi-level variables, like:

``` python
def setvar_onelevel(self, varname, val, level):
"""
Sets one level of a multi-level field
"""
    self.file[varname][level,:,:] = self.file[varname][level,:,:].where(self.not_rectangle, other=val)
```

- [ ] And similarly for a version that sets one slice of a 4-d field (as needed by `set_lai_sai_heights`); I'm not sure off-hand what the best naming would be here

One other advantage of redefining the function in this way is: We'll need this to work on surface datasets that are on unstructured as well as structured grids. The former will imply that there is only one spatial dimension, not two. You should then be able to make changes in just a small number of places to handle files with either one or two spatial dimension(s).

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # dictionary of 4d variables to loop over
+        vars_4d = {'MONTHLY_LAI': lai,
+                   'MONTHLY_SAI': sai,
+                   'MONTHLY_HEIGHT_TOP': hgt_top,
+                   'MONTHLY_HEIGHT_BOT': hgt_bot}
+        for var_name, var in vars_4d.items():
+            if var is not None:
+                self.set_lai_sai_hgts(dom_nat_pft=dom_nat_pft,
+                                      var_name=var_name, var=var)
```
billsacks, line 161: Thanks for this code consolidation using the dictionary.

python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -0,0 +1,288 @@
...
+        max_inundated = 0  # max inundated fraction
+        std_elev = 0  # standard deviation of elevation
+        slope = 0  # mean topographic slope
+        pftdata_mask = 1
+        landfrac_pft = 1
+        # if pct_nat_veg had to be set to less than 100, then each special
+        # landunit would have to receive a unique pct value rather than the
+        # common value used here in pct_not_nat_veg = 0
+        pct_nat_veg = 100  # do not change; works with pct_not_nat_veg = 0
+        pct_not_nat_veg = 0  # do not change; works with pct_nat_veg = 0
```
billsacks, line 239: - [ ] Thanks for the above comments; this one should say "works with pct_nat_veg = 100" (not 0)

python/ctsm/test/test_sys_fsurdat_modifier.py
```diff
@@ -0,0 +1,105 @@
...
+        self.assertFalse(fsurdat_out_data.equals(fsurdat_in_data))
+
+        # compare fsurdat_out to fsurdat_out_baseline
+        fsurdat_out_baseline = fsurdat_in[:-3] + '_modified' + fsurdat_in[-3:]
+        fsurdat_out_base_data = xr.open_dataset(fsurdat_out_baseline)
+        # assert that fsurdat_out equals fsurdat_out_baseline
+        self.assertTrue(fsurdat_out_data.equals(fsurdat_out_base_data))
+
+
+    def prepare_cfg_file(self, case):
```
billsacks, line 82: Thanks for pulling the common code into this function.

billsacks (changes requested):
Once again, this is getting closer, though I have some new comments on your new changes. I have checked off the old requests that you have done; there are still a few left there that I have commented on. Here is one other outstanding request that I wasn't able to reply to inline for some reason:

https://github.com/ESCOMP/CTSM/pull/1448#discussion_r748742122

python/ctsm/utils.py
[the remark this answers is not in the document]
slevisconsulting: Ok, did this now.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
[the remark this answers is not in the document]
slevisconsulting: Thanks, @billsacks 
I had originally tried what you're suggesting here but opted for a way to make the function shorter. Now I'm trying your suggestion, which works. Let me know if you see an elegant way to shorten the function in this context.

python/ctsm/modify_fsurdat/modify_fsurdat.py
slevisconsulting: thank you for catching

python/ctsm/modify_fsurdat/fsurdat_modifier.py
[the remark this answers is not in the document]
slevisconsulting: Scrolling through your comments, I see that you have further suggestions regarding this function which may answer my question.

python/ctsm/modify_fsurdat/modify_fsurdat.py
slevisconsulting: I think I have addressed the comments in this post now.

I did not check whether my code the way I had it would work for a 3d variable. I just moved all 3d and 4d vars into for loops.

python/ctsm/modify_fsurdat/modify_fsurdat.py
slevisconsulting: Responded to this elsewhere.

billsacks:
@slevisconsulting - Brian Vanderwende in CISL gave me guidance as to how we can ensure that we are using the correct python version in the call to fsurdat_modifier in `fsurdatmodifyctsm.py`. His suggestion is to do something like:

```python
subprocess.check_call(['/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/pkg-library/20201220/bin/python', tool_path, self._cfg_file_path])
```

I played around with this a bit to remove the hard-coding of a specific python version (which feels fragile) and came up with https://github.com/slevisconsulting/ctsm/pull/3 – which is still a bit fragile and hard-coded to cheyenne, but hopefully less fragile than the simple thing above.

I tested this and it appears to get past the 'import xarray' but fails now with:

```
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/pkg-library/20201220/lib/python3.7/site-packages/xarray/core/dataset.py", line 1259, in _construct_dataarray
    variable = self._variables[name]
KeyError: 'F0'

During handling of the above exception, another exception occurred:

  File "/glade/work/sacks/ctsm_code/ctsm/tools/modify_fsurdat/fsurdat_modifier", line 66, in <module>
    main()
  File "/glade/work/sacks/ctsm_code/ctsm/tools/modify_fsurdat/../../python/ctsm/modify_fsurdat/fsurdat_modifier.py", line 111, in main
    modify_fsurdat.set_idealized()  # set 2D variables
  File "/glade/work/sacks/ctsm_code/ctsm/tools/modify_fsurdat/../../python/ctsm/modify_fsurdat/modify_fsurdat.py", line 259, in set_idealized
    self.setvar_lev0('F0', max_inundated)
  File "/glade/work/sacks/ctsm_code/ctsm/tools/modify_fsurdat/../../python/ctsm/modify_fsurdat/modify_fsurdat.py", line 203, in setvar_lev0
    self.file[var] = self.file[var].where(
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/pkg-library/20201220/lib/python3.7/site-packages/xarray/core/dataset.py", line 1359, in __getitem__
    return self._construct_dataarray(key)
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/pkg-library/20201220/lib/python3.7/site-packages/xarray/core/dataset.py", line 1262, in _construct_dataarray
    self._variables, name, self._level_coords, self.dims
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/pkg-library/20201220/lib/python3.7/site-packages/xarray/core/dataset.py", line 171, in _get_virtual_variable
    ref_var = variables[ref_name]
KeyError: 'F0'
  File "./case.build", line 167, in <module>
    _main_func(__doc__)
  File "./case.build", line 147, in _main_func
  File "/glade/work/sacks/ctsm_code/ctsm/cime_config/SystemTests/fsurdatmodifyctsm.py", line 50, in __init__
    self._run_modify_fsurdat()
  File "/glade/work/sacks/ctsm_code/ctsm/cime_config/SystemTests/fsurdatmodifyctsm.py", line 78, in _run_modify_fsurdat
    subprocess.check_call([python_path, tool_path, self._cfg_file_path])
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/pkg-library/20201220/bin/python', '/glade/work/sacks/ctsm_code/ctsm/tools/modify_fsurdat/fsurdat_modifier', '/glade/scratch/sacks/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.20211130_143652_slthwu/modify_fsurdat.cfg']' returned non-zero exit status 1.
```

I assume that may be coming from some other recent changes you have made?

If you are happy with this, feel free to merge https://github.com/slevisconsulting/ctsm/pull/3 into this branch whenever you're read.

billsacks:
@slevisconsulting - I have made the changes needed to store netCDF files via git LFS - see https://github.com/slevisconsulting/ctsm/pull/4 . If you're happy with that, please merge it into your branch. Please especially read through the README file and make sure it makes sense; if it doesn't make sense, let's work on improving it. (I think I have put that in the right directory - your new directory where you plan to put netcdf files.) Then, once you merge that into your branch, let's go ahead and commit the netcdf files you're using for testing.

python/ctsm/utils.py
```diff
@@ -207,9 +207,8 @@ def _handle_config_value(var, default, item, is_list, convert_to_type,
     if allowed_values is not None:
         for v in var:
             if v not in allowed_values:
-                print('v in var not in allowed_values', v)
-                errmsg = "{} is not an allowed value for {} in .cfg file\n" \
-                         "Allowed values: {}".format(v, item, allowed_values)
+                print('v = ', v, ' in var not in allowed_values')
+                errmsg = '{} is not an allowed value for {} in .cfg file. Check allowed_values'.format(v, item)
```
slevisconsulting, line 211: Changed this because I could not get the unit test to work when `allowed_values` was part of `errmsg`. I hope you consider this fine, too, and I'm also open if not.

slevisconsulting:
@billsacks I will send you a calendar invite for a time to finalize or discuss how to finalize this PR. Pls feel free to reschedule if I pick a bad time.

billsacks:
Thanks. Please let me know when you are done making changes and I will look back at this.

slevisconsulting:
I have stopped making changes.

slevisconsulting:
@billsacks 
prompted by today's email thread titled
_Meeting tomorrow at 10am to talk about user-mods for subset data script_
I have successfully tested the `fsurdat_modifier` on izumi by running `./run_ctsm_py_tests` and also by running the tool directly `./fsurdat_modifier modify_slevis.cfg` as an additional sanity check.

billsacks (changes requested):
Thank you very much for all of your continued work on this. You have addressed all of my previous comments really well. And for the most part I am quite happy with all of the new code you have added (mainly the unit test code). I really appreciate all of your work on that! In particular, your tests of `_handle_config_value` are great (though see a few small comments inline). And, while I haven't looked at them carefully, at a glance the tests of `_get_not_rectangle` also seem really good – thank you for adding all of these tests!

In addition to my inline comments on your new test code, I can think of the following things that still need to be done before this PR is finalized:

- [ ] Need to add a new test to the aux_clm test list leveraging the new system test
- [ ] Need to commit the files in testinputs (and I'll confirm that they're committed via lfs; can see this when viewing the file on github)
- [ ] Need to get pylint passing

python/ctsm/test/test_unit_fsurdat_modifier.py
```diff
@@ -0,0 +1,471 @@
...
+        can_be_unset = False
+        allowed_values = None
+        errmsg = 'More than 1 element found for .cfg file variable: {}'.format(item)
+
+        with self.assertRaisesRegex(SystemExit, errmsg):
+            val = _handle_config_value(var=val, default=default, item=item,
+                is_list=is_list, convert_to_type=convert_to_type,
+                can_be_unset=can_be_unset, allowed_values=allowed_values)
+
+    def test_setvarLev(self):
```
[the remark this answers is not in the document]
slevisconsulting: @billsacks I think that I have addressed this based on our meeting yesterday. Let me know if you see issues. Thank you for all the help.

slevisconsulting:
Confirmed:
The regional FSURDATMODIFYCTSM test
`FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel`
points to the right fsurdat and this fsurdat was modified as expected.

Confirmed for a global case:
`FSURDATMODIFYCTSM_D_Ld1.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_intel`
Not adding the global test to aux_clm.

python/ctsm/test/test_unit_fsurdat_modifier.py
```diff
@@ -0,0 +1,471 @@
...
+        comp_lev0[lat_1-min_lat:lat_2-min_lat+1,
+                  lon_1-min_lon:lon_2-min_lon+1] = val_for_rectangle
+        comp_lev1[...,lat_1-min_lat:lat_2-min_lat+1,
+                    lon_1-min_lon:lon_2-min_lon+1] = val_for_rectangle
+        comp_lev2[...,lat_1-min_lat:lat_2-min_lat+1,
+                      lon_1-min_lon:lon_2-min_lon+1] = val_for_rectangle
+
+        # test setvar
+        ModifyFsurdat.setvar_lev0(ModifyFsurdat, 'var_lev0', val_for_rectangle)
+        self.assertTrue(ModifyFsurdat.file.var_lev0.equals(comp_lev0))
```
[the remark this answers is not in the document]
slevisconsulting: I found an alternative, and let me know if you like it less.

python/ctsm/test/test_unit_fsurdat_modifier.py
```diff
@@ -0,0 +1,471 @@
...
+            longxy=longxy, latixy=latixy)
+        compare = np.ones((rows,cols))
+        # assert this to confirm intuitive understanding of these matrices
+        self.assertEqual(np.size(not_rectangle), np.size(compare))
+
+        # Hardwire where I expect not_rectangle to be False (0)
+        # I have chosen the lon/lat ranges to match their corresponding index
+        # values to keep this simple
+        compare[lat_1-min_lat:lat_2-min_lat+1, lon_1-min_lon:lon_2-min_lon+1] = 0
+        self.assertIsNone(np.testing.assert_array_equal(not_rectangle, compare))
```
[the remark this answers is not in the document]
slevisconsulting: I replaced with an alternative, and LMK if you don't like it.

slevisconsulting:
@billsacks 
Next iteration; should be close to done at this point. I have stopped making changes.
I will submit the aux_clm test-suite on cheyenne.

billsacks:
I noticed that "make lint" wasn't checking the new files. The problem was that there was no `__init__.py` file in the `modify_fsurdat` directory. I will add that (and fix the one outstanding pylint issue).

billsacks (approved):

Issue #1198, opened 2020-10-29
Single point simulations


wwieder:
We are discussing deprecating the PTCLM toolchain (as the toolchain is being remade).  This will be replaced with the faster scripts that pull atmospheric forcing, surface dataset and domain file with python scripts from @swensosc (currently in tools/contrib) that  can be written over / modified as  needed.

Are there features / functions that we'll lose by doing this?   Specifically, @olyson how do you set up single point urban simulations?
If / when we go down this route we'll need to:
- [x] modify singlept.py script, #1352 
  - update singlept.py script to python3  (currently python2)
  - develop testing to  bring this out of /tools/contrib?
- [x] Create new modify_singlept_NEON script that reads in NEON observations, #1353 
   - modify dominant PFTs & soil properties
   - eventually modify parameter file (foliar C:N, allocation, others?)
- [x] Batch process multiple sites, #1354 
  - communicate with PLUMBER2 & NEON for distributing tower forcing data with CTSM
  - improve @olyson .csh script that can simulate multiple sites at once. This may need  to be done more generically to run  in  the cloud.
- [x] Provide jupyter notebooks for sample analyses, #1355  
- [ ] Update users  guide to reflect this new workflow, #1494

*This can be addressed by the SE supported by  NCAR-NEON funding*

swensosc:
my understanding was that the reason ptclm takes as long as it does is due
to its use of mapping files.  Another way forward would be to write a
script that would go to the raw data files and just grab the values at the
closest point.  This would be a little work, but it would preclude the need
to point to an existing surface data file.  It would also represent our
'best guess' for the site, rather than getting values that have been
averaged over a much larger area.

olyson:
For urban sites, I generally just start with an existing surface dataset/domain file and edit using NCL.

billsacks:
> my understanding was that the reason ptclm takes as long as it does is due to its use of mapping files. Another way forward would be to write a script that would go to the raw data files and just grab the values at the closest point. This would be a little work, but it would preclude the need to point to an existing surface data file. It would also represent our 'best guess' for the site, rather than getting values that have been averaged over a much larger area.

[...]

PR #1467, opened 2021-08-23
improve ui for neon script

### LINKED ISSUE #1550: In perl code replace glob with bsd_glob
Reported by @jedwards4b (https://github.com/ESCOMP/CTSM/issues/585#issuecomment-969002883)

errput: File::Glob::glob() will disappear in perl 5.30. Use File::Glob::bsd_glob() instead. at /work/02503/edwardsj/cesm2_x.ihesp/components/clm/bld/CLMBuildNamelist.pm line 4193.

I'm not sure if we can simply replace glob with bsd_glob or if the interface is different. We should also confirm that bsd_glob is available in older perl versions so that this will work for both older & newer versions.

### LINKED ISSUE #1563: increase skip_steps for balance checks by one to permit new CAM physics re-ordering
BalanceChecks are only applied at nstep=3 or later, since the balance checks don't pass for nstep=0,1,2. This is controlled by the ```skip_steps``` variable in BalanceCheckMod.F90:
https://github.com/ESCOMP/CTSM/blob/679f1eba01a27d03536a3e924484f12f83ca1622/src/biogeophys/BalanceCheckMod.F90#L85

The importance of nstep=3 is that this is when the first non-initialization phase radiation is passed from CAM to the surface components. CAM is rearranging it's physics time-loop (see https://github.com/ESCOMP/CAM/pull/387), such that the first non-initialization phase radiation will be sent one time-step later, at nstep=4. Hence, we need to increase the ```skip_steps``` variable from 2->3, or else the BalanceChecks will end the run at nstep=3. (It's possible I'm wrong on the details; this is what I am able to infer from lots of debugging ... and the fact remains that the new physics ordering only works when I increase ```skip_steps``` to 3).

An aside. If someone can explain why the balance checks fail for the first few time-steps, that would be good to know. Specifically, the balance checks fail when this ```errsol``` number is large:
https://github.com/ESCOMP/CTSM/blob/679f1eba01a27d03536a3e924484f12f83ca1622/src/biogeophys/BalanceCheckMod.F90#L884-L885

Which basically says the ```fsa+fsr``` (absorbed + reflected radiation, which is computed earlier in clm time-loop via SurfaceRadiationMod) is very different from the sum of the incident radiation from the ```atm2lnd_inst``` object/array.

@cacraigucar @mvertens (<- whom I know is on vacation, but is also interested in understanding this issue further)

### LINKED ISSUE #1574: Trouble with 1x1_brazil for mksurfdata_map because of negative longitude in SCRIP grid file
### Brief summary of bug

This is in ctsm5.1.dev066, but would have been a problem for well before that.

### General bug information

**CTSM version you are using:** ctsm5.1.dev066

**Configurations affected:** mksurfdata_map for 1x1_brazil

### Details of bug

The SCRIP grid file for 1x1_brazil has a negative longitude which causes it to fail. It should work with the longitude converted to 0-360 rather than -180-180. I didn't see other grid files with this same problem.

mksurfdata.pl dies as follows. Note, also that mksurfdata.pl seems to go on without recognizing that an error was thrown.

>  mksrf_gridtype = regional
>  Output ALL data in file as 64-bit
>  Set wetland to 0% over land
> In mkpftMod::mkpftInit()...
>  calling domain_read
>  finished domain_read
>  fsurdat is 2d lat/lon grid
>  nlon=            1  nlat=            1
>   Output domain must be on a 0 to 360 longitude grid rather than a -180 to 180 g
>  rid as it is required for CESM
> 
> ===========================================
> 
> Successfully created fsurdat files

jedwards4b:
### Description of changes
Redo options list to remove positional arguments that were difficult to input correctly.
Transient runs now use run_type startup and get finidat from s3 server unless --run-from-postad option is used (or finidat is not available).   Use mpi instead of mpi-serial, this mod was recommended for container use.   Add a new script neon_finidat_upload which allows authorized users to upload finidat files to the s3 server.  

### Specific notes

  Fixes #1563 
  Fixes #1550
  Fixes #1574

Any User Interface Changes (namelist or namelist defaults changes)? Yes for NEON

Testing performed, if any: regular testing and tools testing will be done
  cheyenne: -- PASS
  izumi: -- OK


tools/site_and_regional/run_neon.py
```diff
@@ -150,99 +138,52 @@ def get_parser(args, description, valid_neon_sites):
...
-    #            action="store_true",
-    #            dest="sasu_flag",
-    #            required = False,
-    #            default = False)
+                type = datetime.date.fromisoformat,
+                default = datetime.datetime.strptime("2020-01-01",'%Y-%m-%d'))
+
+    parser.add_argument('--run-from-postad',
+                        help='''
+                        For transient runs only - should we start from the postad spinup or finidat?
```
wwieder, line 182: I'm not sure what this note is asking, but a `run-from-postad` the initial conditions should be from the postad simulation, as opposed to `run-from-latest` which would run from the last restart file of a transient run that's available.

tools/site_and_regional/run_neon.py
```diff
@@ -150,99 +138,52 @@ def get_parser(args, description, valid_neon_sites):
...
-    #            help='''
-    #            Matrix (SASU) spin-up
-    #            [default: %(default)s]
-    #            ''', 
-    #            action="store_true",
-    #            dest="sasu_flag",
-    #            required = False,
-    #            default = False)
+                type = datetime.date.fromisoformat,
+                default = datetime.datetime.strptime("2020-01-01",'%Y-%m-%d'))
```
wwieder, line 178: I guess I'd assume that the end date for ad spinup and postad should default to 2021-01-01, unless there are missing data in 2020?  Will this work with leap years correctly?

tools/site_and_regional/run_neon.py
```diff
@@ -254,20 +195,45 @@ def get_parser(args, description, valid_neon_sites):
...
+def get_isosplit(s, split):
+    if split in s:
+        n, s = s.split(split)
+    else:
+        n = 0
+    return n, s
+
+def parse_isoduration(s):
+    '''
+    simple ISO 8601 duration parser, does not account for leap years and assumes 30 day months
```
wwieder, line 224: what is this code being used to do?  I'm assuming it's not used for transient runs?

tools/site_and_regional/run_neon.py
```diff
@@ -368,18 +333,28 @@ def run_case(self, base_case_root, run_type, run_length, overwrite=False):
...
         if not os.path.isdir(case_root):
             # read_only = False should not be required here
             with Case(base_case_root, read_only=False) as basecase:
                 logger.info("---- cloning the base case in {}".format(case_root))
                 basecase.create_clone(case_root, keepexe=True, user_mods_dirs=user_mods_dirs)
 
-
         with Case(case_root, read_only=False) as case:
-            case.set_value("STOP_OPTION", "nyears")
+            case.set_value("STOP_OPTION", "ndays")
```
wwieder, line 357: do we want this as nyears for 'official' runs?

tools/site_and_regional/run_neon.py
```diff
@@ -368,18 +333,28 @@ def run_case(self, base_case_root, run_type, run_length, overwrite=False):
...
             case.set_value("STOP_N", run_length)
             case.set_value("REST_N", 100)
```
wwieder, line 359: This seems OK for ad runs, but for postad runs that are 50 years do we want REST_N = run_length?  It doesn't look like this is a runtype == "postad" loop, but I can't tell

tools/site_and_regional/run_neon.py
```diff
@@ -389,19 +364,25 @@ def run_case(self, base_case_root, run_type, run_length, overwrite=False):
...
+                case.set_value("JOB_WALLCLOCK_TIME", "12:00:00", subgroup="case.run")
             else:
                 case.set_value("CLM_FORCE_COLDSTART","off")
                 case.set_value("CLM_ACCELERATED_SPINUP","off")
                 case.set_value("RUN_TYPE", "hybrid")
                 
             if run_type == "postad":
                 self.set_ref_case(case)
+                case.set_value("REST_OPTION","nmonths")
+                case.set_value("REST_N", 12)
```
wwieder, line 376: see comment above, but we just need one restart file, that's written out at the end of postad.

tools/site_and_regional/run_neon.py
```diff
@@ -389,19 +364,25 @@ def run_case(self, base_case_root, run_type, run_length, overwrite=False):
...
             if run_type == "transient":
-                self.set_ref_case(case)
+                if self.finidat:
+                    case.set_value("RUN_TYPE","startup")
+                else:
+                    self.set_ref_case(case)
                 case.set_value("STOP_OPTION","nmonths")
                 case.set_value("STOP_N", self.diff_month())
-                case.set_value("REST_N", "12")
+                case.set_value("REST_OPTION", "end")
```
wwieder, line 385: does end need to be defined, or does this just take us through the last datm file that's found?

tools/site_and_regional/run_neon.py
jedwards4b: Right - I've changed this to 2021-01-01.  The model doesn't know about leap years.

tools/site_and_regional/run_neon.py
jedwards4b: By default the transient run uses the finidat file found on the s3 server.  But if you want to redo the spin-up and rerun from the postad you would use this flag to trigger that.

tools/site_and_regional/run_neon.py
jedwards4b: This is used to set the run length.  I don't think that it will be a very commonly used option, but it accept options in the ISO8601 format, so for example 3Y (3 years) or 3Y4M (3 years 4 months), or 7D (7 days).

tools/site_and_regional/run_neon.py
jedwards4b: The problem with nyears is understanding leap years.   This is still defaulting to 200 years for ad and 50 for postad just expressed in terms of days.

tools/site_and_regional/run_neon.py
jedwards4b: We can change this to REST_OPTION=end to just get restarts at the end of the run.

tools/site_and_regional/run_neon.py
jedwards4b: I'll change to REST_OPTION='end'

tools/site_and_regional/run_neon.py
jedwards4b: end is the end of the run, it's defined in terms of STOP_N and STOP_OPTION.
This is set for transient to the end of the last datm file found.

tools/site_and_regional/run_neon.py
wwieder: maybe some a comment on the code to this effect would help clarify

tools/site_and_regional/run_neon.py
wwieder: OK, if you run transient case from 2018-2020 will there be 365 days / year (with 366 in leap years)?

tools/site_and_regional/run_neon.py
wwieder: again, maybe a comment to this effect would be helpful. I'm thinking of users who may be making code modifications and having to run a whole new spinup and transient case.

wwieder:
Jim, this looks good.  The only thing that may be helpful is some comments in the code, help functions, or a README that explains what the workflow and how users should use this script.

negin513:
Thanks @jedwards4b. The changes look great. 
Overall, I think we need to modify the instructions on the top of the code to clarify its use for different cases. 

This will be part of the README for our future use.

tools/site_and_regional/run_neon.py
```diff
@@ -254,20 +195,45 @@ def get_parser(args, description, valid_neon_sites):
...
+    if "CIME_OUTPUT_ROOT" in args.output_root:
+        args.output_root = None
+
+    if args.run_length == '0Y':
+        if args.run_type == 'ad':
+            run_length = '200Y'
+        elif args.run_type == 'postad':
+            run_length = '50Y'
+        else:
+            run_length = '4Y'
```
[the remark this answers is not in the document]
wwieder: oh, maybe not,  but can't users set this?  I think it's set to `end` below?

negin513:
When I am running the code, I receive the following error without any meaningful help:
I think we should set the defaults somehow so the code works for some default cases.
@ekluzek might have a better insight on this but it seems like the CTSM SE team like the codes to work with some defaults. 
```
./run_neon.py
  File "./run_neon.py", line 607, in <module>
    main(__doc__) 
  File "./run_neon.py", line 577, in main
    site_list, output_root, run_type, overwrite, run_length, base_case_root, run_from_postad = get_parser(sys.argv, description, valid_neon_sites)
  File "./run_neon.py", line 213, in get_parser
    return neon_sites, args.output_root, args.run_type, args.overwrite, run_length, base_case_root, args.run_from_postad
UnboundLocalError: local variable 'base_case_root' referenced before assignment
```

Again, I think this looks great.

tools/site_and_regional/run_neon.py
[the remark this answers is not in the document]
jedwards4b: The transient case default is computed in the datm buildnml and is set to the length of the available data.   This 4Y is just a placeholder.  I can add documentation to clarify.

tools/site_and_regional/run_neon.py
```diff
@@ -368,18 +333,28 @@ def run_case(self, base_case_root, run_type, run_length, overwrite=False):
...
             # read_only = False should not be required here
             with Case(base_case_root, read_only=False) as basecase:
                 logger.info("---- cloning the base case in {}".format(case_root))
                 basecase.create_clone(case_root, keepexe=True, user_mods_dirs=user_mods_dirs)
 
-
         with Case(case_root, read_only=False) as case:
-            case.set_value("STOP_OPTION", "nyears")
+            case.set_value("STOP_OPTION", "ndays")
             case.set_value("STOP_N", run_length)
```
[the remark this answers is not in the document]
jedwards4b: Whatever the user specifies is converted to days in the code.   I can add more documentation regarding this.

tools/site_and_regional/run_neon.py
```diff
@@ -150,99 +138,52 @@ def get_parser(args, description, valid_neon_sites):
...
                 [default: %(default)s]
                 ''',               
                 action="store", 
-                dest="start_year",                                                                                                                                                 
+                dest="start_date",                                                                                     
                 required = False,
-                type = int,        
-                default = 2018)
+                type = datetime.date.fromisoformat,
+                default = datetime.datetime.strptime("2018-01-01",'%Y-%m-%d'))
```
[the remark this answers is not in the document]
jedwards4b: For ad and postad cases the default is updated below.

cime_config/usermods_dirs/NEON/defaults/user_nl_clm
```diff
@@ -19,7 +19,7 @@
 !----------------------------------------------------------------------------------
 
 flanduse_timeseries = ' '   ! This isn't needed for a non transient case, but will be once we start using transient compsets
-fsurdat = "$DIN_LOC_ROOT/lnd/clm2/surfdata_map/NEON/surfdata_hist_78pfts_CMIP6_simyr2000_${NEONSITE}_c210720.nc"
+fsurdat = "$DIN_LOC_ROOT/lnd/clm2/surfdata_map/NEON/surfdata_hist_78pfts_CMIP6_simyr2000_${NEONSITE}_c210805.nc"
 model_year_align_urbantv = 2018
 stream_year_first_urbantv = 2018
 stream_year_last_urbantv = 2019
```
wwieder, line 25: Don't we want ndep and popdens to update to the last year of the simulation (or at least the spinup)?  this would be 2020 for spinup and 2021 for transient cases.  How do we keep updating this moving forward?

ekluzek:
@wwieder added some notes about leap years. The model CAN run with leap years when flipped to GREGORIAN calendar mode. Just need to

`./xmlchange CALENDAR=GREGORIAN`

This could be added as a default setting for NEON sites. I actually thought we already were doing that, but it looks like we aren't yet. Should that be done?

wwieder:
- For transient simulations I think we should use `./xmlchange CALENDAR=GREGORIAN`, this is critical for aligning results with the visualization script.  
- For spinup let's keep using `noleap`, or what ever @jedwards4b is doing to ignore leap days and facilitate this now.

bld/CLMBuildNamelist.pm
```diff
@@ -4254,7 +4254,7 @@ sub check_input_files {
                 my $pathname = $nl->get_variable_value($group, $var);
                 # Need to strip the quotes
                 $pathname =~ s/['"]//g;
-
+                next if ($pathname =~ /UNSET$/);
```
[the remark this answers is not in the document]
ekluzek: Yes, this is a good start. I'd also like it to not show up in the namelist, as having a file set to "$DIN_LOC_DIR/UNSET" is confusing and looks wrong. I'll likely work on this in addition to what you are doing here.

ekluzek:
This updates externals, and gets to a version where nuopc is the default driver. As such the following tests have issues:

> ================================================================================
> These tests are pending (some tests may fail in the pending state)
> ================================================================================
> SMS_Ld2_D.f09_g17.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic_interp		
> SMS_Lm1.f10_f10_mg37.I1850Clm50BgcCropCmip6waccm.cheyenne_gnu.clm-basic		
> SMS_Lm1.f19_g17.I1850Clm50BgcCropCmip6waccm.cheyenne_intel.clm-basic		
> ================================================================================
> These tests failed
> ================================================================================
> DAE_N2_D_Lh12.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_intel.clm-DA_multidrv		
> DAE_N2_D_Lh12_Vnuopc.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_intel.clm-DA_multidrv		EXPECTED
> ERI_N2_Ld9.f19_g17.I2000Clm51BgcCrop.cheyenne_intel.clm-default		
> ERS_Ly3.f10_f10_mg37.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic		
> SMS_D_Lm1_Mmpi-serial.CLM_USRDAT.I1PtClm50SpRs.cheyenne_intel.clm-USUMB_mct		
> SMS_N2_D_Lh12.f09_g17.I2000Clm50Sp.cheyenne_intel.clm-pauseResume

mvertens:
@erik - my new PR to CMEPS will fix all the tests that have `clm-basic` in the test name. That should cover about half of these failed tests.

ekluzek:
@mvertens yes I'm thinking that your PR and the update to make nuopc the default should come in before this one, because of that.

bld/CLMBuildNamelist.pm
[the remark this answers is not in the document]
ekluzek: I also did work in 017a7ed1ed to prevent blank files being listed. I think @jedwards4b is also a good protection although I think it really should trigger an error because something is screwy if you have a namelist item set to UNSET. 

The comment that generated this was...

https://github.com/ESCOMP/CTSM/issues/1198#issuecomment-833615077

cime_config/usermods_dirs/NEON/GUAN/shell_commands
```diff
@@ -1,3 +1,4 @@
 ./xmlchange NEONSITE=GUAN
 ./xmlchange PTS_LON=293.13112
 ./xmlchange PTS_LAT=17.96882
+./xmlchange RUN_STARTDATE=2018-06-01
```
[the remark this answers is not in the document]
ekluzek: @wwieder It will just start on June/1st and use that as it's reference date. This means for purposes of spinup dynamics June/1st is the date when spinup variables are updated. It's sort of like making June/1st the start of your fiscal year. Somethings are still tied to the calendar year (like dynamic land use change), but anything that is measuring time since start of run or doing running averages will use June/1st as the marker. Does that help?

cime_config/usermods_dirs/NEON/defaults/user_nl_clm
ekluzek: @wwieder these don't get autoupdated by any means. Right now the only way to change it is to keep changing the last year in this file (or in each case run). The last year update could also be handedl in run_neon.py though. I don't know that's any better though.

ekluzek (approved):

wwieder:
> Jim, this looks good. The only thing that may be helpful is some comments in the code, help functions, or a README that explains what the workflow and how users should use this script.

cime_config/usermods_dirs/NEON/defaults/user_nl_clm
wwieder: I think this is fine for now, but as we move to making more of the real-time hindcasts they should be updated for new years of data?  Updating urban likely ins't too important and for short runs the deposition files shouldn't changethat much?

cime_config/usermods_dirs/NEON/GUAN/shell_commands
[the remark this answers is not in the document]
wwieder: ok, but for spinup, when you cycle over years 2018-2020 repeatedly does the model get to May 30 2020 of the last year and then restart on June 1 2018 or does it go all the way until Dec 31 2020 and then restart on June 1 2018?

wwieder (approved):
These changes look good

wwieder:
@ekluzek  I have one outstanding question about how the model handles spinup when data are not available Jan 1 for a particular year, but otherwise I think my questions have been addressed for this PR.

jedwards4b:
@wwieder consider the log file here: /glade/scratch/jedwards/NEON/archive/GUAN.ad/logs/atm.log.290175.chadmin1.ib0.cheyenne.ucar.edu.210901-142537   It starts the spinup at model date 180101 and with ATM data file GUAN_atm_2020-01.nc and spins up using 2019 and 2020 data only.    This is controlled by DATM_YR_START which is set via variable self.start_year in the run_neon.py script using the first year for which JAN 1 data is available.

wwieder:
thanks for clarifying @jedwards4b

wwieder:
sorry, one more on these lines
``` 
model_year_align_urbantv = 2018
stream_year_first_urbantv = 2018
stream_year_last_urbantv = 2019
...
```
Should all of these at least go to 2020 for the spinups, which is what we're doing now?  They'll get out of date for the transient runs, but I'm less worried about this (for now), as it will be a while untile we have many years of data past 2020

ekluzek:
@wwieder do all sites have data that goes through 2020 now?

The would be problems with cases that don't have 2020 data as the case will point out the missing a files. Bit that could be considered a feature not a big...

wwieder:
Yep we're (mostly) spinning all sites up by cycling over 2018-2020 data and then running a full 'transient' from 2018-2021

ekluzek:
OK @wwieder I changed the end year to 2020 in commit 411c321 and also brought in @negin513 update of surface datasets from #1539.

I've run the standard testing at this point. The other thing I'm going to do is run the tools testing and make the ChangeLog and then this tag will be done.

Issue #1479, opened 2021-09-07
Suggested cleanup in CropPhenology(): Removing duplicated sowing code


samsrabin:
[At tag ctsm5.1.dev054 (dfe9d66)]

In the `CropPhenology()` subroutine in `CNPhenologyMod.F90`, the code in the `if` block beginning at line 1925 and the following `else if` (beginning at line 1988) both sow crops. The difference in the conditionals is that the former looks for two temperature conditions (lines 1926–1927) to be satisfied during the sowing window, whereas the latter relaxes that on the last day of the window.

There's only one difference between the code of those blocks. It happens for C4 crops (corn, sugarcane, _Miscanthus_, and switchgrass); i.e.:
```
if (ivt(p) == ntmp_corn .or. ivt(p) == nirrig_tmp_corn .or. &
    ivt(p) == ntrp_corn .or. ivt(p) == nirrig_trp_corn .or. &
    ivt(p) == nsugarcane .or. ivt(p) == nirrig_sugarcane .or. &
    ivt(p) == nmiscanthus .or. ivt(p) == nirrig_miscanthus .or. &
    ivt(p) == nswitchgrass .or. ivt(p) == nirrig_switchgrass) then
```
Here's what happens in the `if` block (lines 1953–1954):
```
gddmaturity(p) = max(950._r8, min(gdd820(p)*0.85_r8, hybgdd(ivt(p))))
gddmaturity(p) = max(950._r8, min(gddmaturity(p)+150._r8, 1850._r8))
```
And here's what happens in the `else if` block (line 2005):
```
gddmaturity(p) = max(950._r8, min(gdd820(p)*0.85_r8, hybgdd(ivt(p))))
```
(I'm assuming this difference in behavior is intentional.) 

This code section would benefit from some consolidation. For instance: have logicals representing the outcomes of the `if` and `else if` conditions, combine those blocks (into, e.g. `if (logical_1 .or. logical_2)`), and then only adding code to distinguish between the behavior for C4 crops:
```
if (ivt(p) == ntmp_corn .or. ivt(p) == nirrig_tmp_corn .or. &
    ivt(p) == ntrp_corn .or. ivt(p) == nirrig_trp_corn .or. &
    ivt(p) == nsugarcane .or. ivt(p) == nirrig_sugarcane .or. &
    ivt(p) == nmiscanthus .or. ivt(p) == nirrig_miscanthus .or. &
[...]

PR #1482, opened 2021-09-08
1479 crop sowing dedupe

samsrabin:
### Description of changes

De-duplicated code in CropPhenology subroutine relating to planting of non-winter-cereal crops.

### Specific notes

CTSM Issues Fixed (include github issue #): #1479 

### DIFF
--- a/src/biogeochem/CNPhenologyMod.F90
+++ b/src/biogeochem/CNPhenologyMod.F90
@@ -1927,7 +1927,6 @@ contains
                     a10tmin(p) > minplanttemp(ivt(p))          .and. &
                     jday       >= minplantjday(ivt(p),h)       .and. &
                     jday       <= maxplantjday(ivt(p),h)       .and. &
-                    t10(p) /= spval .and. a10tmin(p) /= spval  .and. &
                     gdd820(p) /= spval                         .and. &
                     gdd820(p) >= gddmin(ivt(p))) then
 
--- a/src/biogeochem/CNPhenologyMod.F90
+++ b/src/biogeochem/CNPhenologyMod.F90
@@ -1860,7 +1860,7 @@ contains
                   hdidx(p)       = 0._r8
                   vf(p)          = 0._r8
                   
-                  call PlantCrop(p, jday, crop_inst, &
+                  call PlantCrop(p, leafcn(ivt(p)), jday, crop_inst, cnveg_state_inst, &
                                  cnveg_carbonstate_inst, cnveg_nitrogenstate_inst, &
                                  cnveg_carbonflux_inst, cnveg_nitrogenflux_inst, &
                                  c13_cnveg_carbonstate_inst, c14_cnveg_carbonstate_inst)
@@ -1891,7 +1891,7 @@ contains
 
                if (do_plant_normal .or. do_plant_lastchance) then
 
-                  call PlantCrop(p, jday, crop_inst, &
+                  call PlantCrop(p, leafcn(ivt(p)), jday, crop_inst, cnveg_state_inst, &
                                  cnveg_carbonstate_inst, cnveg_nitrogenstate_inst, &
                                  cnveg_carbonflux_inst, cnveg_nitrogenflux_inst, &
                                  c13_cnveg_carbonstate_inst, c14_cnveg_carbonstate_inst)
@@ -2209,8 +2209,9 @@ contains
   end subroutine CropPhenologyInit
 
     !-----------------------------------------------------------------------
-  subroutine PlantCrop(p, jday, &
-       crop_inst, cnveg_carbonstate_inst, cnveg_nitrogenstate_inst, &
+  subroutine PlantCrop(p, leafcn_in, jday, &
+       crop_inst, cnveg_state_inst,                                 &
+       cnveg_carbonstate_inst, cnveg_nitrogenstate_inst,            &
        cnveg_carbonflux_inst, cnveg_nitrogenflux_inst,              &
        c13_cnveg_carbonstate_inst, c14_cnveg_carbonstate_inst)
     !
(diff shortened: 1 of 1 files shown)


samsrabin:
Also need to de-dupe winter cereals block.

ekluzek:
Note that we don't run winter cereals so I'm not sure how important it is to do. But in principle removing duplication is always helpful.

dlawrenncar:
@ekluzek Note that a goal of @samsrabin is to get the winter cereals (winter wheat) going again.

billsacks (changes requested):
This looks great. Thanks so much for submitting this nice cleanup. I'm happy with your choice of logical names. Once you make a similar change for the winter cereals, we can test & merge this. (I'm marking this as "request changes" just to note the request for that additional change: the changes you have already made look great to me.)

ekluzek (approved):
This is pretty simple and a nice removal of duplication.

samsrabin:
I've pushed two new commits. The first is a very simple change to the winter cereals block, along the lines of what I did for other crops yesterday. The second is bigger: A new subroutine that performs planting-day operations used in both winter cereals and other crops. I'm not sure whether what I wrote even compiles, let alone results in bit-for-bit identical changes. (I'm still getting my feet under me with an editing-testing workflow.) Feel free to edit or revert as you feel necessary.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -1839,104 +1841,62 @@ subroutine CropPhenology(num_pcropp, filter_pcropp                     , &
...
-                     endif
-                  endif
-
-                  ! latest possible date to plant winter cereal and after all other 
-                  ! crops were harvested for that year
-
-               else if (jday       >=  maxplantjday(ivt(p),h) .and. &
-                    gdd020(p)  /= spval                   .and. &
-                    gdd020(p)  >= gddmin(ivt(p))) then
+               ! srabin 2021-09-09. (No new behavior added; just allows for de-duplication of code.)
```
billsacks, line 1845: - [ ] Here and one other similar place: Please remove this comment with your name, date and note about the change you made. Although there are some instances of comments like this in the code, noting the history of changes, we try to avoid them now. (The next line's comment is fine and good. The distinction is whether the comment is an explanatory comment or one that is just describing a change; for the latter, we try to put the comment in a commit message but not in the code.)

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -1839,104 +1841,62 @@ subroutine CropPhenology(num_pcropp, filter_pcropp                     , &
...
 
                   cumvd(p)       = 0._r8
                   hdidx(p)       = 0._r8
                   vf(p)          = 0._r8
-                  croplive(p)    = .true.
-                  cropplant(p)   = .true.
-                  idop(p)        = jday
-                  harvdate(p)    = NOT_Harvested
+                  
+                  PlantCrop(p, jday, crop_inst, &
```
billsacks, line 1864: This needs to be `call PlantCrop` (and similarly for the other call)

billsacks (changes requested):
Very nice, thank you! This is some very helpful cleanup. I really appreciate that you went the next step of extracting some more common code into a shared subroutine.

I have a couple of small comments, one cosmetic and one needed for it to compile. (I haven't tried compiling or running this yet myself, but just noticed this change that's needed.) How would you like to move forward at this point? Would you like me to suggest a test case that you can run to ensure it compiles and is bit-for-bit the same as baselines, or would you like to hand this over at this point so you can get back to other stuff? Either way is fine with me – it just depends how much you would like to use this to learn some of the testing techniques.

samsrabin:
Okay, I've made those changes—thanks!

I think this would indeed be a great chance for me to learn how test cases work. I'm not yet set up to run anything yet, but I should be by early next week. So if you want to go ahead and suggest a test case, I'll give it a shot when I'm ready. I'm not in any rush to get this all merged; I have other stuff I can work on in the meantime.

billsacks (approved):

billsacks:
Thanks for these edits, @samsrabin . This looks good to me now. For these changes, I think the clm_short test list would be the right thing to run. This runs 5 short t

For these changes, I would suggest running the following tests:
```
ERP_D_P36x2_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_intel.clm-default
ERP_Ly3_P72x2.f10_f10_mg37.IHistClm50BgcCrop.cheyenne_intel.clm-cropMonthOutput
```

The first one is a 3-day debug-mode test, and the second one is a 3-year non-debug test. Both of them test a number of things, including exact restart and bit-for-bit results when changing processor count and threading.

The easiest way to run these is:

(1) If you haven't already done so, specify a default account number to use for cheyenne jobs. There are a few ways to do this; one is to set the PROJECT environment variable in your shell startup file (e.g., `.bashrc`).

(2) Clone your branch on cheyenne and run `./manage_externals/checkout_externals` to get all of the required externals.

(3) Create a text file (e.g., named `testlist`, as is assumed in the command in step (4)) in the top-level ctsm directory containing the above two test names, one per line.

(4) Run the following command from the top level of your CTSM clone:
```
./run_sys_tests -f testlist --compare ctsm5.1.dev054 --skip-generate
```

In addition to running the above tests, that command will compare the test results with the appropriate baselines to ensure that your changes are bit-for-bit with the baselines.

For more information on running system tests like this, including how to interpret the test results, see https://github.com/ESCOMP/CTSM/wiki/System-Testing-Guide . However, I know that walking through this live (in person or by zoom) can be a lot easier than trying to wade through documentation. So please let me know if you'd like to schedule a time to do this together.

Once you have done this, I can run final testing and bring this to master.

samsrabin:
With my latest commit, those tests completed successfully. As discussed, I'll leave this in your hands :-)

billsacks:
I ran the full test suite on this last night; all tests passed and are bit-for-bit. I will bring this to master shortly. Thanks again for this contribution @samsrabin !

PR #1495, opened 2021-09-21
Changes needed to get aux history output with cmeps

### LINKED ISSUE #1238: In nuopc cap: change logic for when glc fields are sent to facilitate writing cpl aux hist files
The current logic in the nuopc cap only sends lnd -> glc fields if glc is present and evolving. However, a common use case for the LIWG is that they want the lnd -> glc fields from runs with a non-evolving CISM or no CISM at all (i.e., a SGLC) so that the model will write cpl aux hist files that can be used to force a later T compset run with a prognostic CISM forced by dlnd.

I think the ideal thing to do would be to have CTSM also send the lnd -> glc fields if the namelist flag controlling the relevant aux hist file (`histaux_l2x1yrg` in the mct driver) is set to true. If that's challenging, we may want to consider always sending the lnd -> glc fields, though I realize that could be expensive because these fields all have an elevation class dimension, so there are a lot of fields that are sent.

mvertens:
### Description of changes
Changes needed to get aux history output with cmeps

### Specific notes
These changes are needed to create auxiliary land history files when the new cmeps configuration variable is set to true.
`histaux_l2x1yrg = .true.`
The changes needed in CTSM enable lnd2glc fields to be sent to the mediator even if glc is not present - based on querying the 
`histaux_l2x1yrg  `setting.

CTSM Issues Fixed:
- Resolves #1238 

Testing performed, if any: 
testlist_drv.xml using what will be cmeps0.13.28 with cesm2_3_alpha05d externals on cheyenne with intel.


src/cpl/nuopc/lnd_import_export.F90
```diff
@@ -295,11 +297,19 @@ subroutine advertise_fields(gcomp, flds_scalar_name, glc_present, cism_evolve, r
...
+    call NUOPC_CompAttributeGet(gcomp, name="histaux_l2x1yrg", value=cvalue, &
+         isPresent=isPresent, isSet=isSet, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    if (isPresent .and. isSet) then
+       read(cvalue,*) send_lnd2glc
+    else
+       send_lnd2glc = .false.
+    end if
 
+    if ((send_lnd2glc) .or. (glc_present .and. cism_evolve)) then
```
billsacks, line 309: - [ ] I find it confusing that send_lnd2glc is set entirely based on the value of histaux_l2x1yrg. That means that, if histaux_l2x1yrg is false, then send_lnd2glc will be false even if you're running with an active glc model. This seems error-prone if someone introduces more logic later based on send_lnd2glc. Would it work to have logic like:
  - if (glc_present .and. cism_evolve) then send_lnd2glc = .true.
  - else set lnd_2glc based on the new logic based on histaux_l2g1yrg

- [ ] Also, I see that lnd -> glc fields aren't sent if cism_evolve is false. I can't remember if we discussed this before, but this strikes me as wrong: it seems like lnd2glc fields should be sent whenever glc_present is true, because land ice folks will sometimes set up a non-evolving ice sheet just to get the downscaled lnd2glc fields. Or is there some reason why this won't work right now with cmeps?

src/cpl/nuopc/lnd_import_export.F90
```diff
@@ -1076,22 +1100,25 @@ subroutine state_getimport_2d(state, fldname, ctsmdata, rc)
...
+    use ESMF          , only : ESMF_LOGERR_PASSTHRU, ESMF_END_ABORT, ESMF_LogFoundError
+    use ESMF          , only : ESMF_Finalize
+    use shr_const_mod , only : shr_const_spval
 
     ! input/output variabes
     type(ESMF_State) , intent(in) :: state
     character(len=*) , intent(in) :: fldname
     real(r8)         , intent(in) :: ctsmdata(:)
     logical, optional, intent(in) :: minus
+    logical, optional, intent(in) :: init_spval
```
billsacks, line 1116: - [ ] Could you please add a comment describing when you should and shouldn't set init_spval to true, to provide guidance for anyone adding fields in the future?

- [ ] [OPTIONAL] I would prefer init_spval to be a required argument, so that anyone adding a new field needs to explicitly think about whether it should be true or false. Otherwise I think it is likely to be left as implicitly false in situations where it really should be true.

billsacks (changes requested):
Thanks a lot for your work on this @mvertens . I do have some comments / requests. Let me know if you'd like to discuss this by zoom.

src/cpl/nuopc/lnd_import_export.F90
mvertens: @sacks - good points. I think the right thing to do would be the following:
- if (glc_present) then send_lnd2glc = .true.
- else (glc_present .and. cism_evolve)
I think this takes care of the two concerns you had. Do you agree? In this case if glc_present you will always send the lnd2glc fields otherwise you simply queyr the histaux_l2g1yrg

billsacks: I'm not sure what you mean by the condition on the else (`else (glc_present .and. cism_evolve)`): is that a mistake? From your description in the last sentence, I agree.

src/cpl/nuopc/lnd_import_export.F90
billsacks: Sounds good; thanks. So I guess the long-term plan would be to not bother having this argument at all, but instead implicitly assuming it is true for all fields. In that case, I don't have strong feelings about whether this is left as an optional argument or turned into a required argument for now (it all depends on how long we expect it to be before we're able to achieve the long-term plan).

mvertens: @billsacks - I've incorporated the changes you suggested and am currently testing them. Please hold of on more testing until I've done that.

billsacks: Thanks, this looks good now, with the one remaining thing adding a brief comment about init_spval (though I'm okay skipping that if you don't feel you have a good enough understanding of the issue yet to add a meaningful comment). Once you give me the go ahead, I can do final testing.

billsacks:
@mvertens testing looks good here. There are FILLDIFFs for many of the lnd -> mediator fields in nuopc tests. I think this is expected based on the changes you made, but I wanted to see if you wanted to take a quick look to make sure this looks as you expect. I have spot-checked a couple of tests, and it looks okay to me: most fields now have fill value over ocean, whereas they used to have some value (0 I think) over ocean. I think this is what you intended, but I wanted to double check.

If you want to take a quick look, the cheyenne tests are in /glade/scratch/sacks/tests_1018-161929ch and the izumi tests are in /scratch/cluster/sacks/tests_1018-162018iz . This cheyenne test is probably as good as any to check: /glade/scratch/sacks/tests_1018-161929ch/ERS_Ly3_P72x2_Vnuopc.f10_f10_mg37.IHistClm50BgcCropG.cheyenne_intel.clm-cropMonthOutput.GC.1018-161929ch_int or if you want a higher resolution test (but without ice sheet coupling): /glade/scratch/sacks/tests_1018-161929ch/ERP_D_Ld3_Vnuopc.f09_g17.I2000Clm50Sp.cheyenne_intel.clm-prescribed.GC.1018-161929ch_int

mvertens:
@sacks - thanks for carrying out the tests. The point was to have fill value over ocean over most fields. I am fine with proceeding with the merge.

PR #1497, opened 2021-09-23
Miscellaneous small answer changes

### LINKED ISSUE #1356: Remove the CN (non CENTURY) version of Carbon-Nitrogen soil biogeochemistry code and it's testing
Remove the option to run the non-CENTURY and vertically resolved BGC code. This code isn't really being used, validated, or kept up to date. None of the current supported versions of the physics use it the default (clm4_5, clm5_0, nor clm5_1). There are also concerns for its scientific validity especially as compared to the full BGC version. The only advantage is it's simpler and faster, but because it isn't and hasn't been tuned in any of the current model physics versions it likely is just adding complexity.

This could be as simple as removing the following file from soilbiogeochemistry:

SoilBiogeochemDecompCascadeCNMod.F90

as well as the "CN" compsets and tests:

I2000Clm50Cn
I1850Clm45Cn

ERP_D_Ld9" grid="f19_g17" compset="I2000Clm50Cn" testmods="clm/drydepnomegan
ERP_P36x2_D_Ld5" grid="f10_f10_mg37" compset="I1850Clm45Cn" testmods="clm/default
ERP_P36x2_D_Ld5" grid="f10_f10_mg37" compset="I2000Clm50Cn" testmods="clm/default
SMS" grid="f19_g17" compset="I2000Clm50Cn" testmods="clm/default
SMS_D_Ld1" grid="f19_g17" compset="I1850Clm45Cn" testmods="clm/default
SMS_P48x1_D_Ld5" grid="f10_f10_mg37" compset="I2000Clm50Cn" testmods="clm/default

Doing that would be quite straightforward.

### LINKED ISSUE #1394: Max partial pressure of CO2 be should be 1.e-6 rather than 10.e-6 which it is for ci_func_PHS
In looking at the changes for Medlyn photosynthesis I noticed a difference between the PHS code and the non-PHS code for calculation of "cs". 

In ci_func_PHS for PHS you have the following for cs_sun (and the same for cs_sha)

``` fortran
       cs_sun = cair - 1.4_r8/gb_mol * an_sun(p,iv) * forc_pbot(c)
       cs_sun = max(cs_sun,10.e-06_r8)
```

While for ci_func for non-PHS you have this...

``` fortran
      cs = cair - 1.4_r8/gb_mol * an(p,iv) * forc_pbot(c)
      cs = max(cs,1.e-06_r8)
```

Note, also that 1.e-6_r8 is used for the copy of the same code in PhotosynthesisHydraulicStress. This is also true of the non-PHS main Photosynthesis subroutine. Which makes me think that 1.e-06 is the correct one to use.

So the constant is 10.e-6 for PHS (in ci_func_PHS) while it's 1.e-6 for non-PHS. Is this intentional? Or should they both be the same? And if so which one?

This is the nature of having code that should be identical, but copied rather than used in one place. It's too easy for things to be slightly modified even unintentionally.

I'm assuming this will make a small difference in the model, but likely hard to tell the difference in climate, but likely larger than roundoff.

@djk2120

### LINKED ISSUE #1460: Initialize t_soisno to 272K instead of 274K to avoid deep soil carbon in permafrost regions
### Brief summary of bug

More of a undesirable model feature than a bug.  We found that initializing t_soisno to 272K instead of 274K (for cold start) solves the problem of significant deep soil carbon in permafrost regions.

### General bug information

**CTSM version you are using:** master/main (specific tag I'm using for testing is branch_tags/PPE.n08_ctsm5.1.dev023)

**Configurations affected:** All BGC

### Details of bug

There is too much soil carbon deep in the soil column in permafrost regions.  Since t_soisno is initialized above freezing (274K) in a cold start, the model diffuses carbon deep into the soil column in the first couple of decades.  Once the soil column eventually freezes that carbon is locked in.  If t_soisno is initialized below freezing (272K) this doesn't happen.  The effect on this on other models quantities appears to be minimal.

See the following section of the cnmatrix google document (https://docs.google.com/document/d/1yhcymw3FgGLlwAuLZxl7t-iQ021HbUlmAQlm9Vo7ewE/edit) for links to results/diagnostics):

"Initialize soil temperature at 272K instead of 274K.  Timeseries plot of AD for a representative gridcell. Survivability timeseries plots for AD for default (t_soisno=274) and new (t_soisno=272). Diagnostics for AD with t_soisno=272 compared to t_soisno=274.  Vertical profile plot for single point. Vertical profile plot for global. Diagnostics for pAD are here.  (complete, seems to work, deep soil carbon doesn’t get frozen in in permafrost regions)."

The code modification required is in TemperatureType.F90, changing this line:

this%t_soisno_col(c,1:nlevgrnd) = 274._r8

to this:

this%t_soisno_col(c,1:nlevgrnd) = 272._r8

See /glade/work/oleson/PPE.n08_ctsm5.1.dev023/cime/scripts/ctsm51c6_PPEn08ctsm51d023_2deg_GSWP3V1_Sparse400_cs_ts_tsoisno272_2000AD/SourceMods/src.clm/TemperatureType.F90

### LINKED ISSUE #1462: GSSUN and GSSHA are not set on history files
### Brief summary of bug

@djk2120 noticed this and has a fix for it.

### General bug information

**CTSM version you are using:** ctsm5.1.dev053

### Details of bug

GSSHA and GSSUN are set to _FillValue everywhere.

Most configurations output these two.

### LINKED ISSUE #1465: Change FATES to run with use_nitrif_dentrif TRUE
By default FATES is being run with use_nitrif_dentrif=FALSE, we'd like to change this so that use_nitrif_dentrif is hardcoded to true (see #1356). This does mean answers will change for FATES cases.

@rgknox @ckoven @rosiealice do you need to do any testing to ensure that FATES simulations are fine to run with use_nitrif_dentrif to TRUE?

### LINKED ISSUE #1496: SP NoAnthro case fails...
### Brief summary of bug

The SP NoAnthro compset fails on trying to build the namelist

### General bug information

**CTSM version you are using:** ctsm5.1.dev057

**Configurations affected:** 
  Compsets: I1850Clm50BgcNoAnthro, I1850Clm50SpNoAnthro, I1850Clm51SpNoAnthro

### Details of bug

SMS_D_Ld9.f09_g17.I1850Clm50SpNoAnthro.cheyenne_intel.clm-decStart1851_noinitial

fails on building of the namelist (see message below). This goes back to the original version of the use-case for NoAnthro.

```
        Building test for SMS in directory /glade/scratch/erik/ctsm5.1.dev057/cime/scripts/SMS_D_Ld9.f09_g17.I1850Clm50SpNoAnthro.cheyenne_intel.clm-decStart1851_noinitial.20210922_124433_72bzn4
        WARNING: CLM is starting up from a cold state
        ERROR: Command /glade/scratch/erik/ctsm5.1.dev057/bld/build-namelist failed rc=255
        out=
        err=ERROR : CLM build-namelist::CLMBuildNamelist::setup_logic_cnfire() : CN is off which implies that cnfire is off and yet a fire constant (pot_hmn_ign_counts_alpha) is being set, which contradicts that
```

ekluzek:
### Description of changes

Change ceta from 450 to 358 for CTSM5.1
Turn Medlyn on for PHS off for clm51 and clm50 (3 dynroot tests will change answers)
Make max CO2 partial pressure consistent
Change surface datasets to only use 78PFT versions (so SP results may change)
Fix GSSUN and GSSHA history variables so now not just missing values
Changes FATES to run with use_nitrif_dentrif=T by defaul;t
Change dynlakes test to methane/use_nitrif_=T
(This will allow use_nitrif and methane flags to be removed)

### Specific notes

Contributors other than yourself, if any: @djk2120 

  Fixes #1462
  Fixes #1394
  Fixes #1460
  Fixes #1465
  Fixes #1356
  Fixes #1496

  Small changes to a few defaults will change answers in most configurations

Any User Interface Changes (namelist or namelist defaults changes)? Changes some defaults

Testing performed, if any: Limited so far, will do regular testing

### DIFF
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -2232,10 +2232,20 @@ sub setup_logic_surface_dataset {
   if ($flanduse_timeseries ne "null" && &value_is_true($nl_flags->{'use_cndv'}) ) {
      $log->fatal_error( "dynamic PFT's (setting flanduse_timeseries) are incompatible with dynamic vegetation (use_cndv=.true)." );
   }
-  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'fsurdat',
+  # Always get the crop version of the datasets now and let the code turn it into the form desired
+  my $var = "fsurdat";
+  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var,
               'hgrid'=>$nl_flags->{'res'}, 'ssp_rcp'=>$nl_flags->{'ssp_rcp'},
-              'sim_year'=>$nl_flags->{'sim_year'}, 'irrigate'=>$nl_flags->{'irrigate'},
-              'use_crop'=>$nl_flags->{'use_crop'}, 'glc_nec'=>$nl_flags->{'glc_nec'});
+              'sim_year'=>$nl_flags->{'sim_year'}, 'irrigate'=>".true.",
+              'use_crop'=>".true.", 'glc_nec'=>$nl_flags->{'glc_nec'}, 'nofail'=>1);
+  # If didn't find the crop version check for the exact match
+  if ( ! defined($nl->get_value($var) ) ) {
+     $log->verbose_message( "Crop version of $var NOT found, searching for an exact match" );
+     add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var,
+                 'hgrid'=>$nl_flags->{'res'}, 'ssp_rcp'=>$nl_flags->{'ssp_rcp'},
+                 'sim_year'=>$nl_flags->{'sim_year'}, 'irrigate'=>$nl_flags->{'irrigate'},
+                 'use_crop'=>$nl_flags->{'use_crop'}, 'glc_nec'=>$nl_flags->{'glc_nec'});
+  }
 }
 
 #-------------------------------------------------------------------------------
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -870,12 +870,13 @@ sub setup_cmdl_bgc {
   my $var = "soil_decomp_method";
   add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var,
               'phys'=>$nl_flags->{'phys'}, 'use_cn'=>$nl_flags->{'use_cn'}, 'use_fates'=>$nl_flags->{'use_fates'} );
+  my $soil_decomp_method = remove_leading_and_trailing_quotes( $nl->get_value( $var ) );
 
   if ( &value_is_true($nl_flags->{'use_cn'}) ||  &value_is_true($nl_flags->{'use_fates'}))  {
-     if ( remove_leading_and_trailing_quotes( $nl->get_value($var)) eq "None" ) {
+     if ( $soil_decomp_method eq "None" ) {
         $log->fatal_error("$var must NOT be None if use_cn or use_fates are on");
      }
-  } elsif ( remove_leading_and_trailing_quotes($nl->get_value($var)) ne "None" ) {
+  } elsif ( $soil_decomp_method ne "None" ) {
      $log->fatal_error("$var must be None if use_cn or use_fates are not");
   }
   #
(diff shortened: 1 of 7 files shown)


ekluzek:
I've fixed #1503 and #1504. Need to do some study in order to get #1505 to work for the FATES cases.

ekluzek:
This is now passing all tests as expected on cheyenne. izumi seems to be down, so will test as soon as I can.

Issue #1429, opened 2021-07-15
Modified NEON surface datasets have errors 


wwieder:
### Brief summary of bug

It looks like a few sites that use the modified NEON surface dataset end up with errors in the soil sand, clay, and ORGANIC profiles.  

### General bug information
A quick look at some sites suggest this is an issue for HEAL, PUUM, TOOL and maybe others?  We don't even have met data for most of these sites, so it's likely not a big issue at this point, but something to be aware of as we start running more sites

@negin513 confirmed this is not an issue with how we're handling `zbedrock`, but results from issues in the raw data from NEON.

**Does this bug cause significantly incorrect results in the model's science?** YES

@ekluzek I don't want to delay bringing in #1375 and #1411, so let's bring in these PRs and @negin513 can address this later.

**Configurations affected:** NEON tower simulations

negin513:
Will, thanks for bringing this up to my attention. 
It seems like the issue of missing data in the updated surface dataset is not related to `zbedrock` and **it is caused by missing data in the provided neon surface dataset.** 

As we discussed earlier, you made [an issue](https://github.com/NEONScience/NCAR-NEON/issues/44) about this on NEON repo.

negin513:
@wwieder:
 Since this does not seem to be caused by any bugs related to the code, should we close this issue and continue the discussion in the corresponding [NEON issue](https://github.com/NEONScience/NCAR-NEON/issues/44).

wwieder:
@negin513 let's leave this open to discuss how we want to handle this on the NCAR side.

It seems like there are three issue here

1. **Missing values**, I wonder what others think about @ekluzek's suggestion to fill in missing data from NEON with the default values from the orig. surface dataset
2. **Organic horizons**, especially in some forest sites this gives really large ORGANIC surface soil layers (e.g. HARV or ORNL). Do we leave this feature, or replace O horizon data with A horizon data for consistency?   I suggest we leave these data and evaluate impacts on moisture and temperature dynamics with the two surface datasets.
3. **Crazy values**,  especially at depth (e.g. ONAQ). These will take a case-by case evaluation to identify potential issues with raw data (e.g. bulk density) in the data NEON are providing, but it's likely important to QC the NEON data.

wwieder:
[...]

PR #1539, opened 2021-11-02
Neon modify surfurface dataset 

negin513:
### Description of changes
This PR includes the scripts for modification of the surface dataset for neon sites to address #1429.
Specifically, the following has been addressed in this PR:
1) Update the calculation of `ORGANIC` to use the new field (`CaCO3`) from NEON data. 
For this calculation if `CaCO3` data is available, we first calculate inorganic carbon by:
inorganic carbon = (caco3 /100.0869)*12.0107
Next, we calculate organic carbon by subtracting inorganic carbon from the total carbon:
 [organic carbon = carbon_tot - inorganic carbon]
If the `CaCO3` is not available then the code uses `carbonTot` and `estimatedOC` by NEON. 
 
2) Discussed here (https://github.com/ESCOMP/CTSM/issues/1429#issuecomment-924895184) 
For the Ag sites (KONA and STER), it changes the `PCT_NATVEG`, `PCT_CROP`, and `PCT_NAT_PFT` to avoid the error that we previously had in spin-up:  `surfrd_veg_all ERROR: sum of wt_nat_patch not    1.00000000000000       at nl=  1          sum is:   0.000000000000000E+000 `

3) Discussed here (https://github.com/ESCOMP/CTSM/issues/1429#issuecomment-931648966)
There was a typo previously in the NEON data for ABBY sites caused by mix of sample measurements. Please note that this was updated by hand once data was downloaded from NEON site. 

The following plot shows the results of the above changes in the `ORGANIC` field for all sites comparing it with the original and previous version of the surface dataset:

allsite_organic_sf_profile_v9
 
### Specific notes

Contributors other than yourself, if any: @wwieder  and @danicalombardozzi 

CTSM Issues Fixed (include github issue #): #1429

Are answers expected to change (and if so in what way)? N/A

Any User Interface Changes (namelist or namelist defaults changes)? N/A


wwieder:
This is fine to come to main, but does it also need to be merged with the NEON updates branch for now so we can get the most up to date NEON code modification?  https://github.com/ESCOMP/CTSM/tree/jedwards/neon_updates

wwieder:
Minor point, but the ORGANIC profiles at MLBS and HEAL look kind of strange. At some point we should double check these for accuracy.  HEAL won't run (too many gaps in datm data), but MLBS may and will certainly look odd with that organic bulge towards the top of the profile.

ekluzek:
We'll keep this as a separate PR, but I'll add it to #1467, and it'll come to CTSM main with that PR.

jedwards4b (approved):

jedwards4b:
@negin513 I'm confused that there are 57 files changed here and the one file I expect to be changed has conflicts.

PR #1540, opened 2021-11-02
Implement `fates_levleaf` as a FATES 1D dimension variable 

glemieux:
### Description of changes

This update adds the fates `nlevleaf` variable as a standalone 1D dimension in histFileMod to provide history file metadata.

### Specific notes

This PR is associated with https://github.com/NGEET/fates/pull/752.  The fates PR renames `dinc_ed` and as such the use statement to this variable needs to be removed from clmfates_interfaceMod.F90.

Testing performed, if any:


ekluzek (approved):

src/main/histFileMod.F90
```diff
@@ -3076,7 +3078,7 @@ subroutine htape_timeconst(t, mode)
              call ncd_defvar(varname='fates_scmap_levscpf',xtype=ncd_int, dim1name='fates_levscpf', &
                   long_name='FATES size index of the combined pft-size class dimension', units='-', ncid=nfid(t))
              call ncd_defvar(varname='fates_levcacls', xtype=tape(t)%ncprec, dim1name='fates_levcacls', &
-                  long_name='FATES cohort age class lower bound', units='years', ncid=nfid(t))
+                  long_name='FATES cohort age class lower bound (yr)', units='-', ncid=nfid(t))
```
ekluzek, line 3081: Wait why are the units being moved to the long name?

ekluzek:
I'm going to bring this in as part of ctsm5.1.dev062.

src/main/histFileMod.F90
ckoven: Hi @ekluzek the reason for this is that metadata-aware data analysis programs (ferret in particular) get confused when there are two different time dimensions in a given dataset.  The new versions of ferret won't even let us open the history files unless we change the metadata. So this change prevents that from happening and makes it clear to the programs which the relevant time dimension is.

ekluzek: Hmmm. OK. That's annoying. It seems like it should realize that there is a different coordinate for time, since the dimension names are different, and the time dimension variables all use the time dimension.

It does look like the CF convention way to handle it would be to add the "coordinate" attribute as follows...

https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#alternative-coordinates

But, that shouldn't even be necessary since it's a different dimension.

So in the end I think I agree to this change. I might add a note about it though. Thanks for the explanation.

ekluzek:
@glemieux OK so one problem with our plan is that I need an updated FATES version to point to that has the fates NGEET/fates#752 PR in it. Otherwise I get a compiler error, because something in FatesInterfaceMod isn't defined "fates_hdim_levelem".

The other way around it would be to back out just that part of the change and have that come in another tag with a fates update.

glemieux:
> @glemieux OK so one problem with our plan is that I need an updated FATES version to point to that has the fates [NGEET/fates#752](https://github.com/NGEET/fates/pull/752) PR in it. Otherwise I get a compiler error, because something in FatesInterfaceMod isn't defined "fates_hdim_levelem".

Is it `fates_hdim_levelem` or `fates_hdim_levleaf`?  `fates_hdim_levleaf` is new to the fates-side PR; I should have realized this would cause a compiler error.
 
> The other way around it would be to back out just that part of the change and have that come in another tag with a fates update.

I think since the `dinc_ed` change is breaking wrt the same PR that has `fates_hdim_levleaf` coming in, we can keep these two things together.  I think it's probably better just to coordinate the integrations as usual instead of trying to roll this into dev tag 062.

ekluzek:
Sorry I didn't see your feedback before this mornings meeting. As we talked about, I'll include this PR in dev062, but I'll comment out the two lines that need a new fates version. Then your PR that brings in the fates update can uncomment them. I think this will be the easiest way forward for both of since, I've already merged this PR into the dev062 PR.

Issue #1569, opened 2021-12-01
History output for DISPLA is incorrect when vegetation is buried by snow


olyson:
### Brief summary of bug

History output for DISPLA is incorrect when vegetation is buried by snow because the variable displa is declared as a local variable within BareGroundFluxesMod.F90 and hence the global variable used for history output is not updated.  DISPLA should be 0 in this case but instead remains at the value set in CanopyTemperatureMod.F90.
Reported by Ronny Meier and confirmed by myself.

### General bug information

**CTSM version you are using:** release-clm5.0.34-2-ga2989b04 (but master/main tags are affected as well)

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** All when history variable DISPLA is active and vegetation is buried by snow

olyson:
There is a similar issue with lakes in that a local variable displa is set to zero.  Since displa for lakes isn't set elsewhere presumably this means that lakes aren't included in the grid cell average since they will be spval.  Urban has a bit of a different issue in that it has its own definition of displacement height (z_d_town) and so presumably it doesn't get included in the grid cell average either.  I'm wondering if the right thing to do is to restrict the output to vegetated landunits only (set_spec=spval, l2g_scale_type='veg'). And include "(vegetated landunits only)" in the long_name.

billsacks:
@dlawrenncar suggests looking through BareGroundFluxes for other variables that might have the same issue. Someone pointed out a similar seeming issue with roughness length (z0 and variations).

In terms of the history attributes: if you set `l2g_scale_type='veg'`, you don't need `set_spec=spval`.

ekluzek:
@olyson is this still a problem in the latest development versions? This seems like an important thing to make sure is fixed. We don't need to worry about if for past release versions, but it does seem like it should be fixed for CESM3. The #1570 PR sounds like maybe it was corrected. Although another suggestion was to look at this for other variables and maybe that needs to be done?

olyson:
I believe this was all fixed by #1570 , including looking at other variables.

ekluzek:
OK, nice I'm closing then...

PR #1570, opened 2021-12-06
Fix issue #1569 (History output for DISPLA is incorrect when vegetation is buried by snow)

olyson:
### Description of changes
This sets history output for DISPLA (displa_patch) to zero when vegetation is buried by snow.

In addition, z0mv_patch (Z0MV), z0hv_patch (Z0HV), z0qv_patch (Z0QV) are set to zero when vegetation is buried by snow.

Z0M (z0m_patch) has been renamed to Z0MV_DENSE since it appears to be momentum roughness length for vegetation with a dense canopy (in SetZ0mDisp; z0m(p)    = pftcon%z0mr(patch%itype(p)) * htop(p)) 

Updated the column values of z0hg (Z0HG) and z0qg (Z0QG), which are used for history output, to the new patch values after they are calculated in BareGroundFluxesMod.F90.

Added l2g_scale_type='veg' to the history calls for DISPLA, Z0MG, Z0HG, Z0QG, Z0MV, Z0HV, Z0QV,  Z0MV_DENSE.

### Specific notes

Contributors other than yourself, if any:  N/A

CTSM Issues Fixed (include github issue #): #1569

Are answers expected to change (and if so in what way)? Only for these history fields (which were and still are all off by default): DISPLA, Z0MG, Z0HG, Z0QG, Z0MV, Z0HV, Z0QV, Z0M (renamed to Z0MV_DENSE)

Testing performed, if any:
Ran the clm_short test suite with all eight fields active.  All pass (FIELDLIST field lists differ (otherwise bit-for-bit)).  Also ran with all eight fields inactive.  All pass.
Another option would be to regenerate baselines with these fields active and then run the test suite on these changes also with the fields active.


ekluzek:
I verified that the test SMS_Lm1.f19_g17.I1850Clm50BgcCropCmip6waccm.cheyenne_intel.clm-basic is marked as bit-for-bit and it is. I also verified that none of these fields affected at in testmods and they aren't. 

As such I'm going to treat this as a bit-for-bit change as it's changing diagnostic fields that aren't on for our normal testing.

ekluzek (approved):

Issue #1132, opened 2020-08-31
Spinup stability script is out of date


olyson:
The spinup stability script .../tools/contrib/SpinupStability.ncl is out of date and should be replaced by the scripts in the CLM5 data tools directory:

/glade/p/cgd/tss/CLM50_data_tools/SpinupStability_SP_v9.ncl  (the SP version)
/glade/p/cgd/tss/CLM50_data_tools/SpinupStability_v10.ncl     (the BGC version)

olyson:
Just a reminder that this should be taken care of as soon as possible.  Users are using the old script in the .../tools/contrib directory which has the equilibrium bug in it.

PR #1576, opened 2021-12-13
Update spinup stability scripts

olyson:
### Description of changes

Update the BGC spinup stability script and add an SP spinup stability script.

### Specific notes

CTSM Issues Fixed (include github issue #): #1132

Are answers expected to change (and if so in what way)?  Only for spinup stability script output

Testing performed, if any:
Ran scripts against existing data

### DIFF
--- a/tools/contrib/README
+++ b/tools/contrib/README
@@ -29,9 +29,15 @@ singlept
         extracting data from global datasets
 	v1 - Sean Swenson 8/2015
 
-SpinupStability.ncl
-        This script assesses the equilibrium state of a spinup run
-        works on either monthly or annual mean history files - Keith Oleson 7/2016
+SpinupStability_SP_v9.ncl
+        This script assesses the equilibrium state of a Satellite Phenology (SP) 
+        spinup run, works on either monthly or annual mean history files - Keith
+        Oleson 12/2021
+
+SpinupStability_BGC_v10.ncl
+        This script assesses the equilibrium state of a Biogeochemistry (BGC) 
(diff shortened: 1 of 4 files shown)


ekluzek (commented):
Thanks for making a PR for this @olyson. We want to make contrib scripts really lightweight to bring in, hence I was planning on just approving it. But, alas I have questions. I'm really happy you did testing with this and as the original author you are the best one to evaluate changes to the script.

So my only question is to understand why you added the previous version as "v9" and updated the name of the new one to "_v10"? Are both scripts useful to have at the same time? I wonder if the name shouldn't reflect  the usage difference. Also we should add notes on them to the README file in the tools/contrib directory, especially if the usage is different between them.

olyson:
SpinupStability_SP_v9.ncl is for SP simulations (this wasn't in the repository previously).
SpinupStability_v10.ncl is for BGC simulations.

olyson:
Sorry, I didn't see the README.  I'll update it and push.

olyson:
Ok I've pushed changes to the README and renamed the scripts appropriately.  Thanks for looking at this.

ekluzek (approved):
OK, thanks for changing that. In the longrun, since there is so much duplication between these scripts, I think we'll want to have a core python module that's shared between them. But, I think it's OK to have two separate scripts for now. We really do need the ability to check both SP and BGC cases, that's a requirement. So thanks for adding that.

Since, this is a tools/contrib thing and you've done the testing/validation, I'm just going to approve as is, and it will be included in the next tag that's made.

PR #1577, opened 2021-12-13
Apply black to python directory

### LINKED ISSUE #1471: Add support for using "black" as a python formatter to our python coding process
We'd like to transition to using "black" as a python code formatter to our python coding process. Using it as a formatter will make sure we comply with PEP8, and keep our code in a consistent format. We talked about this in our CSEG meeting and CAM and others thought that there was a need for such a tool. pylint is still worthwhile in checking for issues, but we need a separate tool for formatting the appearance of code. Having a tool to do the formatting means we don't have to waste time working on appearance and can just let the formatter take care of consistent look and feel. We want to apply to this to the python modules under our "python" directory in CTSM. We won't require it of other python code, especially right now, and especially don't think the code in "tools/contrib" needs to worry about this.

There are two steps we think are important to accomplish this.

- Add a target to the python directory Makefile for "black" to run the python codes through black.
- Add a github action to check that python code has been run through black and if not trigger an error.

The first one should be straight forward as it will be similar to the pylint check in the python directory.

The second we aren't quite sure how it will work. What we'd like it to do is to prevent a PR from being approved if the "black check" operation isn't clear. This would be similar to the checkbox PR checker we already have in place. We think looking at how the action is done for the checkbox checker should help in figuring it out. Documentation on github for actions should also help. And black has some advice on github actions here...

https://black.readthedocs.io/en/stable/integrations/github_actions.html

I also see a lot of help on this from doing a google search for this, so there's likely many different examples of how to do this and we just need to pick the one for the way we do it.

@billsacks also suggests checking with Jesse in CAM and Jason Boutte from LBL.

ekluzek:
### Description of changes

Run black on python directory, and add autorun of check with black for python directory.

### Specific notes

 Fixes #1471

Testing performed, if any: python tests


billsacks:
@ekluzek since this will interfere with git blame pretty significantly, what would you think about adding a file like `.git-blame-ignore-revs` in the root of the ctsm repository listing the commit(s) that ran reformatting operations. See https://akrabat.com/ignoring-revisions-with-git-blame/ for details.

ekluzek:
@billsacks that is a trick that I read about from the "black" documentation. I didn't know exactly what it meant, but in principle sounds like a really good idea. So I'll figure that out and add it to the PR.

billsacks:
As discussed in today's standup: plan is: after bringing in #1448 and upcoming PRs for subset_data.py, we'll be at a cleaner break point. Then @ekluzek will redo the black-ification of the python and we'll bring this in.

The general principle here is: If possible, we'll try to only need to do a full-rewriting run of black once (to avoid messing with git blame too much).

ekluzek:
This should be coordinated with #1768. We should run black (and any pre-commit hooks) we want in place on our code before we bring in pre-commit as a standard tool to use.

Issue #1556, opened 2021-11-17
Remake PCT_PFT raw datasets to NOT force landunit area to 100%, and some other fixes needed to PCT_PFT raw datasets


billsacks:
This is connected with #1555 (and I realized this issue while investigating the mksurfdata_map code related to #1555 ) but this is a somewhat separate issue.

Currently, my understanding is that the PCT_PFT raw datasets are constructed so that the total landunit area of natural veg plus crop adds up to 100%: in source grid cells where the original data specify less than 100% cover of vegetation, PCT_PFT values are increased so that the raw data coming in to mksurfdata_map will specify 100% cover of PCT_NATVEG + PCT_CROP.

But as far as I can tell from an initial read through of mksurfdata_map, there is no reason why this needs to be the case, and I realized that there is a problem with this: In this call:

https://github.com/ESCOMP/CTSM/blob/7a850247fd363abacec378159e400b9b4facabaf/tools/mksurfdata_map/src/mkpftMod.F90#L561-L564

I'm pretty sure that the source (and dest) weighting imply that PCT_NATVEG (i.e., the landunit area) is used as a weighting factor in remapping PCT_NAT_PFT from source to dest. That seems like the right thing to do, but the correction of the raw datasets to force vegetated areas to 100% makes this weighting factor not work the way it should.

Consider a destination grid cell that has two source grid cells, with equal coverage:
- Source 1: truly 100% natural veg, with 100% c3 grass
- Source 2: truly 1% natural veg, 99% lake, where the natural veg portion is 100% bare ground

The correct result would be a grid cell with 50.5% natural veg, 49.5% lake, where the natural veg is almost entirely c3 grass (with just a tiny bit of bare ground). However, because the PCT_PFT raw dataset is adjusted to give 100% area in Source 2, the current result in the code would be 50% c3 grass and 50% bare ground.

To get correct PCT_NAT_PFT mappings, I believe the raw datasets should be constructed **without** forcing the total veg area to add to 100%.

See the bold note at the bottom of https://github.com/ESCOMP/CTSM/issues/1555#issue-1056687092 for something to check in the course of doing this. I believe it's safe to make the changes in this issue before or after #1555, though if I had to choose an order, I would do #1555 before updating the raw datasets, because there are less likely to be issues with assuming that PCT_NATVEG + PCT_CROP = 100% once we do #1555.

---

Edit: here is the bold note that I was referring to, just so it's all in one place:

[...]

PR #1586, opened 2021-12-29
Enable the mksurfdata_map to generate landuse_timeseries for dynamic urban and lake

billsacks:
### Note about this PR

This PR is a continuation of #1579 (from @keerzhang1). I approved that PR in terms of the work needed from @keerzhang1, but there is still some work needed on our end, primarily in the perl code, before it can be completely merged. I have moved the branch to the ESCOMP repository so that further work can happen here, and so that @olyson can open a follow-up PR into this branch. (See also https://github.com/ESCOMP/CTSM/pull/1579#issuecomment-1002789374)

The following description is copied from #1579 

### Description of changes
This update includes code changes that enable the mksurfdata_map to generate landuse_timeseries with transient urban variables (PCT_URBAN and HASURBAN).

Update (2021-12-21): The changes in #1073 was combined with this PR. Now the code changes enable the mksurfdata_map to generate landuse_timeseries with dynamic urban and lake variables (PCT_URBAN, PCT_LAKE, PCT_URBAN_MAX, and PCT_LAKE_MAX)

### Specific notes
For now, the changes in the mksurfdata.pl are hardcoded to simply generate a landuse_timeseries*. txt file containing urban raw data locations that I specified.

Contributors other than yourself, if any: @keerzhang1 @olyson @fang-bowen @Face2sea

CTSM Issues Fixed (include github issue #): https://github.com/ESCOMP/CTSM/issues/1445

Are answers expected to change (and if so in what way)?
The surface datasets will be different.
Any User Interface Changes (namelist or namelist defaults changes)?
No.
Testing performed, if any:
(List what testing you did to show your changes worked as expected)
(This can be manual testing or running of the different test suites)
(Documentation on system testing is here: https://github.com/ESCOMP/ctsm/wiki/System-Testing-Guide)
(aux_clm on cheyenne for intel/gnu and izumi for intel/gnu/nag/pgi is the standard for tags on master)

**NOTE: Be sure to check your coding style against the standard
(https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines) and review
the list of common problems to watch out for
(https://github.com/ESCOMP/CTSM/wiki/List-of-common-problems).**

PR #1587, opened 2021-12-30
Incorporate changes from PR #1564 into dynurb_mksurf_part2 branch

### LINKED ISSUE #1555: Change renormalization of landunit areas in mksurfdata_map to be consistent with raw datasets
In recent discussions (spinning off from https://github.com/ESCOMP/CTSM/issues/1445 , including @Face2sea @fang-bowen @keerzhang1 @olyson and @lawrencepj1 ), it has become clear that there are problems with the current renormalization of PCT_NATVEG and PCT_CROP in mksurfdata_map to ensure that all landunits add up to 100%. This refers mainly to the code in `normalizencheck_landuse` and this initial adjustment:

https://github.com/ESCOMP/CTSM/blob/7a850247fd363abacec378159e400b9b4facabaf/tools/mksurfdata_map/src/mksurfdat.F90#L751-L763

My sense is that these adjustments made sense at one point (e.g., maybe for CLM4.0), but have not been updated to remain correct with other recent changes – one big one being the standard placement of crop on its own landunit.

Specific problems with what's currently done are:
- Urban should *not* preferentially replace bare ground: all PFTs should keep their original proportions. (The replacement of bare ground with urban may have made sense originally (see https://github.com/ESCOMP/CTSM/issues/1445#issuecomment-963710297 ), but @lawrencepj1 says that it no longer makes sense with recent raw datasets, where urban is split out from bare ground.) Urban should be treated the same as other special landunits now.
- Both crop and natural veg are decreased to accommodate too-large special landunit areas. This is a problem because, in the creation of the raw datasets, only natural veg is increased beyond what's specified initially (to accomplish PCT_CROP + PCT_NATVEG = 100%), so it is inconsistent that we then decrease crop to accommodate extra special landunit areas

My proposal, based on discussions with @lawrencepj1 and @olyson, is:
- PCT_NATPFT and PCT_CFT should never be adjusted beyond what's on the raw datasets. In particular, this means that urban should **not** adjust PCT_NATPFT.
- The areas of all landunits other than natural veg should be treated symmetrically; this should include the crop landunit, which should now be treated similarly to special landunits in places where we assure sums don't exceed 100% and in places where we adjust all other landunits proportionally. In particular, do not decrease crop to accommodate too-large special landunit areas. The justification is that it feels equally important to capture the correct areas of all of these non-natveg landunits (including crop!).
- PCT_NATVEG should be the residual after all other landunits. So, effectively, too-large areas of non-natveg landunits will first decrease the area of natural veg; only after that will the non-natveg landunits themselves decrease below what is specified. It's arguable if this makes the most sense for a present-day surface dataset (it would arguably make more sense to decrease /everything/ proportionally – though if the input datasets were consistent, then this would be a non-issue, so it's hard to say what is really right), but this feels most correct for transient cases, particularly when the natural veg area is an unknown that is just the residual of other areas.
    - Note, though, that we'll still need to read PCT_NATVEG initially for at least one purpose: giving the weighting of each source point in the remapping of PCT_NATPFT: e.g., a source cell that is 100% natural veg should have twice the weight of a source cell thath is 50% natural veg in remapping the PFT fractions from source to model resolution. However, for determining landunit areas, PCT_NATVEG will be set to 100 *minus* the sum of the other landunits.
- For the reason given above (the need for PCT_NATVEG in order to provide source weightings), the raw dataset creation should be changed to *not* rescale PCT_NATVEG such that it plus PCT_CROP equal 100%. I'll open a separate issue for this.

To accomplish this:
- Look at the sum of all landunits other than natural veg (but including crop!): If this sum is > 100%, scale down all landunits other than natural veg (but including crop!) proportionally. This will replace the `suma` rescaling code that is done just before calling `normalizencheck_landuse`.
- Set PCT_NATVEG to 100 *minus* the sum of the other landunits. This will replace the current rescaling code in `normalizencheck_landuse`.

Note that this scaling down process differs from what's done in dynLandunitAreaMod: in the latter, landunits are decreased in a priority order. There is some argument for keeping the two consistent, although there is also some argument for allowing them to remain inconsistent. The argument for the latter is: The corrections in mksurfdata_map relate to reconciling raw datasets that disagree but should (in principle) be consistent; the corrections in dynLandunitAreaMod are more about handling dynamic areas from other components (e.g., dynamic glacier). It may be that the best solution is to just save all of the normalization to 100% to runtime (see my comment near the end of https://github.com/ESCOMP/CTSM/issues/1445#issuecomment-963681306 ), but it feels like that introduces a lot of potential for new issues, for uncertain benefit, so we probably won't go there for now.

**When making these changes, we should double-check that there isn't any code that assumes that PCT_NATVEG + PCT_CROP adds up to 100% (or forces that to be the case). I didn't see anything like this in my initial, quick read-through, but we should double-check that. One specific thing to look for is: will code work correctly if the input PCT_NATVEG = PCT_CROP = 0%? (Or would such grid cells incorrectly be considered ocean, for example?)**
- Edit (2021-12-20): I think this last (bold) paragraph is mainly or entirely relevant in the context of #1556, not for this issue, so I have copied it to there.

olyson:
### Description of changes

Incorporate changes from PR #1564 (Change renormalization of landunit areas in mksurfdata_map to be consistent with raw datasets) into dynurb_mksurf_part2 branch.

### Specific notes

Completely resolves #1555 (note that part of #1555 was split out into its own issue: #1556 )

Are answers expected to change (and if so in what way)? Yes, surface datasets will change.

Testing performed, if any:
See description of testing in PR #1564

### DIFF
--- a/tools/mksurfdata_map/src/mkpftUtilsMod.F90
+++ b/tools/mksurfdata_map/src/mkpftUtilsMod.F90
@@ -24,7 +24,6 @@ module mkpftUtilsMod
   ! !PUBLIC MEMBER FUNCTIONS:
   !
   public :: convert_from_p2g      ! Convert a p2g array into pct_pft_type objects
-  public :: adjust_total_veg_area ! Adjust the total vegetated area (natural veg & crop) to a new specified total
 
   !
   ! !PRIVATE MEMBER FUNCTIONS:
@@ -212,45 +211,6 @@ contains
     default_cft(c3cropindex) = 100._r8
 
   end function get_default_cft
-
-
-  !-----------------------------------------------------------------------
-  subroutine adjust_total_veg_area(new_total_pct, pctnatpft, pctcft)
-    !
-    ! !DESCRIPTION:
-    ! Adjust the total vegetated area on the grid cell (natural veg & crop) to a new
-    ! specified total.
-    !
-    ! If the old areas are 0%, then all the new area goes into pctnatpft.
-    !
-    ! !USES:
-    use mkpctPftTypeMod, only : pct_pft_type
-    !
-    ! !ARGUMENTS:
-    real(r8), intent(in) :: new_total_pct ! new total % of natural veg + crop landunits
-    class(pct_pft_type), intent(inout) :: pctnatpft ! natural veg cover information
-    class(pct_pft_type), intent(inout) :: pctcft    ! crop cover information
-    !
-    ! !LOCAL VARIABLES:
-    real(r8) :: natpft_l2g ! grid cell % cover of nat. veg.
-    real(r8) :: cft_l2g    ! grid cell % cover of crop
-    real(r8) :: old_total  ! old total % cover of natural veg + crop landunits
-    
-    character(len=*), parameter :: subname = 'adjust_total_veg_area'
-    !-----------------------------------------------------------------------
-    
-    natpft_l2g = pctnatpft%get_pct_l2g()
-    cft_l2g = pctcft%get_pct_l2g()
-    old_total = natpft_l2g + cft_l2g
-    if (old_total > 0._r8) then
-       call pctnatpft%set_pct_l2g(natpft_l2g * new_total_pct / old_total)
-       call pctcft%set_pct_l2g(cft_l2g * new_total_pct / old_total)
-    else
-       call pctnatpft%set_pct_l2g(new_total_pct)
-    end if
-
-  end subroutine adjust_total_veg_area
-
 
 end module mkpftUtilsMod
 
--- a/tools/mksurfdata_map/src/mksurfdat.F90
+++ b/tools/mksurfdata_map/src/mksurfdat.F90
@@ -1120,13 +1120,14 @@ program mksurfdat
              write(6,*) subname, ' error: year for urban not equal to year for PFT files'
              call abort()
           end if
-      write(6,*)'input urban dynamic dataset for year ', year2, ' is : ', trim(furbname) 
+          write(6,*)'input urban dynamic dataset for year ', year2, ' is : ', trim(furbname)
+          ! Read input lake data
           read(nfdyn, '(A195,1x,I4)', iostat=ier) flakname, year2
           if ( year2 /= year ) then
              write(6,*) subname, ' error: year for lake not equal to year for PFT files'
              call abort()
           end if
-      write(6,*)'input lake dynamic dataset for year ', year2, ' is : ', trim(flakname)    
+          write(6,*)'input lake dynamic dataset for year ', year2, ' is : ', trim(flakname)
            
           ntim = ntim + 1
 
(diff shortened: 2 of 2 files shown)


billsacks (approved):
I finally got around to looking over this new version of #1564 . Thank you very much @olyson for doing this. It looks great to me. The only request for a change that I have is to edit the initial PR comment to clarify that this completely resolves #1555 other than the part that was split out into its own issue (#1556). You sort of say that, but let's state it more directly (if you agree that that's true), so that we can be sure to close #1555 when this is merged.

The original plan was to merge #1586 into the ctsm5.2.mksurfdata branch, then change the base of this PR to that branch, then do testing and merge this (see https://github.com/ESCOMP/CTSM/pull/1586#issuecomment-1002819166 for details). But that plan has changed with the work @mvertens and @slevisconsulting have been doing to completely rewrite mksurfdata_map.

So, @mvertens and @slevisconsulting, I think the new plan is that you will make sure that all of the changes in #1586 and this PR have been incorporated into the new version of mksurfdata_map, at which point we can close this PR. Do you agree?

olyson:
I've updated the initial PR comment.

slevisconsulting:
> So, @mvertens and @slevisconsulting, I think the new plan is that you will make sure that all of the changes in #1586 and this PR have been incorporated into the new version of mksurfdata_map, at which point we can close this PR. Do you agree?

@mvertens alerted me to this when I met with her today, and I'm fine with incorporating #1586  and #1587 in #1663. I will get to that in the next few days.

slevisconsulting:
...to explain why I asked the above question:
As with #1586 , I manually incorporated the code changes from #1587 in mksurfdata_esmf.

ekluzek:
@slevisconsulting all of the work with mksurfdata should be rebased into the ctsm5.2.mksurfdata branch. That's where they'll actually get merged into and turned into alpha-ctsm5.2 tags. Since, mksurfdata_esmf is new code a lot of that work will need to be done manually. This one could still merge into dynurb_mksrf_part2, but dynurb_mksrf_part2 itself should be rebased to merge into ctsm5.2.mksurfdata.

billsacks:
@slevisconsulting says this has been manually incorporated into the 5.2 mksurfdata branch. @olyson will confirm. Assuming these changes are all on the 5.2 mksurfdata branch, we can close #1555.

olyson:
I've reviewed this.  The only difference I see is that I believe this line in the ctsm5.2 mksurfdata branch:

               write (6,*) subname, 'Error: sum of special land units nearly 100% but none is >= 1% at ', &

should be:

               write (6,*) subname, 'Error: sum of special plus crop land units nearly 100% but none is >= 1.0% at ', &

given that suma is the sum of special plus crop.
@slevisconsulting do you agree?

slevisconsulting:
> I've reviewed this. The only difference I see is that I believe this line in the ctsm5.2 mksurfdata branch:
> 
> ```
>            write (6,*) subname, 'Error: sum of special land units nearly 100% but none is >= 1% at ', &
> ```
> 
> should be:
> 
> ```
>            write (6,*) subname, 'Error: sum of special plus crop land units nearly 100% but none is >= 1.0% at ', &
> ```
> 
> given that suma is the sum of special plus crop. @slevisconsulting do you agree?

@olyson thank you for checking and good catch! I will add this to a list of small corrections needed in the ctsm5.2 branch. At this morning's meeting I agreed to open a ctsm5.2 PR. I will address items on that list there.

And otherwise, feel free to close this PR (#1587).

olyson:
Sam is fixing the item I identified on the branch.  So I am closing this.

PR #1589, opened 2022-01-02
Remove reference to LND_DOMAIN_FILE and LND_DOMAIN_PATH for nuopc cases

mvertens:
### Description of changes
Remove reference to LND_DOMAIN_FILE and LND_DOMAIN_PATH for nuopc

### Specific notes
Using NUOPC, the only need for having the xml variables LND_DOMAIN_FILE and LND_DOMAIN_PATH was for the LILAC smoke test. This PR removes that dependency by hard-wiring the path to the domain file for the test itself in lilacsmoke.py.
When LILAC is used in an external model (e.g. WRF), users will have to manually add the correct fatmlnd_frc and fsurdat entries for their particular resolution, so this will not come from CIME anyway. 

Contributors other than yourself, if any: @billsacks 

CTSM Issues Fixed: None

Are answers expected to change (and if so in what way)? No

Any User Interface Changes (namelist or namelist defaults changes)?

Testing performed, if any: ran LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.cheyenne_intel.clm-lilac successfully.

PR #1591, opened 2022-01-03
Remove reference to LND_DOMAIN_FILE and LND_DOMAIN_PATH for nuopc cases

billsacks:
### Description of changes

Remove reference to LND_DOMAIN_FILE and LND_DOMAIN_PATH for nuopc

**These changes are from @mvertens . This is essentially a duplicate of #1589 but with fixed git history.**

### Specific notes
Using NUOPC, the only need for having the xml variables LND_DOMAIN_FILE and LND_DOMAIN_PATH was for the LILAC smoke test. This PR removes that dependency by hard-wiring the path to the domain file for the test itself in lilacsmoke.py.
When LILAC is used in an external model (e.g. WRF), users will have to manually add the correct fatmlnd_frc and fsurdat entries for their particular resolution, so this will not come from CIME anyway. 

Contributors other than yourself, if any: @mvertens 

CTSM Issues Fixed: None

Testing performed, if any: ran LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.cheyenne_intel.clm-lilac successfully.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -37,7 +37,7 @@ required = True
 local_path = cime
 protocol = git
 repo_url = https://github.com/ESMCI/cime
-tag = cime6.0.11
+tag = cime6.0.12
 required = True
 
 [cmeps]
@@ -48,7 +48,7 @@ local_path = components/cmeps
 required = True
 
 [cdeps]
-tag = cdeps0.12.32
+tag = cdeps0.12.34
 protocol = git
 repo_url = https://github.com/ESCOMP/CDEPS.git
 local_path = components/cdeps
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -37,7 +37,7 @@ required = True
 local_path = cime
 protocol = git
 repo_url = https://github.com/ESMCI/cime
-tag = cime6.0.12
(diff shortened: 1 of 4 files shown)


billsacks:
Since I opened this new PR, I can't officially approve it – but I have reviewed @mvertens 's changes and I do approve them.

@ekluzek would you like to look at this as well? Note that this ties the LILACSMOKE test to this given 10x15 resolution, but I don't see an easy way around that for now, so we have just tried to call this out in comments.

@mvertens if I remember correctly, this is needed to support a recent change in CMEPS, right? If so, can you tell us which CMEPS change that is so that we can coordinate those changes?

ekluzek (approved):
This looks fine to me. It has a check to make sure that 10x15 is chosen for the LILAC test, so that's pretty robust to me.

mvertens:
@billsacks - actually this PR is independent of any CMEPS PR but will facilitate bringing in a new CMEPS PR that removes the reference for the LND_DOMAIN_FILE and LND_DOMAIN_PATH. The order of PRs is CTSM, CIME (for split grids) and then CMEPS.

Issue #1590, opened 2022-01-03
Gridcell water balance errors for some PLUMBER2 sites with branch_tags/PPE.n11_ctsm5.1.dev030


olyson:
### Brief summary of bug

Some PLUMBER2 sites throw gridcell water balance errors with branch_tags/PPE.n11_ctsm5.1.dev030.

### General bug information

**CTSM version you are using:** branch_tags/PPE.n11_ctsm5.1.dev030

**Does this bug cause significantly incorrect results in the model's science?** Unknown

**Configurations affected:** Single point PLUMBER2 tower site simulations.  --compset I1PtClm51SpRs --res CLM_USRDAT

### Details of bug

I'm attempting to run PLUMBER2 sites in preparation for developing scripts to run PPE for these sites.  Most of the sites run to completion but several crash with gridcell water balance errors ranging from e-4 to e-2.  The cause of this seems to be that the beginning water balance (begwb_grc) of the problematic time step is different from the ending water balance (endwb_grc) of the previous time step.  The corresponding column water balance checks are fine.  There is only 1 column for the site I'm investigating so there is obviously something different in the terms for the grc and col checks.

### Important details of your setup / configuration so we can reproduce the bug

The setup scripts and input files for these sites were ported from a release branch tag (release-clm5.0.34) version of these simulations.  Other than some modifications to accommodate infrastructure changes, I don't think there is anything uniquely different about these simulations.  I'm curious whether anyone has reported issues with water balance in NEON simulations.
The case directory for the site I'm currently investigating is:

/glade/work/oleson/PPE.n11_ctsm5.1.dev030/cime/scripts/CTL2010SP/basecases/PPEn11_CTL2010SPSpin_AU-Tum

ekluzek:
I did some checking to see if there might have been a fix on CTSM main-dev that would cover this, and I didn't see anything that looks like it. So I don't think that updating to the latest would help the issue. I also searched for other issues like it either open or closed and didn't find anything.

In looking at the case, what happens when you run with irrigate off? And for that matter what happens when irrigate is on with a stub-river model? Do we expect that to function correctly?

olyson:
I didn't see any reference to this type of problem either.
The particular site I'm looking at is 100% BET temperate so I assume irrigate doesn't have an effect.
This seems to be associated with a dribble term at the first of a new year: qflx_liq_dynbal_left_to_dribble.
[...]

PR #1593, opened 2022-01-06
Fix get_curr_yearfrac and get_prev_yearfrac with Gregorian calendar

billsacks:
### Description of changes

Fix get_curr_yearfrac and get_prev_yearfrac with Gregorian calendar.

Also add unit tests of get_prev_yearfrac with a Gregorian calendar.

### Specific notes

Contributors other than yourself, if any: @olyson gets most of the credit here

- Resolves ESCOMP/CTSM#1590

Are answers expected to change (and if so in what way)? YES, but very limited: From a search through the code, I think this will only impact CNDV cases when run with the Gregorian calendar. (The changed code is also used by the AnnualFluxDribbler, but only to set some values in gridcell-level balance checks; this can cause a run failure as in #1590 but I don't think it will actually cause any answer changes. The changed code is also used for time-interpolated prescribed dynamic landuse (dyn_var_time_interp_type), but we currently don't use that for any variables.) **I think (but am not 100% sure) that these changes won't show up in any tests, so I'm marking this PR as bfb, since I think it will be bfb from a testing standpoint.**

Testing performed, if any: Only unit tests so far

### DIFF
--- a/cime_config/testdefs/testlist_clm.xml
+++ b/cime_config/testdefs/testlist_clm.xml
@@ -1258,6 +1258,16 @@
       <option name="tput_tolerance">0.5</option>
     </options>
   </test>
+  <test name="SMS_Ly5_Mmpi-serial" grid="1x1_smallvilleIA" compset="IHistClm50BgcCropQianRs" testmods="clm/gregorian_cropMonthOutput">
+    <machines>
+      <machine name="cheyenne" compiler="gnu" category="aux_clm"/>
+    </machines>
+    <options>
+      <option name="wallclock">01:00:00</option>
+      <option name="comment"  >Multi-year Gregorian test with transient and crop to test code that might break in leap years.</option>
+      <option name="tput_tolerance">0.5</option>
+    </options>
+  </test>
   <test name="LII_D_Ld3" grid="f19_g17" compset="I2000Clm50BgcCrop" testmods="clm/default">
     <machines>
       <machine name="cheyenne" compiler="intel" category="aux_clm"/>
--- a/cime_config/testdefs/testmods_dirs/clm/gregorian_cropMonthOutput/include_user_mods
+++ b/cime_config/testdefs/testmods_dirs/clm/gregorian_cropMonthOutput/include_user_mods
@@ -1,0 +1,1 @@
+../cropMonthOutput
--- a/cime_config/testdefs/testmods_dirs/clm/gregorian_cropMonthOutput/shell_commands
+++ b/cime_config/testdefs/testmods_dirs/clm/gregorian_cropMonthOutput/shell_commands
@@ -1,0 +1,1 @@
+./xmlchange CALENDAR=GREGORIAN
--- a/src/biogeochem/CNAnnualUpdateMod.F90
+++ b/src/biogeochem/CNAnnualUpdateMod.F90
@@ -28,7 +28,7 @@ contains
     ! On the radiation time step, update annual summation variables
     !
     ! !USES:
-    use clm_time_manager, only: get_step_size_real, get_days_per_year
+    use clm_time_manager, only: get_step_size_real, get_curr_days_per_year
     use clm_varcon      , only: secspday
     use SubgridAveMod   , only: p2c
     !
@@ -51,7 +51,7 @@ contains
     !-----------------------------------------------------------------------
 
     dt = get_step_size_real()
-    secspyear = get_days_per_year() * secspday
+    secspyear = get_curr_days_per_year() * secspday
 
     do fc = 1,num_soilc
        c = filter_soilc(fc)
--- a/src/biogeochem/CNC14DecayMod.F90
+++ b/src/biogeochem/CNC14DecayMod.F90
@@ -5,7 +5,7 @@ module CNC14DecayMod
   !
   ! !USES:
   use shr_kind_mod                       , only : r8 => shr_kind_r8
-  use clm_time_manager                   , only : get_step_size_real, get_days_per_year
+  use clm_time_manager                   , only : get_step_size_real, get_curr_days_per_year
   use clm_varpar                         , only : nlevdecomp, ndecomp_pools
   use clm_varcon                         , only : secspday
   use clm_varctl                         , only : spinup_state
@@ -87,7 +87,7 @@ contains
 
       ! set time steps
       dt = get_step_size_real()
-      days_per_year = get_days_per_year()
+      days_per_year = get_curr_days_per_year()
 
       half_life = 5730._r8 * secspday * days_per_year
       decay_const = - log(0.5_r8) / half_life
--- a/src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
+++ b/src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
@@ -5,7 +5,7 @@ module CIsoAtmTimeseriesMod
   !
   ! !USES:
   use shr_kind_mod     , only : r8 => shr_kind_r8
-  use clm_time_manager , only : get_curr_date,get_days_per_year, get_curr_yearfrac
+  use clm_time_manager , only : get_curr_date, get_curr_yearfrac
   use clm_varcon       , only : c14ratio, secspday
   use shr_const_mod    , only : SHR_CONST_PDB                    ! Ratio of C13/C12
   use clm_varctl       , only : iulog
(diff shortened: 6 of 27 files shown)


billsacks:
@olyson feel free to limit your review to the small change in clm_time_manager if you don't want to look through all the unit test changes.

olyson (approved):
Looks good to me thanks!

billsacks:
@olyson @ekluzek the commit I just pushed (c432b6d0721dfb5b05b5bd417d64c0aaad8aff84) is NOT yet the fix needed for @olyson 's issue: I first wanted to make the simpler change of removing a similar hack from get_calday, which is used in very limited places in the code and so it seems totally safe to remove the hack from there. The only place this is used is to convert crop planting dates in the format mmdd to a real-valued day-of-year. I did some unit testing (with some temporary tests that I deleted, because they didn't seem valuable long-term) to show that this change won't change answers in this use case, even for a planting date of 1231 (this translates to 366.0 which was left as is by the "hack" code). However, it did raise a different issue that I'll open an issue for shortly.

billsacks:
@olyson I have a pushed a fix that I expect will solve your latest problem, as we discussed yesterday. Do you want to see if the changes in this branch indeed fix all of your problems now? I have tested the build but nothing more than that yet.

@ekluzek do you want to take a quick look at these changes before I start final testing?

Note that I ended up going with the solution of adding an optional argument to get_curr_calday. This is because I would otherwise have needed to make 8 identical changes to apply the hack to 2 variables in each cap plus 2 more in clm_initializeMod, and that felt like too much duplication.

ekluzek (approved):
This looks good. The use of that argument "reuse_day_365_for_day_366 " makes the hack put into place more obvious. Hacks like that often cause problems if they aren't more obvious, and that's what happened here. It was OK previously, but because get_curr_calday was assumed to function correctly for all calendars thing built on it could break.

olyson:
Sorry for the delay.  The simulations I'm running only work with the PPE tag so I had to bring in your changes manually as SourceMods.  I put in changes to clm_initializeMod.F90, clm_time_manager.F90, and lnd_comp_mct.F90 (I'm not using nuopc or lilac and I figured I didn't need the testing changes).
These did fix the problems I was having, thanks!

billsacks:
Awesome, thanks @ekluzek and @olyson ! Now I just need to get things working with cime6.0.12 – I'm having problems running the pFUnit unit tests with that cime tag and am looking into why....

billsacks:
@ekluzek @olyson - I'm working on solving the remaining issue(s) with the Gregorian test. One issue is that there is at least one call to `get_days_per_year` (in the AnnualFluxDribbler) that really should be looking at the number of days per year in the year corresponding to the *start* of the time step, not the *end* of the time step. To encourage callers to think about which one is appropriate for each use, I thought it would be best to make two functions, `get_curr_days_per_year` and `get_prev_days_per_year`. I don't love the naming of these functions: I don't feel like the `curr` and `prev` are very intuitive in this context, but I thought it would be best to stick with the same naming convention that we use for similar pairs of functions like `get_curr_yearfrac` vs. `get_prev_yearfrac` – where `curr` means "as of the end of the time step" and `prev` means "as of the start of the time step" (and I have tried to add documentation in the functions to make it clear how they differ). However, please let me know if you feel like these names are too confusing and you have better suggestions.

I'm planning to go through and see if there are other uses of `get_days_per_year` (i.e., now `get_curr_days_per_year`) that should actually be using `get_prev_days_per_year`. My plan is that, for cases where it's not entirely clear what is right, I'll leave it as is, but I want to see if there are other situations where we should really be using `get_prev_days_per_year`.

ekluzek:
I agree that you should use curr and prev, since that's what's currently being done. If we change the name for it, we should change it throughout clm_time_manager. And I think it's reasonably clear, so I don't know that we really need to do that wholesale change. But, adding the new function for prev does sound necessary to make this more explicit.  It's especially more explicit than the hidden hack that was done previously. So I support your description of what you are doing. Thanks for working through this tricky and subtle aspect of the code!

olyson:
I also agree that although it's not ideal, we should use curr and prev for consistency. I think your comments will make it clear what the differences are.  And again, thanks for your thorough examination of this!

billsacks:
With the latest changes on this branch, the test `SMS_Ly5_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianRs.cheyenne_gnu.clm-gregorian_cropMonthOutput` now passes. In addition to the issues found by @olyson, I also had to fix the call to `get_days_per_year` in `AnnualFluxDribbler` so that it looked at the year as of the start of the time step, not the end of the time step.

Suspecting that there might be other, similar issues in other calls to `get_days_per_year`, I have looked through all of the calls to this function to see what else should possibly be changed. Here are my notes on the various calls:
- Things I have changed
    - AnnualFluxDribbler: This one should be changed (done)
- Things for which we should probably be using a fixed constant, because the point is to convert a per-year parameter into a per-second value. I'm thinking we should have a function like get_average_days_per_year, which will return 365 for a NOLEAP calendar and 365.2425 for a GREGORIAN calendar.
    - C14Decay: This should probably use a fixed constant, converting the decay rate from a per-year value to a per-second value: as currently written, C14 decays more slowly on leap years.
    - CNGapMortality: This should probably use a fixed constant, converting the mortality rate parameter from a per-year value to a per-second value: as currently written, mortality is a bit slower on leap years.
    - CNPhenologyMod.F90: CNEvergreenPhenology and CNStressDecidPhenology and the multiplier of leaf_long in CropPhenology: This should probably use a fixed constant, converting a litterfall rate from a per-year value to a per-second value: as currently written, litterfall happens more slowly on leap years.
    - decomp_rate_constants_bgc: As for the others, this should probably use a fixed constant
    - ndep_interp: This one is questionable, and may depend in part on whether the ndep input file and the model are consistent in their use of a GREGORIAN vs. NOLEAP calendar. But since I think this will typically come from model run with a NOLEAP calendar, I'm thinking that it should use a fixed constant (so that a constant annual ndep rate on the file would result in a constant annual ndep rate in CTSM, rather than a rate that varies slightly in leap vs. non-leap years). (But from spot-checking some recent files, it looks like rates are typically already specified as per-second, so this conversion from a per-year rate to a per-second rate is never done. I think that this implies that, in a GREGORIAN run, there is slightly more ndep than what occurred in the run that created the forcing data, since a year is slightly longer.)
        - Actually, upon further reflection, I think it's best *not* to change this one.
- Things I'm not sure about:
    - CNAnnualUpdateMod.F90:CNAnnualUpdate: It feels like this should probably be changed, but I'm not positive. And I don't think it will matter hugely: it looks like problems would just cause this to trigger at slightly the wrong time: For example, in a non-leap year transitioning to a leap year, I have a feeling this might not trigger as intended at the end of Dec 31 (because on the last time step of the year, secspyear will be `366*86400`, but the counter will only be `365*86400`), but it should trigger at the end of the next Jan 1. And then I think things will work themselves out at the end of the next year to get back in sync.
    - CNFUNInit: I'm not sure what the intent of this code is. It seems like this is supposed to be consistent with some other logic governing when FUN is run, but I'm not sure where that logic is. So I'm not sure what's correct here.
    - CNFireArea and CNFireFluxes (in various CNFire modules): Irrelevant, I think, because it's in a conditional that doesn't get executed on the last day of the year (which is weird and maybe wrong)... if it weren't in that conditional then my intuition is that at least the uses in CNFireFluxes should be get_prev_days_per_year; I'm not sure about the uses in CNFireArea.
    - CNNDynamicnMod.F90: CNFreeLivingFixation and CNNFixation: I'm not sure about these.
    - CNPhenologyMod.F90: CNPhenologyClimate: I'm not sure, but I think it should be consistent with CNAnnualUpdateMod.
    - ch4_annualupdate: I haven't looked carefully at this
    - clmfates_interfaceMod.F90: I haven't looked into this. It's possible that the answer differs for different uses within FATES.
- Other:
    - CropPhenology: determining idpp: see https://github.com/ESCOMP/CTSM/issues/1595

So the main changes that I'm pretty sure should be made are those under "Things for which we should probably be using a fixed constant". I am going to open a separate issue for this. I think the changes needed for that will be relatively simple, so I'll take a crack at them this week, but I'd like to do that in a separate tag, partly so those changes can be reviewed separately and partly because they will be more answer-changing than the ones in this PR (for Gregorian cases, at least).

billsacks:
I did a final pass through the code, looking for any other uses of `get_curr_calday`, `get_curr_yearfrac` or `get_prev_yearfrac` that should be changed. As far as I can tell, everything looks good now; in particular, it looks correct to me *not* to use `reuse_day_365_for_day_366=.true.` for any of the other calls.

billsacks:
Starting final testing now....

Issue #1316, opened 2021-03-30
New parameterization of vegetation surface roughness for forests


RonnyMeier:
The vegetation surface roughness (z0) of deciduous trees is close to the z0 of the ground when the trees shed their leaves. This is in contradiction to observations, which even observe an increase of z0 during the dormant phase. Therefore, we intend to introduce a new parameterization of the vegetation z0 for the forest PFTs, probably using the proposed parameterization of Nakai et al. (2008) (https://doi.org/10.1016/j.agrformet.2008.03.009). Steps to be taken:

1. Implement parameterization of Nakai et al. (already done). Could test other parameterizations later on too.
2. Use data of Hu et al. (2020) (https://doi.org/10.1016/j.agrformet.2020.107956) to tune parameters in parameterization.
3. Assess how new parameterization affects model performance.
4. Potentially do additional tuning.

wwieder:
@RonnyMeier this looks like exciting work (sorry I missed your presentation last week).
We'd welcome these changes as on option that users can chose in CTSM simulations and we'd like to help you bring this into the CTSM code base.  To do this it seems to me like you kind of have two options here:
1. Submit a pull request (PR) with your progress on implementing the Nakai parameterization     

OR

2. You can wait until you have the Hu data to parameterize and evaluate your code modifications before opening up the PR.

Either option is fine. Once you start the PR you can start working with the software engineering team to put namelist switches around your code so we can turn on/off your new roughness parameterization. The [ctsm wiki](https://github.com/ESCOMP/CTSM/wiki/Quick-start-to-CTSM-development-with-git) has helpful suggestions for getting ready for and submitting your PR.     

Please reach out if you have questions or concerns along the way.

wwieder:
To follow up, @RonnyMeier, you'll notice that @rgknox linked this issue to one in FATES.  The FATES team is also interested in this work and (I think) would welcome contributions to the FATES code base too.  Unfortunately, this is somewhat duplicative at this stage. BUT once you have the code up, running, and incorporated into the big-leaf OR FATES code it will hopefully be pretty straightforward to migrate it over to the other.  There may need to be additional testing and evaluation of the parameterizations used in each model, however.

RonnyMeier:
[...]

PR #1596, opened 2022-01-07
Surface roughness modifications

RonnyMeier:
This branch contains the surface roughness (z0) modifications published in: https://doi.org/10.5194/gmd-15-2365-2022

When changing the namelist input z0param_method from ZengWang2007 (default) to Meier2022 the following modifications are activated:

- A new parameterization of the vegetation surface roughness based on Raupach (1992) with optimized parameters to match the data collected in Hu et al. (2020) for different types of vegetation. This requires several new PFT-specific input parameters in the parameter file.
- A spatially explicit z0m input field for bare soil based on the data of Prigent et al. (2005). This may be activated specifically by the user through the namelist input use_z0mg_2d. This requires a new input variable in the surfdata file.
- The parameterization of z0m for snow based on accumulated snow melt as proposed in Brock et al. (2006). This may be activated specifically by the user through the namelist input use_z0m_snowmelt.
- The parameterization of Yang et al. (2008) for z0h and z0q over bare soil, snow, and glaciers.

Note that the study in GMD also proposes new globally constant values for the z0m of bare soil, snow, and ice. To "activate" those the parameter file needs to be changed at the moment. The original and modified parameter files and surfdata files will be shared by ftp.

Open issues/questions (already discussed with @ekluzek, @dlawrenncar, and @olyson:

-  How to incorporate the data of Prigent et al. (2005) in the surfdata generation. I will write an email about this to Catherine Prigent.
- One statement marked in CanopyFluxesMod should probably be changed when using Meier2022 (i.e., Yang et al. (2008) formulation should be used instead of Zeng and Dickinson (1998)). 
- At the moment one needs to change the parameter file to switch between the original and proposed globally constant z0m values for bare soil, snow, and ice. This is obviously not very user friendly and prone for mistakes.
- The introduction of Yang et al. (2008) frequently results in z0h and z0q larger than z0m. This is only rarely observed in the field and in contradiction to the theory that z0h and z0q should be smaller because heat and water vapor need to be transported through molecular diffusion in the surface sublayer.

Issues Resolved: #1316


ekluzek:
I know the date for your paper isn't completely known right now (hence the XXXX in the namelist name). But, it seems like we'll want a real year when this comes into main-dev of CTSM. So I propose we go with Meir2022? I think in the context of namelist options for CTSM being off a bit in the namelist name is acceptable. Especially if the exact reference is pointed to in the code. And we should put the reference in appropriate places where it's refereed to.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -861,12 +873,50 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+               delt = abs(U_ustar - U_ustar_prev)
+            end do
+
+            U_ustar = 4._r8 * U_ustar / lt / z0v_c(patch%itype(p))
+
+            z0mv(p) = htop(p) * (1._r8 - displa(p) / htop(p)) * exp(-vkc * U_ustar + &
+                      log(z0v_cw(patch%itype(p))) - 1._r8 + z0v_cw(patch%itype(p))**(-1._r8))
+
+
+            ! --> Use this for CLM-Ya08
```
ekluzek, line 906: Remove the commented out code.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1510,6 +1563,7 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
          ! function that goes to zero as LAI (ELAI + ESAI) go to zero.
 
          t_skin_patch(p)  =  emv(p)*t_veg(p)  +  (1._r8 - emv(p))*sqrt(sqrt(lw_grnd))
+         !t_skin_patch(p) = (ulrad(p) / 5.670367e-8_r8)**0.25_r8
```
ekluzek, line 1566: Remove commented out code.

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -516,26 +610,73 @@ subroutine SetRoughnessLengthsAndForcHeightsNonLake(this, bounds, &
...
+          else
+             if(use_z0mg_2d) then
+                z0mg(c) = z0mg_2D(c)
+             else
+                z0mg(c) = this%zlnd
+             end if
+                  
+          end if
+       case ('MeierXXXX')           ! Bare ground and ice have a different value
+          l = col%landunit(c)
```
ekluzek, line 647: Looks like this is largely repeated with above option. Only the part that is different should be in the case statement.

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -516,26 +610,73 @@ subroutine SetRoughnessLengthsAndForcHeightsNonLake(this, bounds, &
...
-          z0mg(c) = this%zsno
-       else
-          z0mg(c) = this%zlnd
-       end if
+
+       select case (z0param_method)
+       case ('ZengWang2007')
+          if (frac_sno(c) > 0._r8) then
+             if(use_z0m_snowmelt) then
+                z0mg(c) = exp(1.4_r8 * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8))-0.31_r8) / 1000._r8
```
ekluzek, line 634: If some of these constants are going to be repeated, they should be made into parameters.

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -516,26 +610,73 @@ subroutine SetRoughnessLengthsAndForcHeightsNonLake(this, bounds, &
...
+          end if
+       case ('MeierXXXX')           ! Bare ground and ice have a different value
+          l = col%landunit(c)
+          if (frac_sno(c) > 0._r8) then ! Do snow first because ice could be snow-covered
+             if(use_z0m_snowmelt) then
+                z0mg(c) = exp(1.4_r8 * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8))-0.31_r8) / 1000._r8 
+             else
+                z0mg(c) = this%zsno
+
+                ! --> Use this for CLM-VEG and CLM-Ya08
```
ekluzek, line 654: Remove the commented out code

src/biogeophys/LakeFluxesMod.F90
```diff
@@ -324,12 +336,47 @@ subroutine LakeFluxes(bounds, num_lakec, filter_lakec, num_lakep, filter_lakep,
             z0qg(p) = max(z0qg(p), minz0lake)
             z0hg(p) = max(z0hg(p), minz0lake)
          else if (snl(c) == 0) then    ! frozen lake with ice
-            z0mg(p) = z0frzlake
-            z0hg(p) = z0mg(p) / exp(params_inst%a_coef * (ust_lake(c) * z0mg(p) / 1.5e-5_r8)**params_inst%a_exp) ! Consistent with BareGroundFluxes
+            select case (z0param_method)
+            case ('MeierXXXX') 
+               z0mg(p) = params_inst%zglc
+
+               ! --> Use this for CLM-VEG and CLM-Ya08
```
ekluzek, line 343: Remove commented out code.

src/biogeophys/LakeFluxesMod.F90
```diff
@@ -324,12 +336,47 @@ subroutine LakeFluxes(bounds, num_lakec, filter_lakec, num_lakep, filter_lakep,
...
+            select case (z0param_method)
+            case ('MeierXXXX') 
+               if(use_z0m_snowmelt) then
+                  z0mg(p) = exp(1.4_r8 * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8))-0.31_r8) / 1000._r8 
+
+                  
+               else
+                  z0mg(p) = params_inst%zsno
+
+                  ! --> Use this for CLM-VEG and CLM-Ya08
```
ekluzek, line 366: Remove commented out code

src/biogeophys/LakeFluxesMod.F90
```diff
@@ -545,12 +593,39 @@ subroutine LakeFluxes(bounds, num_lakec, filter_lakec, num_lakep, filter_lakep,
...
             else if (snl(c) == 0) then
                ! in case it was above freezing and now below freezing
-               z0mg(p) = z0frzlake
-               z0hg(p) = z0mg(p) / exp(params_inst%a_coef * (ustar(p) * z0mg(p) / 1.5e-5_r8)**params_inst%a_exp) ! Consistent with BareGroundFluxes
-               z0qg(p) = z0hg(p)
+               select case (z0param_method)
+               case ('MeierXXXX') 
+                  z0mg(p) = params_inst%zglc
+
+                  ! --> Use this for CLM-VEG and CLM-Ya08
```
ekluzek, line 600: Remove commented out code

src/biogeophys/LakeFluxesMod.F90
```diff
@@ -545,12 +593,39 @@ subroutine LakeFluxes(bounds, num_lakec, filter_lakec, num_lakep, filter_lakep,
...
             else ! Snow layers
-               ! z0mg won't have changed
-               z0hg(p) = z0mg(p) / exp(params_inst%a_coef * (ustar(p) * z0mg(p) / 1.5e-5_r8)**params_inst%a_exp) ! Consistent with BareGroundFluxes
+               if(use_z0m_snowmelt) then
+                  z0mg(p) = exp(1.4_r8 * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8))-0.31_r8) / 1000._r8      
+               end if
+
+               select case (z0param_method)
+               case ('MeierXXXX')
+                  z0hg(p) =  70._r8 * 1.5e-5_r8 / ustar(p) * exp( -7.2_r8 * ustar(p)**(0.5_r8) * (abs(tstar))**(0.25_r8)) ! Consistent with BareGroundFluxes
```
ekluzek, line 620: Remove commented out code.

ekluzek (changes requested):
@RonnyMeier thanks so much for your work on this. This looks really good to me overall. I like the namelist control you added in, as well as the use of case statements for choices. You did a great job with the build-namelist work as well.

One thing that would help with this is that this currently includes changes outside of surface roughness. It also includes running PFT's on their own column, which is in a different PR to come in (#1249), and local time history averaging which already came in PR #1374. It would help to see this without both of those.

It'll likely help prevent conflicts in merging forward, if the local time history changes are removed from here. The update to a newer version will automatically bring it back in anyway.

I also think it's good to have the PFT on a column as a separate PR as well. I remember discussing that one and I don't remember what our conclusion was on including it. I'll make a point of having the software group discuss it more again next week. So removing that would allow us to bring it in separately.

I have some detailed questions that I've put in the code as well. Mainly there is some commented out code that looks like it's for a different option. If you had time to develop it into a third option I think that would be the best path forward. But, because we only have a week, I think the best would be to just remove the commented out code.

I also noticed repeated hardcoded constants. Some of these existed before your work in the code, so aren't really your fault, but still would be good to change into parameters. If they are used only in one file, making them local parameters would be fine. If they are used in more than one file adding them to clm_varcon might be better. If they are something that someone might want to tune, they could be put on the paramsfile. Just elevating them to a local parameter helps though in any case.

Separately I also suggest that MeirXXXX become Meir2022, but we can discuss that. And feel free to discuss any of this, there's always room for talking this over.

But, good work here. Thanks for your contribution. Best of luck on your future endeavors, outside of academia.

RonnyMeier:
Thanks @ekluzek for reviewing the code. I don't have any strong opinions regarding the tag. Meier2022 should most likely be appropriate. We can also go for MeierDavin2022 or GMD2022. Not sure if CESM has any guidlines on how to choose such tags. 

The commented-out code sections marked with "! --> Use this for CLM-YY" were used for the sensitivity tests that are presented in the appendix of the GMD study. Those can all be removed (sorry I forgot to do this). What could be considered to be included as an additional option is the formulation of Owen and Thomson (1963) to compute z0hg and z0qg. 

Also, I realized that we actually opened an issue for this development a couple of months ago: New parameterization of vegetation surface roughness for forests #1316

ekluzek:
We talked about this a bit this morning. 

One aspect that we should keep in mind about this is how this interacts with FATES. Our testing ensures we won't break FATES with it. Two questions are can part of it be turned with FATES, and secondly can/should FATES be updated to bring these kind of changes to the FATES side of things? Before we do that I'd like to make a little more progress so the code is a little more clearer for a FATES scientist to evaluate.

Specifically, the next steps for this are:
- Fix the subscript overflow in DEBUG mode
- Change the MeierXXXX namelist tag to Meier2022
- Tag that as a branch tag
- Remove the Local time changes
- Remove the PFT on separate columns option
- Update to the latest model version
- Make sure all model configurations work with the new option (or disable using them together)
- Run more extensive testing to make sure we agree this is good to come in
- We may remove the 2D option that requires changes to surface datasets
- Some of the parameters should be made into parameters or placed on the paramsfile
- Possibly have it turned on by default for clm5_1 physics

ekluzek:
Note that SMS_D_Ly2.f10_f10_mg37.I1850Clm51SpNoAnthro.cheyenne_intel.clm-Meier2022_surf_rough is failing at startup because "Forcing height is below canopy height for patch index".

ekluzek:
@olyson and @swensosc I'm running into some science issues with this branch that I'll need you to look into. I think I should go ahead with our plan to bring it up to the latest tag, and then pass it over to one or both of you to look into. Just FYI for now, unless you or others think you should start looking into it now. I'm adding the "next" label so that we'll talk about this Thursday.

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -329,7 +342,22 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
...
-            z0hg_patch(p) = z0mg_patch(p) / exp(params_inst%a_coef * (ustar(p) * z0mg_patch(p) / 1.5e-5_r8)**params_inst%a_exp)
+
+            select case (z0param_method)
+            case ('ZengWang2007')
+               z0hg_patch(p) = z0mg_patch(p) / exp(params_inst%a_coef * (ustar(p) * z0mg_patch(p) / 1.5e-5_r8)**params_inst%a_exp)
+            case ('MeierXXXX')
+               ! After Yang et al. (2007)
+               z0hg_patch(p) = 70._r8 * 1.5e-5_r8 / ustar(p) * exp( -7.2_r8 * ustar(p)**(0.5_r8) * (abs(tstar))**(0.25_r8))
+               
+               ! After Owen and Thomson (1963)
```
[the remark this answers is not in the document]
ekluzek: Done

src/biogeophys/BiogeophysPreFluxCalcsMod.F90
```diff
@@ -154,8 +160,37 @@ subroutine SetZ0mDisp(bounds, num_nolakep, filter_nolakep, &
...
+
+               U_ustar = 4._r8 * (pftcon%z0v_Cs(patch%itype(p)) + pftcon%z0v_Cr(patch%itype(p)) *  (pftcon%z0v_LAImax(patch%itype(p))) & 
+                         / 2._r8)**(-0.5_r8) /  (pftcon%z0v_LAImax(patch%itype(p))) / pftcon%z0v_c(patch%itype(p))
+
+               z0m(p) = htop(p) * (1._r8 - displa(p) / htop(p)) * exp(-0.4_r8 * U_ustar + &
+                           log(pftcon%z0v_cw(patch%itype(p))) - 1._r8 + pftcon%z0v_cw(patch%itype(p))**(-1._r8))
+
+            end if
+
+            ! --> Use this for CLM-Ya08
```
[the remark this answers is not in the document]
ekluzek: Done

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -861,12 +873,50 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+         select case (z0param_method)
+         case ('ZengWang2007')
+            lt = min(elai(p)+esai(p), tlsai_crit)
+            egvf =(1._r8 - alpha_aero * exp(-lt)) / (1._r8 - alpha_aero * exp(-tlsai_crit))
+            displa(p) = egvf * displa(p)
+            z0mv(p)   = exp(egvf * log(z0mv(p)) + (1._r8 - egvf) * log(z0mg(c)))
+
+         case ('MeierXXXX')
+            lt = max(0.00001_r8,elai(p)+esai(p))
+            displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(7.5_r8 * lt)**0.5_r8)) / (7.5_r8*lt)**0.5_r8)
```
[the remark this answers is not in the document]
ekluzek: This still should happen.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: Done.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: I removed this.

src/biogeophys/FrictionVelocityMod.F90
ekluzek: Done

src/biogeophys/FrictionVelocityMod.F90
ekluzek: This should still be done.

src/biogeophys/LakeFluxesMod.F90
ekluzek: Done

src/biogeophys/LakeFluxesMod.F90
ekluzek: Done

src/biogeophys/LakeFluxesMod.F90
ekluzek: Done

src/biogeophys/LakeFluxesMod.F90
ekluzek: Done.

ekluzek:
I've got most of the updates that needed to happen that were talked about above. It's updated to ctsm5.1.dev077 and other changes removed and the namelist option is Meier2022 now. There are some tests that fail though, including some that don't turn the new changes on.

ERP_D_Ld5_P48x1.f10_f10_mg37.I2000Clm50BgcCru.izumi_nag.clm-noFUN_flexCN		
ERP_D_Ld5_P48x1.f10_f10_mg37.I2000Clm50BgcCru.izumi_nag.clm-reduceOutput
SMS_D_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.izumi_intel.clm-tracer_consistency		
SMS_D_Ld5.f10_f10_mg37.I2000Clm50BgcCrop.izumi_nag.clm-irrig_alternate	
ERP_D_Ly3.f09_f09_mg17.I2000Clm51Bgc.cheyenne_intel.clm-Meier2022_surf_rough_all_f09NonCrop		
ERS_Ly3.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-Meier2022_surf_rough		
LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart		
SMS_D.f10_f10_mg37.I2000Clm51Bgc.cheyenne_gnu.clm-Meier2022_surf_rough		
SMS_D.f10_f10_mg37.I2000Clm51Bgc.cheyenne_intel.clm-Meier2022_surf_rough		
SMS_D_Ln1.f09_g17.I1850Clm50BgcNoAnthro.cheyenne_intel.clm-Meier2022_surf_rough

The surf_rough tests fail on izumi, because the files are only on cheyenne right now.

ekluzek:
OK, one simple thing I needed to do was update the Ronny Meier paramsfile to the latest, and I did that so that fixes some problems. But, now all of the tests are failing with the following problem. I saw this with the testing in ctsm5.1.dev029, but I didn't see all of the tests failing.

>  Error: Forcing height is below canopy height for patch index 
> iam = 0: local  patch    index = 13
> iam = 0: global patch    index = 1935
> iam = 0: global column   index = 1366
> iam = 0: global landunit index = 457
> iam = 0: global gridcell index = 145
> iam = 0: gridcell longitude    =   15.0000000
> iam = 0: gridcell latitude     =   30.0000000
> iam = 0: pft      type         = 5
> iam = 0: column   type         = 1
> iam = 0: landunit type         = 1
>  ENDRUN:
>  ERROR: ERROR in /glade/work/erik/ctsm_worktrees/surf_rough/src/biogeophys/CanopyFluxesMod.F90 at line 975

ekluzek:
I met with @olyson and he is going to verify my changes as well as look into some of the problems I ran into. There could be a mistake in my changes to prevent floating point overflow. Or there could be something with the updated code that's working differently than the version that Ronny ran with.

olyson:
Regarding the forcing height error, I think this statement in biogeophys/BiogeophysPreFluxCalcsMod.F90:

            if ( is_first_step() .or. get_nstep() <= GetBalanceCheckSkipSteps() ) then

should be:

            if ( is_first_step() .or. get_nstep() <= GetBalanceCheckSkipSteps()-1 ) then

because at that point htop is available and displacement height can be calculated.  Otherwise, the forcing height will be calculated as              
forc_hgt_u_patch(p) = forc_hgt_u(g) + z0mg(c) + displa(p) = forc_hgt_u_patch(p) = 30m, since z0mg and displa are zero.  

Then displa is recalculated in CanopyFluxes and that calculation can be greater than 30m and hence zldis(p) = forc_hgt_u_patch(p) - displa(p) is negative.

Tried that and the forcing height errors went away.

ekluzek:
@olyson awesome. Thanks for figuring that out.  I'll add that in and make sure my tests work now.

olyson:
This check you put in is still triggering, so I was going to look at that next:

ustar*thvstart is positive and has to be negative

olyson:
More generally, there is inconsistency between the forcing height which is based on this calculation in BiogeophysPreFluxCalcsMod.F90:

displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(7.5_r8 * (pftcon%z0v_LAImax(patch%itype(p))))**0.5_r8)) &
                           / (7.5_r8*(pftcon%z0v_LAImax(patch%itype(p)) ))**0.5_r8)

and the later calculation of displa in CanopyFluxes:

displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(7.5_r8 * lt)**0.5_r8)) / (7.5_r8*lt)**0.5_r8)

So I think we should be updating the forcing height for this new displa in CanopyFluxes.

Tried this and it fixed the negative ustars in my test that were occurring in nsteps 0,1,2.  Error was occurring because zldis, while positive, was less than the roughness length, so ln(zldis/z0) was negative.

RonnyMeier:
> More generally, there is inconsistency between the forcing height which is based on this calculation in BiogeophysPreFluxCalcsMod.F90:
> 
> displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(7.5_r8 * (pftcon%z0v_LAImax(patch%itype(p))))*_0.5_r8)) & / (7.5_r8_(pftcon%z0v_LAImax(patch%itype(p)) ))**0.5_r8)
> 
> and the later calculation of displa in CanopyFluxes:
> 
> displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(7.5_r8 * lt)**0.5_r8)) / (7.5_r8*lt)**0.5_r8)
> 
> So I think we should be updating the forcing height for this new displa in CanopyFluxes.

Hi,

I think that's a quick one to answer: Those displa and z0mv values are a first guess only. In the default version of the model, they are computed as if the LAI+SAI went towards infinity (i.e., the highest possible values with this vegetation height and type). With the proposed formulations, z0mv has a maximum at LAImax. Therefore, I used this value for a first guess, analogues to the default implementation, as indicated by the comment "Compute as if elai+esai = LAImax - LAIoff in CanopyFluxes". Note that LAIoff is now 0. So, the "- LAIoff" could be removed in the comment.

I hope that makes sense,
Ronny

olyson:
Yes, that makes sense, thanks.

ekluzek:
@olyson to check into the branch we would need to get you collaborator access. We can ask @RonnyMeier to do that for you. But, in the meantime if you have proposed changes that you can show me, I can check them in and try it out.

olyson:
You can try:

/glade/work/oleson/ctsm_MeierZ0/src/biogeophys/BiogeophysPreFluxCalcsMod.F90
/glade/work/oleson/ctsm_MeierZ0/src/biogeophys/CanopyFluxesMod.F90

The cheyenne tests you listed above all pass now except for:

LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart
SMS_D_Ln1.f09_g17.I1850Clm50BgcNoAnthro.cheyenne_intel.clm-Meier2022_surf_rough

The first doesn't have Meier2022 on and I haven't found any useful error messages for the second test yet.

olyson:
Actually, what I have now in /glade/work/oleson/ctsm_MeierZ0/src/biogeophys/CanopyFluxesMod.F90 makes more sense, i.e.,

           z0hv(p)   = z0mv(p)
           z0qv(p)   = z0mv(p)

+          ! Update the forcing heights
+          forc_hgt_u_patch(p) = forc_hgt_u(g) + z0mv(p) + displa(p)
+          forc_hgt_t_patch(p) = forc_hgt_t(g) + z0hv(p) + displa(p)
+          forc_hgt_q_patch(p) = forc_hgt_q(g) + z0qv(p) + displa(p)
+
       end do

olyson:
As @ekluzek noted, the tests were also failing due to negative humidity in HumanIndexMod, so he turned that off by default.  I've turned that on now in my latest version of the code and the tests are still failing.
In this particular test:

SMS_D.f10_f10_mg37.I2000Clm51Bgc.cheyenne_intel.clm-Meier2022_surf_rough

there is one pft that is triggering the error.  CanopyFluxes is producing a slightly negative 2-m specific humidity which is then converted to relative humidity and then fed into HumanIndexMod.  This seems to be occurring under very stable conditions, with tall trees with large elai+esai, a displacement height near the canopy top and a very small ustar (numbers below):

zeta = 100
z0h = z0q = 4.29
htop = 36
displa = 31.3
ustar = 0.0114
elai+esai = 7.1

With the stress indices turned off, the test runs to completion even with the negative humidities.
I tried restricting the max zeta to 60 (instead of 100) and that eliminated the negative humidities, at least for that test.
Not a real solution though.

olyson:
Regarding this test that was failing (despite not having Meier2022 on):

LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart

This was failing with:

59:ERROR in CompareBulkToTracer: tracer does not agree with bulk water
59:Called from: after first stage of hydrology
59:Variable: qflx_snomelt_accum_col
59:First difference at index: 15
59:Bulk  :   0.94434303138415165E-04
59:Tracer:   0.98722208432612920E-04
59:ratio:   0.10000000000000000E+01
59:Bulk*ratio:   0.94434303138415165E-04

This is a new variable that was added to WaterFluxType.F90 to be used in the new soil roughness calculation.  After consultation with @billsacks , I've moved this to WaterDiagnosticBulkType since it is a diagnostic variable (diagnostic in the sense that it isn't a fundamental state variable) and only used in the roughness calculation.  We don't need any tracer/isotope quantities associated with this.  I've also renamed it from qflx_snomelt_accum to snowmelt_accum since it isn't a flux.
That test passes now.

olyson:
This test:

SMS_D_Ln1.f09_g17.I1850Clm50BgcNoAnthro.cheyenne_intel.clm-Meier2022_surf_rough

is failing because of:

20220310 182753.068 ERROR            PET0205 src/addon/NUOPC/src/NUOPC_Base.F90:950 Invalid argument  - setClock timeStep=10800s is not a divisor of runDuration=1800s

Found this in the PET log files.  This seems to be a 1 time-step run.  Seems unrelated to the roughness parameterization...

billsacks:
> This test:
> 
> SMS_D_Ln1.f09_g17.I1850Clm50BgcNoAnthro.cheyenne_intel.clm-Meier2022_surf_rough
> 
> is failing because of:
> 
> 20220310 182753.068 ERROR PET0205 src/addon/NUOPC/src/NUOPC_Base.F90:950 Invalid argument - setClock timeStep=10800s is not a divisor of runDuration=1800s

If you are using an active runoff model with standard coupling interval of 3 hours, then the test needs to run for at least 3 hours.

Is there a reason why this test needs to be at f09 resolution? I'm in the process of trying to reduce our high resolution testing because it is expensive to run these high-resolution tests with every run of the test suite, and they can occasionally hold up testing by getting stuck in the queue. Making them _Ln1 doesn't really help because most of the time in these short high-res tests is spent in model initialization, so whether it's one time step or a few days doesn't really matter much.

olyson:
Thanks, that explanation makes sense.  I didn't create these tests so I'll leave it up to @ekluzek to respond regarding the question about resolution.

olyson:
@ekluzek it seems like the ZengWang2007 version of z0m in FrictionVelocity needs some protection for log10 as is done for Meier2022.  

       case ('ZengWang2007')
          if (frac_sno(c) > 0._r8) then
             if(use_z0m_snowmelt) then
                z0mg(c) = exp(1.4_r8 * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8))-0.31_r8) / 1000._r8
             else
                z0mg(c) = this%zsno
             end if

Add an if statement:

                if ( snomelt_accum(c) < 1.e-5_r8 )then
                    z0mg(c) = exp(1.4_r8 * -rpi/2.0_r8 -0.31_r8) / 1000._r8

Another question is whether we should even be using that expression for z0mg for ZengWang2007.  It used to be just zsno right?

ekluzek:
@billsacks the noAnthro test needs to be f09 because it's the only resolution for Potential Vegetation surface dataset. But, for now I'm just adding more tests than are really needed while we are working on the code. I don't think I'll keep that test when it comes to master. So I'll add a note to that effect for some of the tests that should be removed. One thing that I really want to do is to run the whole test suite with the Meier surface roughness on for everything to make sure everything works.

Anyway, @olyson I'll work with this test and add a note to it that we won't keep it long term.

@olyson you probably have more code that could be checked in. I asked @RonnyMeier to add collaborator access, so you should likely checkin your latest changes once you get that.

olyson:
I have access now, thanks Ronny.

olyson:
I think the one remaining problem now is the negative 2-m humidity.  I can solve that by limiting zeta in stable conditions, but not sure if we really want to do that or if there is another way to solve the problem.  I think we need to consult with some of the group at this point.

ekluzek:
@olyson could you call a meeting with the relevant people that should be involved in discussing that last problem?

olyson:
You had also asked me to check the expression for z0mg when snomelt_accum = 0.  I think your formulation is correct.

olyson:
Gordon, Sean, Erik, and I met to discuss the negative 2-m humidity issue.  Short-term plan for now is to try setting a max zeta of 2 (or maybe 20 since that passes in the tests I've been running) in CanopyFluxes with heat storage biomass and Meier2022 on and make the zeta calculation in FrictionVelocity consistent with this and assess differences compared to a control with the current settings.
Longer-term plan is to evaluate the consistency of the stability calculations in CanopyFluxes, FrictionVelocity, BareGroundFluxes, and LakeFluxes and formulate a plan to resolve any inconsistencies.

ekluzek:
@olyson it looks like we enter zetamaxstable from the namelist, but then essentially hardcode it to 100 when biomass heat storage is on. I think what we should do is to always enter it from the namelist (including a different value when BHS is on), which would allow us to more easily change it as needed. It's odd to have it handled differently just for BHS, so it would also improve the code. I'll make this as a change to the code either here or in a BFB code change.

zetamaxstable could also be moved to the paramsfile, but the problem is that it'll be different for different physics versions (clm4_5, clm5_0, etc.) or settings (BHS). So I think it's best if it's left on the namelist file. I don't think this is going to be a parameter that you would optimize the model for, so being on the namelist should be fine.

olyson:
The namelist setting seems like a reasonable way to go.  Right now the way I have the hardcoding setup results in bfb with ZengWang.  The problem is that zetamaxstable also applies to BareGroundFluxes and LakeFluxes (currently set to 0.5 in clm5.0 and clm5.1, and 2.0 in clm4.5).  I think we'll want to make these consistent with whatever we end up with in CanopyFluxes, which would change answers.

ekluzek:
Yes, any part that will change answers I'd like to add to a small answer changing tag. I'll turn this into an issue so it can be tracked.

olyson:
@ekluzek @dlawrenncar @swensosc , I've completed three I2000 simulations, all with Meier22 and BHS on and HumanIndexMod active which catches any occurrences of negative 2-m humidity.  The first is a control (with zetamax=100), the second with zetamax=20 (the maximum zeta required to pass a test that was failing), and the third with zetamax=2 (what we use for CLM4.5).  Regular and ILAMB diagnostics are here:

https://webext.cgd.ucar.edu/I2000/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_zetamax20_2000/lnd/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_zetamax20_2000.171_200-ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_CON_2000.171_200/setsIndex.html

https://webext.cgd.ucar.edu/I2000/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_zetamax2_2000/lnd/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_zetamax2_2000.171_200-ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_CON_2000.171_200/setsIndex.html

https://webext.cgd.ucar.edu/I2000/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_zetamax20_2000/lnd/_build_Meier22/

The main thing I see is an increase in TSA mostly at higher latitudes, with larger increases in daily minimum temperature than daily maximum, and in boreal winter than summer.  A summary can be found on slides 7 and 8 here:

https://docs.google.com/presentation/d/1u6ycr7F97QYYRcRfEdD9yIxH75diUx2r/edit?usp=sharing&ouid=104026900087612515425&rtpof=true&sd=true

I'm inclined to set zetamax to 20 as long as it is robust across the test suite, although see slide 9.

olyson:
Regarding our meeting on this subject, I conducted a quick sensitivity study on the effects of forcing height on zeta.  See slides 14 and 15 of this presentation:

https://docs.google.com/presentation/d/1weBcmdh7VAzbVe-bOmu7K4F9OLlUZDvQ/edit?usp=sharing&ouid=104026900087612515425&rtpof=true&sd=true

Shown is the max monthly zeta for forcing height default (30m) and forcing height 10m, for July and December.  There is quite a bit of sensitivity to forcing height.  In particular, max zeta is around 100 everywhere with forcing height 30m and this is reduced in most land areas with forcing height 10m. @swensosc

wwieder:
Thanks for doing this so quickly, Keith.  I'm also including Gordon on this
thread.

As Sean suggested, the zeta calculation is quite sensitive to the forcing
height.

Back to where we were yesterday, what's the best way to handle this and the
zeta cap in land only and coupled simulations.

olyson:
I'm curious to see what is happening with zeta in a coupled simulation, so I'm going to run a short F-case

dlawrenncar:
Interesting.  Of course, I guess lots more than just zeta is different with
the lower forcing height, but it seems like with the lower forcing height,
the very large zeta is less frequent, but not absent, so you could still
get negative humidities, right?  So, it seems to me that we should probably
should just go ahead and implement the cap at 20.

Then, a separate topic is thinking about what the forcing height should be
for reanalysis datasets and whether or not there is a way to decide what
would be best to do there.

olyson:
Documenting some more findings:
1. Meier22 simulation with forc_hgt=10m crashes in year 7 with a carbon balance error, also triggered some checks for negative aerodynamic resistance that @ekluzek had put in.
2. Meier22 simulation with forc_hgt=2m crashes in the first month with a NaN to the coupler, also some negative aerodynamic resistances
3. CAM6CTSM51 simulation with default roughness has maximum zeta values that are generally much less than in land-only mode.  See slide 16 of the ppt above.

olyson:
I found out that the zeta output is for an average of the vegetated pfts only.  The zeta variable is local in other modules like BareGroundFluxes, LakeFluxes, etc.  That probably explains why there are fill values for some grid cells in the Sahara, Arabian Peninsula, etc.

olyson:
I currently have a cap of 20 on the branch.  If we want to use that cap for all surfaces, then I think we can just use the namelist variable zetamaxstable=20 everywhere.  If we want to maintain a cap on zeta of 0.5 for surfaces other than vegetated then we'll need some way of distinguishing between surface types in FrictionVelocity since all of the flux modules call that module, and perhaps two namelist variables, zetamaxstable_veg and zetamaxstable_other, or something like that.

swensosc:
It would be worth outputting the zeta values for bare ground, so we can see
what's going on over those gridcells.  If I understand correctly, the maps
@oleson showed aren't really telling us what's going on over bare ground
(deserts).

olyson:
I suspect it's maxing at 0.5, but I can setup zeta as pointers in BareGroundFluxes, LakeFluxes, and UrbanFluxes to get that output.

olyson:
Please see slides 17-18.  The dark blues areas, e.g., most of the Sahara, Arabian Peninsula, Antarctica, Greenland have max zeta of 0.5.

swensosc:
Sorry I didn’t say that well. I meant we should see what values occur in the absence of a limit for bare soil patches. Do they also tend to high zeta values or is it mainly occurring in vegetated patches?

olyson:
Ah, ok, makes sense.

olyson:
I set zetamaxstable=100 for all surfaces.  It looks like those surfaces want to go to 100 also.  See slides 19 and 20.

olyson:
A few notes from our meeting:

1) We will use zetamaxstable=2 over all surfaces.
2) Run a new simulation with forc_hgt=30m and zetamaxstable=2 to assess the impact of changing the forcing height to 10m under zetamaxstable=2 conditions.
3) File an issue to put the forcing height for the GSWP3 forcing on the forcing files and streams so that it is read in from the files and not hardcoded in the datm code.  This may end up being either 10m or 30m depending on our analysis of the above comparison.

Note that the negative ustar problem encountered in the forc_hgt=10, zetamaxstable=100 simulation appears to be due to unrealistic behavior in the stabiliity function (the denominator in the equation below) under very stable conditions.  This is similar to what was happening with the negative 2-m specific humidity problem.  

image

olyson:
Diagnostics for a simulation with forc_hgt=10m and zetamaxstable=2 compared to a simulation with forcing height=30m and zetamaxstable=2 are here (showing the impacts of changing the forcing height to 10m from 30m):

https://webext.cgd.ucar.edu/I2000/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_forchgt10zetamax2allsfc_2000/lnd/ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_forchgt10zetamax2allsfc_2000.171_200-ctsm51_ctsm51d77MeierZ0_1deg_GSWP3V1_forchgt30zetamax2allsfc_2000.171_200/setsIndex.html

The model ran successfully for 200 years in an I2000 compset.

dlawrenncar:
Thanks, I've reviewed a bunch of plots and it's not obvious that
anything is wrong.  Pretty significant differences in fluxes in the
tropics, but perhaps hard to distinguish whether or not this better or
worse.

In the context of understanding this change and helping us understand
possible transition to CRUJRA as our 'main' forcing dataset, could we also
see a run with 10m forcing height with CRUJRA.  I think it would be
interesting to try to turn on using downwelling longwave radiation to see
whether or not prior issues with downwelling longwave radiation persist or
not.  So, perhaps two additional runs, both CRUJRA, both with 10m forcing
height and zetamaxstable=2 and one using FLDS and one not.  Though, I guess
the prior runs you looked at were year 2000 runs, not transient.

Hmmmm.  Maybe we should discuss at next opportunity whether now is an
appropriate time to explore this and to define a new baseline configuration
in terms of forcing datasets and forcing height, etc.

wwieder:
I agree @dlawrenncar this quickly gets into questions about preferred configurations for the model moving forward.  Looking more carefully at the transition to CRUJRA-TRENDY seems like a good idea

olyson:
Per the CLM business meeting on 06/09/22, we decided to use zetamaxstable=2 for all surfaces, and further investigate setting the  forcing height to 10m along with implementing CRUJRA as our default dataset but as a separate issue.
I've made a change to the CLMBuildNamelist.pm so that the proper zetamaxstable is chosen based on z0param_method.
I've also fixed a problem with setting zeta to a global variable in UrbanFluxesMod.F90.
I ran the clm_short test suite to compare with ctsm5.1.dev077 with z0param_method='ZengWang2007' and result are bfb.
So I think this is finally ready for review and full testing, although I'm not sure if all of the items on Erik's original list have been completed:

Fix the subscript overflow in DEBUG mode
Change the MeierXXXX namelist tag to Meier2022
Tag that as a branch tag
Remove the Local time changes
Remove the PFT on separate columns option
Update to the latest model version
Make sure all model configurations work with the new option (or disable using them together)
Run more extensive testing to make sure we agree this is good to come in
We may remove the 2D option that requires changes to surface datasets
Some of the parameters should be made into parameters or placed on the paramsfile
Possibly have it turned on by default for clm5_1 physics

ekluzek:
@dmleung made a summary of the differences between the 2005 and the 2016 Prigent datasets that I'm uploading here:

[Summary of Prigent's two roughness datasets.docx](https://github.com/ESCOMP/CTSM/files/9245650/Summary.of.Prigent.s.two.roughness.datasets.docx)

ekluzek:
In @dleung work #1712 has the Prigent data as drag partition factor which relates to the Meir data as follows:

```
f_eff.r = 1 - ( log(z0a/100/z0s) / log(0.7*(10/z0s)^0.8) )
z0s = 130e-6 / 15    # m, smooth roughness length, White (2006) equation
```

We should compare the two datasets to make sure they are reasonably similar. We think they will be different because the @dmlenung data is the minimum.

ekluzek:
The thinking now is that for the 2D data here, we'll use the data from Leung 2022 (see #1712) and convert it back to what's needed here. So we won't use the Meier data that was processed partially because it's only over non-vegetated areas. We also think the choice of Leung to use the minimum makes sense physically.

slevis-lmwg:
TODOs
1) Update to latest tag
2) z0param_method can be ZengWang2007 or Meier2022
zetamax_stable = 2 regardless of method but will depend on physics (clm45=2, clm5=0.5, clm51=2)
use_z0mg_2d should be removed from /testmods
3) Run test-suites
4) Refactor some of the "magic numbers" with parameter names from the corresponding publication

slevis-lmwg:
@RonnyMeier @ekluzek 
I'm ready to push the update to ctsm5.1.dev129 and need permission to do so to this PR.

@RonnyMeier can you make me a collaborator on your fork of CTSM? On your fork go to the "Settings" tab and then "Collaborators" and you should see a way to add a collaborator where you can invite @slevis-lmwg and give me permission to work on your fork.

ekluzek:
@slevis-lmwg we haven't been able to get ahold of Ronny, so we can't push to his forks. What you'll have to do is to branch from his branch and create a branch on your own fork that will include this one.

We'll want to keep both PR's open for at least a little while to make sure everything is transferred over. And this PR will autoclose when the new branch brings it to main-dev. so it's actually OK to leave it open until then.

slevis-lmwg:
Superseded by #2045

ekluzek:
We had problems with this update as we had turned Meier on for FATES (and shouldn't have). In looking over things here we noticed that my update here:

https://github.com/ESCOMP/CTSM/pull/1596#issuecomment-1017989562

Should have been more explicit about getting scientific oversight from FATES about whether it should be turned on for FATES. We plan to make sure there is a FATES scientific review step (at least asking some level of examination by a FATES scientists, as well making sure there is a FATES team approval for something that changes answers for FATES). In the future with calibrated global FATES the LMWG science liaison will run answer changes for FATES and compare to the previous science baselines to make sure results aren't out of whack.

For more on that discussion see this document:

https://docs.google.com/document/d/1Hjry2ew_C6SXoRG7ChISbm-V1J5exwcyQJzxPz1prK8/edit?tab=t.0#heading=h.1sq2v9xtt2vd

Issue #1048, opened 2020-06-18
Rename master branch


billsacks:
The `master` term can be offensive, and so GitHub and others are encouraging switching to a more neutral term. `main` seems to be a good choice; some others would be `latest`, `trunk`, `stable` or `develop`. It sounds like GitHub will be publishing guidance on how to accomplish this soon. Otherwise, a quick search turned up these recent guides:

https://dev.to/rhymu8354/git-renaming-the-master-branch-137b

https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx

I'm calling this a "documentation" issue because the pieces needed are probably more on the documentation side than anywhere else, and I don't think this will require any actual code changes.

billsacks:
@ckoven thanks for suggesting this idea on the FATES side.

wwieder:
Yep, let's make this move & make the transition as smooth as possible for cloned repos.

billsacks:
https://github.com/github/renaming

ekluzek:
As part of this we should also change our documentation (README files and the like) that refer to the master branch. Also likely in our wiki pages as well.

rgknox:
I was reading the second guide you posted @billsacks , thanks.  
It seems like this will only require two commands to make the change.

ekluzek:
I think github.com has done all of the changes they plan to implement to help this change. They make it easy on the web-interface, and it automatically forwards branch protections and even lets people know about the change on a "push". The web-interface also lets you know what you need to do. When you go to your fork it does for example point out that you need to change it. And the first time you go to a repository with the change there's a pop up that tells you about the change.

Clones and forks do have to change. The commands for us in git on a clone are like this (github also tells you about the need to do this)...

```
git branch -m master main
git fetch escomp
git branch -u escomp/main main
```
(Note the details of above does depend on how you have named your remotes, this assumes you use the escomp convention)

I've done this in PTCLM, just to see how easy it was and what had to be done.

billsacks:
Awesome, thanks @ekluzek for looking into this more. I have edited the top-level comment to put a checklist of things I can think of that need to be done. Feel free to edit it and add anything else you can think of.

wwieder:
Is this something we need to help communicate to model users / developers, or can new forks & PRs just merge these changes without too much disruption?

billsacks:
> Is this something we need to help communicate to model users / developers, or can new forks & PRs just merge these changes without too much disruption?

[...]

PR #1605, opened 2022-01-13
Use UNSTRUCT mesh files instead of SCRIP when running mkmapdata.sh

slevisconsulting:
### Description of changes

Switching from SCRIP to UNSTRUCT mesh files saves computation time when running mkmapdata.sh.

I have started by making the switch for the SRC mesh files used by mkmapdata.sh.

Repeating for the DST mesh files requires generating the UNSTRUCT mesh files for every DST resolution. Since the switch to UNSTRUCT changes answers, it seems worth completing the work for both the SRC and the DST mesh files.

As part of this PR, I will add (i.e. rimport) the UNSTRUCT files to the repository.

### Specific notes

@billsacks @mvertens 

#648 

Yes. We expect diffs in the mkmapdata.sh output files (known as mapping or weight files), which means diffs in the resulting fsurdat files. We agreed (2022/1/12) that an eyeball check will tell us whether we must perform more rigorous checks.

Changes should be transparent to users who will use the mesh files that we provide. Even users who use their own raw datasets to generate their own fsurdat files, should not notice a change. They will still generate their own respective mesh files. Ideally they will generate UNSTRUCT mesh files, but codes will still work with SCRIP.

Testing performed, if any:
My original time estimate here was incorrect. Updated time estimate provided later in the PR.

### DIFF
--- a/bld/namelist_files/namelist_defaults_ctsm_tools.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm_tools.xml
@@ -112,27 +112,30 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <scripgriddata hgrid="tx0.1v2"    lmask="tx0.1v2"   >/glade/proj3/cseg/mapping/grids/tx0.1v2_090127.nc</scripgriddata>
 
 <!-- UNSTRUCT grid data files -->
-<!--scripgriddata frm_hgrid="0.5x0.5"   lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.5x0.5_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="0.5x0.5" lmask="nomask">UNSTRUCT</scripgriddata_type>
-<scripgriddata frm_hgrid="0.25x0.25" lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.25x0.25_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="0.25x0.25" lmask="nomask">UNSTRUCT</scripgriddata_type>
-<scripgriddata frm_hgrid="0.125x0.125" lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.125x0.125_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="0.125x0.125" lmask="nomask">UNSTRUCT</scripgriddata_type>
-<scripgriddata frm_hgrid="10x10min" lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_10x10min_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="10x10min" lmask="nomask">UNSTRUCT</scripgriddata_type>
-<scripgriddata frm_hgrid="5x5min"    lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_5x5min_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="5x5min" lmask="nomask">UNSTRUCT</scripgriddata_type>
-<scripgriddata frm_hgrid="3x3min"    lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_3x3min_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="3x3min" lmask="nomask">UNSTRUCT</scripgriddata_type>
-<scripgriddata frm_hgrid="0.9x1.25"  lmask="nomask"
-   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.9x1.25_nomask_c200129.nc</scripgriddata>
-<scripgriddata_type frm_hgrid="0.9x1.25" lmask="nomask">UNSTRUCT</scripgriddata_type-->
+<unstructdata frm_hgrid="1km-merge-10min" lmask="HYDRO1K-merge-nomask"
+   >lnd/clm2/mappingdata/grids/UGRID_1km-merge-10min_HYDRO1K-merge-nomask_c130402.nc</unstructdata>
+<unstructdata_type frm_hgrid="1km-merge-10min" lmask="HYDRO1K-merge-nomask">UGRID</unstructdata_type>
+<unstructdata frm_hgrid="0.5x0.5"   lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.5x0.5_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="0.5x0.5" lmask="nomask">UNSTRUCT</unstructdata_type>
+<unstructdata frm_hgrid="0.25x0.25" lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.25x0.25_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="0.25x0.25" lmask="nomask">UNSTRUCT</unstructdata_type>
+<unstructdata frm_hgrid="0.125x0.125" lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.125x0.125_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="0.125x0.125" lmask="nomask">UNSTRUCT</unstructdata_type>
+<unstructdata frm_hgrid="10x10min" lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_10x10min_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="10x10min" lmask="nomask">UNSTRUCT</unstructdata_type>
+<unstructdata frm_hgrid="5x5min"    lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_5x5min_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="5x5min" lmask="nomask">UNSTRUCT</unstructdata_type>
+<unstructdata frm_hgrid="3x3min"    lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_3x3min_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="3x3min" lmask="nomask">UNSTRUCT</unstructdata_type>
+<unstructdata frm_hgrid="0.9x1.25"  lmask="nomask"
+   >lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.9x1.25_nomask_c200129.nc</unstructdata>
+<unstructdata_type frm_hgrid="0.9x1.25" lmask="nomask">UNSTRUCT</unstructdata_type>
 
 <!-- land masks for different types of mksurfdata_map input files -->
 <lmask type="lak"              >nomask</lmask>
--- a/bld/namelist_files/namelist_defaults_ctsm_tools.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm_tools.xml
@@ -22,44 +22,44 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- Finite-Volume grids -->
 <scripgriddata hgrid="0.23x0.31" lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_0.23x0.31_nomask_c110308.nc</scripgriddata>
 <scripgriddata hgrid="0.47x0.63" lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_0.47x0.63_nomask_c170914.nc</scripgriddata>
-<scripgriddata hgrid="0.9x1.25"  lmask="nomask" >lnd/clm2/mappingdata/grids/0.9x1.25_c110307.nc</scripgriddata>
-<scripgriddata hgrid="1.9x2.5"   lmask="nomask" >lnd/clm2/mappingdata/grids/1.9x2.5_c110308.nc</scripgriddata>
+<scripgriddata hgrid="0.9x1.25"  lmask="nomask" >share/meshes/fv0.9x1.25_141008_polemod_ESMFmesh.nc</scripgriddata>
+<scripgriddata hgrid="1.9x2.5"   lmask="nomask" >share/meshes/fv1.9x2.5_141008_ESMFmesh.nc</scripgriddata>
 <scripgriddata hgrid="2.5x3.33"  lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_2.5x3.33_nomask_c110308.nc</scripgriddata>
-<scripgriddata hgrid="4x5"       lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_4x5_nomask_c110308.nc</scripgriddata>
-<scripgriddata hgrid="10x15"     lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_10x15_nomask_c110308.nc</scripgriddata>
+<scripgriddata hgrid="4x5"       lmask="nomask" >share/meshes/fv4x5_050615_polemod_ESMFmesh.nc</scripgriddata>
+<scripgriddata hgrid="10x15"     lmask="nomask" >share/meshes/10x15_nomask_c110308_ESMFmesh.nc</scripgriddata>
 
 <!-- Spectral Eulerian grids -->
 <scripgriddata hgrid="512x1024"  lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_512x1024_nomask_c110308.nc</scripgriddata>
 <scripgriddata hgrid="128x256"   lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_128x256_nomask_c110308.nc</scripgriddata>
-<scripgriddata hgrid="94x192"    lmask="nomask" >lnd/clm2/mappingdata/grids/SCRIPgrid_94x192_nomask_c110308.nc</scripgriddata>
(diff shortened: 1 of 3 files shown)


slevisconsulting:
@mvertens showed me that the DST mesh files have been created already, so I will update https://github.com/ESCOMP/CTSM/pull/1605/files#diff-eb3c4f84552076e5a2ffef3803f49c53336bded5ea474b3db58d6237c2452fad
accordingly.

ekluzek:
@slevisconsulting I think this really is about issue #648 right? There might be some advice in there about doing this.

If this is we should change the introduction to have 648 close on this merge.

slevisconsulting:
@ekluzek in [this post](https://github.com/ESCOMP/CTSM/issues/648#issue-414824349) I see your suggestion to eventually rename `scripgriddata` to `esmfgriddata`. I will take your suggestion a step further and split `scripgriddata` into two groups:
- `srcmeshdata` for the short list of grids associated with the SRC files (ultimately the raw datasets)
- `dstmeshdata` for the long list of grids associated with the DST files (ultimately fsurdat)

Does this sound reasonable?

bld/namelist_files/namelist_defaults_ctsm_tools.xml
```diff
@@ -18,49 +18,48 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 
 <!-- Set default values for grid attributes -->
 <scripgriddata_lrgfile_needed>none</scripgriddata_lrgfile_needed>
-<scripgriddata_type>SCRIP</scripgriddata_type>
```
slevisconsulting, line 21: Unnecessary because this information is now picked up from the files on the fly.

bld/namelist_files/namelist_defaults_ctsm_tools.xml
```diff
@@ -105,34 +104,29 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
...
-<scripgriddata hgrid="5x5_amazon" lmask="navy"      >lnd/clm2/mappingdata/grids/SCRIPgrid_5x5_amazon_navy_c111207.nc</scripgriddata>
-<scripgriddata hgrid="gx1v6"      lmask="gx1v6"     >/glade/proj3/cseg/mapping/grids/gx1v6_090205.nc</scripgriddata>
-<scripgriddata hgrid="gx3v7"      lmask="gx3v7"     >/glade/proj3/cseg/mapping/grids/gx3v7_090903.nc</scripgriddata>
-<scripgriddata hgrid="tx1v1"      lmask="tx1v1"     >/glade/proj3/cseg/mapping/grids/tx1v1_090122.nc</scripgriddata>
-<scripgriddata hgrid="tx0.1v2"    lmask="tx0.1v2"   >/glade/proj3/cseg/mapping/grids/tx0.1v2_090127.nc</scripgriddata>
+<scripgriddata hgrid="5x5_amazon" lmask="navy"      >share/meshes/5x5pt-amazon_navy_ESMFmesh_cd5_c20210107.nc</scripgriddata>
+<scripgriddata hgrid="gx1v6"      lmask="gx1v6"     >share/meshes/gx1v6_090205_ESMFmesh.nc</scripgriddata>
+<scripgriddata hgrid="gx3v7"      lmask="gx3v7"     >share/meshes/gx3v7_120309_ESMFmesh.nc</scripgriddata>
+<scripgriddata hgrid="tx1v1"      lmask="tx1v1"     >share/meshes/tx1v1_ESMFMesh_cd5_c20210105.nc</scripgriddata>
+<scripgriddata hgrid="tx0.1v2"    lmask="tx0.1v2"   >share/meshes/tx0.1v2_ESMFmesh_cd5_c20210105.nc</scripgriddata>
```
slevisconsulting, line 111: Everything that says `scripgriddata` (above) I suggest changing to `dstmeshdata` because it corresponds to DST mesh files (ultimately fsurdat).

Everything that says `unstructdata` (below) I suggest changing to `srcmeshdata` because it corresponds to SRC mesh files (ultimately the raw datasets).

ekluzek: Hmmm, the one thing I wonder there is that whether a mesh is a source or destination depends on context. There is for example a f09 grid that used as a source grid, for raw data in mksurfdata, but of course there's also the model destination grid. But, whether it's a src or dst grid depends on context, not on anything intrinsic in the grid itself.

So I would suggest just using meshdata.

slevisconsulting: > [...] but I think it will be clearest if we go with @slevisconsulting 's proposal of listing them separately and explicitly for each. It seems okay to me if the same mesh file is listed in two places – both as a src mesh and a dst mesh – and my gut feeling is that long-term maintenance will be easier that way: e.g., if we tweak the mesh for the f09 model resolution, we would want to change the dst mesh, but NOT the src meshes. Perhaps most importantly, I think @slevisconsulting 's proposal gives a smoother transition to the rework that @mvertens is doing (which, @ekluzek , you probably haven't seen yet, so I know it's hard for you to comment on this with that goal in mind).

@ekluzek @billsacks 
Bill's comment is correct: In this PR I'm already pointing to different SRC vs. DST mesh files for the same resolution. To explain:
- The UNSTRUCT SRC mesh files -- I generated those from our SCRIP SRC files.
- The DST mesh files -- I'm pointing to existing files that @mvertens suggested I use from here:
https://github.com/ESMCI/ccs_config_cesm/blob/main/component_grids_nuopc.xml

ekluzek:
Ahh, OK going to explicit SRC and DST names is good then. There's only a few files that overlap right now, and I don't expect that to change. @mvertens has a good point that we want to use the standard DST files being used in ccs_config_cesm. But, the SRC files we want to store in our own location.

slevisconsulting:
I feel this is ready for testing.
@billsacks we talked about comparing 1-degree fsurdat files before and after.
Do you recommend any additional testing?

slevisconsulting:
I get an error when generating the 1-degree fsurdat with the new mapping files:

```
Attempting to make PFTs .....
 Creating surface datasets with extra types for crops; total pfts =          78
 domain_read_dims_2d read lon and lat dims from lon/lat
 domain_read initialized domain
 domain_read read LANDFRAC
 domain_read read LANDMASK
 Open PFT file:
 /glade/p/cesm/cseg/inputdata/lnd/clm2/rawdata/pftcftlandusedynharv.0.25x0.25.MODIS.simyr1850-2015.c170412/mksrf_landuse_histclm50_LUH2_2000.c170412.nc
(gridmap_map_read) reading mapping matrix data...
(gridmap_map_read) * file name                  : /glade/work/slevis/git/mksurfdata_toolchain/tools/mkmapdata/0.9x1.25_from_unstruct/map_0.25x0.25_nomask_to_0.9x1.25_nomask_aave_da_c220113.nc
 * matrix dimensions rows x cols :      259200  x       55296
 * number of non-zero elements:       554400
 domain_checksame ERROR: input domain size and gridmap source size are not the s
 ame size
  domain size =      1036800
  map src size=       259200
abort:
```

In the new mapping file I see:
```
dimensions:
        n_a = 259200 ;
        n_b = 55296 ;
        n_s = 554400 ;
        nv_a = 4 ;
        nv_b = 4 ;
        num_wgts = 1 ;
        src_grid_rank = 1 ;
        dst_grid_rank = 1 ;
```

In the old one that works I see:
```
dimensions:
        n_a = 1036800 ;
        n_b = 55296 ;
        n_s = 1572480 ;
        nv_a = 4 ;
        nv_b = 4 ;
        num_wgts = 1 ;
        src_grid_rank = 2 ;
        dst_grid_rank = 2 ;
```

It didn't occur to me earlier. Will I need to modify to the ./mksurfdata_map (fortran) error checking to account for 1d rather than 2d mapping files?

billsacks:
Thank you very much for this work @slevisconsulting!

I am happy to mainly defer to @ekluzek in terms of any reviewing / approval of this, since I think he has a better sense than I do of what it might impact. That said, I do have a lot of thoughts that I'm sharing below... sorry this got so long.

**I'm thinking we'll want to change the base of this PR to go into the ctsm5.2.mksurfdata branch.** @ekluzek, do you agree? I think we agreed that we'd put anything that changes answers for surface datasets into that branch instead of directly onto master; this one is a little ambiguous because, with what's in here so far, I think it won't immediately change answers for surface datasets for existing resolutions, but I think it probably still makes sense to follow that same plan. Also, as I suggest below, I think we'll want to update the out-of-the-box mapping files as part of this PR, which would imply that surface datasets *will* change for out-of-the-box resolutions, which implies more strongly that this should go into ctsm5.2.mksurfdata rather than into master.

I have been thinking a little more about what we should do to finalize and test this PR. I think there are two possible approaches at a high level:

(1) Bring in this PR now (whether to master or the ctsm5.2.mksurfdata branch; probably the latter), with the other toolchain rework following later

(2) Bring these changes in along with the other toolchain rework

When I talked with @slevisconsulting and @mvertens earlier this week, I argued for approach (1), which is why this PR is being done now. This is because (a) since this should be the only answer-changing part of the toolchain rework, bringing this in now will ensure that the answer-changing parts of the toolchain rework are done, and so the ctsm5.2 surface dataset regeneration can theoretically happen at any time; (b) as I am a Bear of Very Little Brain, trying to combine too many things makes my head explode, so I thought it would be easier to separate this answer-changing piece from the rest of the toolchain rework, thus providing a baseline against which the rest of the toolchain rework can be compared (I am still having a hard time wrapping my head around how we would test the toolchain rework if we folded this PR in with everything else: I'm sure it would be possible, but I think it would be harder).

**However, I am now realizing that this approach (1) means that we should probably remake all of the src -> dest mapping files for all grid combinations and update the xml database to point to these new mapping files, so that surface datasets generated at this point actually use the mappings that were generated from the mesh files rather than the scrip grid files.** Otherwise it feels like we'll be in a confusing half-done state here. This is an unfortunate aspect of (1), because if we went with (2) we wouldn't need these mapping files anymore (since they will be generated on the fly, as part of generating the surface dataset). I still feel like (1) is probably the best approach, but I'm less sure of this if others feel differently – especially @slevisconsulting , who will be doing most of the actual work. Assuming we go with (1) I would propose that the mapping files be put in a temporary place on glade – probably the TSS project space – and *not* checked in to the inputdata repository (to avoid clogging the inputdata repository permanently with these temporary files). Then we should make a note somewhere (probably an issue) to delete these temporary new mapping files when we move to the new toolchain, which won't require them. (I'm thinking it will be acceptable to not have these in the inputdata repo because they will only exist in this state on the ctsm5.2.mksurfdata branch, not on master, since the plan is to move to the new toolchain before we move the ctsm5.2.mksurfdata branch to master. @ekluzek do you agree?)

In terms of the question of what to check:

**First, I realized that you will probably need to generate new baseline surface datasets for comparison (from the ctsm5.2.mksurfdata branch), because I think the out-of-the-box surface datasets are not consistent with what you would get from running the latest version of mksurfdata_map.**

Once you have done that, I would suggest the following tests, since you are pointing to all new source and destination mesh files:

(1) Running mksurfdata_map's `Makefile.data` to ensure that it works to make surface datasets for all standard resolutions using the mapping files created from the new source & dest mesh files – assuming we expect `Makefile.data` to work right now (@ekluzek can possibly comment on this). (These surface datasets can be thrown away after this testing is done: my understanding is that we'll do the real generation of the new surface datasets later.)

(2) A visual check of all fields on a surface dataset at one production resolution – either f09 or f19 – generated using the new mapping files as noted above. i.e., compare a surface dataset generated with this PR and its new mapping files against a baseline surface dataset. The point of this would be to verify that there are no glaring issues with the new source mesh files.

(3) Optionally, a cursory check of surface datasets covering a few different classes of grids – e.g., one regional grid, one single point grid, maybe one global unstructured grid (though it will be hard to compare that one visually); and anything else that I'm not thinking of.

billsacks:
Regarding the issue with trying to make an f09 surface dataset: I'm hopeful that this is an easy one:

Based on `ncdump -h /glade/work/slevis/git/mksurfdata_toolchain/tools/mkmapdata/0.9x1.25_from_unstruct/map_0.25x0.25_nomask_to_0.9x1.25_nomask_aave_da_c220113.nc` I see:

```
		:domain_a = "/glade/p/cesmdata/cseg/inputdata/lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.5x0.5_nomask_c200129.nc" ;
```

It looks like you made this mapping file using a 0.5 deg grid, when in fact it should have been made using a 0.25 deg grid. Could this be a carry-over from a time when the pft data were at 0.5 deg rather than the new 0.25 deg?

> Will I need to modify to the ./mksurfdata_map (fortran) error checking to account for 1d rather than 2d mapping files?

That's a good question. I'm not sure. I hope not. It looks like the above error is for something simpler than that. If it does look like you'll need to make a change to account for 1d rather than 2d mapping, let's talk first: I wonder if we can get the mapping files to understand that these are 2d grids even when building them from the mesh files, maybe by having appropriate metadata on the mesh files themselves.

ekluzek:
I agree with @billsacks there should be a way of making the mesh files in a way that's perfectly compatible with the SCRIP grid files. That's what we found in initial tests with #648. Bill might be right it's just a difference in the half degree to quarter degree grid.

ekluzek:
I've thought about whether this should be on the ctsm5_2 branch or master. I definitely see @billsacks point about that. But, the main point of the ctsm5_2 branch is for mksurfdata_map changes that change mksurfdata and CTSM in fundamental ways so that it can't go on master. For example, the urban change that didn't allow you to use surface datasets that you created with master.

The ctsm5_2 is going to be kept up to date with master so changes on master will come to ctsm5_2. But, anything that goes on ctsm5_2 won't be on master until the whole thing is ready. And it'll be easier to handle the branch if it's as small as possible. So anything that can go on master should.

So I'd say things that aren't answer changing in mksurfdata -- or only roughly roundoff changing should go ahead and go on master. And the ctsm5_2 branch will update to them, so it'll be on both. There might be other things that there's more of a judgement call for, but this isn't any kind of fundamental change to how mksurfdata works. This is similar to your insight that @slevisconsulting earlier change about unit tests should go on master.

How does that sound?

ekluzek:
I should replace all my use of "master" with "main-dev". I'm looking forward to #1048

ekluzek:
@billsacks and @slevisconsulting I do expect Makefile.data to work on main-dev. And for it to create the right datasets, although maybe that needs to be adjusted. So if you find any problem with Makefile.data to create all the datasets, it's something that would need to be fixed.

billsacks:
@ekluzek your rationale makes a lot of sense: this will be answer changing but (hopefully!) not substantially so. So as long as people don't expect to get bit-for-bit results from running master's mksurfdata_map against the current surface datasets (which it sounds like they don't / can't anyway) then I agree that it makes sense to put this directly on master. Thanks for sharing your rationale about this.

mvertens:
@slevisconsulting @billsacks @erik - the mesh representation cannot be 2d. It inherently can only be represented as a 1d. I'm happy to discuss this in more detail next week.

slevisconsulting:
> Regarding the issue with trying to make an f09 surface dataset: I'm hopeful that this is an easy one:
> 
> Based on `ncdump -h /glade/work/slevis/git/mksurfdata_toolchain/tools/mkmapdata/0.9x1.25_from_unstruct/map_0.25x0.25_nomask_to_0.9x1.25_nomask_aave_da_c220113.nc` I see:
> 
> ```
> 		:domain_a = "/glade/p/cesmdata/cseg/inputdata/lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.5x0.5_nomask_c200129.nc" ;
> ```
> 
> It looks like you made this mapping file using a 0.5 deg grid, when in fact it should have been made using a 0.25 deg grid.

@billsacks thank you for noticing (saved me some digging). My last commit corrects this and I can generate both the mapping files and the fsurdat for 0.9x1.25. HOWEVER, now my previous time estimate comparing UNSTRUCT to SCRIP was wrong because it used the 0.5x0.5 SRC file for all the SRC resolutions!

**My new time estimate (for DST resolution 0.9x1.25) shows UNSTRUCT as **slower** than SCRIP!** The numbers represent the total (all the SRC resolutions) except 1km. From the system emails that I got for the jobs, "real" = wallclock:
**UNSTRUCT**
real    32m6.685s
user    63m13.675s
sys     0m27.584s
**SCRIP**
real    24m34.674s
user    48m14.524s
sys     0m26.715s

Here's a breakdown for files that took >= 5 minutes:

| SRC | UNSTRUCT | SCRIP |
| --- | --- | --- |
| 3x3min | 21 min | 15 min |
| 5x5min | 6 min | 5 min |

The 1km file remains unchanged (was unstructured already), so I left it out of the timing estimate.

Is it possible that UNSTRUCT saves time for very high DST resolutions? I am happy to look into this more if you want me to. I also wonder whether this new information is a vote in favor of @billsacks option (2) above, i.e. we don't spend the time generating all new mapping files just to trash them a couple of months from now...

I'm open to all suggestions.

mvertens:
@slevisconsulting - thanks for the update. I am really perplexed by this result. 
I did started wondering if it would be worth exploring changing the mksurfdata to simply generate the route handles at run time - and thereby totally skipping the step of generating offline files. This would imply that mksurfdata would be a parallel utility and potentially run a lot faster. The ESMF generation of route handles scales is done in parallel and scales very well with processor count. This could potentially also help the high resolution cases as well as totally simplifying the workflow. I used this in the CTSM nuopc cap to generate the land fractions at run time and I believe it would be a fairly straightforward exercise to incorporate this into mksurfdata.
I would be happy to help with this and prototype it.

billsacks:
@slevisconsulting I also don't understand this result, but I don't think I was ever very involved with the timing comparisons of scrip vs. unstructured files. My impression is that there are a few different esmf unstructured / mesh formats, and I wonder if the timings you're doing now are using a different format than what was used when these timings were last done? Or could the results depend on how many mpi tasks you give it???

However, I just got off a call with @mvertens where she explained her latest thinking (summarized in the comment above). It sounds very appealing to me. If we go with her suggestion, that might make it unnecessary to try to dig in to what you're seeing here. It would also most likely change answers, because mapping would be done online in mksurfdata_map rather than via the offline ESMF regrid weight generator – so my above arguments about separating answer-changing from answer-preserving modifications would probably no longer hold. The main reasons I could then see for moving ahead with this PR would be to separate one source of answer changes and to separately verify that the source and destination meshes are all good... I'm not really sure if it makes sense to do this extra work for those reasons.

It seems like we should have a call early next week to talk about a tentative path forward given @mvertens 's proposal.

slevisconsulting:
> It seems like we should have a call early next week to talk about a tentative path forward given @mvertens 's proposal.

I sent an invite with permissions for you to change the time if needed.
I duplicated the zoom info from our Wednesday mtg. Will that work?

slevisconsulting:
@ekluzek 
With the work on the parallel mksurfdata_map, this PR (and corresponding issue #648) will also likely be closed without coming in because the parallel mksurfdata_map already uses UNSTRUCT mesh files.

PR #1613, opened 2022-01-25
Get FATES working for NEON

### LINKED ISSUE #1609: Need to get FATES working for NEON
We need to make it easy to run with FATES for a NEON site. Currently there isn't a compset alias for FATES with tower data. So we need to add that. Second, the user mods for FATES currently assume that BGC is active. So for example ndep and fire data are set in the namelist, things that FATES normally either doesn't use or only under certain conditions.

ekluzek:
### Description of changes

Proposal for changes to get FATES working for NEON

### Specific notes

Contributors other than yourself, if any: Whole team in discussing this

  Fixes #1609

  Added new use-cases: 2018-PD_NEON_transient, PD_control, 2018_control

Testing performed, if any:
   So far just the build-namelist testing

### DIFF
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -3393,12 +3393,12 @@ sub setup_logic_nitrogen_deposition {
                   'sim_year_range'=>$nl_flags->{'sim_year_range'});
     }
     add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_fldfilename_ndep', 'phys'=>$nl_flags->{'phys'},
-                'use_cn'=>$nl_flags->{'use_cn'}, 'lnd_tuning_mode'=>$nl_flags->{'lnd_tuning_mode'},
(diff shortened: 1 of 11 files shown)


ekluzek:
At the NEON meeting @adrifoster pointed out that we'll likely need a different list of output history fields for FATES than BGC (and possibly SP as well). I could do that in the user-mod directory, with bash logic so that we don't have to create more user-mod directories. The other way would be to possibly put it in the use case, or more embedded into build-namelist (probably namelist_defaults_ctsm.xml).

wwieder:
yep, good idea here Erik we're using user_mod_dirs to set this up for the NEON case, but a more general way of defining the default history file output for different configurations (SP/BGC/FATES) would be helpful.

wwieder:
@ekluzek and @adrifoster we still may need some documentation, but is this issue resolved?

adrifoster:
> @ekluzek and @adrifoster we still may need some documentation, but is this issue resolved?

I guess it depends on our definition of done.... we still don't have anything for the crop sites, which is maybe fine. I am working on a solution to update the surface datasets... and `run_neon` doesn't work for FATES.... but you *can* run FATES at most of the NEON sites.

Issue #1606, opened 2022-01-18
NEON tools not handling crop weights correctly


wwieder:
### Brief summary of bug

@negin513 and I noticed that the current [subset_data](https://github.com/ESCOMP/CTSM/blob/master/tools/site_and_regional/subset_data.py) and [modify_singlept_site_neon](https://github.com/ESCOMP/CTSM/blob/master/tools/site_and_regional/modify_singlept_site_neon.py) scripts don't set 'PCT_CFT' correctly.  

### General bug fix

#1461 should be modified with the simple example below, where a use defines a single dominant PFT, which gets 100% weight on respective land units.

if dompf < 16:
  PCT_NAT_PFT(dompft) = 100 
else:
  PCT_CFT(dompft-16) = 100 

Land unit weights should also be adjusted defined by `PCT_NATPFT` &`PCT_CROP` 
if dompf < 16:
  PCT_NATPFT = 100 
  PCT_CROP = 0 
else:
  PCT_NATPFT = 0  
  PCT_CROP = 100

wwieder:
Maid questions:

1. Are we OK with only allowing one land unit (natpft or crop) if users are defining dom_pft in the subset_data.  This makes 2 simpler.
2. Does our proposed fix make sense? 
3. Is indexing of cfts correct based on this plan?

ekluzek:
@wwieder and @negin513 thanks for digging into this.

Iit's PCT_NATVEG rather than PCT_NATPFT, and I think you want the equality to be

if dompft <= 16:

You want all PFTS from 0 to 16 to be expressed as natural PFT's. The crops start at "PFT" 17 and up. The PCT_NATPFT array is zero based (with 0 as bare soil), but the PFT_CFT array has temperate_rainfed_corn as index 0 (in python). It's easy to get this off by 1. But, that's what I got going through it.

ekluzek:
Several of us met last Friday to talk about this and go over the details (@negin513 @wwieder @danicalombardozzi  and I). I had thought that my answer above was correct, but in our discussion I found I was off. The two generic crops are still considered CFT's (since they will be on their own crop landunit). So the natural veg PFT's are 0 to 14 (with 15 PFT's). So the inequality should be <= 14 or < 15. We should also make sure that the magic number 15 is encapsulated as a parameter in the code. I'll edit my answer above to give the correct one.

[...]

PR #1615, opened 2022-01-25
modify_fsurdat: Add dom_cft as alternative user-choice to dom_nat_pft

slevisconsulting:
### Description of changes
This PR addresses Isla Simpson's CSSI [example (2b)](https://docs.google.com/document/d/1wc17F5i_x8MP0XBwqgUAcY_fDbruLaseHVxdIddJMbk/edit?usp=sharing), which requires replacement of the vegetation with crop in a rectangle of North America.

I am adding this capability to the modify_fsurdat tool.

### Specific notes

@billsacks @ekluzek 

None

No. This is a new feature.

modify_template.cfg now includes the line
`dom_cft = UNSET`
to allow users to set the cft of their choice to replace the non-crop vegetation.

Testing performed, if any:
I have modified a few fsurdat files to confirm that the code does what I want.


python/ctsm/modify_fsurdat/modify_fsurdat.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, my_data, lon_1, lon_2, lat_1, lat_2, landmask_file):
 
         self.file = my_data
 
-        self.not_rectangle = self._get_not_rectangle(
+        self.rectangle = self._get_rectangle(
```
slevisconsulting, line 28: New function `get_dom_cft` needs `rectangle` rather than `not_rectangle` so I did a minor rewrite here and a few lines down.

ekluzek (changes requested):
@slevisconsulting this is actually something that we might want to be consistent with subset_data. In subset_data, you can pick either a crop or a nat-veg with the same --dom_pft option, by using PFT numbering. See the issue #1606 which goes over this, and if you go with that you'd need to do the same here. Here you have one way to set a Crop PFT and another for a natural-veg PFT. We likely want to have this consistent in the way we handle it. An advantage of the former is that you only need one option rather than two, and I think that's helpful.

The other thing I noticed is that there are a few magic numbers in here (15, 79, and 12 is what I saw), these should be changed to variables so the context of the meaning behind them is given.

And finally I didn't see the change added to a unit test and if you can it would be good to do that.

Thanks for putting this together. Feel free to reach out to me to go over it.

billsacks:
Thank you very much for your work on this @slevisconsulting . I don't feel a need to review it.

slevisconsulting:
> And finally I didn't see the change added to a unit test and if you can it would be good to do that.

Thank you for your comments, @ekluzek 
Next I will work on your unit test suggestion.

python/ctsm/modify_fsurdat/fsurdat_modifier.py
```diff
@@ -65,10 +65,15 @@ def fsurdat_modifier(cfg_path):
     landmask_file = get_config_value(config=config, section=section,
         item='landmask_file', file_path=cfg_path, can_be_unset=True)
 
+    # Create ModifyFsurdat object
+    modify_fsurdat = ModifyFsurdat.init_from_file(fsurdat_in,
+        lnd_lon_1, lnd_lon_2, lnd_lat_1, lnd_lat_2, landmask_file)
```
slevisconsulting, line 70: Moved this up from a few lines down so as to eliminate some "magic" numbers that I could instead read from the fsurdat file.

slevisconsulting:
Python tests PASS
clm_pymods tests PASS

tools/modify_fsurdat/modify_template.cfg
```diff
@@ -56,13 +56,13 @@ lnd_lon_2 = 360
 # user-defined mask in a file, as alternative to setting lat/lon values
 landmask_file = UNSET
 
-# Non-crop PFT to be set to 100% according to user-defined mask.
-# If idealized = True and dom_nat_pft = UNSET, the latter defaults to 0
-# (bare soil). Valid values 0 to 14 (int).
-dom_nat_pft = UNSET
+# PFT/CFT to be set to 100% according to user-defined mask.
+# If idealized = True and dom_plant = UNSET, the latter defaults to 0
+# (bare soil). Valid values 0 to 78 (int).
```
ekluzek, line 61: This is another magic number, although it's also only in a comment. And I think this might be similar to an issue that @negin513 and I discussed where we didn't see a great way to eliminate a magic number for a specific thing. So I'd like to discuss this with a bit...

ekluzek (approved):
This looks good to me. You addressed the suggestions I had earlier. 

I'd still like to meet with you to go over it a bit. It would help me to understand how the testing for this works. There is one comment I made about a magic number, but I'm not sure it's something we can really change and might be OK as it is...

tools/modify_fsurdat/modify_template.cfg
slevisconsulting: Maybe instead of specifying the values 0 and 78, I could send the user to the surface dataset and get the valid values there.

python/ctsm/test/test_sys_fsurdat_modifier.py
```diff
@@ -59,6 +59,32 @@ def test_minimalInfo(self):
         self.assertTrue(fsurdat_out_data.equals(fsurdat_in_data))
 
 
+    def test_crop(self):
+        """
+        This version replances the vegetation with a crop
```
slevisconsulting, line 64: spelling

slevisconsulting: Also add comments to explain what's happening, starting with setup and making a cfg file etc...

slevisconsulting:
I linked the dev081 tests to dev082.

ekluzek:
@slevisconsulting OK is this ready to go from your perspective? If so I'll do a final review and approve it and then make the tag. Also are there new clm_pymod tests or are answers different for them than before? If so we should link each specific test directory, rather than the top level, and then remove the softlinks for the clm_pymod tests and then rerun them. Could you do that?

slevisconsulting:
> @slevisconsulting OK is this ready to go from your perspective? If so I'll do a final review and approve it and then make the tag. Also are there new clm_pymod tests or are answers different for them than before? If so we should link each specific test directory, rather than the top level, and then remove the softlinks for the clm_pymod tests and then rerun them. Could you do that?

I think this is ready.
No new clm_pymod tests.
Existing clm_pymod tests PASSed against baseline.
Thanks, @ekluzek !

tools/modify_fsurdat/modify_template.cfg
negin513: One thing I did was setting `MAX_PFT` and `NAT_PFT` as constants in the code and used those constants, when referring to them both in comments and codes:

https://github.com/ESCOMP/CTSM/blob/04a0348573689d8d9bb9aa75fe3004adcacb787f/python/ctsm/site_and_regional/single_point_case.py#L21-L27

If you want, you can import these constants directly and reuse them.

negin513:
I have not reviewed this code but I have a quick suggestion about the user interface.


I've noticed here in the namelist the user can choose either `dom_nat_pft` or `dom_cft` to modify the surface dataset. In the subset_data code which is really similar, the user can choose `dom_pdf` that they want to modify and the code calculates if it means changing `dom_nat_pft` or `dom_cft` based on the value of `dom_pft`. So for example it automatically modifies `dom_cft` if it is above 15. 

I was thinking maybe having a more uniform user-interface options would be a good idea between fsurdat_modifier and subset_data. I was wondering what @ekluzek and @slevisconsulting think about using `dompft` here instead of two namelist options of `dom_nat_pf` and `dom_cft`. 
Again, I am mentioning this here as a suggestion to have a uniform user-interface options for our codes which has a lot of similarities. Please let me know if I am missing something.

ekluzek:
@negin513 this is a really valuable comment here. I agree that tools are
easier to understand if the user interfaces are similar between them. I
actually made a similar comment at the top of...

https://github.com/ESCOMP/CTSM/pull/1615#pullrequestreview-862838606

But, since I didn't make a checkbox out of it, it's easy to miss. So I
agree with your comment. I think based on where we are at with this
specific PR, we should go ahead as is, but I'll make an issue to address
that for a future one that I'll assign to @slevisconsulting. I think we
should have a few of us meet next week to discuss the user interfaces of
the tools to make sure we are building them similarly. This also tells me
that I should use checkboxes more often. I have a love/hate relationship
with them, but I'm realizing they are important to ensure things don't get
lost. :-) @billsacks is the King of using them, and I think the one that
originally endorsed the idea. I think he's right they really do help.

Also @negin513 when I refreshed the issue in github I didn't see your
comment, so I'm sending this from my email. I haven't figured out why I
didn't see your comment in github itself.

*Erik Kluzek, (CGD at NCAR)*

*National Center for Atmospheric Research*

*Boulder CO, *

*(off) (303)497-1326 (fax) (303)497-1348*

*(skype) ekluzek    (cell) (303)859-0183*

*(github) ekluzek   **Pronouns: he/his/him **(*Why do I list my personal
pronouns? <https://www.mypronouns.org/what-and-why/>)
*------------------ Home page ------------------------*

*          https://staff.ucar.edu/users/erik
<https://staff.ucar.edu/users/erik>*



*!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!*

tools/modify_fsurdat/modify_template.cfg
slevisconsulting: Thank you @negin513 !

In talking with @ekluzek we decided to let users figure out the valid range by using ncdump on their specific fsurdat file. This way we're not tied to any hardwired values.

slevisconsulting: For example we still have fsurdat files with max_pft = 16 instead of 78

ekluzek:
On @negin513 point about making the user interface more similar. I was about to make an issue, but then I looked and the difference is that this tool has it as a config file, and most of the contents of that file follow what's on the surface dataset. 

> I've noticed here in the namelist the user can choose either dom_nat_pft or dom_cft to modify the surface dataset. In the subset_data code which is really similar, the user can choose dom_pdf that they want to modify and the code calculates if it means changing dom_nat_pft or dom_cft based on the value of dom_pft. So for example it automatically modifies dom_cft if it is above 15.

Since subset_data is using a simpler user-interface I think it's OK that the user interfaces are different. This one more closely reflects what's on the surface dataset being modified. While subset_data is making a command line option that's easier for the user. So I'm siding with the idea that it's OK they are different now. But, I'm also happy to be convinced back the other way again. So I'm open to still having a discussion on this next week. So @negin513 what do you think should we schedule a meeting to discuss?

slevisconsulting:
@negin513 I also couldn't see your comment here in github. As @ekluzek said, he made the same comment early in the PR. If you look at my cumulative code modifications, I did make that change. But instead of using `dompft` I chose `dom_plant` to avoid confusion between the terms pft and cft.

negin513:
I agree with @ekluzek. Overall I think it might be good to have our options relatively similar, but maybe we can worry about them in future.

For example. previously subset_data had the option of `zero_nonveg` which is also used here. However, in #1461 based on reviewers comments on the names being confusing we changed it to `include_nonveg` which is the opposite of `zero_nonveg`. But we are still using `zero_nonveg` here...

> @negin513 I also couldn't see your comment here in github. As @ekluzek said, he made the same comment early in the PR. If you look at my cumulative code modifications, I did make that change. But instead of using `dompft` I chose `dom_plant` to avoid confusion between the terms pft and cft.

I intentionally removed my original comment, because upon further looking into the newer version of the code I've noticed you have made this change from `dam_nat_pft` and `dom_cft` to `dom_plant` which is great.

tools/modify_fsurdat/modify_template.cfg
negin513: Thanks for your response. Overall, @ekluzek and I also found it difficult to get rid of the hard-coding of 16 vs 78. But we tried it using the constants mentioned above. 

> For example we still have fsurdat files with max_pft = 16 instead of 78
In #1461, I put in explicit error checking to take care of this exact scenario. So if the user chose a file with 16 pfts on them and `dom_pft` is higher than 16 it gives an error.

slevisconsulting:
> For example. previously subset_data had the option of `zero_nonveg` which is also used here. However, in #1461 based on reviewers comments on the names being confusing we changed it to `include_nonveg` which is the opposite of `zero_nonveg`. But we are still using `zero_nonveg` here...

Oh, interesting. Would you like to open an issue @negin513 to remind us that the `zero_nonveg` function could be replaced in this tool with the `include_nonveg` function to avoid duplication?

slevisconsulting:
Unless @ekluzek you'd like me to revisit this PR with @negin513 's comments about `zero_nonveg` and `dom_plant`.

PR #1625, opened 2022-02-02
Use average days per year when converting parameter units

### LINKED ISSUE #1612: Some uses of get_days_per_year should use the average number of days in a year, not the number of days in the current year
### Brief summary of bug

There are a number of places in the code where a parameter that is expressed in units of something per-year is converted to units of per-second using the result of the `get_days_per_year` function. However, when running with a calendar that uses leap years, it looks like these should really be using the *average* number of days per year, not the number of days in the current year.

### General bug information

**Does this bug cause significantly incorrect results in the model's science?** No – just small errors

**Configurations affected:** BGC/CN runs with a Gregorian calendar

### Details of bug

This emerged from my analysis documented in https://github.com/ESCOMP/CTSM/pull/1593#issuecomment-1020429625 . The impact is minor, but this looks easy enough to fix that I'm going to go ahead and fix it.

Currently, for things like decay rates and mortality rates that are initially expressed in per-year units, the decay or mortality rate is slightly slower in a leap year than in a non-leap year – or, if the conversion happens just once at model initialization time, then the resulting rate would depend on whether the model starts in a leap year or a non-leap year. Even though this doesn't cause large errors, it's weird....

This impacts the following subroutines:
- C14Decay: This should probably use a fixed, hard-coded constant, converting the decay rate from a per-year value to a per-second value: as currently written, C14 decays more slowly on leap years.
- CNGapMortality: This should probably use a fixed, hard-coded constant, converting the mortality rate parameter from a per-year value to a per-second value: as currently written, mortality is a bit slower on leap years.
- CNPhenologyMod.F90: CNEvergreenPhenology and CNStressDecidPhenology and the multiplier of leaf_long in CropPhenology: This should probably use a fixed, hard-coded constant, converting a litterfall rate from a per-year value to a per-second value: as currently written, litterfall happens more slowly on leap years.
- decomp_rate_constants_bgc: As for the others, this should probably use a fixed, hard-coded constant
- ndep_interp: This one is questionable, and may depend in part on whether the ndep input file and the model are consistent in their use of a GREGORIAN vs. NOLEAP calendar. But since I think this will typically come from model run with a NOLEAP calendar, I'm thinking that it should use a fixed, hard-coded constant (so that a constant annual ndep rate on the file would result in a constant annual ndep rate in CTSM, rather than a rate that varies slightly in leap vs. non-leap years). (But from spot-checking some recent files, it looks like rates are typically already specified as per-second, so this conversion from a per-year rate to a per-second rate is never done. I think that this implies that, in a GREGORIAN run, there is slightly more ndep than what occurred in the run that created the forcing data, since a year is slightly longer.)
    - Actually, upon further reflection, I think it's best *not* to change this one.

My plan is to introduce a new `get_average_days_per_year` function that returns the average number of days per year for the given calendar (which can be obtained via a query to an appropriate ESMF calendar routine), and then use that function in the above subroutines rather than using the number of days in the current year.

billsacks:
### Description of changes

There are a number of places in the code where a parameter that is expressed in units of something per-year was converted to units of per-second using the result of the `get_curr_days_per_year` function. However, when running with a calendar that uses leap years, these should really be using the *average* number of days per year, not the number of days in the current year.

See #1612 for details.

### Specific notes

Resolves #1612 

- Small changes when running with Gregorian calendar
- With a NOLEAP calendar, it's possible that there will be roundoff-level changes for some compilers

Testing performed, if any:
ERP_Ly3_P72x2.f10_f10_mg37.IHistClm50BgcCrop.cheyenne_intel.clm-cropMonthOutput passes and is bit-for-bit with baseline


billsacks:
@ekluzek @olyson - I'm assigning both of you as reviewers, though feel free to pass if you don't feel a need to review.

@ekluzek - I'd welcome your review of the SE aspects of this change.

@olyson - I'd welcome your review for confirmation that these changes make sense scientifically – i.e., using the average number of days per year rather than the current number of days per year in these places in the code.

That said, I'm reasonably confident that this is correct; I wanted to give you a chance to review it if you want, but you shouldn't feel a need to.

olyson (commented):
I *think* I understand the distinction between when get_average_days_per_year is needed and when get_curr_days_per_year is ok.  For example, I see that this section of code in  cpl/mct/ndepStreamMod.F90 still uses get_curr_days_per_year:

   if ( divide_by_secs_per_yr )then
      ig = 0
      dayspyr = get_curr_days_per_year( )
      do g = bounds%begg,bounds%endg
         ig = ig+1
         atm2lnd_inst%forc_ndep_grc(g) = sdat%avs(1)%rAttr(1,ig) / (secspday * dayspyr)
      end do

But that is appropriate because we want to distribute the same total amount of nitrogen for the given year regardless of whether it is a leap year or not?
Another example is the continued use of get_curr_days_per_year in biogeochem/CNAnnualUpdateMod.F90 because it is being used to determine the end of the year...
Also, I see that one of the new mimics modules in ctsm5.1.dev074 uses get_curr_days_per_year so I assume that will need to be changed to get_average_days_per_year as well since it is being used to translate to a per-second time constant.  For development code in progress, I guess we'll have to watch out for instances of this...

billsacks:
@olyson - very good questions.

The use in ndepStreamMod was the one that I was least sure about. Originally I thought it should be changed, but then decided it should probably be left as is (see notes in https://github.com/ESCOMP/CTSM/issues/1612 ). I think it partly comes down to a question of how the forcing data were generated and used. But I checked some recent forcing files and it looks like this is irrelevant, because now ndep is already specified in units of per-sec, so the conversion from a per-year rate shouldn't be invoked. Since I could convince myself that either way is right for that use, and since it seems not to matter, I decided to leave it as is. I can change that one, though, if you or @ekluzek think it should be changed.

Yes, the CNAnnualUpdateMod one felt like it should be left as `get_curr_days_per_year` because it's being used to determine the end of the year.

Good catch with the new use in MIMICS. I just changed that one.

olyson:
After reading your discussion about ndep, I'm fine with leaving it as you now have it, thanks.

PR #1643, opened 2022-02-09
Get FATES and carbon_only test-suite tests to pass with MIMICS active

### LINKED ISSUE #1636: FATES tests fail with MIMICS active
### Brief summary of bug

On izumi I get the error:
SoilBiogeochemDecompCascadeMIMICSMod.F90 line 852: Reference to undefined POINTER

On cheyenne the error feels more helpful:
SoilBiogeochemDecompCascadeMIMICSMod.F90 line 1111: Index '1' of dimension 1 of array 'ligninnratioavg' above upper bound of 0

ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesRs.**cheyenne**_gnu.clm-FatesColdDef
ERS_D_Ld5.f10_f10_mg37.I2000Clm50Fates.**izumi**_nag.clm-FatesColdDef.0202-170916iz_nag

### General bug information

**CTSM version you are using:** ctsm5.1.dev073-79-geba82445d

**Configurations affected:** soil_decomp_method = 'MIMICSWieder2015'

slevisconsulting:
### Description of changes

To pass with MIMICS active, FATES and carbon_only tests need calculations of:
`ligninNratioAvg` and `annsum_npp_col`, both currently declared and calculated in `CNVegCarbonFluxType.F90`.

`ligninNratioAvg` is a function of `leafc_to_litter_patch, frootc_to_litter_patch, soilbiogeochem_cwdc_col, soilbiogeochem_cwdn_col, soilbiogeochem_decomp_cascade_ctransfer_col`.

I got both FATES and carbon_only tests to pass by declaring local copies (in MIMICS) of `ligninNratioAvg` and `annsum_npp_col`.

### Specific notes

@wwieder @rgknox @ckoven @ekluzek 

 Fixes #1636 

Mostly not applicable. MIMICS is a new option and we're testing it here as if it were the default option. However, actual MIMICS tests introduced in dev074 should be bfb, unless they end up with roundoff diffs due to changes in the order of certain calculations.

No.

**Update:**
Turns out that carbon_only was failing due to annsum_npp < 0, which caused the MIMICS sqrt of this quantity to fail. The fix for this is different than I originally thought, but I am still addressing it in this PR along with the fix for the FATES failures.

**Issue discovered while working on this PR:**
For what I can tell, c3psn for sorghum and millet are set to 1 in the param files, and I'm pretty sure they should be 0 (ie not c3 but c4 photosynthesis). I suspect I made this mistake when I first introduced these cfts. I will open an issue to get confirmation from others that this is a mistake.

Testing performed, if any:
With MIMICS active, these tests now pass with the code changes committed this far:
`ERP_D_P36x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.cheyenne_intel.clm-cn_conly`
`SMS_D_Ld5.f10_f10_mg37.I2000Clm50FatesRs.cheyenne_gnu.clm-FatesColdDef`


src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1100,21 +1108,73 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
+      ! Local column-level ligninNratioAvg and annsum_npp_col because these are
+      ! not available when use_fates = .true. or when carbon_only = .true.
+      if (use_fates .or. cnallocate_carbon_only()) then
+         ! Initialize
+         do fc = 1,num_soilc
+            c = filter_soilc(fc)
+            ligninNratioAvg_local(c) = 0._r8
+            annsum_npp_col_local(c) = 0._r8
+            counter(c) = 0
+         end do
```
slevisconsulting, line 1120: The above loop is all I've really done for now:
I set local copies of `ligninNratioAvg` and `annsum_npp_col` to zero for FATES or carbon_only and these get used below.

If we decide to go with this solution for carbon_only and for FATES, then I will uncomment the next few lines, where the local copies of `ligninNratioAvg` and `annsum_npp_col` get values by pft.

slevisconsulting: @ckoven thank you for this helpful pointer.

I've looked at this some more and found that I misspoke in one of my earlier posts:
Turns out that carbon_only has everything it needs and was failing for a different reason. So code mods in this PR are for FATES only. 

I plan to do the following (unless the group decides that we should go back to @ckoven 's original suggestion to make FATES and MIMICS mutually exclusive for now):
1) When running FATES, use annsum_npp from FATES in SoilBiogeochemDecompCascadeMIMICSMod.F90 as done in ch4Mod.
2) Follow @wwieder 's suggestion for ligninNratioAvg when running FATES.

src/biogeochem/EDBGCDynMod.F90
```diff
@@ -266,7 +269,7 @@ subroutine EDBGCDynSummary(bounds, num_soilc, filter_soilc, num_soilp, filter_so
        c13_soilbiogeochem_carbonflux_inst, c13_soilbiogeochem_carbonstate_inst, &
        c14_soilbiogeochem_carbonflux_inst, c14_soilbiogeochem_carbonstate_inst, &
        soilbiogeochem_nitrogenflux_inst, soilbiogeochem_nitrogenstate_inst, &
-       clm_fates, nc)
+       nc)
```
slevisconsulting, line 272: Elsewhere I added clm_fates as an argument to subroutine calls, ultimately for use by MIMICS. Here, however, I removed it because it seemed unused. Let me know if I shouldn't have.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -846,12 +855,16 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
+    real(r8):: ligninNratio(bounds%begp:bounds%endp)  ! local patch-level lignin to nitrogen ratio
+    real(r8):: annsum_npp(bounds%begp:bounds%endp)  ! local annual sum of NPP at the patch level
+    real(r8):: ligninNratioAvg_scalar  ! lignin to nitrogen ratio, scalar in column-level loop
+    real(r8):: annsum_npp_col_scalar  ! annual sum of NPP, scalar in column-level loop
 
     !-----------------------------------------------------------------------
 
     associate(                                                           &
-         ligninNratioAvg => cnveg_carbonflux_inst%ligninNratioAvg_col  , & ! Input:  [real(r8) (:)     ]  column-level lignin to nitrogen ratio
-         annsum_npp_col => cnveg_carbonflux_inst%annsum_npp_col        , & ! Input:  [real(r8) (:)     ]  annual sum of NPP at the column level (gC/m2/yr)
```
slevisconsulting, line 854: Moved above two lines from here to inside an if statement where they get used only when not use_fates.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
slevisconsulting: In today's commit/push I completed item (1).
Next I will address item (2).

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1100,21 +1113,71 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
+      ! Also get FATES copy of annsum_npp for FATES cases
+      if (use_fates) then
+         ! Initialize
+         do fc = 1,num_soilc
+            c = filter_soilc(fc)
+            ligninNratioAvg_local(c) = 0._r8
+            annsum_npp_col_local(c) = 0._r8
+         end do
+
+         ! Sum over p
```
slevisconsulting, line 1127: Incorrect comment in line 1127 from before. I will change it.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
slevisconsulting: Latest commit addresses item (2).

slevisconsulting:
Ran cheyenne test suite:
FATES and carbon_only tests all PASS

slevisconsulting:
@rgknox thank you for alerting me to the fact that annsum_npp gets passed from FATES to the HLM only if `use_lch4 = .true.` as you indicated here:
https://github.com/NGEET/fates/blob/master/main/FatesInterfaceMod.F90#L613-L625

At your recommendation I have disallowed FATES-MIMICS with `use_lch4 = .false.` and tested as follows:

Default test `ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesRs.cheyenne_gnu.clm-FatesColdDef` passes.
Same test with `use_lch4 = .false.` stops w error:
```
err=Warning : CLM build-namelist::CLMBuildNamelist::setup_cmdl_bgc() : If MIMICS is on and use_fates = .true. then use_lch4 must be .true. and currently it's not
```

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1100,21 +1126,107 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
+            pf = p - col%patchi(c)
+            s  = clm_fates%f2hmap(nc)%hsites(c)
+            annsum_npp(p) = clm_fates%fates(nc)%bc_out(s)%annsum_npp_pa(pf)
+
+            ! Values by pft from Table 1 in Brovkin et al. (2012):
+            ! https://bg.copernicus.org/articles/9/565/2012/bg-9-565-2012.pdf
+            ! TODO Move to the params files? Though this hardwiring will become
+            ! unnecessary when we calc. ligninNratioAvg the same way for FATES
+            ! and non FATES cases.
+            if (patch%itype(p) == noveg) then
```
slevisconsulting, line 1157: @ekluzek recommends using a switch / case statement instead of if-statement for computational efficiency.

slevisconsulting: also add a simple/short 1x1_brazil test of FATES-MIMICS on izumi to ensure that it keeps working.

ekluzek (changes requested):
@slevisconsulting and I went over this in a Google Meet call. I made a few suggestions, that he already added in.

I also think that @rgknox or @glemieux should review this as well.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
slevisconsulting: > @ekluzek recommends using a switch / case statement instead of if-statement for computational efficiency.

I'm still trying to get this to work because the compiler does not consider the integer options that I'm using as constants. I can't change them to parameters in pftconMod because they get calculated at runtime. So my current thought is to convert them to strings for the case statement to work.

ekluzek: Oh, these are the integers in pftcon right? There wouldn't be a way to make them into parameters in that case. If you convert to string that makes the evaluations much less efficient. So you should probably go back to the if, else if, structure. Switch statements are good to use if you can, but it's not a requirement. I was worth looking into though.

slevisconsulting: No, instead I will try using `pftname` which is a string already.

ekluzek: Yeah, the thing is that the string comparison will be slower than the integer comparison. So although the switch statement might be faster, overall it might be slower, because it has to do a string comparison. I suspect there will be a bigger slowdown in using a string comparison over the gain from using the switch statement. These strings are pretty short so it shouldn't be horrible, but you do potentially have to compare each character individually rather than the value of the integer itself. Another consideration would be readability, but I'm not sure that either one is going to be that much better than the other.

slevisconsulting: ...sorry @ekluzek I didn't see that you had responded (github hadn't refreshed).

Since you brought up readability, I do think that the select statement is easier to follow because the strings describe the pfts/cfts explicitly, while in the if-statement they are in shorthand, which isn't 100% intuitive.

Also with the select statement I was able to omit the long use statement listing all the pfts/cfts because now I'm checking explicit strings.

One more thing to consider: This piece of code is temporary. Once FATES and non FATES cases use the same calculation of `ligninNratioAvg`, this code will be deleted.

If you still prefer that we go back to the if-statement, I'm happy to do so. Let me know.

slevisconsulting: > also add a simple/short 1x1_brazil test of FATES-MIMICS on izumi to ensure that it keeps working.

To avoid adding a new test, I'm opting for changing this existing izumi test
`ERS_Ld30.f45_f45_mg37.I2000Clm50FatesRs.clm-FatesColdDefSizeAgeMort`
to running with MIMICS. Let me know if I shouldn't do that.

slevisconsulting:
> I also think that @rgknox or @glemieux should review this as well.

@rgknox @glemieux 
A look at what I wrote in the ChangeLog would help.
The code mods aren't much different from the last time you looked, so that might be secondary.
Thanks for your input on this PR!

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
ekluzek: We talked about this and it would be OK to use an existing test that includes MIMICS. The important thing to do there is to make sure MIMICS is included in the testmod name. We do expect that there shouldn't be strange interactions that MIMICS would only work for some FATES options, but not others. In general we expect that in our testing, and you should inspect the code to see if there are unique interactions with other options that would require separate tests for some options. I don't think that's the case for MIMICS though.

ekluzek (approved):

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
slevisconsulting: > make sure MIMICS is included in the testmod name. 

That's taken care of now.

> We do expect that there shouldn't be strange interactions that MIMICS would only work for some FATES options, but not others. In general we expect that in our testing, and you should inspect the code to see if there are unique interactions with other options that would require separate tests for some options. I don't think that's the case for MIMICS though.

With this PR's mods (which allow FATES and MIMICS to work together), I'm pretty sure that you're right @ekluzek

src/biogeochem/EDBGCDynMod.F90
ekluzek: If it's not currently used you should remove it.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
ekluzek: Right because now there is a local version of the array used for FATES. And there is a scalar that's assigned for either fates on/or off that actually gets used. This version is only used for the non-FATES case.

ekluzek: I think this is OK. Technically the local version of the array is only used when FATES is on, but will be allocated while the subroutine is active even without FATES active. But, to get around that you'd need to add explicit allocate/deallocate statements which would add more infrastructure.

ekluzek:
This is ready for either @rgknox or @glemieux to review. We have a couple other tags scheduled that will come before this, but this should come in pretty quickly.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
ekluzek: @rgknox and/or @glemieux one of you should validate that this is correct.

bld/CLMBuildNamelist.pm
```diff
@@ -901,6 +901,12 @@ sub setup_cmdl_bgc {
            $log->warning("$var normally use_nitrif_denitrif should only be FALSE if FATES is on, it has NOT been validated for being off for BGC mode" );
         }
      }
+     # if MIMICS is on and use_fates = .true. then use_lch4 must = .true.
+     if ( (! &value_is_true($nl_flags->{'use_lch4'})) && &value_is_true($nl_flags->{'use_fates'}) ) {
+        if ( $soil_decomp_method eq "MIMICSWieder2015" ) {
+           $log->warning("If MIMICS is on and use_fates = .true. then use_lch4 must be .true. and currently it's not" );
```
rgknox, line 907: @slevisconsulting @ekluzek will this trigger a warning with or without a graceful fail?  Seems we need the former right?

ekluzek: Yes, I agree and that's what we have. The warning will abort by default with the above message. You can get around it if you tell build-namelist to ignore-warnings. It'll then still print the warning, but will build the namelist and you can run the model. I'm not sure what the code will do though if you run with FATES and MIMICS without methane? Will it just give bad results or will it abort in some strange way? Do we also need an abort in the FORTRAN code as well?

cime_config/testdefs/testlist_clm.xml
```diff
@@ -2033,13 +2033,13 @@
       <option name="comment">60 day exact restart test providing coverage for the FATES logging mode on an f45 grid.</option>
     </options>
   </test>
-  <test name="ERS_Ld30" grid="f45_f45_mg37" compset="I2000Clm50FatesRs" testmods="clm/FatesColdDefSizeAgeMort">
+  <test name="ERS_Ld30" grid="f45_f45_mg37" compset="I2000Clm50FatesRs" testmods="clm/mimicsFatesColdDefSizeAgeMort">
```
rgknox, line 2036: The ColdDefSizeAgeMort is a very specific way to run FATES with a tree mortality function added by @JessicaNeedham. 

I like the grid and that this is an ERS_Ld30.  But perhaps we should create a new test, and base the configuration off of FatesColdDef and not FatesColdDefSizeAgeMort.  Otherwise, we will have trouble disentangling any failure attribution to the size-age-mort stuff vis-a-vis the mimics coupling.

Also, we should give this an aux_clm-izumi, and a category="fates" with cheyenne.

@slevisconsulting , others, sound good?

ekluzek: That makes sense to me @rgknox. This is an example of why we wanted you to look at this.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1100,21 +1114,100 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
+            pf = p - col%patchi(c)
+            s  = clm_fates%f2hmap(nc)%hsites(c)
+            annsum_npp(p) = clm_fates%fates(nc)%bc_out(s)%annsum_npp_pa(pf)
+
+            ! Values by pft from Table 1 in Brovkin et al. (2012):
+            ! https://bg.copernicus.org/articles/9/565/2012/bg-9-565-2012.pdf
+            ! Not moving to the params files at this time: Hardwiring will
+            ! become unnecessary when we calc. ligninNratioAvg the same way for
+            ! FATES and non FATES cases.
+            select case (pftname(patch%itype(p)))
```
rgknox, line 1145: If use_fates is true, then none of these patch level PFTs are actually valid.  I need to read through this code more to figure out how non-fates mimics arrives at this ligninRatio to give any recommendations thought.  brb

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1100,21 +1114,100 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
       mimics_cn_r = params_inst%mimics_cn_r
       mimics_cn_k = params_inst%mimics_cn_k
 
+      ! Local column-level ligninNratioAvg because
+      ! not available when use_fates = .true.
+      ! Also get FATES copy of annsum_npp for FATES cases
+      if (use_fates) then
```
rgknox, line 1120: recommend using a label here: a la   "fates_if:  if (use_fates) then"

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1100,21 +1114,100 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
+
       ! calculate rates for all litter and som pools
       do fc = 1,num_soilc
          c = filter_soilc(fc)
 
+         if (use_fates) then
+            ligninNratioAvg_scalar = ligninNratioAvg_local(c)
+            annsum_npp_col_scalar = max(0._r8, annsum_npp_col_local(c))
+         else
+            ligninNratioAvg_scalar = cnveg_carbonflux_inst%ligninNratioAvg_col(c)
```
rgknox, line 1197: I was tracking down how mimics w/o fates arrives at this column level cnveg_carbonflux_inst%ligninNratioAvg_col(c)

Which took me here:

https://github.com/ESCOMP/CTSM/blob/master/src/biogeochem/CNVegCarbonFluxType.F90#L4498-L4503

It looks like its taking an average of the leaf, fineroot and cwd pools.  But I was expecting that the terms ligninNratio_leaf_col(c), ligninNratio_froot_col(c) and ligninNratio_cwd  would all be multiplied by their respective carbon weighting factors (ie leafc_to_litter_col, etc) before being divided through by the sum of those factors.  I probably missed something though, maybe they already are, or I don't understand the units.

rgknox: I think I understand it now, looking here at the leaf term:
https://github.com/ESCOMP/CTSM/blob/master/src/biogeochem/CNVegCarbonFluxType.F90#L4422-L4424
I see that it is pre-multiplied by leafc_to_litter_patch(p). 
Alright, I see no issue.

rgknox: However, I really think there should be units for the definition of the column level average, it wasn't clear to me exactly what this ratio is.  It is defined here:

https://github.com/ESCOMP/CTSM/blob/master/src/biogeochem/CNVegCarbonFluxType.F90#L338

And I think it should have the same units as the multiple of these two factors (using leaf as a guide):

real(r8), allocatable :: lf_flig       (:)   ! leaf litter lignin fraction
real(r8), allocatable :: lflitcn       (:)   ! leaf litter C:N (gC/gN)

So it seems like this column level average is the mean C:N ratio of all lignin tissues in the column.  [gC/gN]

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
rgknox: @slevisconsulting and @wwieder , I see two ways to handle this:
1) just use a single representative constant for any and all FATES patches.
2) we could have FATES pass in this ratio right now. Even though we don't have the dynamic FATES N model coupled with CLM yet, we do have the stoichiometries and the lignin fractions of all of our fates generated litter.

Perhaps we should use item 1 for right now, and we can have a follow up PR to add in item 2? (I could help with that one)

rgknox (changes requested):
See comments in line.  My only serious concern was how we calculate the ligninNratio, and I think the simple solution of using a single constant value for right now is the way to go, and then follow up with a FATES API changing PR that passes in this value from FATES.

bld/CLMBuildNamelist.pm
slevisconsulting: I see what you're saying @ekluzek ...if use_lch4 = .false.:
- This will abort with the error gracefully by default, which I have confirmed, BUT
- if user bypasses the error with `ignore-warnings`, the message will only appear as a warning. I don't know off the top of my head whether the run will abort, but it will certainly give bad results because it won't have access to the annsum_npp variable from FATES.
Let me know how you prefer to resolve this. Do we have error checks in the fortran for all error checks that we bypass with `ignore-warnings`? Or do we assume that users take matters in their own hands when they `ignore-warnings`?

ekluzek: We sometimes both have an exit in the build-namelist as well as the FORTRAN code. And sometimes we just have it in the build-namelist, because it's easier. There are also things that the build-namelist can't really catch, so it's only in the FORTRAN code.

But, yes I think it's reasonable to have the stance that if the user sets "ignore-warnings" that they are on their own and get what they deserve. I do think that the worst kind of problem to have though is silently giving bad results. That's expensive in terms of human and computer time. So I can go either way on this in terms of adding a check in the FORTRAN code.

slevisconsulting:
> See comments in line. My only serious concern was how we calculate the ligninNratio, and I think the simple solution of using a single constant value for right now is the way to go, and then follow up with a FATES API changing PR that passes in this value from FATES.

Thank you for your thorough review @rgknox !

@ekluzek I think this PR comes off the ready-to-merge list for a few weeks because I'm working on other priorities today and then mostly unavailable through March 15th.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
wwieder: This plan seems reasonable.  @slevisconsulting how do we keep track of this suggestion # 2 from @rgknox? 

Alternatively, if it's not too difficult maybe we just do this right from the start?

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
wwieder: Thanks for diving in here @rgknox.  @slevisconsulting maybe some additional comments in the code are warranted to clarify what's going on?

bld/CLMBuildNamelist.pm
slevisconsulting: I decided to add an error check in the fortran.

cime_config/testdefs/testlist_clm.xml
slevisconsulting: @rgknox @ekluzek 
I reverted the existing test to the way it was and added the recommended test as
`name="izumi" compiler="intel" category="aux_clm"`

Were you recommending that I repeat this test on cheyenne with category="fates"? I'm double checking to avoid unnecessary proliferation of tests :-)

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
slevisconsulting: @rgknox let's follow up to pursue option 2. How's your availability this wk on Weds, Thurs, or Friday afternoons?

slevisconsulting: In our meeting @rgknox and I agreed that he would make the necessary changes on the fates side to accomplish item 2 above. Timeframe about 2-4 weeks.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
```diff
@@ -1138,44 +1136,7 @@ subroutine decomp_rates_mimics(bounds, num_soilc, filter_soilc, &
...
-                        'tropical_corn', 'irrigated_tropical_corn', &
-                        'temperate_corn', 'irrigated_temperate_corn', &
-                        'sugarcane', 'irrigated_sugarcane', &
-                        'sorghum', 'irrigated_sorghum', &
-                        'millet', 'irrigated_millet')  ! all c4
-                     ligninNratio(p) = 24.1_r8  ! 23.4 / 0.97
-                  case default  ! all pfts/cfts that do not fit above categories
-                     ligninNratio(p) = 12.1_r8  ! 16.6 / 1.37
-               end select
+               ligninNratio(p) = clm_fates%fates(nc)%bc_out(s)%litt_flux_ligc_per_n(pf)
```
slevisconsulting, line 1139: @rgknox @ekluzek 
This is the change I made so as to use the new variable `litt_flux_ligc_per_n` calculated by FATES. I followed the template of what I had done with the other FATES variable used here `annsum_npp_pa`.

**Two problems that I need to go over with you:**
1) I had to make some modifications to FATES for the code to build.
2) The test that I introduced in this PR runs, but COMPARE_base_rest fails.

**The details:**
1) I added 1 line and changed 2 lines in FATES for the build to complete:
`FatesInterfaceMod.F90`
@@ -305,8 +305,8 @@ contains
```
     fates%bc_out(s)%rootr_pasl(:,:) = 0.0_r8
     fates%bc_out(s)%btran_pa(:)     = 0.0_r8

-    ! MIMIC litter quality, always initialize to unset
-    fates%bc_out(s)%litt_flux_ligc_per_n = fates_unset_r8
+    ! MIMICS litter quality, always initialize to unset
+    fates%bc_out(s)%litt_flux_ligc_per_n(:) = fates_unset_r8
```
@@ -589,6 +589,8 @@ contains
```
       allocate(bc_out%ftid_parb(maxPatchesPerSite,hlm_numSWb))
       allocate(bc_out%ftii_parb(maxPatchesPerSite,hlm_numSWb))

+      ! MIMICS
+      allocate(bc_out%litt_flux_ligc_per_n(maxPatchesPerSite))
```

`FatesInterfaceTypesMod.F90`
@@ -609,8 +609,7 @@ module FatesInterfaceTypesMod
```
       ! MIMICS Boundary Conditions
       ! -----------------------------------------------------------------------------------
-      real(r8) :: litt_flux_ligc_per_n  ! lignin carbon per total nitrogen
-                                        ! in the fragmentation flux, per square meter [g/g]
+      real(r8), allocatable :: litt_flux_ligc_per_n(:)  ! lignin carbon per total nitrogen in the fragmentation flux, per square meter [g/g]
```

2) Regarding the failing restart, I wanted to check in before trying to solve this:
@rgknox I suspect that some part of the calculation leading to the new variable `litt_flux_ligc_per_n` must go to restart. I will start looking, but I wanted your input, in case your familiarity with the calculation leads you to a quick answer.

slevisconsulting: @rgknox looking at the code more:
I started with the assumption of a patch-level variable, but I think you're calculating `litt_flux_ligc_per_n` at the Site level. If so, I need to revisit my changes and see it I get things to work without additional mods on the FATES side.

slevisconsulting: That was the problem. The test now passes.

Feel free to review my mods for further comments/suggestions/corrections.

Meanwhile I will kick off the test suites.

rgknox: ok, great, nevermind, I see you did exactly this in commit dd24018

slevisconsulting: Perfect, thank you for that confirmation, @rgknox .

Good news with the test suites:
Cheyenne PASS
Izumi OK (i.e. 2 expected failures)

src/soilbiogeochem/SoilBiogeochemCarbonFluxType.F90
```diff
@@ -31,6 +31,7 @@ module SoilBiogeochemCarbonFluxType
      real(r8), pointer :: decomp_cascade_ctransfer_vr_col           (:,:,:) ! vertically-resolved C transferred along deomposition cascade (gC/m3/s)
      real(r8), pointer :: decomp_cascade_ctransfer_col              (:,:)   ! vertically-integrated (diagnostic) C transferred along decomposition cascade (gC/m2/s)
      real(r8), pointer :: cn_col                                    (:,:)   ! (gC/gN) C:N ratio by pool
+     real(r8), pointer :: litr_lig_c_to_n_col                       (:)     ! (gC/gN) lignin C:N ratio; FATES equivalent of cnveg_carbonflux_inst%ligninNratioAvg_col because FATES doesn't have access to cnveg_carbonflux_inst; could reduce duplication by consolidating cnveg_carbonflux_inst%ligninNratioAvg_col into this one
```
slevisconsulting, line 34: What I want to try before we merge this code is reduce the duplication explained in the above comment.

slevisconsulting: The next commit handles the duplication reduction successfully. 
The CTSM-MIMICS test passes, including compared to the baseline. 
The CTSM-MIMICS-FATES test passes.

I have submitted the test suites and will report results in the morning.

slevisconsulting: A short list of tests FAIL on cheyenne and izumi, so I need to investigate.

slevisconsulting:
Test suites:
Cheyenne OK (expected failures due to new variable added to history; otherwise bit-for-bit)
Izumi ONE FAILURE: With my latest changes, the restart test fails in the new MIMICS-FATES test...

src/utils/clmfates_interfaceMod.F90
```diff
@@ -1111,13 +1111,26 @@ subroutine wrap_update_hlmfates_dyn(this, nc, bounds_clump,      &
...
+       ! (this will happen in the first timestep of any restart)
+       ! then skip this variable because a more accurate value was obtained
+       ! from the restart file.
+       ! I had hoped that is_first_restart_step() alone would suffice here, but
+       ! is_first_restart_step() remained true for the whole first day in my
+       ! test. Hence I added the check for copy_fates_var which changes to
+       ! .true. the first time that we come through this code in a restart.
+       !------------------------------------------------------------------------
+       if (is_first_restart_step() .and. .not. copy_fates_var) then
+          copy_fates_var = .true.
```
slevisconsulting, line 1126: @rgknox @ekluzek 
This gets the MIMICS-FATES restart test to pass by not copying the variable from FATES when `is_first_restart_step = .true.` and using instead the corresponding CTSM variable that I sent to restart. For more, pls read the comment above the if statement.

My question is whether you consider this solution acceptable or would you rather determine why the FATES variable isn't restarting correctly?

Meanwhile I have submitted the test suites and will report results in the morning.

slevisconsulting: So this PR is ready to merge unless anyone has concerns about my solution to the failing restart MIMICS-FATES test (my last commit).

ekluzek: Sounds like we should maybe meet to go over this code. Sometime tomorrow work?

ekluzek: This should be threadsafe as different threads would have different instances of the entire wrap_update_hlmfates_dyn subroutine. So that should be OK as long as it's not saved. We should maybe make sure we have a threading restart test though.

There is a save statement on the initialized variable and I wonder if that is threadsafe.

ekluzek: Nope, @rgknox is correct, this is global so needs to be a module array based on the number of threads. Thanks for pointing that out!

slevisconsulting: We met and decided that I will:
- confirm that other fates variables getting copied to ctsm variables have same values at restart as in a continuous run
- if so, look for the variable responsible for the failing restart in the calculation of our new variable `litt_flux_ligc_per_n`

slevisconsulting: ...I take that back. Follow-up with @rgknox suggests that I pursue the workaround for now.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -541,21 +542,19 @@ subroutine init(this, bounds_proc )
       allocate(this%fates(nclumps))
       allocate(this%f2hmap(nclumps))
 
-
       if(debug)then
          write(iulog,*) 'clm_fates%init():  allocating for ',nclumps,' threads'
       end if
 
 
-      nclumps = get_proc_clumps()
```
slevisconsulting, line 550: I removed this line because it's repeated a few lines above.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -541,21 +542,19 @@ subroutine init(this, bounds_proc )
...
-
       !$OMP PARALLEL DO PRIVATE (nc,bounds_clump,nmaxcol,s,c,l,g,collist,pi,pf,ft)
       do nc = 1,nclumps
-
          call get_clump_bounds(nc, bounds_clump)
          nmaxcol = bounds_clump%endc - bounds_clump%begc + 1
 
          allocate(collist(1:nmaxcol))
+         allocate(copy_fates_var(1:nmaxcol))
+         copy_fates_var(:) = .false.  ! .false. when starting any run
```
slevisconsulting, line 557: Now allocating and initializing `copy_fates_var` in subroutine init.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -1106,6 +1109,33 @@ subroutine wrap_update_hlmfates_dyn(this, nc, bounds_clump,      &
...
+       ! Note 2. copy_fates_var is a column-level array to make thread-safe.
+       ! Note 3. This if statement is a workaround for restart tests to pass.
+       ! Ideally, the fates variable would have the correct value at restart,
+       ! but rgknox explains that accomplishing this is more complex given the
+       ! current FATES treatment of other litter fluxes passed to the CTSM.
+       !------------------------------------------------------------------------
+       do s = 1, this%fates(nc)%nsites
+          c = this%f2hmap(nc)%fcolumn(s)
+          if (is_first_restart_step() .and. .not. copy_fates_var(c)) then
+             copy_fates_var(c) = .true.
```
slevisconsulting, line 1132: Moved the if statement inside the column loop now.

ekluzek: @slevisconsulting I'm not sure we should add this to the test suite. But, I would like you to run a threaded test for this. So run

ERP_P144x2_Ld30.f45_f45_mg37.I2000Clm50FatesRs.cheyenne_intel.clm-mimicsFatesColdDef

It looks like currently the only threaded FATES test we run is:

ERP_D_P32x2_Ld3.f19_g17.I2000Clm50FatesCru.cheyenne_intel.clm-FatesColdDef

But, it is passing. We should likely add some more...

slevisconsulting: Good call @ekluzek 
this threaded test fails, so I'm investigating.

ekluzek: You might try some other threaded tests to see what works and what breaks. I'd try both changing the above ERP to ERS and SMS. And you should try some ones that are closer to the test that works in the baseline. So try the baseline test of:

ERP_D_P32x2_Ld3.f19_g17.I2000Clm50FatesCru.cheyenne_intel.clm-FatesColdDef

and then also do...

ERP_D_P32x2_Ld3.f19_g17.I2000Clm50FatesCru.cheyenne_intel.clm-mimicsFatesColdDef

That should help to figure out if it's threading and FATES in general or if it's just in MIMICS/FATES. And the SMS, ERS, and ERP tests tell if it's a problem with threading in general, with restarts in general, or only when changing PE's on restart. All of that would be good to know.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -583,6 +580,8 @@ subroutine init(this, bounds_proc )
                end if
             endif
 
+            col%copy_fates_var(c) = .false.
```
slevisconsulting, line 583: @ekluzek I ended up using col%is_fates(c) as a template to get copy_fates_var to work in the threaded test.

You recommended "staying local" if possible, but none of my corresponding experiments succeeded:
- I tried allocating locally in this subroutine init with
`copy_fates_init(bounds_clump%begc:bounds_clump%endc)`
but failed with error "trying to allocate an already allocated variable."
- I tried allocating locally in subroutine Init2, and the continuous run finished, but the restart run failed with a segfault. This may mean that restarts don't run through subroutine Init2.

If you have another suggestion that you would like me to try, pls let me know, though I probably won't get to it before Monday. In the mean time, I think I'll update to the latest tag and start the test suites, so as to be ready by Monday if no tests fail :-)

slevisconsulting: > In the mean time, I think I'll update to the latest tag and start the test suites

cheyenne Ok
izumi Ok
...expected failures + FIELDLIST field lists differ (otherwise bit-for-bit)

slevisconsulting:
@ekluzek recommends adding the threading test that I used above to the test-suite, likely in aux-clm.

ekluzek:
@slevisconsulting @glemieux and I went over this, this morning. Since the copy_fates_var is really specific to the clmfates_interface model I think it's important to keep it local there. So we went over how to make it a local module level variable, and it looks like that's working now. Since, fates-SP mode has the biggest changes to the CTSM side of running FATES it makes sense to have a threading test specific for it.

slevisconsulting:
Test suites:
cheyenne OK
izumi OK

I think the PR is ready. Pls let me know if I'm forgetting something...

cime_config/testdefs/testlist_clm.xml
ekluzek: @slevisconsulting I think this is one thing that is worth doing. Change the name of the testmod and also remove the setting for size-age-mort stuff. So it just turns on mimics. And then redo the tests that have this test mod.

With that this should be ready to go. So let me know when you have that ready and I can make it into a tag.

slevisconsulting: @ekluzek I'm pretty sure you're talking about a change that I made in https://github.com/ESCOMP/CTSM/pull/1643/commits/d95a71623115455b223e3aa5387a646045dc757c
where I backed my changes out of any existing tests.

If you look at the cumulative changes from all commits in this PR and search for testlist_clm.xml, you'll see only the new test that we talked about today.

src/soilbiogeochem/SoilBiogeochemDecompCascadeMIMICSMod.F90
slevisconsulting: Did do for lines 1118, 1119 where the if statements are long:

```
fates_if: if (use_fates) then
         lch4_if: if (use_lch4) then
```

True, didn't do elsewhere.

cime_config/testdefs/testlist_clm.xml
ekluzek: Ahhh, OK. For some reason looking at it here, I was seeing the earlier commit and not the latest. The latest looks fine. Thanks for doing that renaming.

Issue #1646, opened 2022-02-13
Time evolving parameter values


wwieder:
I've been in several discussions with groups asking about the feasibility of implementing time evolving parameterizations for CTSM that can be used for sensitivity tests in the model.  For example, there seems to be good evidence that SLA and tissue stoichiometry changes under elevated CO2, but these are effectively fixed by values in the parameter file.  Is this something that can be modified (e.g. as a function of CO2) at one place in the code, or would it have to be modified in individual modules where a parameter is used?

e.g.  `pftcon%leafcn = pftcon%leafcn  * ln(CO2now/360)`  

Interested to hear ideas about how this may be possible to explore.

wwieder:
@ekluzek suggested this could be something in pftconMod.f90 in a hacky way

wwieder:
The parameter would also have to be updated in clm_driver.f90 (and maybe added to history)

slevisconsulting:
clm_driver would call a new subroutine that's located in pftconMod.f90 and the new subroutine would contain the function that makes the parameter(s) time-varying.

slevisconsulting:
CO2_base for now will be hardwired in the new subroutine.

slevisconsulting:
Initially test with single-pt with and without time-varying leafcn.

Later will follow up with global runs.

slevisconsulting:
Met with @ekluzek 
Motivated by errors that I'm getting when I place the new subroutine in pftconMod, we agreed that I should place it in CanopyStateMod instead.

ekluzek:
@wwieder I've got some thoughts on the philosophy of this and how to organize it that we should perhaps discuss. The specific solution that @slevisconsulting is pursing makes sense for this specific case. And if it's impactful and interesting and we want to bring it into the main code, we'll go over the implementation details, and may make some changes.

The philosophical part though is what does it mean for "parameters" to change in time? It seems like in general parameters should be constants that aren't going to change in time, and if they do the parameterization itself should change to have additional parameters that define the change in time. So like in this example you have new parameters that define how something changes relative to CO2 (or something else that may change in time).

[...]

Issue #1332, opened 2021-04-10
I think some additional atm2lnd terms need to be downscaled


billsacks:
### Brief summary of bug

I came across a number of atm2lnd terms that have units of Pa, which I think should be included in the atm -> lnd downscaling.

### General bug information

**CTSM version you are using:** master; this bug (if I'm right in thinking it's a bug) would have been in place for a long time

**Does this bug cause significantly incorrect results in the model's science?** Possibly

**Configurations affected:** I think this impacts any configuration with CISM – where we downscale to vegetated columns as well as glacier columns over Greenland. It may also impact any configuration with multiple glacier elevation classes (which is typically the case over Greenland and Antarctica), though I'm not sure if any of the problematic quantities have an impact on glacier columns.

### Details of bug

Looking through the code in lnd_import and atm2lndType, I realized that there are a number of variables that have units of Pa and are derived through a multiplication by the gridcell-level pbot value. In columns where we downscale atmospheric quantities to the column level, pbot gets adjusted for each column. Some of these other variables are also downscaled, but not all.

Potentially impacted quantities that I see from a quick look are co2, ch4 and variables that depend on pbot in this subroutine:

https://github.com/escomp/CTSM/blob/master/src/cpl/utils/lnd_import_export_utils.F90#L24-L99

There may be others, too, but those are the ones that catch my eye.

My guess is that these variables mostly / entirely impact vegetated columns. Most of the current downscaling is done over glacier columns, but we also downscale to vegetated columns in the CISM domain (typically Greenland), so I think this would at least have an impact there. This may have more widespread impacts in @swensosc 's branch where I think he has extended the downscaling to operate over hillslope columns.

### Solving the problem

A relatively straightforward solution would be to add these extra variables to the downscaling done in subroutine downscale_forcings, creating downscaled versions that depend on the column-level pbot.

An alternative, which may help prevent errors like this in the future, would be: Do not create gridcell-level derived quantities that depend on pbot in lnd_import / derive_quantities. Instead, wait until after downscale_forcings has been done and then create the derived quantities at the column level, based on the downscaled pbot. (However, ch4 and possibly other variables seem to already come in with units of pressure, so would need to be added to the downscaling in any case.)

In either case, references to these variables in the science code will need to be changed to refer to downscaled, column-level quantities.

billsacks:
Probably bring this in with or before hillslope.

wwieder:
[...]

PR #1654, opened 2022-02-18
Introduce LEAFCN_TARGET as time-evolving parameter

slevisconsulting:
### Description of changes

From clm_driver I call new `subr. time_evolv_params` located in `CanopyStateType.F90` where I calc. the new variable `leafcn_patch` (history field LEAFCN_TARGET).

### Specific notes

@wwieder @ekluzek 

#1646 

Yes, once I replace all pftcon%leafcn with canopystate_inst%leafcn_patch(p)

New history field LEAFCN_TARGET

Testing performed, if any:
In /glade/scratch/slevis, both of these tests PASS:
SMS_Ly3_Mmpi-serial.1x1_numaIA.IHistClm50BgcCropQianRs.cheyenne_intel.clm-default.20220218_111830_bodlmh
SMS_Ly3_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.cheyenne_intel.clm-default.20220218_105852_zphwfb
I will plot timeseries of LEAFCN_TARGET to compare the IHist case to the I2000 case.
For now I have only looked at a single hist. file (SMS...I2000...clm2.h0.2002-02-28-00000.nc) and found:
leafcn_target = 26.0
leafc / leafn = 55.4
leafcn = 34.7

### DIFF
--- a/src/main/clm_driver.F90
+++ b/src/main/clm_driver.F90
@@ -165,6 +165,11 @@ contains
     call get_proc_bounds(bounds_proc)
     nclumps = get_proc_clumps()
 
+    ! -------------------------------------------------
+    ! Obtain updated values of time-evolving parameters
+    ! -------------------------------------------------
+    call canopystate_inst%time_evolv_params(bounds_proc, atm2lnd_inst)
+
     ! ========================================================================
     ! In the first time step of a startup or hybrid run, we want to update CLM's glacier
     ! areas to match those given by GLC. This is because, in initialization, we do not yet
--- a/src/biogeophys/CanopyStateType.F90
+++ b/src/biogeophys/CanopyStateType.F90
@@ -173,6 +173,11 @@ contains
     begp = bounds%begp; endp= bounds%endp
     begc = bounds%begc; endc= bounds%endc
 
+    this%leafcn_patch(begp:endp) = spval
+    call hist_addfld1d (fname='LEAFCN_TARGET', units='gC/gN', &
+        avgflag='A', long_name='Target leaf C:N; compare against leafC/leafN', &
+        ptr_patch=this%leafcn_patch)
+
     this%elai_patch(begp:endp) = spval
     call hist_addfld1d (fname='ELAI', units='m^2/m^2', &
         avgflag='A', long_name='exposed one-sided leaf area index', &
@@ -233,11 +238,6 @@ contains
 
 
     endif !fates or CN
-
-    this%leafcn_patch(begp:endp) = spval
-    call hist_addfld1d (fname='LEAFCN_TARGET', units='gC/gN', &
-        avgflag='A', long_name='Target leaf C:N; compare against leafC/leafN', &
-        ptr_patch=this%leafcn_patch)
 
     if(use_fates_sp)then
       this%tlai_hist_patch(begp:endp) = spval
@@ -609,8 +609,8 @@ contains
          dim1name='pft', long_name='fraction of vegetation not covered by snow (0 or 1)', units='', &
          interpinic_flag='interp', readvar=readvar, data=this%frac_veg_nosno_alb_patch)
 
-!   call restartvar(ncid=ncid, flag=flag, varname='LEAFCN', xtype=ncd_double,  &
-!        dim1name='pft', long_name='leaf C:N', units='gC/gN', &
+!   call restartvar(ncid=ncid, flag=flag, varname='LEAFCN_TARGET', xtype=ncd_double,  &
+!        dim1name='pft', long_name='Target leaf C:N; compare against leafC/leafN', units='gC/gN', &
 !        interpinic_flag='interp', readvar=readvar, data=this%leafcn_patch)
 
     call restartvar(ncid=ncid, flag=flag, varname='tlai', xtype=ncd_double,  &
--- a/src/biogeophys/CanopyStateType.F90
+++ b/src/biogeophys/CanopyStateType.F90
@@ -235,8 +235,8 @@ contains
     endif !fates or CN
 
     this%leafcn_patch(begp:endp) = spval
-    call hist_addfld1d (fname='leafcn', units='gC/gN', &
-        avgflag='A', long_name='Leaf C:N', &
+    call hist_addfld1d (fname='LEAFCN_TARGET', units='gC/gN', &
+        avgflag='A', long_name='Target leaf C:N; compare against leafC/leafN', &
         ptr_patch=this%leafcn_patch)
 
     if(use_fates_sp)then
--- a/src/biogeophys/CanopyStateType.F90
+++ b/src/biogeophys/CanopyStateType.F90
@@ -7,6 +7,7 @@ module CanopyStateType
   use shr_log_mod     , only : errMsg => shr_log_errMsg
   use abortutils      , only : endrun
   use decompMod       , only : bounds_type
+  use atm2lndType     , only : atm2lnd_type
   use landunit_varcon , only : istsoil, istcrop
   use clm_varpar      , only : nlevcan, nvegwcs
   use clm_varcon      , only : spval
@@ -23,6 +24,8 @@ module CanopyStateType
 
      integer  , pointer :: frac_veg_nosno_patch     (:)   ! patch fraction of vegetation not covered by snow (0 OR 1) [-]
      integer  , pointer :: frac_veg_nosno_alb_patch (:)   ! patch fraction of vegetation not covered by snow (0 OR 1) [-]
+
+     real(r8) , pointer :: leafcn_patch             (:)   ! patch leafcn
 
      real(r8) , pointer :: tlai_patch               (:)   ! patch canopy one-sided leaf area index, no burying by snow
      real(r8) , pointer :: tsai_patch               (:)   ! patch canopy one-sided stem area index, no burying by snow
@@ -70,6 +73,7 @@ module CanopyStateType
      procedure, public  :: InitAccBuffer
      procedure, public  :: InitAccVars
      procedure, public  :: UpdateAccVars
+     procedure, public  :: time_evolv_params  ! updates time-evolving params
      procedure, public  :: Restart
 
   end type CanopyState_type
@@ -120,6 +124,7 @@ contains
     allocate(this%tlai_hist_patch          (begp:endp))           ; this%tlai_hist_patch          (:)   = nan
     allocate(this%tsai_hist_patch          (begp:endp))           ; this%tsai_hist_patch          (:)   = nan
     allocate(this%htop_hist_patch          (begp:endp))           ; this%htop_hist_patch          (:)   = nan
+    allocate(this%leafcn_patch             (begp:endp))           ; this%leafcn_patch             (:)   = nan
     allocate(this%tlai_patch               (begp:endp))           ; this%tlai_patch               (:)   = nan
     allocate(this%tsai_patch               (begp:endp))           ; this%tsai_patch               (:)   = nan
     allocate(this%elai_patch               (begp:endp))           ; this%elai_patch               (:)   = nan
@@ -228,6 +233,11 @@ contains
 
 
     endif !fates or CN
+
+    this%leafcn_patch(begp:endp) = spval
+    call hist_addfld1d (fname='leafcn', units='gC/gN', &
+        avgflag='A', long_name='Leaf C:N', &
+        ptr_patch=this%leafcn_patch)
 
     if(use_fates_sp)then
       this%tlai_hist_patch(begp:endp) = spval
@@ -511,6 +521,7 @@ contains
     do p = bounds%begp, bounds%endp
        l = patch%landunit(p)
 
+       this%leafcn_patch(p)      = 0._r8
        this%tlai_patch(p)        = 0._r8
        this%tsai_patch(p)        = 0._r8
        this%elai_patch(p)        = 0._r8
@@ -537,6 +548,42 @@ contains
 
   end subroutine InitCold
 
+  !-----------------------------------------------------------------------
+  subroutine time_evolv_params(this, bounds, atm2lnd_inst)
+    !
+    ! !DESCRIPTION:
+    ! Update time-evolving parameters
+    !
+    ! !USES:
+    use pftconMod, only : pftcon
+    !
+    ! !ARGUMENTS:
+    class(canopystate_type)        :: this
+    type(bounds_type) , intent(in) :: bounds
+    type(atm2lnd_type), intent(in) :: atm2lnd_inst
+    !
+    ! !LOCAL VARIABLES:
+    integer :: g, c, p  ! indices
+    real(r8) :: co2_ppmv  ! atm co2 concentration
+    ! Could use 355 value from clm_varctl.F90 to avoid this hardwiring?
+    real(r8), parameter :: co2_base = 350._r8  ! units ppmv
+    real(r8), parameter :: cn_slope = 10._r8
+
+    character(len=*), parameter :: subname = 'time_evolv_params'
+    !-----------------------------------------------------------------------
+
+    ! Loop to get leafcn_col
+    do p = bounds%begp, bounds%endp
+       c = patch%column(p)
+       g = patch%gridcell(p)
+       co2_ppmv = 1.e6_r8 * atm2lnd_inst%forc_pco2_grc(g) / &
+                            atm2lnd_inst%forc_pbot_downscaled_col(c)
+       this%leafcn_patch(p) = pftcon%leafcn(patch%itype(p)) + &
+          max(cn_slope * log(co2_ppmv / co2_base), 0._r8)
+    end do
+
+  end subroutine time_evolv_params
+
   !------------------------------------------------------------------------
   subroutine Restart(this, bounds, ncid, flag)
     !
@@ -561,6 +608,10 @@ contains
     call restartvar(ncid=ncid, flag=flag, varname='FRAC_VEG_NOSNO_ALB', xtype=ncd_int,  &
          dim1name='pft', long_name='fraction of vegetation not covered by snow (0 or 1)', units='', &
          interpinic_flag='interp', readvar=readvar, data=this%frac_veg_nosno_alb_patch)
+
+!   call restartvar(ncid=ncid, flag=flag, varname='LEAFCN', xtype=ncd_double,  &
+!        dim1name='pft', long_name='leaf C:N', units='gC/gN', &
+!        interpinic_flag='interp', readvar=readvar, data=this%leafcn_patch)
 
     call restartvar(ncid=ncid, flag=flag, varname='tlai', xtype=ncd_double,  &
          dim1name='pft', long_name='one-sided leaf area index, no burying by snow', units='', &


wwieder:
Hi Sam,

Thanks for getting started on this.  I don't really understand what you're reporting with  this test, but I really don't understand why `leafcn_target < leafcn` given the code mods you made.    Even more surprising is the really high values for `leafc / leafn`. Maybe there are some details here to discuss?
```
leafcn_target = 26.0
leafc / leafn = 55.4
leafcn = 34.7
```

slevisconsulting:
The time series may clarify...
@wwieder the high C:N may have been specific to the model day (Feb 28) that I randomly looked at.

1x1_numa_I2000
leafcn_vars_I2000

1x1_numa_IHist
leafcn_vars_IHist

slevisconsulting:
As I replace `leafcn(ivt(p))` (the parameter) with `leafcn_patch(p)` throughout the code, I now see that it makes even more sense to declare `leafcn_patch` in **CNVegNitrogenStateType.F90** rather than in **CanopyStateType.F90**. I say this because the model uses `leafcn_patch` in **CNVegNitrogenStateType.F90** right after reading the restart file. 

If I make this change, FATES will not have access to the time-evolving leafcn, but that probably doesn't matter while FATES is C-only.  @wwieder and/or @ekluzek pls confirm, thanks.

ekluzek:
Moving it to CNVegNitrogenStateType makes sense to me, as it's really a BGC variable and not a canopy variable that's always used (including in SP mode). By appearing in CanopyState you get the impression that it's also used in SP mode and it's not. You are right that means it's not available to FATES, but I think for the purposes you are using LEAFCN_TARGET it's OK that it won't be available with FATES. I don't think FATES has a similar variable that it could operate on, so this whole thing would need to be rethought for working inside of FATES. And you are right the main FATES code is Carbon only, with nutrients coming in as an extension. So Nitrogen is being worked on, but it's new development.

slevisconsulting:
Sanity check after moving `leafcn_patch` from **CanopyStateType** to **CNVegNitrogenStateType** (I will push corresponding commit soon):
Got bfb identical history from `SMS_Ly3_Mmpi-serial.1x1_numaIA.IHistClm50BgcCropQianRs.cheyenne_intel.clm-default` as before.

slevisconsulting:
@wwieder mods were more widespread than I had originally expected, but I think I got them all.

This far, testing tells me that the new code is working.

How would you like to proceed:
- Review the code?
- Suggest some runs? Longer and/or global?

src/biogeochem/CNVegNitrogenStateType.F90
```diff
@@ -1057,6 +1081,42 @@ subroutine Summary_nitrogenstate(this, bounds, num_allc, filter_allc, &
 
   end subroutine Summary_nitrogenstate
 
+  !-----------------------------------------------------------------------
+  subroutine time_evolv_params(this, bounds, atm2lnd_inst)
```
slevisconsulting, line 1085: Given the coversation about this, what if I renamed to `time_evolv_leafcn`?

wwieder:
wow, there were a lot of changes @slevisconsulting  Thanks for doing all of this.

Can we try a few fully transient single point historical runs (maybe for Harvard Forest, 42.5° N, 72.2° W, where we have a good foliar C:N record).  
- Let's do this with `co2_base = 310`, which means foliar stoichiometry will start changing around 1940.
- It's likely easiest to have just a single pft for this run (temp decid tree, 7) and no LULCC.      
- A 400 y AD spinup and 200y postAD  for 1850 would be enough, unless you can interpolate from initial conditions. 
- Then let's do two historical simulations (starting in 1850).  One each with:
   - `cn_slope = 0`, which should give identical results to the current parameterization and 
   - `cn_slope = 30`, so we get a pretty big change.

slevisconsulting:
cn_slope = 0 during spinup
400 yrs AD_SPINUP
transient CO2 and climate but nothing else
use subset_data to get fsurdat
question about whether will need mesh domain file for this 1x1 site
./create_newcase with same pts_lat and pts_lon that I select in subset_data and don't mess w datm streams
(OR ./run_neon for harv but then I have to change datm streams but don't have to create fsurdat)

billsacks (changes requested):
I was curious what you did here, so looked through it. I have a few comments inline. In addition:

- [ ] I think some more thought needs to be given to the placement of the call to `time_evolv_leafcn` in the driver loop: I think you need to do one of two things (pick one):
1. Since this depends on a downscaled atmospheric forcing, and downscale_forcings is called later in the driver loop, this call to `time_evolv_leafcn` should be moved to somewhat later in the driver loop, near the start of the physics – perhaps right after the call to downscale_forcings. In this case, I think you'll need to keep leafcn on the restart file (as you currently have it), because it looks like this is needed in dynSubgridDriver, which is called before this point.
2. Alternatively, you could keep this call early in the driver loop (e.g., where it is now), but then it can only depend on a limited set of variables, such as the non-downscaled atmospheric forcings. For the current function, I think that would be okay: I think what's important is that pbot and pco2 are either both downscaled or neither downscaled (currently pco2 isn't downscaled at all, but that needs to change: see #1332 ); either way I think you'll get the same result (within roundoff), because the downscaling should be done consistently for both pbot and pco2 (once #1332 is done). In fact, you might consider adding co2_ppmv to atm2lnd_inst and filling it in in lnd_import (in place of the current local variable in that subroutine) rather than recomputing the previously-held-but-discarded co2_ppmv value (and this would side-step the issue of whether to use the downscaled or non-downscaled version). In this case, I think you could remove leafcn from the restart file, because it should be set each time step before it is referenced.

I feel like either approach would be reasonable. (2) might be best if you feel pretty confident that this new function will only depend on a limited set of variables, whereas (1) would give more flexibility at the cost of needing to add another variable to the restart file.

Let me know if you think I'm missing something here, though.

wwieder:
Thanks for this feedback @billsacks.   Before spending any time on this @slevisconsulting I think we need to decide if this is a feature we want to include in the model, or if it's a one-off for the project that Emma's doing.  We'll be discussing the science on this PR at the CLM meeting next week.

billsacks:
Note that https://github.com/ESCOMP/CTSM/pull/1705 (which I plan to merge soon) will result in more conflicts with this PR.

wwieder:
To make this even more complicated, I used mct for the historical simulation (and source mods to set cn_slope = 20).
@emhauser and I have also been discussing putting an upper limit on the changes in `leafcn_patch` that we're calculation, but haven't finalized what we want to do here.

slevisconsulting:
> Sam, I'm not sure we need single point runs here

I canceled those.

wwieder:
One comment on the current approach we're taking.

Under the SSP3-70 scenario co2 gets to ~1000 ppm by 2100.  with the current "additive" parameterization all pfts will have their targetCN increase by 23.  

This would be a nearly doubling of foliar C:N for a PFT that's parameterized with a target C:N of 25 (deciduous forests?, solid), but only a 40% increase for a PFT with a target C:N of 58 (evergreen boreal forests, dashed), black lines below. 

This makes me wonder if we should use a different function that scales the % change in C:N?  For example: `cnMulti = 1+log(CO2now/CO2base`)  This gives just over a doubling of leafCN, shown in the red lines below.  We can figure out ways to modify this formula, but it gives pretty different results for PFTs that start with high C:N values.   

image


From what you've seen in the data, is there any way to tell which approach may be more correct Emma?

emhauser:
The model data from the global runs through 2015 are like the "add" lines in black above: 
Screen Shot 2022-06-06 at 4 55 37 PM

My sense from reading is that % change in foliar C:N (and the multiplicative model) is more likely than additive change although the % change varies somewhat by ecosystem. This plot from Du et al (2019) reports % change in C:N between ambient and elevated CO2 scenarios for several different types of ecosystem :
Screen Shot 2022-06-06 at 5 08 37 PM

Based on this it doesn't seem like the % change in CN is all that different among systems although wetlands are a little lower and grasslands/croplands a little higher (although error bars overlap in all cases). Based on this it seems like we might want to have a comparable % change across systems rather than an absolute value.  I'm still looking around for other helpful datasets but maybe this could be a starting place?

slevis-lmwg:
> @slevis-lmwg is this still relevant? And is it something to get into the release or for afterwards (the ctsm7.0.0 milestone)?

Question for @wwieder, though I would guess that we will close this as "nice to have but won't merge" since Emma -- I'm pretty sure -- published the relevant paper a little while ago.

wwieder:
I agree. Emma published her paper and there was significant reservation from the LMWG about the empirical function we used to generate time evolving Leaf_CN ratios.  On one hand, I'd suggest we just close this PR with a won't fix.  

At the same time, it seems like there was enough work that went into this PR that could serve as an example if others wanted to explore the impact of empirically derived parameterizations for plant functional response to environmental perturbations.  If we merge the PR, but set the new parameters controlling the CO2-LeafCN function to zero, it offers the opportunity for others to investigate the model sensitivity further (e.g. using CLM as a hypothesis testing tool).  Since Emma is currently working on a proposal that potentially build on this work, and that integrates with MIMICS), I'm more inclined to aim for merging this PR to b4b-dev and have it available in CLM6.  

Anyone else have thoughts here?

slevis-lmwg:
I'm completely open to your alternate suggestion @wwieder

dlawrenncar:
That's sound good to me.  Maintaining the flexibility would be useful, I
think.

ekluzek:
Move to beyond ctsm6.0.0 as I recall that being the case.

src/biogeochem/CNVegNitrogenStateType.F90
```diff
@@ -1057,6 +1081,42 @@ subroutine Summary_nitrogenstate(this, bounds, num_allc, filter_allc, &
...
+    class(cnveg_nitrogenstate_type) :: this
+    type(bounds_type) , intent(in)  :: bounds
+    type(atm2lnd_type), intent(in)  :: atm2lnd_inst
+    !
+    ! !LOCAL VARIABLES:
+    integer :: g, c, p  ! indices
+    real(r8) :: co2_ppmv  ! atm co2 concentration
+    ! Could use 355 value from clm_varctl.F90 to avoid this hardwiring?
+    real(r8), parameter :: co2_base = 310._r8  ! units ppmv
+    real(r8), parameter :: cn_slope = 0._r8
```
slevis-lmwg, line 1103: These will be added to the paramfile in #2365 
so make changes accordingly to finish this PR and merge to b4b-dev.

See this comment
https://github.com/ESCOMP/CTSM/pull/2365#issuecomment-2691527404

ekluzek:
@wwieder I agree and like your suggestion here...

> At the same time, it seems like there was enough work that went into this PR that could serve as an example if others wanted to explore the impact of empirically derived parameterizations for plant functional response to environmental perturbations. If we merge the PR, but set the new parameters controlling the CO2-LeafCN function to zero, it offers the opportunity for others to investigate the model sensitivity further (e.g. using CLM as a hypothesis testing tool). Since Emma is currently working on a proposal that potentially build on this work, and that integrates with MIMICS), I'm more inclined to aim for merging this PR to b4b-dev and have it available in CLM6.

Since, the main point of CTSM is a scientific tool to do experiments in. And because this will enable continued work in this space. I agree with the suggestion. This is essentially bringing some refactoring into the code to enable a particular line of experiments. Whenever we can enable that kind of flexibility for reasonable experiments that doesn't add excess complexity, I think that's a good thing to do. There can be times when the flexibility added is going to be too complex, but that's not the case here. leafcn is still an input in the expected places, it just can be time-evolving rather than from the parameter file.

The only two suggestions I have for doing that are:
- [ ] Add more comments that say what's going on, that leafcn can be time-evolving, but the default is fixed by the parameter file
- [ ] Add a test that exercises the code for a simple time-evolving case just to show that it continues to work

I'll more specifics to a review of the code.

wwieder:
@ekluzek we have a plan to get this PR in sooner rather than later, although it's not @slevis-lmwg's top priority. 

We think this can be accomplished in the upcoming sprint (unless CMIP7 data come in this month).  Can review next week.

src/biogeochem/CNVegNitrogenStateType.F90
slevis-lmwg: - [ ] Updating to ctsm5.3.030 raised numerous conflicts. After resolving them, I will run aux_clm for sanity before I make further changes.
- [ ] Start reading the params _leafcn_co2_base_ and _leafcn_co2_slope_ from the paramfiles introduced in #2365.
- [ ] Look at code reviews: Bill's and Erik's

src/biogeochem/CNAllocationMod.F90
```diff
@@ -503,7 +505,7 @@ subroutine calc_allometry(num_soilp, filter_soilp, &
        else
           g1 = grperc(ivt(p))
        end if
-       cnl  = leafcn(ivt(p))
+       cnl  = leafcn_t_evolving(p)
```
slevis-lmwg, line 508: In the original (2022) implementation, either I missed this change or that version of the code didn't need it.

slevis-lmwg: Looking back before my 2025 updates at ctsm5.1.dev074-9-g0fe849e20, I have confirmed that we did not have a separate CNAllocationMod file, so I do not think that I missed anything in the original implementation.

src/biogeochem/CNVegComputeSeedMod.F90
```diff
@@ -118,6 +121,7 @@ subroutine ComputeSeedAmounts(bounds, &
           my_deadstem_seed = 0._r8
 
           pft_type = patch%itype(p)
+          leafcn = leafcn_t_evolving(p)
```
slevis-lmwg, line 124: from meeting with @ekluzek 
the local variable is degrading efficiency, though we checked and found that this happens once per year in transient simulations, so it's not a huge deal in this case.

slevis-lmwg: If I decided to revert this change, then it may make sense to remove all (or most, e.g. not pft_type) the local variables.

src/biogeochem/CNVegNitrogenStateType.F90
```diff
@@ -17,19 +17,27 @@ module CNVegNitrogenStateType
...
+  type :: cnveg_nstate_params_type
+     real(r8) :: leafcn_co2_base  ! ppm CO2 for target leaf CN scaling
+     real(r8) :: leafcn_co2_slope  ! CO2 effects on target leaf CN scaling (unitless)
+  end type cnveg_nstate_params_type
   !
+  type(cnveg_nstate_params_type), public, protected :: params_inst  ! params_inst is populated in readParams
+
   type, public :: cnveg_nitrogenstate_type
 
+     real(r8), pointer :: leafcn_patch(:) ! time evolving leafcn calculated in subroutine time_evolv_leafcn from co2_ppmv and three paramfile parameters; of the three parameters, leafcn_co2_slope is set to 0 (as of ctsm5.4) so leafcn_patch defaults to the values of paramfile parameter leafcn
```
slevis-lmwg, line 40: from meeting with @ekluzek 
```suggestion
     real(r8), pointer :: leafcn_patch(:)  ! (gC/gN) time evolving leafcn
```

src/biogeochem/CNVegNitrogenStateType.F90
```diff
@@ -1815,11 +1833,20 @@ subroutine Restart ( this,  bounds, ncid, flag, leafc_patch, &
...
              else
-                this%leafn_patch(p)                           = leafc_patch(p)         / pftcon%leafcn(patch%itype(p))
-                this%leafn_storage_patch(p)                   = leafc_storage_patch(p) / pftcon%leafcn(patch%itype(p))
+                ! max in next line ensures leafcn_patch equals at least the
+                ! leafcn for bare ground (noveg) to avoid div by 0 when the
+                ! initial file doesn't include leafcn_patch. In that situation
+                ! it seems, however, like a better initial condition to set
+                ! this%leafcn_patch(p) = pftcon%leafcn(patch%itype(p))
+                ! ...but leafcn_patch does get updated before anything else
+                ! in clm_driver, so either option may give the same answer.
```
slevis-lmwg, line 1842: This may be unnecessary because I eliminated leafcn_patch from restart files.

So this backwards compatibility note and line 1843 are unnecessary.

src/main/clm_driver.F90
```diff
@@ -169,6 +169,17 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
     call get_proc_bounds(bounds_proc)
     nclumps = get_proc_clumps()
 
+    ! -------------------------------------------------
+    ! Obtain updated values of time-evolving parameters
+    ! As of ctsm5.4
+    ! - We have this capability for leafcn by calculating this parameter in subroutine time_evolv_leafcn
+    ! - We default the time evolving leafcn calculation to leafcn's paramfile values
```
slevis-lmwg, line 176: Add comment that this needs to be called before Dynpatch...

src/biogeochem/CNVegNitrogenStateType.F90
```diff
@@ -2276,6 +2304,77 @@ subroutine Summary_nitrogenstate(this, bounds, num_soilc, filter_soilc, num_soil
...
+    type(bounds_type) , intent(in)  :: bounds
+    type(atm2lnd_type), intent(in)  :: atm2lnd_inst
+    !
+    ! !LOCAL VARIABLES:
+    integer :: g, l, c, p  ! indices
+    real(r8) :: co2_ppmv  ! atm co2 concentration
+    !-----------------------------------------------------------------------
+
+    if (is_first_step()) then
+       allocate(this%leafcn_patch(bounds%begp:bounds%endp)); this%leafcn_patch(:) = pftcon%leafcn(patch%itype(:))
```
slevis-lmwg, line 2361: Replace the allocate with an endrun as a test to confirm whether I really need to allocate here, because I shouldn't have to.

src/biogeochem/CNVegComputeSeedMod.F90
slevis-lmwg: ...but that's a refactor outside of this PR's scope

slevis-lmwg: so i'm resolving

src/main/clm_driver.F90
```diff
@@ -165,6 +165,12 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
     call get_proc_bounds(bounds_proc)
     nclumps = get_proc_clumps()
 
+    ! -------------------------------------------------
+    ! Obtain updated values of time-evolving parameters
+    ! -------------------------------------------------
+    call bgc_vegetation_inst%cnveg_nitrogenstate_inst%time_evolv_leafcn( &
```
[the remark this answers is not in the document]
slevis-lmwg: Testing with 
```
PASS ERS_D_Ld7_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropRs.derecho_intel.clm-decStart1851_noinitial
PASS ERS_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.derecho_intel.clm-default
PASS ERP_P64x2_D_Ld5.f10_f10_mg37.I2000Ctsm50NwpBgcCropGswp.derecho_intel.clm-default
```
against the ctsm5.3.033 baseline.

src/biogeochem/CNVegNitrogenStateType.F90
```diff
@@ -1057,6 +1081,42 @@ subroutine Summary_nitrogenstate(this, bounds, num_allc, filter_allc, &
...
+    real(r8) :: co2_ppmv  ! atm co2 concentration
+    ! Could use 355 value from clm_varctl.F90 to avoid this hardwiring?
+    real(r8), parameter :: co2_base = 310._r8  ! units ppmv
+    real(r8), parameter :: cn_slope = 0._r8
+
+    character(len=*), parameter :: subname = 'time_evolv_leafcn'
+    !-----------------------------------------------------------------------
+
+    ! Loop to get leafcn_col
+    do p = bounds%begp, bounds%endp
```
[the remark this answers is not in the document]
slevis-lmwg: Testing with
`ERP_P64x2_D_Ld5.f10_f10_mg37.I2000Ctsm50NwpBgcCropGswp.derecho_intel.clm-default`
against the ctsm5.3.033 baseline.

slevis-lmwg:
From Bill's code review
- [ ] Open new issue for better handling of co2_ppmv?
- [ ] Open new issue about this inconsequential test failure: https://github.com/ESCOMP/CTSM/pull/1654#discussion_r2017731179

slevis-lmwg:
@billsacks 
merging this PR requires your approval since you requested changes. I think I have addressed all suggestions.

billsacks (approved):
To be honest, I don't remember this and don't have time to look at it right now, but I'm marking it as approved to clear the block on merging. Thanks for your work on this @slevis-lmwg !

Issue #1572, opened 2021-12-08
Improve mechanism for determining where to allocate memory for urban when running with dynamic urban


billsacks:
(Moving this from some discussion in #1546 - see https://github.com/ESCOMP/CTSM/pull/1546#pullrequestreview-809461860 )

In the soon-to-be-merged PR to enable dynamic urban (#1546 ), the temporary mechanism for allocating memory for urban is to set run_zero_weight_urban to .true. That seems reasonable for development, but I think what we want long-term is something similar to lake; this is less of a performance and memory hit than the current `run_zero_weight_urban` mechanism (I'm not sure if you want to keep `run_zero_weight_urban` in addition to this mechanism). (although @olyson showed in https://github.com/ESCOMP/CTSM/issues/1445#issuecomment-936773113 that the performance hit isn't too great for the current mechanism). Here's what I think should be done:
- Add a hasurban variable, similar to haslake (do we want a separate one for each urban landunit, or just one overall for urban – so if there is ever any urban in a given grid cell in a transient run, we'll allocate memory for all urban landunits in that grid cell?)
- Add a subroutine urban_landunit_exists, similar to the lake_landunit_exists function added in ctsm5.1.dev003
- I think we can (and should) change the urban logic in is_active_l to be similar to lakes: if memory is allocated for an urban column, then make it active, period. I think what we want is to have some logic that determines whether a given urban landunit should exist in memory, and then as long as the landunit exists in memory, it should be active so that it is spunup. (I think that's what we want, but I'm open to discussion here.)
- Add code like surfrd_lakemask for urban in surfrdMod

@olyson replied to my original comment:

> We'd like to do a separate PR for the hasurban implementation, so if you could open a separate issue for that, that would be great. We also think it would be beneficial to have separate hasurban for each urban landunit (likely a 3d variable instead of 3 separate variables). For example, the tall building district class is generally sparse and is likely to generally remain that way in the future, so we could probably save some memory by considering it separately. We plan on keeping the run_zero_weight_urban option for now.

@olyson @fang-bowen @Face2sea @keerzhang1

This is connected with #1445

olyson:
Thanks for opening this PR.
A couple of questions: 
1) It looks like HASLAKE is a double precision variable on the surface dataset that is used to test dynamic lake.  So I've put a double precision HASURBAN on the surface dataset that is used to test dynamic urban.  However, @keerzhang1 points out that we could also use an integer for this variable.  Is there a reason that this has to be a double?
[...]

PR #1661, opened 2022-02-25
Improve mechanism for determining where to allocate memory for urban when running with dynamic urban

fang-bowen:
### Description of changes

Implement a 'pct_urban_max' mechanism that reads 'PCT_URBAN_MAX' (maximum urban percentage throughout timeseries) from landuse.timeseries and initialize urban landunits only where urban ever exists in the transient run. 

### Specific notes

As discussed in #1572, in the previous dynamic urban settings, the mechanism for allocating memory for urban is to set run_zero_weight_urban to .true., which causes performance and memory issues.

This PR includes the following changes to address the issue:

1. Add a subroutine surfrd_urbanmask to read 'PCT_URBAN_MAX' from landuse.timeseries, which indicates the maximum percentage of urban in the timeseries.

2. Add a 'pct_urban_max' variable and a function 'urban_landunit_exists'. If urban exists or will grow in a grid cell, memory is allocated.

3. Modify the urban logic in is_active_l. If memory is allocated for an urban column, then make it active.

@olyson. Also referenced #1109 ('haslake').

CTSM Issues Fixed (include github issue #): #1572 

No. For dynamic urban runs, user can leave 'run_zero_weight_urban' as .false. and supply landuse.timeseries file which includes PCT_URBAN_MAX (3 urban density types) and the model will only initialize urban landunits where urban exists or will grow.

Testing performed, if any:
Please see https://github.com/ESCOMP/CTSM/pull/1661#issuecomment-1051070789 (Modified dynamic urban test; aux_clm testing).


olyson:
Thanks Bowen!
A couple of other things; 
1) the dynamic urban test (ERS_Lm25.1x1_smallvilleIA.IHistClm50BgcCropQianRs.cheyenne_gnu.clm-smallville_dynurban_monthly) has been modified so that it uses PCT_URBAN_MAX (instead of HASURBAN).  It also now sets one of the urban landunits to zero percent throughout the timeseries in order to test that the landunit is indeed not active throughout the simulation.  The landuse timeseries file (/glade/p/cesm/cseg/inputdata/lnd/clm2/surfdata_map/landuse.timeseries_1x1_smallvilleIA_hist_78pfts_simyr1850-1855_dynUrban_c211206.nc) has been updated (/glade/p/cgd/tss/people/oleson/modify_surfdata/landuse.timeseries_1x1_smallvilleIA_hist_78pfts_simyr1850-1855_dynUrban_c220223.nc) and needs to be moved to inputdata.
2) ran aux_clm testing, all pass and all bfb (expected as do_transient_urban is .false. by default).

src/main/subgridWeightsMod.F90
```diff
@@ -334,9 +336,17 @@ logical function is_active_l(l, glc_behavior)
...
+       ! through the run_zero_weight_urban setting which runs all urban landunits in each grid cell or
+       ! through pct_urban_max which is the maximum percent urban for each density type in a transient run.
+       ! By doing this, urban land units are also run virtually in grid cells which will grow
+       ! urban during the transient run.
+
+       if ( lun%itype(l) >= isturb_MIN .and. lun%itype(l) <= isturb_MAX ) then 
+          dens_index = lun%itype(l) - isturb_MIN + 1
+          if (run_zero_weight_urban .or. pct_urban_max(g,dens_index) > 0._r8) then
+             is_active_l = .true.
+          end if
```
billsacks, line 349: I'm not positive, but I think this could safely be changed to just return `.true.` for all urban landunits, similarly to what is done for lakes: this would say, if we have bothered to allocate memory for an urban landunit for any reason, then we want to make it active. I'm pretty sure that's what is intended here. However, the current code seems fine, too, so I'll leave it up to you whether you want to change it. The main advantages of changing it would be (1) consistency with lake, and (2) future-proofing in case the logic in subgridMod for when to allocate urban changes (currently the logic here needs to be kept consistent with the logic there).

billsacks (approved):
This looks great - thanks a lot @fang-bowen and @olyson !

I pushed a small commit where I point to a new location for the landuse timeseries file in the smallville test: I have moved this to the inputdata repository.

See my one inline comment below, but I'll leave it up to you whether you want to change it or if you prefer to keep this more explicit code in subgridWeightsMod even if it means a bit of duplication.

Other than that, this is ready for final testing and integration.

src/main/subgridWeightsMod.F90
billsacks: Given the current logic, I'm wondering why such an urban landunit would be allocated in the first place: if it has zero-weight throughout the time series, shouldn't the logic in `urban_landunit_exists` lead it to not getting allocated?

olyson: Hmm, no, I think you are right.  In both cases an ncdump of the h1 file gives:

data:
 land1d_wtgcell = 1, 0, 0, 0 ;
 land1d_ityplunit = 1, 2, 7, 8 ;
 land1d_active = 1, _, 1, 1 ;

which is correct since the landuse.timeseries has zero percent urban for MD (land1d_ityplunit=9).

I agree that it would be good to be consistent, so I'll make the change you suggested.

billsacks: Sounds good, thanks for checking.

olyson:
Testing on cheyenne and izumi have passed as expected.
Two expected failures on izumi:

    FAIL SMS_D.f10_f10_mg37.I2000Clm51BgcCrop.izumi_pgi.clm-crop MODEL_BUILD time=1 (EXPECTED FAILURE)
    FAIL SMS.f10_f10_mg37.I2000Clm50BgcCrop.izumi_pgi.clm-crop MODEL_BUILD time=1 (EXPECTED FAILURE)

Two expected failures on cheyenne due to changing the dynamic urban test:

    FAIL ERS_Lm25.1x1_smallvilleIA.IHistClm50BgcCropQianRs.cheyenne_gnu.clm-smallville_dynurban_monthly NLCOMP
    FAIL ERS_Lm25.1x1_smallvilleIA.IHistClm50BgcCropQianRs.cheyenne_gnu.clm-smallville_dynurban_monthly BASELINE ctsm5.1.dev082: DIFF

I've updated and committed the ChangeLog (please review).

billsacks:
Looks great, thanks a lot! I'm about to merge this. (I just made a couple minor tweaks to the ChangeLog and ChangeSum; one is that I have started putting "multiple" in the ChangeSum when I'm bringing in a tag that had substantial contributions from multiple people, as this one seems to.)

olyson:
Thanks for the tweaks, good catches.

PR #1663, opened 2022-02-28
New parallel surface dataset generation with online regridding

### LINKED ISSUE #1664: Remove override options from mksurfdata process
mksurfdata previously as part of PTCLM included some override options for creating single point tower sites. We've moved our process over from being inside of mksurfdata to outside tools that do this functionality in subset_data and modify_fsurdat.

So the following options in mksurfdata.pl can be removed (as well as the corresponding changes in the mksurfdata_map code). It looks like they were already removed in toolchain/gen_mksurf_namelist, since we realized they didn't need to be moved to the new toolchain process.

> OPTIONS to override the mapping of the input gridded data with hardcoded input
> 
>      -pft_frc "list of fractions"  Comma delimited list of percentages for veg types
>      -pft_idx "list of veg index"  Comma delimited veg index for each fraction
>      -soil_cly "% of clay"         % of soil that is clay
>      -soil_col "soil color"        Soil color (1 [light] to 20 [dark])
>      -soil_fmx "soil fmax"         Soil maximum saturated fraction (0-1)
>      -soil_snd "% of sand"         % of soil that is sand

mvertens:
### Description of changes

Implement a parallel version of mksurfdata (mksurfdata_esmf) that uses ESMF regridding directly in mksurfdata so that offline mapping files don't have to be created as a separate step.

### Specific notes

Contributors other than yourself, if any: @slevisconsulting 

  Fixes #1664
CTSM issues added as a result of thiis: #1719 #1718 #1717

Any User Interface Changes (namelist or namelist defaults changes)? Behavior of gen_mksurf_namelist changes

Testing performed, if any: @slevisconsulting ran the mksurfdata_esmf makefile to make sure all resolutions are created
@slevisconsulting and @mvertens also compared to the previous version for one-degree to make sure the changes make sense.


ekluzek:
This should be rebased to come in on the ctsm5.2.mksurfdata branch. And we plan to address #1665 and #1664 in this PR as well.

ekluzek:
@mvertens and @slevisconsulting another case I want to have a way to keep working is the 5x5_amazon case. Its original purpose was to test a small regional grid. But, it's also become an important part of the mizuRoute testing, so I want the same grid. 

Based on the problems you saw with 10x15, I suspect the 5x5_amazon will fail in mksurfdata_esmf as well. So I'll either have to create a larger regional grid that can then be subset from -- or we have to do something like not run the 1km elevation grid, and just assign a constant to STD_ELEV over the region. The later would probably be fine.

Possibly that would be the way to get this to work for small grids -- to enter a constant value to use for STD_ELEV so that you don't have to mess with the 1km elevation file? Currently for 10x15 you enter a lower resolution dataset, and I suppose I could enter a dataset with a single point as an alternative. It's just more effort to do that rather than enter a single value.

wwieder:
@ekluzek is this also an example of when we can use the regional feature in subset_data?

ekluzek:
@wwieder for the 5x5_amazon grid? Yes, but the issue is that the 5x5_amazon grid isn't subset from one of the global grids, it was its own defined regional area. I could use subset_data to create the 5x5_amazon grid -- but I'll have to create a larger global or regional grid that mksurfdata_esmf can run on and that I can subset the 5x5 amazon area from. That's the first part of the second sentence in the middle paragraph above that I talk about that. I'm thinking it would be easier to assign a constant value though. Technically there is a huge variation in the current file (goes from 10 to 1430, with avg of 183, and median of 50. Since, it's a test dataset though and the STD_ELEV only applies to snow, it's probably OK to use a constant value.

slevisconsulting:
> @ekluzek is this also an example of when we can use the regional feature in subset_data?

@ekluzek I think I understand your response to @wwieder and particularly that STD_ELEV can be a constant in the 5x5_amazon fsurdat file. Also that setting this to a constant may be a cleaner solution for the 10x15 grid, as well. (Rather than pointing to a different raw file just for 10x15.)

So then, here is my preference for the simplified workflow that we're hoping to achieve:
- New mksurfdata tool generates **global** fsurdat files
- subset_data generates **regional/1x1**

As such, I am not in favor of adding a regional case to the new mksurfdata tool. Generating 5x5_amazon from a global fsurdat file with subset_data should be straightforward, I think.

ekluzek:
We do need the new mksurfdata_esmf to work with WRF regional grids though. So regional grids do need to work. It would be good to understand the limitations though do we just need the total number of grid points greater than 3000? Does the physical extent of the grid matter? I think most WRF grids are going to be bigger than that. But, we might need an alternative process for small WRF regional grids, which is where the setting of STD_ELEV to a single value might be a nice workaround. The 5x5_amazon and 10x15 are truly test cases, so it's OK to have less ideal solutions. But, we do want the WRF regional grids to work robustly. There are also people who do regional cases that are just subset from the global grids, and they are of course what subset_data is designed for. But, we can't do that for WRF regional cases.

It does sound like we should have the three of us to discuss this a bit though. Is today possible? Or should we schedule something in a couple weeks?

mvertens:
@ekluzek - Regional grids should work fine. And in fact based on a conversation with Bob Oehmke this afternoon - single points should not have the performance hit I was expecting. I plan to verify this. It seems to be very low resolution global grids that are currently the issue.

ekluzek:
@mvertens and @slevisconsulting OK, perfect, that's really great to hear. I'll wait to hear back what you find then. Sounds like this won't be a problem.

@negin513 by the way do you have a sample WRF grid that they could do some testing with for the new mksurfdata_esmf so we can know that WRF grids will work? Thanks for whatever you can find.

mvertens:
@erik - I was able to totally get around the performance problem of the 10x15 my reading in an already calculated STD_ELEV and SLOPE. Everything else was very fast.

mvertens:
@ekluzek @slevisconsulting - I have verified that I can successfully construct the wrf surface dataset that duplicates the surface dataset that @olyson and @negin513 were using: `/glade/work/slevis/git_wrf/ctsm_surf/surfdata_conus_hist_16pfts_Irrig_CMIP6_simyr2000_c210119.nc`
The timing was under 3 minutes on 64 nodes. I am using the input mesh that Keith created - /glade/p/cgd/tss/people/oleson/CLM5_urban/wrf2ctsm_land_conus_ESMFMesh_c20201110.nc.

ekluzek:
@mvertens so you use the STD_ELEV and SLOPE from the previous 10x15 files right? In general that seems like a workable solution especially right ow.

The long term thing I wonder about though is if we update the elevation dataset, or have a new low resolution grid that we need to run -- how would handle that? Technically we could pull up the current mksurfdata_map to do it, but that might be challenging in the distant future. I wonder if we could throw together a simple offline python script to create the low resolution STD_ELEV/SLOPE? But, we don't need it right now, so may not be worth it. The other way would be if mksurfdata_esmf could be used for the 10x15 STD_ELEV/SLOPE even if it's slow -- but it still creates it. As long as it does create it within the queue wallclock limit -- that would be fine. Have you tried just letting the 10x15 case run to completion even if it's terribly slow?

mvertens:
@erik - @slevis was about to get that 10x15 completed but it took 2 hours. We are submitting a whole set of resolutions in one queue submission - and all the other resolutions took on the order of 5 minutes or so. So it's really not feasible to do this as part of the job submission. At this point I think the our current short term solution is the best. Bob Oehmke will also be looking at this once the upcoming ESMF release is done.

ekluzek:
OK, @mvertens and I went over this a bit. She had missed the point I was asking about, but we were able to clarify with direct communication. The concern I have with the 10x15 grid is that we will need to be able to update the 10x15 STD_ELEV and SLOPE when a new dataset comes in, or we have a new low resolution dataset. I wanted to make sure there was a way to do that even if it was slow. That's a long term need we'll have. It's also possible that Bob will come up with a solution before long anyway. For normal operation we'll use the existing 10x15 data, it'll be a rare event that we'll need to update it.

tools/mksurfdata_esmf/README
```diff
@@ -0,0 +1,100 @@
...
+NOTES
+================
+
+------------------------------------------------------------
+slevis HAS RUN THESE CASES and HAS LISTED ISSUES ENCOUNTERED
+------------------------------------------------------------
+1) 1x1_brazil good except:
+ time = -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 0 ;
+ instead of
+ time = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 ;
```
slevisconsulting, line 67: TODO Need to write out time as a variable

slevisconsulting, line 67: TODO Need to write out time as a variable

tools/mksurfdata_esmf/README
```diff
@@ -0,0 +1,100 @@
...
+- LAI in .log comes out shifted by one
+- Soil color, ag fire peak month outputs too high in .log
+- Soil txt, peat, fmax, lakes, harvest too low in .log unless raw data changed
+- Slope, stddev, soildepth, urban, glaciers not appearing in .log
+- VOC em factors from input not appearing in .log
+
+3) --merge_gis, --nocrop work
+4) --hires_pft works, but must abort for years /= 2005 (and 1850?)
+5) --potveg_flag needs file(s) converted to cdf5
+6) --fast doesn't do anything as far as I can tell; do we need it?
```
slevisconsulting, line 82: TODO Need to remove --fast option

This list in the README includes more TODOs.

slevisconsulting, line 82: TODO Need to remove --fast option

This list in the README includes more TODOs.

slevisconsulting:
TODO Need to update `gen_mksurfdata_namelist.py` where it writes out for example `landuse_timeseries_hist_78pfts_simyr1850-1899.txt` to now include a new line for transient urban and a new line for transient lake.

tools/mksurfdata_esmf/README
ekluzek: @slevisconsulting what the fast option did before was to set STD_ELEV and SLOPE to fixed values so that you didn't have to process the 1km topo dataset. This didn't matter much for mksurfdata_map, but it did have a huge impact on the creating of the mapping files, and if you didn't create a 1km mapping file -- then you had to set STD_ELEV and SLOPE to constants. So if it was triggered in the first it had to be triggered the same in mksurfdata_map.

We are adding special handling of the 1km dataset which is different now, so you either need a flag to trigger it (or as @mvertens suggests) it looks for file availability and uses other datasets if they are available.

ekluzek:
@mvertens and I talked yesterday about this a bit. A new idea she is looking into is to speed everything up and require less resources by using a preprocessed 1km dataset for each standard resolution. For new resolutions you always need to process it, but for standard resolutions you wouldn't need to replicate it each time (unless the topo dataset changes). The best impact of this is to relax memory requirements (currently we are memory limited and can't use a full node of processors on cheyenne).

ekluzek:
We had our CTSM software meeting, so I brought this idea up (just above). The one thing we can add here is that we continue to think that we won't update surface datasets very often. It's expensive to do that in terms of disk space and storage on inputdata. So unless there are critical bugs to fix we likely won't update surface datasets except when we do a larger overhaul of the surface datasets (like the CTSM5.2 work which is a significant upgrade in CTSM and surface datasets). Our last update of surface datasets was 3 years ago. Before that was 2 years previous (which did include a few iterations to get the final files).

slevisconsulting:
Regarding the limited years that transient lake and urban span, @billsacks mentioned a suggestion that may have been @ekluzek's from a while ago:
Reuse the first and last years for all years before and after. E.g. for yr 1850 use 1850 lake and 2000 urban and for yr 2100 use 2017 lake and 2100 urban. Currently this PR's code triggers a file-not-found error that would need to be changed accordingly.

Before that though, I have a more immediate problem getting errors (different ones) running the new and the old mksurfdata codes with transient lake and transient urban. I had hoped to start from a working old mksurfdata, so that I could then replicate the behavior with the new. @olyson have you run the old mksurfdata with transient lake and urban successfully? If so, could we set up a time to discuss?

olyson:
Sam, see my email to you.

slevis:
Please remove me from this thread, I was accidentally cc'd

ekluzek:
Sorry @slevis I corrected the mention to @slevisconsulting. I'm not sure how to remove you from the thread other than have you press the "unsubscribe" button?

slevisconsulting:
Huh, interesting. I thought I was the only slevis! This shattered my world view :-))

olyson:
I have a new location for the transient urban data:

/glade/p/cgd/tss/people/oleson/Dynamic_Urban_Data/BNU_NoAdjust

We've copied year 2000 to generate the 1850-1999 files.

I just generated landuse timeseries file and surface dataset for transient urban the other day using the older code here:

/glade/work/oleson/ctsm_dynurb_mksurf_ctsm5.2.mksurfdata_I1564/tools/mksurfdata_map

The relevant files are:

landuse_timeseries_SSP5-8.5_78pfts_CMIP6_simyr1850-2100.txt
surfdata_0.9x1.25_SSP5-8.5_78pfts_CMIP6_simyr1850_c220323.namelist
[surfdata_0.9x1.25_SSP5-8.5_78pfts_CMIP6_simyr1850_c220323.nc](http://surfdata_0.9x1.25_ssp5-8.5_78pfts_cmip6_simyr1850_c220323.nc/)
surfdata_0.9x1.25_SSP5-8.5_78pfts_CMIP6_simyr1850_c220323.log
[landuse.timeseries_0.9x1.25_SSP5-8.5_78pfts_CMIP6_simyr1850-2100_c220323.nc](http://landuse.timeseries_0.9x1.25_ssp5-8.5_78pfts_cmip6_simyr1850-2100_c220323.nc/)

And I successfully ran 2015-2100 with this case:

/glade/work/oleson/ctsm_dynurb_mksurf_ctsm5.2.mksurfdata_I1564/cime/scripts/ctsm51_dynurbI1564_1deg_CPLHIST_SSP585

I'm happy to meet anytime Monday afternoon.

slevisconsulting:
Thank you, @olyson 

I got the new mksurfdata to work now, so I will not attempt to resolve my issues with the old mksurfdata. For the new mksurfdata to work with the transient lake and urban data I did the following:
- code modification to skip over lakedepth when reading transient lake data
- `nccopy -k cdf5 oldfile newfile` to get lake and urban files to cdf5. I placed the new files in
`/glade/p/cgd/tss/people/slevis/transient_lake_data`
`/glade/p/cgd/tss/people/slevis/transient_urban_data`

I will push changes shortly.

tools/mksurfdata_esmf/src/mksurfdata.F90
```diff
@@ -0,0 +1,1273 @@
...
+  if (fsurdat /= ' ') then
+     rcode = pio_inq_varid(pioid, 'natpft', pio_varid)
+     rcode = pio_put_var(pioid, pio_varid, (/(n,n=natpft_lb,natpft_ub)/))
+     if (num_cft > 0) then
+        rcode = pio_inq_varid(pioid, 'cft', pio_varid)
+        rcode = pio_put_var(pioid, pio_varid, (/(n,n=cft_lb,cft_ub)/))
+     end if
+     ! time is months for LAI, SAI, and pft heights
+     rcode = pio_inq_varid(pioid, 'time', pio_varid)
+     rcode = pio_put_var(pioid, pio_varid, (/(n,n=n_jan,n_dec)/))
```
slevisconsulting, line 339: Added these two lines to put values of 1 to 12 in the time variable (following cft just above as a template) but all I get is 12 zeros in my fsurdat's time variable.

@mvertens do you possibly see what I'm doing wrong?

slevisconsulting: @jedwards4b provided a solution in #1695 with the explanation that time defined with an unlimited dimension needs to be written one value (ie one month) at a time.

@jedwards4b I brought the mods in manually. Also, just checking: You added a -g in the makefile. That's to compile in debug mode, no?

Thanks again!

jedwards4b: Right the -g was added because the way the build currently works the individual files are compiled with -g
but the link step is not.   I would like to work with you to improve the cmake build so that the second (src directory)
build step is no longer required.

slevisconsulting: > Right the -g was added because the way the build currently works the individual files are compiled with -g but the link step is not.

So keeping the -g then, as well.

> I would like to work with you to improve the cmake build so that the second (src directory) build step is no longer required.

Sounds good. Shall we meet as a first step? Or could we even accomplish that during a meeting?

slevisconsulting: > Right the -g was added because the way the build currently works the individual files are compiled with -g but the link step is not.

So keeping the -g then, as well.

> I would like to work with you to improve the cmake build so that the second (src directory) build step is no longer required.

Sounds good. Shall we meet as a first step? Or could we even accomplish that during a meeting?

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
```diff
@@ -0,0 +1,317 @@
...
+
+<rawdata_files>
+
+  <!-- ======================== -->
+  <!-- lai -->
+  <!-- ======================== -->
+
+  <mksrf_flai>
+    <entry>
+      <data_filename>lnd/clm2/rawdata/pftcftlandusedynharv.0.25x0.25.MODIS.simyr1850-2015.c170412/mksrf_lai_78pfts_simyr2005.cdf5.c220206.nc</data_filename>
```
ekluzek, line 13: - [ ] We need to update this to the new LAI file that @lawrencepj1 created

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
```diff
@@ -0,0 +1,317 @@
...
+  <!-- <mksrf_firrig> -->
+  <!--   <entry> -->
+  <!--     <data_filename>lnd/clm2/rawdata/mksrf_irrig_2160x4320_simyr2000.c110527.nc</data_filename> -->
+  <!--     <mesh_filename>lnd/clm2/mappingdata/grids/UNSTRUCTgrid_5x5min_nomask_c200129.nc</mesh_filename> -->
+  <!--   </entry> -->
+  <!-- </mksrf_firrig> -->
+
+  <mksrf_fsoicol>
+    <entry>
+      <data_filename>lnd/clm2/rawdata/pftcftlandusedynharv.0.25x0.25.MODIS.simyr1850-2015.c170412/mksrf_soilcolor_CMIP6_simyr2005_cd5.c220126.nc</data_filename>
```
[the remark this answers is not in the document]
slevisconsulting: @lawrencepj1 just to confirm, should this be updated to
`pftcftdynharv.0.05x0.05.LUH2.histsimyr2005.c190116/mksrf_soilcolor_histclm52deg005_earthstatmirca_2005.cdf5.c220228.nc`?

mvertens:
@wwieder @slevisconsulting - we need to determine how we want to use the new soil texture dataset in this PR. This involves a new mksoiltexnewMod routine - but we can have things so they are backwards compatible if that is the best way forwards.

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
slevisconsulting: @lawrencepj1 to confirm, should this be updated to
`pftcftdynharv.0.05x0.05.LUH2.histsimyr2005.c190116/mksrf_lai_histclm52deg005_earthstatmirca_2005.cdf5.c220228.nc`
or a different file?

lawrencepj1: Thanks @ekluzek and @slevisconsulting yes this is the correct file for updated monthly lai, sai, and canopy height top and bottom values

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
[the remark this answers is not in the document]
lawrencepj1: Thanks @ekluzek and @slevisconsulting yes this is the correct file for updated soil color values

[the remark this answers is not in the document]
slevisconsulting: ...actually, let me clarify further:
We're already pointing to the new one for `<entry pft_years="2005" hires_pft='on'>` and otherwise we're pointing to the file shown in line 42. Should we now always point to the new file?

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
slevisconsulting: ...same clarification needed for this file as for the soil color file. Thank you, @lawrencepj1 and @ekluzek

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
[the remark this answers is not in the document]
ekluzek: Actually that isn't the correct file, that's the file only for use for the hirespft option. @lawrencepj1 is going to update that one with new LAI and soil color datasets that are all at the 0.05x0.05 resolution. For the standard resolution case there is a 0.25x0.25 resolution dataset that goes with the PFT files.

slevisconsulting:
@mvertens ...regarding your code updates with mksoiltexnewMod.F90:
I'm ready to push some modifications pertaining to .log file diagnostics. Would you rather I waited for you to push your work first?

slevisconsulting:
Submitted job to generate "all" fsurdat files, 66 in total:

- 32 completed successfully
- Ran out of time in 12-hr queue, so submitting to complete the rest
- 1 aborted with an error (unrelated to running out of time), so I will investigate

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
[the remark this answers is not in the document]
ekluzek: @slevisconsulting this is still an issue, but I'm creating an issue for it to make sure we resolve it. Since, we are getting many new datasets we just need to make sure the final datasets are the right ones.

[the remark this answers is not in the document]
ekluzek: @slevisconsulting scratch that -- it's correct.

PR #1677, opened 2022-03-05
Modify mesh mask with new mesh_mask_modifier tool

### LINKED ISSUE #1644: Prospective new tool for user to easily modify ocn/lnd mask in their CESM cases
In this list of [CSSI use-cases](https://docs.google.com/document/d/1wc17F5i_x8MP0XBwqgUAcY_fDbruLaseHVxdIddJMbk), only case 2 can be handled solely by the new fsurdat_modifier tool that allows users to modify existing fsurdat files (#1448). In cases with active land where users want to replace ocean with land, users must modify two more files. E.g. in the 1-degree gx1v7 resolution:

```
mesh_mask = /glade/p/cesmdata/cseg/inputdata/share/meshes/gx1v7_151008_ESMFmesh.nc
mesh_atm = /glade/p/cesmdata/cseg/inputdata/share/meshes/fv0.9x1.25_141008_polemod_ESMFmesh.nc
mesh_lnd = /glade/p/cesmdata/cseg/inputdata/share/meshes/fv0.9x1.25_141008_polemod_ESMFmesh.nc
```

To discuss (@mvertens @billsacks @ekluzek others?):
- Where should this tool reside?
- From discussion with @mvertens I got the sense that it may be non-trivial to modify the mesh-mask file.

slevisconsulting:
### Description of changes

mesh_mask_modifier is a tool that modifies mesh_mask files. It reads the mesh_mask file and outputs a modified copy of the file.

mesh_mask_modifier will complement the fsurdat_modifier tool, which modifies fsurdat files. When a user introduces land in place of ocean grid cells in their fsurdat file, they need to modify the mesh_mask file for their fsurdat modifications to take effect in a simulation.

### Specific notes

@ekluzek 

 Fixes #1644 

N/A ...new tool

Testing performed, if any:
Test run in progress. Typed:
`./mesh_mask_modifier modify_slevis.cfg`
in /glade/work/slevis/git/mksurfdata_toolchain/tools/modify_mesh_mask


slevisconsulting:
@billsacks @ekluzek @adrifoster and I talked at stand-up about the slowness of my code. A few ideas came up:
- fortran can run nested loops like these much faster than python
- xarray/pandas may have non-loop solutions (xarray is built on top of pandas); pick Anderson Banihirwe's brain!
- does it make more sense to use the existing tool that generates mesh files rather than developing a new tool for modifying mesh files...

adrifoster:
Hey all,

I was thinking about this more and it might be easier to think about this outside of xarray. 

I would think (?) there should be a way to do something faster if you just pull out the latitudes and longitudes into np arrays. Or perhaps, as @billsacks suggested as I was chatting with him, 1 big np array where the rows (for example) are the lats/lons from 1 file and the columns are lats/lons from the other file.

Then you may be able to do some whole-array operations to get a matrix of T/F.

I was messing around with an example of this but am having to get back to my regularly scheduled manuscript writing for the week... 

I will think on this more if I have time and report back with any other ideas! But since @slevisconsulting you know much more about the requirements for this function, perhaps these ideas aren't tractable.

slevisconsulting:
Thank you @adrifoster @billsacks for your comments. I expect to return to this around the middle of next week. (Still distracted with the move.)

slevisconsulting:
@ekluzek I'm initially asking for your review and would also be happy if you have additional reviewers in mind. This is a relatively simple tool and development has stopped at this point. It's probably close to ready.

slevisconsulting:
The tool already has user instructions, but I just now decided to also add detailed instructions for a specific use-case, using text that I previously included in Isla Simpson's corresponding google doc.

ekluzek:
@slevisconsulting I have two thoughts here after starting to use it.

The first is:

- Move to a shared directory (modify_inputdatasets)?
Currently this tool is in. it's own directory (as is modify_fsurdat). Personally it seems like normally directories should have more than one thing in them. So a shared directory makes more sense to me.
- Add capability to handle other variable names besides the standard names for surface datasets?
I'm using it to get a mask from a domain file, so I'd like to be able to customize the variable names, which should be straightforward to do.

slevisconsulting:
> * Add capability to handle other variable names besides the standard names for surface datasets?
>   I'm using it to get a mask from a domain file, so I'd like to be able to customize the variable names, which should be straightforward to do.

Follow-up from mtg:
@ekluzek suggests allowing the user to specify alternate variable names in the .cfg file instead of hardwiring default ones in the code (currently landfrac, lat, lon). To clarify, these are the variables that get read from the landmask_file. Also obtain the 1d lats/lons from the 2d rather than reading from the file.

python/ctsm/modify_mesh_mask/modify_mesh_mask.py
```diff
@@ -0,0 +1,100 @@
...
+                # All else in this function supports error checking
+
+                # lon and lat from the landmask file
+                lat_new = float(self.lat_2d[row])
+                lon_new = float(self.lon_2d[col])
+                # ensure lon range of 0-360 rather than -180 to 180
+                lon_new = lon_range_0_to_360(lon_new)
+                # lon and lat from the mesh file
+                lat_mesh = float(self.file['centerCoords'][ncount, 1])
+                lon_mesh = float(self.file['centerCoords'][ncount, 0])
```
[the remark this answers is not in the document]
billsacks: I just looked into this and I think I was wrong: it looks like this should probably work for both structured and unstructured grids (based on looking at a few examples).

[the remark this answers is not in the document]
slevisconsulting: On second look, regular and non-reg grids appear the same in esmf mesh files.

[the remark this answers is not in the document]
slevisconsulting: ...so then the limitation currently is that the user-provided landmask/landfrac file is expected in 2D.

ekluzek:
@slevisconsulting I'm having trouble getting a sample landmask file in the right format. I thought it needed to be a history file, but it looks like the format is slightly different. Do you have a sample file in the correct format?

Also this makes me realize the importance of being able to set the file variable and dimension names in the config file.

slevisconsulting:
> @slevisconsulting I'm having trouble getting a sample landmask file in the right format. I thought it needed to be a history file, but it looks like the format is slightly different. Do you have a sample file in the correct format?
> 
> Also this makes me realize the importance of being able to set the file variable and dimension names in the config file.

@ekluzek this is a sample file that works:
`/glade/work/slevis/git/mksurfdata_toolchain/tools/modify_fsurdat/fill_indian_ocean/fill_indianocean_slevis.nc`

I agree with your suggestion of increasing the flexibility of this tool, by not hardwiring the variable and dimension names.

ekluzek:
@slevisconsulting thanks for the sample file, that really helps. I'm still having trouble with my case, so I'd like to meet with you sometime next week to go over this.

slevisconsulting:
Meeting with @ekluzek 

Some of these notes apply to fsurdat_modifier.

Suggestion to modify the landmask file to include 3 variables:
- landfrac as is in my current file (not required anymore so back to 2 variables in the file)
- landmask same as landfrac but zeros and ones (no fractions) should be read by mesh_modifier
- landmask_diff which is the user's area of change (my current file calls this landmask) read by fsurdat_modifier

Shared optional error check by both tools that compares landmask_diff to quantity calculated on the fly by the tool:
- for error check both tools need to read both variables
- tell it variable and dimension names that you assume are same as in the file you're comparing against
- discourage user from using fsurdat (var PFTDATA_MASK not same as landmask) as the starting point
- domain file also becoming obsolete...
- so clm history from vanilla case is the way to go

slevisconsulting:
@ekluzek as we discussed, I have placed this PR on hold:
Alper's jupyter-based bathymetry tool allows interactive calculation/assignment of bathymetry. In the CSSI group we agreed that Alper would add the option of reading Isla’s modified landmask file to generate a mesh file. His tool will also perform any regridding between the landmask file’s grid and the ocean grid. This will likely cover the needs of the CSSI project.

@ekluzek if we decide that this PR is still relevant for use-cases not involving an ocean mesh and/or Alper's solution doesn't work out for whatever reason, then we can revisit this PR.

slevisconsulting:
Made these updates to the landmask input file:
* landmask (was landfrac before) but zeros and ones (no fractions) should be read by mesh_modifier
* landmask_diff which is the user's area of change (was landmask before) read by fsurdat_modifier
* ncrenamed the variables lat to lsmlat and lon to lsmlon

Next I'm proceeding with other suggestions listed in [this earlier post](https://github.com/ESCOMP/CTSM/pull/1677#issuecomment-1115122373).

slevisconsulting:
> Shared optional error check by both tools that compares landmask_diff to quantity calculated on the fly by the tool:
> 
> * for error check both tools need to read both variables
> * tell it variable and dimension names that you assume are same as in the file you're comparing against
> * discourage user from using fsurdat (var PFTDATA_MASK not same as landmask) as the starting point
> * domain file also becoming obsolete...
> * so clm history from vanilla case is the way to go

We agreed that the above error-check was optional, and I am leaning strongly against implementing it because it requires a clm history file that may not exist at the time the user wishes to use these tools. I realize that a user who's modifying fsurdat and/or meshes should also feel comfortable generating a vanilla-case history file. However, such a requirement changes the simplicity of this tool from easy-to-use to a-bit-complicated, I think. Regardless, I am open to hearing other opinions.

slevisconsulting:
I'm considering a different consistency check that doesn't require a history file:
In mesh_mask_modifier, I can check that landmask = 1 for all landmask_diff = 1. This will ensure that the variable used in mesh_mask_modifier (landmask) is consistent with the variable used in fsurdat_modifier (landmask_diff).

slevisconsulting:
> I'm still having trouble with my case, so I'd like to meet with you sometime next week to go over this.

@ekluzek will confirm that this use-case now works.

slevisconsulting:
> > I'm still having trouble with my case, so I'd like to meet with you sometime next week to go over this.
> 
> @ekluzek will confirm that this use-case now works.

Ok, I think this is now ready other than the test that @ekluzek may perform.

slevisconsulting:
TODO
Add comment in README or other documentation that
`conda activate ctsm`
is required for fsurdat_modifier to generate a correctly formatted netcdf file on izumi.

ekluzek:
@slevisconsulting something new we are doing is to add commits that are just black reformatting to the top level file:

.git-blame-ignore-revs

Just add your hash for your black reformat commit to that file. That way you can do this in a clone and those commits will be ignored in git blame

git config blame.ignoreRevsFile .git-blame-ignore-revs

You commit [e44dc46](https://github.com/ESCOMP/CTSM/pull/1677/commits/e44dc469439e02e9ee582dab274d890ebdfab104)[e44dc46](https://github.com/ESCOMP/CTSM/pull/1677/commits/e44dc469439e02e9ee582dab274d890ebdfab104) applies for this

ekluzek:
OK I got my case setup appropriately (I think) it's in 

/glade/work/erik/ctsm_worktrees/modify_meshes/tools/modify_mesh_mask

Using the crujra_mesh.cfg file.

It's giving an error that the landmask and landmask_diff don't match. But, my understanding is that landmask should be the new landmask that I want and landmask_diff should be zero everywhere the mask doesn't change, and 1 where the mask does change.

@slevisconsulting could you take a look at my case and let me know what I need to do to get it to work?

ekluzek (changes requested):
This is a nice tool that will be really helpful to have. I'm mostly good with this. But, realize there should be some testing for it. So I've added a couple checkboxes to think about testing and get an example to work out of the box.

And I think possibly there is an error when you want to change the mask from 1 to 0 for a location, which means landmask_diff should be "1" because it's changing. It would work if you are changing the mask from 0 to 1.

tools/modify_fsurdat/fsurdat_modifier
```diff
@@ -28,10 +28,9 @@ This wrapper script accesses
 
 Instructions
 ------------
-To run on Cheyenne/Casper
-1) Load the following to your local environment:
-module load python
-ncar_pylib
+To run on cheyenne/casper/izumi
+1) Activate the following environment:
```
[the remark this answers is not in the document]
slevisconsulting: > This would be similar to what you have for the mesh modifier below.

From looking at this branch versus the reduce_redundancy branch, I see that the README for modify_fsurdat and your recommended changes to it are coming in with the #1714 PR.

slevisconsulting:
> And I think possibly there is an error when you want to change the mask from 1 to 0 for a location, which means landmask_diff should be "1" because it's changing. It would work if you are changing the mask from 0 to 1.

@ekluzek I think I understand the confusion. I may also be confused myself, so let's try to figure this out:
1) If I were to rename `landmask_diff` to `landmask_new_land` then it would be clearer that this variable represents **land-filling**.
2) Your intention with `half_landmaks.nc` appears to be to fill all ocean with land. Hence the 1s over ocean and the 0s over land. However, you will not accomplish that without changing the variable `landmask`. In fact, `landmask_diff` is used by this tool ONLY for consistency checking and has no effect on the mesh file.

Let me know if I have misunderstood your intention with `half_landmaks.nc`. If not, then I will go ahead and rename landmask_diff to landmask_new_land or landmask_landfill or whatever you think is most intuitive. Then I will also update the README and .cfg instructions to be clearer.

slevisconsulting:
I should clarify that in grid cells where you wish to remove land, you should set landmask and landmask_diff to 0. So to summarize:
- Where you're adding land, set landmask and landmask_diff to 1
- Where you're removing land, set landmask and landmask_diff to 0
- Where nothing changes, leave landmask unchanged and set landmask_diff to 0

Hence the suggestion to rename landmask_diff to landmask_new_land (or similar).

ekluzek:
@slevisconsulting I looked at my case to figure out what's going on. And what's going on is the starting mesh file has mask==1 everywhere, and I want to have a land/mask with ocean, but with Antarctica as ocean as well. So it's removing land over ocean and Antarctica, and as such nothing is being filled in.

I copied landmask to landmask_diff and I got it to go further, although then it complained about a problem with latitude

```
    assert isclose(lat_new, lat_mesh, abs_tol=1e-5), errmsg
AssertionError: Must be equal:  lat_new = 0.0 lat_mesh = -89.75 (at ncount = 0)
```

ekluzek:
It would be good to discuss landmask_diff further. We should make sure it has a useful purpose. I think it would be useful to either be a mask of the area's where the mask is changing from the original grid (either being filled in or removed) or as the difference with the original mask (so it could be -1, 0, or +1). Or another approach would be for it to be the original mask (which should agree with the mask on the mesh file). You could end up putting the wrong landmask file with your mesh file, so there is utility in reproducing the original mask, and making sure they agree in the ways you expect them to.

I'm not sure I understand what landmask_diff is currently trying to function as, but there is an important utility in making sure that your landmask and mesh files agree with each other. We should perhaps discuss this?

slevisconsulting:
> I copied landmask to landmask_diff and I got it to go further, although then it complained about a problem with latitude
> 
> ```
>     assert isclose(lat_new, lat_mesh, abs_tol=1e-5), errmsg
> AssertionError: Must be equal:  lat_new = 0.0 lat_mesh = -89.75 (at ncount = 0)
> ```

This one I solved easily. Your lat_varname/lon_varname variables were indices rather than latitude/longitude values. I changed them like this in the .cfg file:
```
lat_varname = lat
lon_varname = lon
```
I also added these new lines in the .cfg file:
```
lat_dimname = lsmlat
lon_dimname = lsmlon
```
to permit different variable and dimension names (as is the case in your file).

However, now I get a similar error but for the longitude:
```
assert isclose(lon_new, lon_mesh, abs_tol=1e-5), errmsg
AssertionError: Must be equal:  lon_new = 180.25 lon_mesh = 0.25 (at ncount = 0)
```
This one I will have to pursue tomorrow.

Let's definitely discuss all this when you're available. Thanks, @ekluzek!

slevisconsulting:
> [...] now I get a similar error but for the longitude:
> 
> ```
> assert isclose(lon_new, lon_mesh, abs_tol=1e-5), errmsg
> AssertionError: Must be equal:  lon_new = 180.25 lon_mesh = 0.25 (at ncount = 0)
> ```

This one is more complicated. The calls on each longitude value
```
lon_new = lon_range_0_to_360(lon_new)
lon_mesh = lon_range_0_to_360(lon_mesh)
```
...don't help because the code expects the whole grid in the mesh file to line up identically to the grid in the landmask file. I will try to come up with a solution that doesn't require the user to recreate their file.

slevisconsulting:
> ...the code expects the whole grid in the mesh file to line up identically to the grid in the landmask file. I will try to come up with a solution that doesn't require the user to recreate their landmask file.

@ekluzek 
with my latest updates, this tool can work with a landmask.nc file containing longitude values from -180 to 180 such as the one you that you were working with. I have now generated the modified mesh file that you need, here:
`/glade/work/slevis/git/mksurfdata_toolchain/tools/modify_mesh_mask/crujra_case/clmforc.cruncep.V7.c2016.0.5d.ESMFmesh_c220421.nc`

slevisconsulting:
> I'm not sure I understand what landmask_diff is currently trying to function as, but there is an important utility in making sure that your landmask and mesh files agree with each other. We should perhaps discuss this?

I'm open to discussing this further and here's an update.

It makes sense that the name "landmask_diff" was a source of confusion, and I apologize for that. In the most recent commit I renamed it to the more descriptive "mod_lnd_props" (i.e. modify lnd properties):
- mod_lnd_props is used by the modify_fsurdat tool.
- mod_lnd_props = 1 where the user wishes to modify certain land properties in the fsurdat file and 0 elsewhere.
- mod_lnd_props does not contain information about adding or removing land grid cells. That's not helpful information for the tool's purposes.
- mod_lnd_props is not used by the modify_mesh tool except in some consistency checks.

The variable used by the modify_mesh tool is called landmask:
- landmask = 1 where the user wants land and 0 elsewhere.
- landmask needs to be on exactly the same grid as the input mesh file that it's modifying (the tool checks for that).
- landmask does not need to know whether it differs from the mask in the input mesh file. That's not helpful information for the tool's purposes.
- landmask is not used by the modify_fsurdat tool.

ekluzek:
@slevisconsulting thanks for that explanation. This makes a lot more sense to me. And I like the fact that the same name is used in the modify_fsurdat tool. Your explanation says that mod_lnd_props serves the role I suggested of "a mask of the area's where the mask is changing", which I think is a great way to go. I don't think I have any further questions or feedback right now, but appreciate your changing to this form. I'll look at this more and try it for my case and get back to you.

slevisconsulting:
`make all` PASS
`clm_pymods` PASS

@ekluzek I have implemented a python system test for the mesh modifier tool, so I would like to review this PR with you. My calendar is current if you'd like to set a time. Thanks.

doc/ChangeLog
```diff
@@ -1,4 +1,76 @@
...
+
+Bugs fixed or introduced
+------------------------
+CTSM issues fixed (include CTSM Issue #): #1644
+
+Notes of particular relevance for developers:
+---------------------------------------------
+Changes to tests or testing:
+ Adding test_sys_mesh_modifier.py to python system tests. This requires
+ "module load nco" before running "make stest" or "make all" in the /python
```
slevisconsulting, line 45: @ekluzek question for when we go over this PR:
Could we include the `module load nco` in py_env_create, so that users won't have to think about it separately? If that's not an option, then do we have other existing infrastructure (eg in cime) that I could point to? @jedwards4b I'm copying you in case you have insight. Currently I run `module load nco` interactivly before running this new ctsm python system test. I would prefer to make the `module load` transparent to the user.

python/ctsm/test/test_sys_mesh_modifier.py
```diff
@@ -0,0 +1,158 @@
...
+        except subprocess.CalledProcessError as e:
+            err_msg = (
+                f"{e} ERROR using ncks to generate {scrip_file} from "
+                + f"{fsurdat_in}; MOST LIKELY SHOULD INVOKE module load nco"
+            )
+            sys.exit(err_msg)
+        # TODO How to invoke 'module load nco' behind the scenes?
+        # TODO How to make the esmf command generic?
+        # TODO How to write the PET0 file in /tempdir or suppress entirely?
+        esmf_cmd = f"/glade/u/apps/ch/opt/esmf-netcdf/8.0.0/intel/19.0.5/bin/bing/Linux.intel.64.mpiuni.default/ESMF_Scrip2Unstruct {scrip_file} {self._mesh_mask_in} 0"
```
slevisconsulting, line 72: @ekluzek another question for when we review this PR:
How to make the esmf command generic?

@billsacks I'm copying you here in case you have insight or in case you have a suggestion of another person to talk to.

doc/ChangeLog
ekluzek: Hmmm. I could add that for our NCAR systems. So it would work at NCAR, but fail elsewhere. It's not really part of the CTSM python environment though. So I'd like to run this by people and see what they think.

python/ctsm/test/test_sys_mesh_modifier.py
ekluzek: The only way I can think of is to use the cime configure and make this setup for cime machines, which will include setting up for ESMF. This would be similar to what you've done in mksurfdata_esmf. It's a bit overkill in this case, but I think that might be OK. Possibly as part of that you could add the module load for nco that you talk about above.

tools/modify_mesh_mask/README
```diff
@@ -0,0 +1,64 @@
...
+tools/modify_mesh_mask/mesh_mask_modifier
+tools/modify_mesh_mask/modify_template.cfg
+
+Instructions
+------------
+To run on Cheyenne/Casper/Izumi
+1) (Un)load, execute, and activate the following:
+module unload python
+module load conda
+./manage_python_env
```
slevisconsulting, line 17: @ekluzek another item to confirm when we go over this PR:
Should all the README documentation change to say ./py_env_create instead of ./manage_python_env?

doc/ChangeLog
jedwards4b: I guess I am confused - are you building tools specific to a given system (eg cheyenne)
modules are different on every HPC system, about the best you can do is provide a README file.

tools/modify_mesh_mask/README
ekluzek: Yes.

ekluzek:
@slevisconsulting there are still some checkboxes that should be addressed. You did add some unit testing, so I'll check off that one, but the other two should be handled.

doc/ChangeLog
ekluzek: @jedwards4b we want this tool to work everywhere you might run CTSM. So yes you are right we could maybe add the module load for a few NCAR systems, but otherwise we'll have to rely on the README file you talk about, and some error checking that fails if NCO isn't available on your system.

ekluzek: - [ ] Add a README that talks about the dependence of NCO
- [ ] Add error checking so that the script fails with a clear error if NCO isn't in the users path

python/ctsm/test/test_sys_mesh_modifier.py
slevisconsulting: I'm having an interesting experience where
- in the ctsm5.2.mksurfdata branch I can see /cime/tools/configure (git describe returns cime6.0.15)
- in my modify_meshes branch I do not see configure in the /cime/tools directory though I still see Readme.configure (git describe returns cime6.0.40)

slevisconsulting: oh, and the current PR's work is in the modify_meshes branch...

jedwards4b: cime directory structure changed, configure is in
cime/CIME/scripts/configure

python/ctsm/test/test_sys_mesh_modifier.py
```diff
@@ -0,0 +1,158 @@
...
+        metadata_file = os.path.join(self._tempdir, "metadata.nc")
+
+        # Generate mesh_mask_in from fsurdat_in
+        ncks_cmd = f"ncks --rgr infer --rgr scrip={scrip_file} {fsurdat_in} {metadata_file}"
+        try:
+            subprocess.check_call(ncks_cmd, shell=True)
+        except subprocess.CalledProcessError as e:
+            err_msg = (
+                f"{e} ERROR using ncks to generate {scrip_file} from "
+                + f"{fsurdat_in}; MOST LIKELY SHOULD INVOKE module load nco"
```
slevisconsulting, line 66: @ekluzek in response to your "error checking" [comment](https://github.com/ESCOMP/CTSM/pull/1677#discussion_r934810980):
Pls see lines 63-66 above for such error checking.

python/ctsm/test/test_sys_mesh_modifier.py
slevisconsulting: Note to self:
./configure will fail in ctsm5.2.mksurfdata branch when we update cime due to a hardwired path in gen_mksurfdata_build.sh

slevisconsulting:
I think I have addressed what we discussed.
@ekluzek pls let me know if I have missed anything.
All tests (python and clm_pymod) PASS.

Just for the record: when I'm in the ncar_pylib environment, the new test's nco commands work without a `module load nco` beforehand. But in the recommended ctsm_py environment, I need to invoke `module load nco` before starting the test.

slevisconsulting:
Made links on cheyenne and izumi, so that now both dev105 and dev106 link to dev104:
`ln -s ctsm5.1.dev104 ctsm5.1.dev106`

slevisconsulting:
@ekluzek I thought I'd test on izumi but ran into this problem:
```
>>> module load lang/python
>>> ./py_env_create
Use conda to install the python environment needed to run the CTSM python tools in the conda environment: ctsm_py
Using the file: ./python/conda_env_ctsm_py_cgd.txt
Create ctsm_py
conda create --force -n ctsm_py -q

CustomValidationError: Parameter channel_priority = 'strict' declared in <<merged>> is invalid.
The value 'strict' cannot be boolified.

Error creating conda environment ctsm_py
```

Is this user error on my part or would you like me to open an issue?

python/ctsm/modify_mesh_mask/modify_mesh_mask.py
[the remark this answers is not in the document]
ekluzek: OK, resolving this conversation as we think it's handled.

doc/ChangeLog
ekluzek: OK, the dependence on NCO is actually only for the test, but we still want to have this documented and easily seen. So resolving this.

python/ctsm/test/test_sys_mesh_modifier.py
ekluzek: This is now in place and working.

python/ctsm/test/test_sys_mesh_modifier.py
ekluzek: Perfect.

ekluzek:
@slevisconsulting go ahead and create an issue about this. I just tried on izumi on your branch and I don't see this problem. But, that means there is likely some environment issue that we need to track down. So having your problem clearly identified in an issue would help track it down. We need this to work for everyone everywhere so it's important to track down these user environment issues.

slevisconsulting:
cheyenne testing
clm_pymods PASS
python (make all) PASS

ekluzek:
@slevisconsulting it looks like there is one checkbox that isn't completed. See above for this July/15th comment that has one checkbox that's open.

> @slevisconsulting it looks like there isn't any testing for this new script. We should add some level of testing for it. You could add some unit testing, a test suite test, and/or a test/tools test for it. It also looks like the default config file won't actually work as it is, it would be good to have one that was actually functional.
> 
> * [x]  Add testing for mesh_modifier
> * [ ]  Get config file that will be functional

Can you take care of that one? It just needs a functional example of a config file that can be run to test it.

ekluzek:
Ok @slevisconsulting and I discussed the working config file. This was a requirement just to make sure testing was in place. There is a system test in place that creates a working config file.

So I'm removing the requirement since the test sys covers it.

slevisconsulting:
> Can you take care of that one? It just needs a functional example of a config file that can be run to test it.

@ekluzek here's notes from our discussion just now:
- The system test for mesh_mask_modifier generates a .cfg file and a landmask.nc file on the fly. 
- To supply a working .cfg file, we would need to provide a landmask.nc file, which adds a maintenance consideration. 
- We decided the template .cfg is sufficient.

ekluzek (approved):

ekluzek:
@slevisconsulting one thing I realized we should update at some point. You point to a test file in the mesh_modifier tool and it's README.

/glade/work/slevis/git/mksurfdata_toolchain/tools/modify_input_files/islas_examples/modify_mesh_mask/fill_indian_ocean/fill_indianocean_slevis.nc

Since it's a personal path we can't rely on that staying around. So we should either add it to python/ctsm/test/testinputs or put it in the inputdata repository somewhere. Or remove the mention of it. Personal paths to specific files on specific machines aren't useful in the long run, we should try to avoid mentioning them.

We can do this after I merge this, but I wanted to add a note about it now.

Issue #1655, opened 2022-02-21
gddplant_patch (GDDPLANT) is misleadingly named


samsrabin:
The output variable `GDDPLANT` is described as "Accumulated growing degree days past planting date for crop." Its corresponding variable in `CropType` is `gddplant_patch`: "patch accum gdd past planting date for crop (ddays)." Arguably, these are both misleadingly named and described because of the following in `CropPhenology()` (where `hui => crop_inst%gddplant_patch`):
https://github.com/ESCOMP/CTSM/blob/d9ae4b48d9c55fe3c998019bcf76817da570db87/src/biogeochem/CNPhenologyMod.F90#L2044-L2046
This bumps `gddplant_patch` up to the value where grain filling will begin if the crop has already reached peak LAI before that many GDDs have actually accumulated. Thus, `gddplant_patch` is not an actual measurement of GDD accumulation. (As far as I can tell, this is the only place anything like this happens.)

This might not have been an issue for anyone before, but it means I need to add a new output variable for my work on #519.

How I'm thinking this should be fixed:
- Rename `gddplant_patch` to `hui` ("heat unit index"), as it is more of an index than an actual measurement. `hui` makes sense because that's how `gddplant_patch` is typically `associate`d in the code.
- Rename `GDDPLANT` to `HUI`.
- Make a new variable in `CropType`, `gddaccum_patch`, to actually track the GDDs accumulated.
- Make a new output variable, `GDDACCUM`, to track `gddaccum_patch`.

This provides a complete separation from previous variable and output names and will thus break any postprocessing scripts that rely on `GDDPLANT`, but I consider that a feature, because it will force users to choose whichever of the new outputs is right for them. It's entirely possible that, as for me, `GDDPLANT` wasn't what they were expecting.

For consistency with the first bullet point, the `CNVegStateType` member `gddmaturity_patch` ("patch growing degree days (gdd) needed to harvest (ddays)") should be renamed `huimaturity_patch`. This would also be consistent with the existing `huileaf_patch` and `huigrain_patch`, also in `CNVegStateType`.

As far as I can tell—although I haven't done a very thorough check—it doesn't seem that any parts of the code using `gddplant_patch` or `gddmaturity_patch` were actually expecting true GDDs accumulated. Instead, it all looks like the HUI is expected, so I don't think anything needs to be reworked. But again, someone should check me on that.

billsacks:
[...]

PR #1684, opened 2022-03-14
1655 rename gddplant2

samsrabin:
### Description of changes
1. Rename `gddplant_patch` to `hui_patch` (and `GDDPLANT` to `HUI`).
2. Add new variable `gddaccum_patch` (output `GDDACCUM`).

### Specific notes

CTSM Issues Fixed: #1655

**Any User Interface Changes (namelist or namelist defaults changes)?** Yes. Output variable `GDDPLANT` (not included by default) is now `HUI`, and new output `GDDACCUM` has been added to track _actual_ accumulation of GDDs since planting without any modifications such as occurs here:
https://github.com/ESCOMP/CTSM/blob/d9ae4b48d9c55fe3c998019bcf76817da570db87/src/biogeochem/CNPhenologyMod.F90#L2044-L2046

#### Testing performed
As expected, `GDDACCUM` was never greater than `HUI` in any patch.

Looking at 30 patches with discrepancies between `GDDACCUM` and `HUI`, the time series look as expected. E.g.:
patch547-temperate_corn

Diffs when testing with `SMS_D_Ld10_P72x1.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-ciso_decStart.C.0311-123808ch` and `SMS_P72x1_Lm25.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-monthly.C.0311-123808ch` are due to the rename of `GDDPLANT` to `HUI`.

### DIFF
--- a/src/biogeochem/CNNDynamicsMod.F90
+++ b/src/biogeochem/CNNDynamicsMod.F90
@@ -333,7 +333,7 @@ contains
     associate(                                                                      & 
          wf               =>  waterdiagnosticbulk_inst%wf_col                      ,         & ! Input:  [real(r8) (:) ]  soil water as frac. of whc for top 0.5 m          
 
-         hui              =>  crop_inst%gddplant_patch                    ,         & ! Input:  [real(r8) (:) ]  gdd since planting (gddplant)                    
+         hui              =>  crop_inst%hui_patch                         ,         & ! Input:  [real(r8) (:) ]  hui accumulated since planting (growing degree-days)    
          croplive         =>  crop_inst%croplive_patch                    ,         & ! Input:  [logical  (:) ]  true if planted and not harvested                  
 
          gddmaturity      =>  cnveg_state_inst%gddmaturity_patch          ,         & ! Input:  [real(r8) (:) ]  gdd needed to harvest                             
--- a/src/biogeochem/CNNDynamicsMod.F90
+++ b/src/biogeochem/CNNDynamicsMod.F90
@@ -333,7 +333,7 @@ contains
     associate(                                                                      & 
          wf               =>  waterdiagnosticbulk_inst%wf_col                      ,         & ! Input:  [real(r8) (:) ]  soil water as frac. of whc for top 0.5 m          
 
-         hui              =>  crop_inst%hui_patch                         ,         & ! Input:  [real(r8) (:) ]  hui accumulated since planting (growing degree-days)    
+         hui              =>  crop_inst%hui_patch                         ,         & ! Input:  [real(r8) (:) ]  patch heat unit index (growing degree-days)    
          croplive         =>  crop_inst%croplive_patch                    ,         & ! Input:  [logical  (:) ]  true if planted and not harvested                  
 
          gddmaturity      =>  cnveg_state_inst%gddmaturity_patch          ,         & ! Input:  [real(r8) (:) ]  gdd needed to harvest                             
--- a/src/biogeochem/CropType.F90
+++ b/src/biogeochem/CropType.F90
@@ -33,11 +33,11 @@ module CropType
      logical , pointer :: cropplant_patch         (:)   ! patch Flag, true if planted
      integer , pointer :: harvdate_patch          (:)   ! patch harvest date
      real(r8), pointer :: fertnitro_patch         (:)   ! patch fertilizer nitrogen
-     real(r8), pointer :: gddplant_patch          (:)   ! patch accum gdd past planting date for crop       (ddays)
+     real(r8), pointer :: hui_patch               (:)   ! patch accum hui past planting date for crop       (ddays)
      real(r8), pointer :: gddtsoi_patch           (:)   ! patch growing degree-days from planting (top two soil layers) (ddays)
      real(r8), pointer :: vf_patch                (:)   ! patch vernalization factor for cereal
      real(r8), pointer :: cphase_patch            (:)   ! phenology phase
-     real(r8), pointer :: latbaset_patch          (:)   ! Latitude vary baset for gddplant (degree C)
+     real(r8), pointer :: latbaset_patch          (:)   ! Latitude vary baset for hui (degree C)
      character(len=20) :: baset_mapping
      real(r8) :: baset_latvary_intercept
      real(r8) :: baset_latvary_slope
@@ -194,7 +194,7 @@ contains
     allocate(this%cropplant_patch(begp:endp)) ; this%cropplant_patch(:) = .false.
     allocate(this%harvdate_patch (begp:endp)) ; this%harvdate_patch (:) = huge(1) 
     allocate(this%fertnitro_patch (begp:endp)) ; this%fertnitro_patch (:) = spval
-    allocate(this%gddplant_patch (begp:endp)) ; this%gddplant_patch (:) = spval
+    allocate(this%hui_patch (begp:endp))      ; this%hui_patch      (:) = spval
     allocate(this%gddtsoi_patch  (begp:endp)) ; this%gddtsoi_patch  (:) = spval
     allocate(this%vf_patch       (begp:endp)) ; this%vf_patch       (:) = 0.0_r8
     allocate(this%cphase_patch   (begp:endp)) ; this%cphase_patch   (:) = 0.0_r8
@@ -225,10 +225,10 @@ contains
          avgflag='A', long_name='Nitrogen fertilizer for each crop', &
          ptr_patch=this%fertnitro_patch, default='inactive')
 
-    this%gddplant_patch(begp:endp) = spval
+    this%hui_patch(begp:endp) = spval
     call hist_addfld1d (fname='GDDPLANT', units='ddays', &
-         avgflag='A', long_name='Accumulated growing degree days past planting date for crop', &
-         ptr_patch=this%gddplant_patch, default='inactive')
+         avgflag='A', long_name='Accumulated heat unit index since planting for crop', &
+         ptr_patch=this%hui_patch, default='inactive')
 
     this%gddtsoi_patch(begp:endp) = spval
     call hist_addfld1d (fname='GDDTSOI', units='ddays', &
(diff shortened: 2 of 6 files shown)


src/biogeochem/CropType.F90
```diff
@@ -569,14 +584,22 @@ subroutine CropUpdateAccVars(this, bounds, t_ref2m_patch, t_soisno_col)
...
+    call extract_accum_field ('HUI', rbufslp, nstep)
+    call extract_accum_field ('GDDACCUM', rbufslp2, nstep)
     do p = begp,endp
-      rbufslp(p) = max(0.0_r8,this%gddplant_patch(p)-rbufslp(p))
+      rbufslp(p) = max(0.0_r8,this%hui_patch(p)-rbufslp(p))
+      rbufslp2(p) = max(0.0_r8,this%gddaccum_patch(p)-rbufslp2(p))
     end do
-    call update_accum_field  ('GDDPLANT', rbufslp, nstep)
+    call update_accum_field  ('HUI', rbufslp, nstep)
+    call update_accum_field  ('GDDACCUM', rbufslp2, nstep)
```
billsacks, line 602: - [ ] Is this initial extract/modify/update step needed for GDDACCUM? I think that, if you look for other uses of accumulation fields, you'll find that this is a non-standard thing (GDDTSOI below is a decent example). I'm thinking that it is needed for HUI because of the possible update to this variable that motivated #1655 , but without an update like that, you don't need this extra step. You should double-check me by looking at other uses of accumulation fields and/or removing this code for GDDACCUM and verifying that answers are the same either way.

- [ ] [OPTIONAL] If you make this change, then I think you could safely remove the rbufslp2 variable, just using rbufslp directly in the update_accum_field for GDDACCUM below.

src/biogeochem/CropType.F90
```diff
@@ -33,11 +33,12 @@ module CropType
      logical , pointer :: cropplant_patch         (:)   ! patch Flag, true if planted
      integer , pointer :: harvdate_patch          (:)   ! patch harvest date
      real(r8), pointer :: fertnitro_patch         (:)   ! patch fertilizer nitrogen
-     real(r8), pointer :: gddplant_patch          (:)   ! patch accum gdd past planting date for crop       (ddays)
+     real(r8), pointer :: hui_patch               (:)   ! crop patch heat unit index (ddays)
+     real(r8), pointer :: gddaccum_patch          (:)   ! patch growing degree-days from planting (air) (ddays)
```
billsacks, line 37: - [ ] Can you add a comment here (either on the individual variables or a comment line above hui_patch) describing the distinction between these two variables?

src/biogeochem/CropType.F90
```diff
@@ -335,7 +342,11 @@ subroutine InitAccBuffer (this, bounds)
 
     !---------------------------------------------------------------------
 
-    call init_accum_field (name='GDDPLANT', units='K', &
+    call init_accum_field (name='HUI', units='K', &
```
billsacks, line 345: - [ ] Although this rename is nice, I think we'll need to do something for the sake of backwards compatibility with existing restart files, because each accumulation variable has a few restart variables associated with it. If this were just a matter of changing a couple of history fields, I don't think this would be important, but my impression is that formerly-GDDPLANT-now-HUI impacts the evolution of the model, right? Options I can see are:
  1. Keep the original name for this accumulation field, even though it means that the accumulation field will be named differently from the main variable name.
  2. Back out this entire rename, sticking with GDDPLANT throughout the code. (You could still add a second variable like you do in this PR.)
  3. Keep your full rename, but add some backwards compatibility capabilities in accumulMod. I think this might not be too hard (famous last words): we have a mechanism for specifying old variable names in calls to restartvar by giving a colon-delimited string like "current_varname:old_varname". So I think this would consist of:
    - Add an optional argument to `init_accum_field` like `old_name` (in principle we could allow an array of old names, but I'm thinking we shouldn't go too crazy here)
    - In `accumulRest`, if `accum(nf)%old_name` is set, we'll need to modify the restart var names accordingly. For example, the `_VALUE` name will become: `varname = trim(accum(nf)%name) // '_VALUE' // ':' // trim(accum(nf)%old_name) // '_VALUE'`, and similarly for the `_NSTEPS` and `_PERIOD` variable names.
    - Testing: I *think* the above is all that's needed, but you should test to ensure that:
      - The actual variable names on the restart file are correct (the new names)
      - If you do an `ncdump -h` on a restart file, you should see a piece of metadata on each field called `varnames_on_old_files`. For variables with a backwards-compatibility field like this, you'll see something like: `LIQCAN:varnames_on_old_files = "LIQCAN:H2OCAN"`. You should check to make sure this appears correctly. (This piece of metadata is read by the code to implement the backwards compatibility if you're interpolating initial conditions via init_interp.)
      - A test against baselines should be bit-for-bit (I suspect this is *not* currently the case for this branch because it doesn't have this backwards compatibility mechanism in place).

billsacks (changes requested):
Thank you for taking on this cleanup. I'm happy with the changes you're making here – though I'd also like to make sure that @danicalombardozzi is happy with this. I do have some specific comments below, the most substantial of which is that I think we need to consider backwards compatibility with existing restart files (unless I'm misunderstanding things and in fact that's not needed).

- [x] One other thing that's needed here is changing the following testmod to remove GDDPLANT and add HUI and GDDACCUM in its place:

https://github.com/ESCOMP/CTSM/blob/bf958ac213184e0beb4a15724b7dc971ed4797d1/cime_config/testdefs/testmods_dirs/clm/crop/user_nl_clm#L3

src/main/accumulMod.F90
```diff
@@ -711,7 +718,11 @@ subroutine accumulRest( ncid, flag )
 
     do nf = 1,naccflds
 
-       varname = trim(accum(nf)%name) // '_VALUE'
+       if (accum(nf)%old_name /= "") then
+          varname = trim(accum(nf)%name) // '_VALUE:' // trim(accum(nf)%old_name) // '_VALUE'
+       else
+          varname = trim(accum(nf)%name) // '_VALUE'
+       end if
```
billsacks, line 725: What you did for this backwards compatibility looks great - thanks! However:

- [ ]  I think you also need to do the same backwards compatibility thing for NSTEPS and PERIOD in accumulRest. These may not be needed in this case, but I think it will be important if people try to use this new old_name in other cases where NSTEPS/PERIOD are needed. However, let me know if you think otherwise based on your understanding of this module. (Side note: I'm not sure why PERIOD needs to be on the restart file in the first place, but let's not go down that rabbit hole.)

billsacks (changes requested):
Alright, this is looking really close now! I have a couple of small follow-ups.

src/main/accumulMod.F90
samsrabin: Good point, done (22bda7d).

src/biogeochem/CropType.F90
```diff
@@ -569,14 +588,18 @@ subroutine CropUpdateAccVars(this, bounds, t_ref2m_patch, t_soisno_col)
        write(iulog,*)'update_accum_hist allocation error for rbuf1dp'
        call endrun(msg=errMsg(sourcefile, __LINE__))
     endif
+    if (ier/=0) then
+       write(iulog,*)'update_accum_hist allocation error for rbuf1dp (2)'
+       call endrun(msg=errMsg(sourcefile, __LINE__))
+    endif
```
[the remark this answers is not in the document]
samsrabin: Done! (e8aee47)

billsacks (approved):
Looks great now, thanks! I just want to get approval from Danica when she is available, and then we can do final testing of this PR.

src/biogeochem/CropType.F90
```diff
@@ -570,13 +589,13 @@ subroutine CropUpdateAccVars(this, bounds, t_ref2m_patch, t_soisno_col)
        call endrun(msg=errMsg(sourcefile, __LINE__))
     endif
 
-    ! Accumulate and extract GDDPLANT
+    ! Accumulate and extract HUI and GDDACCUM
     
-    call extract_accum_field ('GDDPLANT', rbufslp, nstep)
+    call extract_accum_field ('HUI', rbufslp, nstep)
```
danicalombardozzi, line 594: Does GDDACCUM also need to be added here? It seems that HUI and GDDACCUM are both added in most other places throughout the code, so I want to make sure it's not an oversight to leave it out here.

src/biogeochem/CNNDynamicsMod.F90
```diff
@@ -333,7 +333,7 @@ subroutine CNSoyfix (bounds, num_soilc, filter_soilc, num_soilp, filter_soilp, &
     associate(                                                                      & 
          wf               =>  waterdiagnosticbulk_inst%wf_col                      ,         & ! Input:  [real(r8) (:) ]  soil water as frac. of whc for top 0.5 m          
 
-         hui              =>  crop_inst%gddplant_patch                    ,         & ! Input:  [real(r8) (:) ]  gdd since planting (gddplant)                    
+         hui              =>  crop_inst%hui_patch                         ,         & ! Input:  [real(r8) (:) ]  patch heat unit index (growing degree-days)
```
danicalombardozzi, line 336: Is HUI accumulated over a certain time period? GDD specifies that it's accumulated since planting, and I think it's the same for HUI. If so, does it make sense to add this description? Would something like "patch accumulated heat unit index (growing degree-days) since planting" be correct?

danicalombardozzi (commented):
Thanks for your work on this @samsrabin! I very much appreciate the separation of HUI, used in the phenology calculations, and GDDACCUM, which accumulates the actual GDD. Two minor comments (questions, really).

src/biogeochem/CropType.F90
billsacks: I'll answer this one, since this is related to something I asked @samsrabin to change: GDDACCUM does not need to be updated here. However, the reason for this is subtle, and the fact that its absence caught your eye (thank you for your careful look that led to that!) suggests that there should probably be a comment here explaining why this is needed for HUI. Something like: "HUI needs this special update (which is non-standard for accumulation fields) because it can be changed outside of the accumulation routines (in CropPhenology), so the accumulation buffer needs to be reset when that happens." @samsrabin what do you think?

src/biogeochem/CNNDynamicsMod.F90
samsrabin: You're right, a bit more info would be helpful. These descriptions now (10032f1) read (new text in bold), "crop patch heat unit index (growing degree-days)**; set to 0 at sowing and accumulated until harvest**".

src/biogeochem/CropType.F90
samsrabin: Good idea, done (66abdbc).

danicalombardozzi:
Thanks @samsrabin! I think these changes look great and I'm happy to sign off on this!

Issue #1691, opened 2022-03-22
Problems with winter wheat sowing criteria


samsrabin:
### Brief summary of bug

The "normal" sowing condition for winter wheat only accounts for whether today is on/after the first day of the planting window; it does not also require that today is on/before the last day of the window. Additionally, the "last chance" condition allows planting anytime on/after the last day of the window.

### General bug information

**CTSM version you are using:** `ctsm5.1.dev083`

**Does this bug cause significantly incorrect results in the model's science?** As far as I know, winter wheat is not currently used, so no—not yet.

**Configurations affected:** Any configurations with winter wheat.

### Details of bug

If either of these conditions are satisfied, winter cereals will be planted:
https://github.com/ESCOMP/CTSM/blob/bf958ac213184e0beb4a15724b7dc971ed4797d1/src/biogeochem/CNPhenologyMod.F90#L1861-L1871
The `a5tmin` bit checks whether the 5-day running mean of min 2-m temperature is ≤ a threshold, and the `gdd020` bits check whether the 20 yr mean of `gdd0` is ≥ a threshold. 

Weirdly, the "normal" condition doesn't check whether today is _in_ the planting window, just whether today is _on or after the start of_ the planting window. And the "last chance" condition, contrary to the comment, actually allows planting _after_ the window.

I thought maybe I had introduced these problems in #1479, but no. From the tag before that was merged in (`ctsm5.1.dev056`):

https://github.com/ESCOMP/CTSM/blob/7761d5bba5d9070b23dbad162d87839309729cdb/src/biogeochem/CNPhenologyMod.F90#L1827-L1846

and

https://github.com/ESCOMP/CTSM/blob/7761d5bba5d9070b23dbad162d87839309729cdb/src/biogeochem/CNPhenologyMod.F90#L1880-L1885

The comments there are all about winter cereals in rotation with some other crop on the same patch, whereas for now they have their own patches. I'm not sure the original code makes any more sense in that light, but maybe I'm missing something. Either way, for now, this seems like a bug.

### Important details of your setup / configuration so we can reproduce the bug

Test `ERS_Ly5_Mmpi-serial.1x1_smallvilleIA.I2000Clm50BgcCropQianRs.izumi_gnu.clm-ciso_monthly` has rainfed winter wheat planted on 2000-12-31; the last day of the planting window is `jday` 334 (Nov. 30).

samsrabin:
Meant to tag @jyoti1singh in this.

[...]

PR #1692, opened 2022-03-24
Fix winter wheat sowing window bugs

samsrabin:
### Specific notes

Contributors other than yourself, if any: **@jyoti1singh**

CTSM Issues Fixed: **#1691**

Are answers expected to change (and if so in what way)? **Only for winter wheat, which isn't currently supported: Some cells will fail to plant.**

Testing performed, if any: **None**


billsacks (approved):
Thank you for this fix!

danicalombardozzi:
Thanks for this fix!

danicalombardozzi (approved):

PR #1705, opened 2022-04-14
Major refactor of NutrientCompetition / CNAllocation to provide hooks for AgSys crop model

billsacks:
### Description of changes

This PR separates the NutrientCompetition modules into pieces based on (1) consolidating duplicate code between the Clm45 and FlexibleCN versions, and (2) separating pieces that will vs. won't be used for crop patches when running with the upcoming AgSys crop model.

### Specific notes

Contributors other than yourself, if any: Substantial consultation with @pengbinpeluo , as well as some with @danicalombardozzi and earlier conversations with @samsrabin 

Testing performed, if any: Just a couple of tests so far


billsacks:
I think this is close to ready. The main thing still needed is introducing a place-holder AgSys version of `calc_npool_to_components_flexiblecn`.

Issue #1707, opened 2022-04-14
Subset data should allow setting to 100% vegetated AND pull the PFTs from the surface file


adrifoster:
### Brief summary of bug

The way subset data currently functions is incorrect, I believe. Right now, you must either pick:

 1) dominant pft(s)    or 
2) --include-nonveg (overrides default setting non-veg units to 0%)

I don't think this should be the way it functions because I believe there are many users (e.g. FATES users) who would not want to set a dominant pft(s) and would also want all 100% vegetated at the single point.

Note: I realize that FATES (currently) does not actually use the surface data to specify PFTs, but I still think it is confusing and incorrect to set one or multiple dominant PFTs when we don't actually want one.

### General bug information

**CTSM version you are using:** ctsm5.1.dev090

**Does this bug cause significantly incorrect results in the model's science?** [No]

**Configurations affected:** [N/A]

### Details of bug

[This](https://github.com/ESCOMP/CTSM/blob/449345ee9a534316cba0d6b4fd696b8d57e60428/python/ctsm/site_and_regional/single_point_case.py#L215) function checks to make sure one of the above arguments is set, which I don't think should be the case.

Users should be able to set the landunit to 100% vegetated AND use the default PFTs/percentages that are in the surface file.

negin513:
Thanks for creating this. This feature was explicitly requested by the reviewers, but I agree with you that it should be disabled if it does not make sense for FATES users. 

As you pointed out, `check_nonveg` explicitly checks for at least one of `--dompft ` or  `--include-nonveg` is chosen by the user . At first, I thought this would be the only modification needed.

However, this is slightly more because there is a  `dompft` needed if we want to zero out non-vegetation land units.

This is where we will have issues:
https://github.com/ESCOMP/CTSM/blob/449345ee9a534316cba0d6b4fd696b8d57e60428/python/ctsm/site_and_regional/single_point_case.py#L425-L440

There are two ways to avoid this:
1. My original suggestion to avoid this was using a default `dompft`, but I remember there were some  issues against using a default pft.

2. Adding another ifs to check if dom_pft exists as #1727:
``` python
            if self.dom_pft is not None:
                max_dom_pft = max(self.dom_pft)
                if max_dom_pft < NAT_PFT:
                    f_mod["PCT_NATVEG"][:, :] = 100 
                    f_mod["PCT_CROP"][:, :] = 0 
                else:
                    f_mod["PCT_NATVEG"][:, :] = 0 
[...]

PR #1727, opened 2022-04-28
Removing checks for dompft and include-nonveg

negin513:
### Description of changes
Previously `subset_data` explicitly checked to make sure the user has chosen at least one of the following options:
 1) dominant pft(s)    or 
 2) --include-nonveg (overrides default setting non-veg units to 0%)

However, @adrifoster pointed out that this limits FATES users who would not want to set a dominant pft(s) and would also want all 100% vegetated at the single point. So I removed this capability. 

 
### Specific notes

CTSM Issues Fixed (include github issue #): #1707

Testing performed, if any:

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,80 @@
 ===============================================================
+Tag name: ctsm5.1.dev094
+Originator(s): negins (Negin Sobhani,UCAR/TSS,303-497-1224)
+Date: Fri May  6 00:04:54 MDT 2022
+One-line Summary: subset_data allows zeroing out nonveg landunits without any dompft selected
+
+Purpose and description of changes
+----------------------------------
+Previously subset_data explicitly checked to make sure the user has chosen at
+least one of the following options:
+
+dominant pft(s) or
+--include-nonveg (overrides default setting non-veg units to 0%)
+
+However, Adrianna pointed out that this limits FATES users who would not
+want to set a dominant pft(s) and would also want all 100% vegetated at the
+single point. So I removed this capability. This needed recalculating
+PCT_CROP and PCT_NATVEG while zero-ing out other land unit types.
+
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+
+Does this tag change answers significantly for any of the following physics configurations?
+(Details of any changes will be given in the "Answer changes" section below.)
+
+[ ] clm5_1
+
+[ ] clm5_0
+
+[ ] ctsm5_0-nwp
+
+[ ] clm4_5
+
+Bugs fixed or introduced
+------------------------
+Issues fixed (include CTSM Issue #):
+ Fixes CTSM/#1707 -- Allowing the user to zero non-veg landunits without any
+ dominant pft given by the user.
+
+
+Notes of particular relevance for users
+---------------------------------------
+Changes to CTSM's user interface (e.g., new/renamed XML or namelist variables):
+ Previously the user needed to choose at least one of the following options
+ for subsetting at single points:
+ --dompft
+ --include-nonveg
+ Now the user is not required to specify either options and can still zero out
+ non-veg landunits without any dominant pft given by the user.
+
+
+Notes of particular relevance for developers:
+---------------------------------------------
+Changes to tests or testing:
+ Added 4 new tests to python/ctsm/test/test_unit_singlept_data_surfdata.py
+ to test recalculation of PCT_CROP and PCT_NATVEG while zero-ing out non-veg
+ land units. 
+
+Testing summary:
+----------------
+here is guidance on different available levels of system testing:
+
+  python testing (if python code has changed; see instructions in python/README.md; document testing done):
+
+    clm_pymods test suite on cheyenne - PASS
+	pylint -- PASS (10.00/10) (cheyenne + python/3.7.9)
+	Python unit tests -- PASS (cheyenne)
+	Python system tests -- FAIL (cheyenne) (CTSM/Issue #1739)
+
+Other details
+-------------
+Pull Requests that document the changes (include PR ids):
+ https://github.com/ESCOMP/CTSM/pull/1727
+
+===============================================================
+===============================================================
 Tag name: ctsm5.1.dev093
 Originator(s): slevis (Samuel Levis)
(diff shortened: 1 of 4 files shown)


adrifoster:
Thank you @negin513!!

negin513:
I think this check was originally proposed in subset_data to avoid an error when we are zeroing-out non-vegetation land units:

https://github.com/ESCOMP/CTSM/blob/449345ee9a534316cba0d6b4fd696b8d57e60428/python/ctsm/site_and_regional/single_point_case.py#L430-L450

Currently I put in a check to avoid this error. However this means that when the user does not specify `dompft` and `include_nonveg`, we are zero-ing out `PCT_LAKE`, `PCT_WETLAND` , `PCT_URBAN` , and ` PCT_GLACIER`, but we are not changing `PCT_CROP` or `PCT_NATVEG`. This means that the sum of PCT landunits would not add up to 100. Wouldn't that be problematic when running CTSM?

ekluzek:
Yes, not having the sum of PCT landunits adding up to 100 will cause the model to abort. That is a requirement of the surface datasets that is checked for.

adrifoster:
My suggestion from the bug is to:

As for the problem with PCT_NATVEG and PCT_CROP:

Could you grab the current percentages of those values and then re-calculate the new percentage? (e.g. if we have 30% nat veg and 10% crop, the new zero-nonveg percentages would be 30/(30 +10)*100 for nat veg and 10/(30 + 10)*100 for crop

EDIT: added *100 for the percentages

negin513:
Thanks @ekluzek and @adrifoster. 

Then there are two possible solution to avoid this:
1- I really like @adrifoster's suggestion to calculate new percentages after zeroing-out not vegetation landunits. 

2-  My original suggestion to avoid this was using a default `dompft`, but I remember there were some  issues against using a default pft. (from #1707)

I like Adrianna's suggestion better and I already implemented it. :grinning: My only concern is that now there are a lot of if else in this part of the code, but it should be okay. I tried clearing it up with a comment.
```python
        if not self.include_nonveg:
            logger.info("Zeroing out non-vegetation land units in the surface data.")
            f_mod["PCT_LAKE"][:, :] = 0.0 
            f_mod["PCT_WETLAND"][:, :] = 0.0 
            f_mod["PCT_URBAN"][:, :, :] = 0.0 
            f_mod["PCT_GLACIER"][:, :] = 0.0 
                    
            if self.dom_pft is not None:
                max_dom_pft = max(self.dom_pft)
                if max_dom_pft < NAT_PFT:
                    f_mod["PCT_NATVEG"][:, :] = 100 
                    f_mod["PCT_CROP"][:, :] = 0 
                else:
                    f_mod["PCT_NATVEG"][:, :] = 0 
                    f_mod["PCT_CROP"][:, :] = 100 
            else:
                #-- recalculate percentages after zeroing out non-veg landunits                                                                                                    
                #-- so they add up to 100%.
                tot_pct = f_mod["PCT_CROP"]+f_mod['PCT_NATVEG']
                f_mod ["PCT_CROP"] = f_mod ["PCT_CROP"]/tot_pct * 100 
                f_mod ["PCT_NATVEG"] = f_mod ["PCT_NATVEG"]/tot_pct * 100 
                    
        else:       
            logger.info(
                "You chose --include-nonveg --> \
                Do not zero non-vegetation land units in the surface data."
            ) 
```

adrifoster:
> My only concern is that now there are a lot of if else in this part of the code, but it should be okay. I tried clearing it up with a comment.

Sometimes unavoidable! I think it's fine (only 3 blocks here). Thanks for doing this!

ekluzek:
@negin513 you are right that complex if-then logic is hard to get right, and it's good to make it simpler if possible. But, as @adrifoster points out it's likely unavoidable here. Comments help as you have done. 

These complex logic parts of the code are things where it's good to have some unit testing in place to make sure that it's right -- because it's so easy to get it wrong. So I would suggest that you have some unit testing around it.

adrifoster:
> To fix the logic, I suppose the check will need to be min_dom_pft, and an extra "elif" added for max_dom_pft< crop, and then the else for the mixed, crop and nat-veg case.

I agree I think this is a good way forward. For the mixed case, can we just use the input `pct_pft` to get the relative percentages?

ekluzek:
Yes, exactly that would be the right thing to do. And then there will need to be a sum over the crop pfts and the nat-veg pft's to figure out how much is in each type. And actually the mixed case is the most general case, so you could get rid of the ifs and always do the mixed case. It might be more clear with the ifs in place though...

negin513:
> For example, if I have my `dompft` list as [1,2,5,20] the max value would be a crop PFT, but we also have some naturally vegetated PFTs - so we can't assume 100% crop. Perhaps this was already discussed and this is not a likely scenario?

Hello Adrianna, Thanks for bringing up this point. Mixed land units is not possible with subset_data; therefore, the user cannot choose landunits that are both above and below 15. For example, if the user choose [1, 2, 5, 20], the code will automatically abort. 

Here is where we explicitly check to make sure this never happens:

https://github.com/ESCOMP/CTSM/blob/82a63cc16a451f06a79a2d4417058cbda819e766/python/ctsm/site_and_regional/single_point_case.py#L201-L213

You should be able to test it and get the same error.

adrifoster:
> Mixed land units is not possible with subset_data;

Ah I missed that okay. Hmm, is this something we want to exclude? It seems like @ekluzek was saying the mixed case might be common.

negin513:
> > Mixed land units is not possible with subset_data;
> 
> Ah I missed that okay. Hmm, is this something we want to exclude? It seems like @ekluzek was saying the mixed case might be common.

I think so. I remember I explicitly asked about this in the previous PR and I heard multiple times that we don't want mixed land units.

adrifoster:
> I think so. I remember I explicitly asked about this in the previous PR and I heard multiple times that we don't want mixed land units.

Ah okay. Well perhaps in the interest of time (upcoming tutorial) we leave it as is for now and revisit later if this is something we want to support.

ekluzek:
Ahhh, that makes sense now. You wouldn't have that mixed case for a tower site. So the only reason to support it would be for a more theoretical case, which could be useful, but, probably isn't really needed.

PR #1728, opened 2022-04-28
Accelerate generation of transient data with mksurfdata_esmf

slevisconsulting:
### Description of changes

Generation of transient data with mksurfdata_esmf was slow.
@mvertens recommended saving the routehandles from the non-transient step and reusing in the transient step.
This is necessary in subroutines:
- mkpft
- mkharvest
- mkurban
- mklake

and it involves these changes:
1. if routehandle is not created - create the route handle
2. if will be doing transient - don't destroy the route handle
3. if will not be doing transient - destroy the route handle

Useful reference in this work:
http://earthsystemmodeling.org/docs/nightly/develop/ESMF_refdoc/node5.html#SECTION050135500000000000000

### Specific notes

@mvertens @ekluzek @jedwards4b 

No

No

Testing performed, if any:
Generated transient (just 1850-1855) and non-transient (1850) data with mksurfdata_esmf before and after my mods and got bit-for-bit same answers. At the time of opening the PR I have only implemented the code changes required for mkpft.


mvertens:
@slevisconsulting - did this speed things up as expected?

slevisconsulting:
> @slevisconsulting - did this speed things up as expected?

This far no... But I've only implemented the changes in mkpft and I've only experimented with --res 1.9x2.5.

tools/mksurfdata_esmf/src/mkharvestMod.F90
```diff
@@ -65,7 +60,7 @@ module mkharvestMod
 contains
 !=================================================================================
 
-  subroutine mkharvest(file_mesh_i, file_data_i, mesh_o, pioid_o, constant, ntime, rc)
+  subroutine mkharvest(file_mesh_i, file_data_i, mesh_o, pioid_o, ntime, routehandle_r8, rc)
```
slevisconsulting, line 63: @mvertens from what I can tell, `mkharvest` was already doing a version of what you suggested to me:
I believe it was declaring routhandle_r8 at the module level to make it available in subsequent calls without recreating every time. I changed it to an in/out argument per your suggestion, but I wonder now which is preferable. Let me know what you think.

slevisconsulting:
**How long it takes to generate the landuse.nc file**
1.9x2.5 1850-1855 default code: 17 minutes
1.9x2.5 1850-1855 accel. code: 6 minutes
I see roundoff differences in the results. Does that make sense to you from the code changes @mvertens ?

@mvertens could you review my code changes? I'm happy to go over things together if you prefer.

mvertens:
@slevisconsulting - the decreased timing is great to see. Thanks for doing this. I'm not sure why we are seeing roundoff level changes. I'd really like to go over things together. Can we try for Monday?

slevisconsulting:
> I'd really like to go over things together. Can we try for Monday?

I sent you an invite. Feel free to reschedule if not convenient.

slevisconsulting:
Good news. I have explained the roundoff diffs:
The executable for my "old" cases (pre #1728 i.e. this PR) was pre #1721 (Jim's cmake mods). I reran my old cases with #1721 included and the results are bfb same now.

mvertens:
@slevisconsulting - that's fantastic!!! Really happy to hear that you figured that out. And that there is a significant speed up!
Do you still want to meet next week?

slevisconsulting:
> @slevisconsulting - that's fantastic!!! Really happy to hear that you figured that out. And that there is a significant speed up! Do you still want to meet next week?

Only to review the code together if you want to. Otherwise we can just meet later in the week and, really, only if problems arise with the new_soiltex code.

tools/mksurfdata_esmf/src/mkharvestMod.F90
```diff
@@ -253,8 +248,7 @@ subroutine mkharvest(file_mesh_i, file_data_i, mesh_o, pioid_o, constant, ntime,
        end if
     end do
 
-    ! If constant model, clean up the mapping
-    if (constant) then
+    if (.not. present(ntime)) then  ! ...else we will reuse it
```
slevisconsulting, line 251: @ekluzek suggests:
Could pass both ntime and end_time so as to also destroy routehandle in last year.
May not be worth the trouble, but it's a good practice.

slevisconsulting: Looking into this:
mksurfdata.F90 doesn't have explicit knowledge of end_time and instead exits the time loop like this:
```
if (root_task) then
           read(nfdyn, '(A195,1x,I4)', iostat=ier) string, year
           if (ier /= 0) end_of_fdynloop = .true.
        end if
        call mpi_bcast(end_of_fdynloop, 1, MPI_LOGICAL, 0, mpicom, ier)
        if (end_of_fdynloop) then
           EXIT
        end if
```
So I think this needs a more extensive refactor to accomplish the suggestion of the previous post.

slevisconsulting:
@ekluzek I think this PR is now ready. Oh, should I remove the `mkmapdata` tool as obsolete while I'm in this PR or would you wait?

As we agreed, I will watch you merge and tag the work that's in #1721 and then you will watch over my shoulder while I merge and tag this one (after a final code review).

tools/mksurfdata_esmf/src/mkharvestMod.F90
ekluzek: I agree, let's not worry about the clean at the end.

ekluzek (approved):
@slevisconsulting and I have gone over this and he went over it with @mvertens so we think this is ready after he updates to the latest tag.

slevisconsulting:
I need to update to the latest tag, resolve conflicts, try out Erik's updated buildscript, and then it will be ready for merge.

ekluzek:
@slevisconsulting let's go over the issue you had with the build script when you and I work on this tag together. So propose a meeting and we'll look at it together then. Hopefully, we can resolve it at that time...

Issue #1703, opened 2022-04-12
Reduce confusion about fsurdat vars LANDFRAC_PFT, PFTDATA_MASK vs. history vars landfrac, landmask, pftmask


slevisconsulting:
@lawrencepj1 brought up that
- fsurdat vars `LANDFRAC_PFT` and `PFTDATA_MASK` include more total land on the planet than supported by high res satellite data, much of it in SE Asia/Indonesia
- fsurdat files do not contain an accurate land mask for use in data processing
- domain files (now obsolete) contain more accurate land mask data.

fsurdat files intentionally omit land mask info (other than the pft-related vars already mentioned) in order to make fsurdat files flexible for use with any land mask provided as the inverse of an ocean mask (supplied in `/ccs_config/component_grids_nuopc.xml`).

In a meeting on 4/12/2022 with @lawrencepj1 @ekluzek @mvertens @slevisconsulting we proposed the following steps toward reducing this confusion:
- Eliminate fsurdat variable LANDFRAC_PFT because it is not used and because it overestimates land area relative to high res satellite data.
- Add a warning to fsurdat variable PFTDATA_MASK stating that it intentionally overestimates land area relative to high res satellite data and should not be used in data processing. This overestimation ensures that we do not miss PFT data in grid cells with tiny fractions of land. CTSM users should use history fields landfrac and landmask for a representation of grid cells used as land vs. ocean in their simulations. Moreover, users should use their own landfrac/landmask data if they find that the model's landfrac/landmask variables disagree with high res satellite data.

@ekluzek determined that fsurdat variable PFTDATA_MASK is currently required in fsurdat files, but only so that it may be written to history files as pftmask. So @ekluzek suggests that we stop writing pftmask to history and possibly remove PFTDATA_MASK from fsurdat, as well. Can anyone think of reasons that we may regret such a choice later?

Update from Bill Sacks (2022-09-08): What we want to do is:
- [x] On the surface datasets, add metadata to LANDFRAC_PFT making it clear that this is NOT the same as landfrac in the simulation
- [x] Possibly remove PFTDATA_MASK from the surface datasets if it is simply a 0/1 version of LANDFRAC_PFT
- [x] Do NOT output pftmask on the history files

slevisconsulting:
Summing up the options presented here:
1) Simplest option is to eliminate fsurdat variables `LANDFRAC_PFT` and `PFTDATA_MASK` in #1663 
and before then to remove pftmask from history.
[...]

Issue #1406, opened 2021-06-24
Update CTSM Glacier Dataset - mountain glaciers and how to handle floating ice shelves


renerwijn:
**Background Issue**
Over High Mountain Asia discrepancies were found in the glacier elevation distribution that were retrieved from the CTSM/CLM glacier dataset (mksrf_glacier_3x3min_simyr2000.c120926.nc). The discrepancies could in particular be found over the southeastern part of High Mountain Asia where the mean glacier elevation has found to be up to 2000 m lower (i.e., around 3000 m) than the mean CLM grid cell elevation (around 5000 m). This gave a reason to find out where these discrepancies could come from, which were eventually found to be related to 1) inaccuracies in the Randolph Glacier Inventory version 1 glacier outlines (relative to RGI version 6) (see screenshot), and 2) potential smoothing issues in the input topography (GLOBE). In addition, the glacier dataset cover elevations up to 6000 m altitude + one extra bin covering altitudes between 6000 m and 10000 m, whereas the HMA grid works with a 36-EC scheme that covers altitudes up to 7000 m + one extra EC covering altitudes between 7000 m and 10000 m. This means a part of the glaciers (between 6000 and 7000 m altitude) are not well captured by CLM. For this reason, we decided to update the existing CLM glacier dataset. 

![Afbeelding2](https://user-images.githubusercontent.com/12967422/123267598-29235b80-d538-11eb-8abe-632be4f78114.png)

**Brief description dataset**
The first version of the glacier dataset, developed by Jeremy Fykes, Alex Gardner, and Bill Sacks, uses glacier outlines retrieved from RGI version 1 (Arendt et al., 2012), vector data from the University of Zurich (Rastner et al., 2012) for the Greenland Icesheet, and vector data from the SCAR Antarctic Digital Database for the Antarctic Icesheet. The topography was retrieved from the 30-arcsec GLOBE topography (Hastings et al., 1999). 

The updated version of the glacier dataset uses glacier outlines retrieved from RGI version 6 (Arendt et al., 2017). The vector data for the GrIS and AIS are retrieved from the masks of BedMachine version 3 and version 2 (Morlighem et al., 2017,2019), respectively. The 30-arcsec topography and land mask are retrieved from the 30-arcsec GMTED2010. BedMachine and GMTED2010 were chosen to keep the datasets consistent with what is already used for the CAM topography (GMTED) and within CISM (BedMachine).  

**Workflow** 
[...]

PR #1732, opened 2022-04-29
New soiltex for ctsm5.2.mksurfdata

### LINKED ISSUE #1303: Update soils data used for surface dataset
It would be nice to update the soils data we're using to generate the surface dataset to something from this century.  This will introduce a number of answer changes to the code, but it seems worth having a discussion about what we need here.

@dlawrenncar suggested using [SoilGrids data,](https://www.isric.org/explore/soilgrids) which just released a version 2.0 of their dataset https://doi.org/10.5194/soil-2020-65.  SoilGrids2.0 contains information on soil texture, OC content, pH, bulk density, coarse fragments, CEC, and soil N at 250 m resolution for 6 soil layers (0-200 cm).  This high resolution data also includes uncertainty estimates!  According to the data providers, v2.0 has changed significantly from previous releases of the dataset, but is currently only available at 250m resolution.

Laura Poggio and Niels Batjes at ISRIC are interested in and willing to provide a coarser resolution data product for our purposes and wondered what we wanted.  I've basically told them we'd like the whole dataset, but to prioritize texture and soil C information.  Is a 5km data product adequate for NWP applications, but not too unwieldy for climate simulations?  **Do we need  1km resolution mapping flies?**

I also wondered if we should think about how to generate soil properties for the hillslope model?  Does this happen in our own tool chain, or could it be generated in the mapping files from ISRIC? This is likely of secondary concern, but may be worth discussion?

slevisconsulting:
### Description of changes
Started from @mvertens branch feature/new_mksurfdata_soiltex
Pushed a few updates directly to that branch:
- Merge remote-tracking branch 'escomp/ctsm5.2.mksurfdata'
- Merge branch 'cmake_bld_branch'

From there I started my branch new_soiltex_slevis and opened this PR.

This PR continues @mvertens and @wwieder 's work to replace the old soil texture and organic raw datasets with new.

### Specific notes

@mvertens @wwieder @dlawrenncar @ekluzek 

Resolves #1303

Yes because new rawdata.

New soil raw dataset consists of two files: a mapunits file and a lookup file.

Testing performed, if any:
@olyson posted diagnostics from running with the new and the old [here](https://github.com/ESCOMP/CTSM/issues/1303#issuecomment-1099208099) and no red flags were raised.


slevisconsulting:
@mvertens I'm adding the --hires_soitex option to the .py and .xml files, but I don't know this: 
Do we have a mesh file for this 30-second grid?

slevisconsulting:
> Do we have a mesh file for this 30-second grid?

@mvertens from this commit
https://github.com/ESCOMP/CTSM/pull/1732/commits/a849f9548cdf227b1db0f42263b94ffb8231004f
I gather that you used the same file to read mapunits and the mesh when working with the 30'' resolution.

@ekluzek and everyone:
I have now renamed and copied the three new files and would like comments/approval before I submit them to the repository with the ./rimport command:
| old name | new name |
|----|----|
| soiltex_mapunits_4320x2160_c220329.nc | mksrf_soitex_mapunits_5x5min_from_wise_30sec.c220329.nc |
| wise_30sec_v1_lookup2.nc | mksrf_soitex.10level.wise_30sec_lookup.c220329.nc |
| wise_30sec_v5_grid.nc | mksrf_soitex_mapunits_wise_30sec.c220330.nc |

I decided to keep the prefix "mksrf_" because it's the only identifier that indicates a CTSM raw dataset. Without the prefix, the file could be anything, while with the prefix the file is clearly intended for use with the mksurfdata tool.

ekluzek:
I think these are good names, I only have a couple tweaks to suggest. You can take or leave as you see fit. I'm happy with above, I just wonder about a few things.

Suggest capitalizing WISE since it's an org abbreviation
Maybe change "soitex" to just "soil" -- since we now include more soil characteristics (like PH) and not just strictly soil texture?
5x5min_from_wise_30sec is maybe confusing for the filename? Maybe just say 5x5min_WISE and inside the file you can see the metadata that says it's from the 30 second resolution file.

slevisconsulting:
> Suggest capitalizing WISE since it's an org abbreviation Maybe change "soitex" to just "soil" -- since we now include more soil characteristics (like PH) and not just strictly soil texture? 5x5min_from_wise_30sec is maybe confusing for the filename? Maybe just say 5x5min_WISE and inside the file you can see the metadata that says it's from the 30 second resolution file.

@ekluzek thank you. Here are the file names that I ended up with based on your suggestions:
`mksrf_soil_mapunits_30sec_WISE.c220330.nc`
`mksrf_soil_mapunits_5x5min_WISE.c220329.nc`
`mksrf_soil_lookup.10level.WISE.c220329.nc`  Here I also moved "lookup" after "soil" as done with "mapunits" and removed "30sec" because this file has resolution in the vertical but not horizontal.

slevisconsulting:
./rimport complete on these three files.

tools/mksurfdata_esmf/README
```diff
@@ -76,6 +78,7 @@ slevis HAS RUN THESE CASES and HAS LISTED ISSUES ENCOUNTERED
 ------------------------------------------------------------
 REMEMBER TO compare against existing fsurdat files in
 /glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/release-clm5.0.18
+0) New 30-sec raw data for soil texture fails. Try requesting more mem.
```
slevisconsulting, line 81: In this README file, this is the only substantial change (others just improve the formatting):
It documents that I haven't run successfully with the new 30-sec soil-texture data.

tools/mksurfdata_esmf/gen_mksurfdata_namelist.py
```diff
@@ -176,7 +176,19 @@ def get_parser():
             [default: %(default)s]
             """,
         action="store_true",
-        dest="hres_flag",
+        dest="hres_pft",
```
slevisconsulting, line 179: Renamed hres_flag to hres_pft to distinguish it from the new hres_soitex.

tools/mksurfdata_esmf/gen_mksurfdata_namelist.py
```diff
@@ -390,6 +410,10 @@ def main ():
                 new_key = f"{child1.tag}_urban"
                 rawdata_files[new_key] = os.path.join(input_path, item.text)
 
+            if item.tag == 'lookup_filename':
+                new_key = f"{child1.tag}_lookup"
+                rawdata_files[new_key] = os.path.join(input_path, item.text)
```
slevisconsulting, line 415: This needed for the newly introduced soil texture lookup file.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -98,229 +103,270 @@ subroutine mksoiltex(file_mesh_i, file_data_i, mesh_o, pioid_o, pctlnd_pft_o, rc
...
-             sand_o(no,l) = sand_i(mapunit_o(no),l)
-             clay_o(no,l) = clay_i(mapunit_o(no),l)
-          end do
+
+       if (pctlnd_pft_o(no) < 1.e-6_r8 .or. mapunit_o(no) == 0) then
+
+          ! Adjust sand and clay be loam if pctlnd_pft is < 1.e-6 or mapunit is 0
+          sand_o(no,:) = 43._r4
+          clay_o(no,:) = 18._r4
+          orgc_o(no,:) = 18._r4  ! TODO slevis: value for loam here?
```
slevisconsulting, line 269: @wwieder these are the model's generic values for sand and clay. What do we want for organic in that case?

Also, I haven't included a units transformation here from gC/kg (lookup file) to kg/m3 (fsurdat), which I think is what you were discussing in #1303. Did you reach a verdict?

wwieder: Sam what are the organic values currently used in oceans?  I'm included to do the same here?

I'm not sure we ever decided how to handle units, let's set up a meeting with @dlawrenncar to discuss after the tutorial.  I'll try setting up a meeting.

slevisconsulting: > Sam what are the organic values currently used in oceans? I'm included to do the same here?

Sounds good. So I changed to 0 here and also in cases when the lookup table returns a negative value. But not when the lookup table returns -4 (sand dunes) where I set organic = 1. This is on the low side compared to what I see from our old dataset. We can address this, too, in our meeting.

slevisconsulting:
This morning's meeting with @wwieder and @dlawrenncar :
1) Keep ORGC units & name unchanged when making fsurdat in mksurfdata_esmf. This allows for easier comparison between fsurdat and raw data and requires unit manipulations to happen in the land model.
2) Add CFRAG, BULK, and PHAQ to fsurdat, also with units & names unchanged.
3) Backwards compatibility:  Modify land model to use the old code when fsurdat contains ORGANIC and new code when fsurdat contains ORGC, CFRAC, BULK.

@wwieder I think the conversion formula you showed was (pls edit if incorrect):
ORGANIC = ORGC * BULK * (1 - CFRAG) * 0.01 / 0.58

I will work on (1) and (2) today.

wwieder:
I need to check units, but I think CFRAG is a fraction (0-100) so the equation would be 
`ORGANIC = BULK * ORGC * (100-CFRAG) * 0.01 / 0.58`

slevisconsulting:
> I need to check units, but I think CFRAG is a percent (0-100) so the equation would be `ORGANIC = BULK * ORGC * (100-CFRAG) * 0.01 / 0.58`

I checked it and you're right: CFRAG is in %

wwieder:
No I just want to check units for the whole calculation to make sure we're generating output thats similar to ORGANIC currently being used gOM/m3.

slevisconsulting:
> No I just want to check units for the whole calculation to make sure we're generating output thats similar to ORGANIC currently being used gOM/m3.

Here's what I'll do:
I can temporarily put both ORGC and ORGANIC (using the above conversion) on the new fsurdat file that I will generate, so that we may compare to an existing old fsurdat file.

wwieder:
The units should be fine, ORGC = gC/kg soil & BULK = g soil / cm^3. After converting units, the product of these = kgC/m3, as intended.  The other calculations  account for rocks and convert to gC to gOM.  Sam, I think going ahead to calculate ORGANIC on the new surface dataset like this would be helpful so we can make progress more quickly.  Good suggestion.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -98,237 +115,428 @@ subroutine mksoiltex(file_mesh_i, file_data_i, mesh_o, pioid_o, pctlnd_pft_o, rc
...
+          end if
+          do l = 2,nlay
+             phaq_o(no,l) = phaq_i(l,1,lookup_index)
+             if (phaq_o(no,l) < 0._r4) then
+                phaq_o(no,l) = phaq_o(no,l-1)
+             end if
+          end do
+
+          ! TODO Rm organic before merging PR #1732 UNLESS we opt for OPTION 2
+          ! organic_o OPTION 1: as we plan to calculate organic in the CTSM
```
slevisconsulting, line 501: As a test I calculated organic_o in two different ways...
- Option 1 as we discussed this morning: calculate organic in the CTSM from orgc_o, cfrag_o, and bulk_o (i.e. after the terms have been regridded).
- Option 2 (a few lines up where orgc_o is also calculated): In this case I regrid organic_i (calculated from orgc_i, cfrag_i, and bulk_i) to organic_o. This may be the more accurate approach. It goes against this morning's idea to keep ORGC the same in fsurdat as in the raw data, but it may be more accurate because it first calculates organic_i and then regrids to organic_o rather than regridding all the terms first and then calculating organic_o.

The fsurdat files from this test are in `/glade/work/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf`:
```
surfdata_1.9x2.5_hist_78pfts_CMIP6_1850_c220527_OPT1.nc
surfdata_1.9x2.5_hist_78pfts_CMIP6_1850_c220527_OPT2.nc
surfdata_1.9x2.5_hist_78pfts_CMIP6_1850_c220527_OPT2-1.nc (the difference file)
```
Also, the old fsurdat that I'm comparing against:
`/glade/p/cesmdata/inputdata/lnd/clm2/surfdata_map/release-clm5.0.18/surfdata_1.9x2.5_hist_78pfts_CMIP6_simyr1850_c190304.nc`
I'm happy to follow up next week. Have a nice holiday weekend!

slevisconsulting: Posting the difference in ORGANIC calculated by method 2 vs. 1.
Very little difference, I think because we regrid by most dominant rather than by averaging.

organic2vs1

slevisconsulting: When I get a chance, I'm curious to understand this small difference. Meanwhile though I've confirmed what should go without saying, i.e. that ORGC, CFRAG, BULK, PHAQ (and all other fields) are identical between these two fsurdats.

slevisconsulting: Oh and old fsurdat shows max organic = 130 as we discussed:

organic_old

dlawrenncar:
This looks encouraging.  In terms of trying to understand what the differences look like between the old fsurdat and the new data, perhaps you could plot with maximum value of 130, so we can see how the two datasets are similar/distinct.  There are big difference, for example in the Canadian Archipelago and around Hudson Bay that I can see.  No reason for me to think that the new data is wrong, though.  I'd also be interested to see how this looks going down into the soil.  That is, what do these datasets look like at 20cm, 50cm depth?

slevisconsulting:
> perhaps you could plot with maximum value of 130, so we can see how the two datasets are similar/distinct. There are big difference, for example in the Canadian Archipelago and around Hudson Bay that I can see. No reason for me to think that the new data is wrong, though.

New versus old layer 1, which goes to 1.75 cm:
organic_new_vs_old

> [...] what do these datasets look like at 20cm, 50cm depth?

Comparing at 16.6 cm
organic_new_vs_old_16 6cm

Comparing at 28.9 cm
organic_new_vs_old_28 9cm

Comparing at 49.3 cm
organic_new_vs_old_49 3cm

wwieder:
There are clear differences between datasets.  Maybe the best way to understand the impacts would be to run simulations that compare permafrost extent, soil temperatures, and hydrology / runoff.

slevisconsulting:
At our 2022/9/8 meeting we agreed on a year-2000 1-degree SP simulation with all the new raw datasets.
Question: How long of a simulation do we want?

To generate the fsurdat file for this simulation with the new mksurfdata_esmf:
- I'm updating the glacier data as discussed in #1406 
- I'm updating the lake data as discussed in #1106 

If I am not mistaken, two issues currently in progress will require pointing to new raw datasets:
- #1556 @lawrencepj1 
- #591 @olyson 

I may try the simulation before these issues complete and repeat it once they complete.

slevisconsulting:
As I have now run simulations with fsurdat files coming out of this branch, and we will likely run more,
@ekluzek recommends that I merge this branch to the ctsm5.2.mksurfdata branch and make a tag.

Once I do that, @ekluzek will merge the latest ctsm tag to the ctsm5.2.mksurfdata branch and make a tag.

slevisconsulting:
But first...
I was reviewing my code changes in this PR and found that we had been comparing two methods of obtaining ORGANIC. The preferred method was "option 1" rather than "option 2", which means that I must rerun the new_fsurdat simulation to use option 1. Once I'm satisfied with the results, I will push another commit to this PR and then merge to ctsm5.2.mksurfdata.

cime_config/testdefs/testmods_dirs/clm/smallville_dynlakes_monthly/user_nl_clm
```diff
@@ -1,7 +1,7 @@
 do_transient_lakes = .true.
 
 ! This file was created with the following command:
-! ncap2 -s 'PCT_LAKE=array(0.0,0.0,PCT_CROP); PCT_LAKE={0.,50.,25.,25.,25.,25.}; HASLAKE=array(1.,1.,AREA); PCT_CROP=array(0.0,0.0,PCT_LAKE); PCT_CROP={0.,25.,12.,12.,12.,12.}' landuse.timeseries_1x1_smallvilleIA_hist_78pfts_simyr1850-1855_c160127.nc landuse.timeseries_1x1_smallvilleIA_hist_78pfts_simyr1850-1855_dynLakes_c200928.nc
+! ncap2 -s 'PCT_LAKE=array(0.0,0.0,PCT_CROP); PCT_LAKE={0.,50.,25.,25.,25.,25.}; PCT_LAKE_MAX=array(1.,1.,AREA); PCT_CROP=array(0.0,0.0,PCT_LAKE); PCT_CROP={0.,25.,12.,12.,12.,12.}' landuse.timeseries_1x1_smallvilleIA_hist_78pfts_simyr1850-1855_c160127.nc landuse.timeseries_1x1_smallvilleIA_hist_78pfts_simyr1850-1855_dynLakes_c200928.nc
```
slevisconsulting, line 4: @billsacks this is the only change in this commit where I didn't feel confident. Could you confirm? Other than the variable name, would you have also changed the array values for the purposes of this test?

slevisconsulting: Thanks, @billsacks 
I will take care of this later since I merged this branch earlier today. 
Probably in a PR that will address #1703

PR #1738, opened 2022-05-05
correct perched water table calculation

swensosc:
### Description of changes
Modifying the calculation of the frost table and perched water table layers.
### Specific notes
* PerchedWaterTable

1) Frost table depth

a) original frost table determination looped from the top 
of the soil column downward to the index of the first layer 
whose temperature was <= freezing, and whose neighbor above 
had a temperature above freezing.  The frost table depth is 
then given by the node of that soil layer, i.e. z(k_frz).  

b) in the new method, the same index is found, but the 
depth of the frost table is given by the depth of the top 
of the frozen layer, i.e. zi(k_frz-1).  Note zi(k_frz) would 
be the bottom of layer k_frz.  

2) Perched water table depth

a) in the original formulation, a loop from k_frz to layer 1 
was used to identify the deepest layer in 1:k_frz whose 
volumetric soil moisture was greater than a threshold given 
by sat_lev (e.g. sat_lev = 0.9).  

b) in the new method, the search is only done if k_frz is 
greater than 1.  The rationale is that if k_frz = 1, then 
zwt_perched has already been initialized to the frost_table 
depth (which is equal to the top of the uppermost soil layer), 
and therefore no search is required.

3) Determining perched water table depth within layer 
identified by index k_perch in 2)

a) in the original formulation, the perched water table 
depth was calculated by linearly interpolating between layers 
k_perch and k_perch+1, with no consideration of their relative 
values.  In the case where the deeper layer was drier than the 
layer above it, this could result in values far outside the 
soil layer.

b) in the new formulation, if the deeper layer is drier than the 
layer above it (s1 > s2), then the perched water table depth is 
simply given by the depth of the upper surface of layer k_perch, 
i.e. zi(c,k_perch-1).

* PerchedLateralFlow

1) Removal of icefrac calculation

a) in the original calculation, the frozen layer was included, 
so an ice impedance factor was calculated using icefrac.  If 
only unfrozen layers are used, no ice impedance factor is needed.

b) in the new formulation, the icefrac variable and loop are 
removed.

2) Move loop calculating frost and perched water table depths

a) in the original formulation, the frost and perched water 
table depths were calculated in the same loop as the calculation 
of the lateral flow from the perched saturated zone.

b) in preparation for the hillslope hydrology branch, this 
calculation is moved into its own loop.

3) q_perch calculation 

a)in the original formulation, q_perch was calculated by summing 
over layers k_perch to k_frost.  However, because k_frost is now 
identified as the frozen layer, and its depth the top of the frozen 
layer, it should not be included in the calculation; only the 
unfrozen layers above it should be included.

b) in the new formulation, the loop is bounded by k_frost-1 instead 
of k_frost.

4) Removal of water from perched saturated zone

a) the in the original method, the frozen layer (k_frost) 
was included in the loop.  Also, the drainage was defined to be 
negative, which was confusing.  

b) in the new method, the frozen layer is not included in the 
loop; water is only removed from the unfrozen layers above k_frost.  
Calculate drainage as positive values, which are then subtracted 
from the soil moisture in each layer.  

yes, small changes to surface climate due to differences originating in soil moisture

no
Testing performed, if any:
two five year simulations w/ and w/o changes to show that variables are broadly the same.  For example, 
ground temperatures (TG) showed maximum mean differences of around 0.05C at some high latitude locations.  
Differences in QDRAI_PERCH were around 1e-10.

### DIFF
--- a/src/biogeophys/SoilHydrologyMod.F90
+++ b/src/biogeophys/SoilHydrologyMod.F90
@@ -2082,7 +2082,7 @@ contains
              
              !--  remove residual rsub_top  --------------------------------
              ! make sure no extra water removed from soil column
-             rsub_top(c) = rsub_top(c) - rsub_top_tot/dtime
+             rsub_top(c) = rsub_top(c) + rsub_top_tot/dtime
           endif
           
           zwt(c) = max(0.0_r8,zwt(c))
--- a/src/biogeophys/SoilHydrologyMod.F90
+++ b/src/biogeophys/SoilHydrologyMod.F90
@@ -1670,7 +1670,6 @@ contains
      character(len=32) :: subname = 'PerchedLateralFlow' ! subroutine name
      integer  :: c,j,fc,i                                ! indices
      real(r8) :: dtime                                   ! land model time step (sec)
-     real(r8) :: dzmm(bounds%begc:bounds%endc,1:nlevsoi) ! layer thickness (mm)
      real(r8) :: wtsub                                   ! summation of hk*dzmm for layers below water table (mm**2/s)
      real(r8) :: h2osoi_vol
      real(r8) :: drainage_tot
--- a/src/biogeophys/SoilHydrologyMod.F90
+++ b/src/biogeophys/SoilHydrologyMod.F90
@@ -1552,6 +1552,7 @@ contains
      associate(                                                            & 
           dz                 =>    col%dz                                , & ! Input:  [real(r8) (:,:) ]  layer depth (m)                                 
           z                  =>    col%z                                 , & ! Input:  [real(r8) (:,:) ]  layer depth (m)                                 
+          zi                 =>    col%zi                                , & ! Input:  [real(r8) (:,:) ] interface level below a "z" level (m)
           t_soisno           =>    temperature_inst%t_soisno_col         , & ! Input:  [real(r8) (:,:) ]  soil temperature (Kelvin)                       
 
           h2osoi_liq         =>    waterstatebulk_inst%h2osoi_liq_col        , & ! Output: [real(r8) (:,:) ]  liquid water (kg/m2)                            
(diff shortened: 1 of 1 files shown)


src/biogeophys/SoilHydrologyMod.F90
```diff
@@ -1705,89 +1709,77 @@ subroutine PerchedLateralFlow(bounds, num_hydrologyc, filter_hydrologyc, &
...
+       enddo
 
        ! compute drainage from perched saturated region
        do fc = 1, num_hydrologyc
           c = filter_hydrologyc(fc)
 
           qflx_drain_perched(c) = 0._r8
+          if (frost_table(c) > zwt_perched(c)) then
 
-          if ((frost_table(c) > zwt_perched(c)) .and. origflag == 0) then
```
billsacks, line 1726: - [ ] Just double-checking the removal of origflag from this conditional, and the removal of `if(origflag == 1) qflx_drain_perched(c) = 0._r8` below: I know we have talked about dropping support for origflag = 1. But since we haven't yet, I wanted to double-check: is it okay to just do this block of code even with origflag = 1, or is now the time to go ahead and remove origflag entirely?

billsacks (approved):
This looks good to me - thanks, @swensosc . 

I do have one specific question on origflag that I want confirmation on - see below.

Also, can you confirm: are all changes here also already on the hillslope_hydrology branch?

(Also note: I pushed a small commit that just removes a now-unused variable.)

billsacks:
@swensosc - I tried running a test of this branch against the latest hillslope hydrology branch (including the commit you pushed there today); on the hillslope hydrology branch I backed out the cdeps update so they would use the same externals, and I rebased this branch to be on top of ctsm5.1.dev090 so they are using the same baseline version.

Unfortunately, I still get answer changes of hillslope against this for a basic test (SMS_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_intel.clm-default). So some more digging may still be needed to figure out the cause of answer changes on the hillslope branch - unless you know already.

That could be done either on this branch or a separate branch... though given that you have created this nice package of changes already, I'm slightly inclined to bring this package to master and then do a separate branch / pr for any additional answer changes from the hillslope branch - unless those other answer changes are closely related to these.

swensosc:
@sacks: I found two small errors/typos in the hillslope branch, and I have
pushed the changes.  Would you mind trying once more with those changes?

swensosc:
it looks like it would be straightforward to remove the rest of the
origflag references; I see it in 3 routines.  Do you want me to take it out
of the hillslope branch entirely now?

billsacks:
Great, thanks. With your latest changes, the one short test is bit-for-bit between this branch and the hillslope branch. I'm going ahead and running the full test suite so we can have a more complete picture; I'll let you know the results of that.

Yes, if you have time to go through and remove references to origflag, that seems helpful at this point. Doing that on the hillslope branch makes sense.

billsacks:
I ran the full test suite on this branch on Friday (rebased onto dev090 so that it could be compared with the hillslope branch). All tests passed, though with answer changes as expected.

I also ran the full test suite on the hillslope_hydrology branch with comparisons against this branch, to make sure this branch now captures all of the answer changes from the hillslope_hydrology branch. I didn't get a chance to check the final results before cheyenne went down, but based on the izumi tests and the tests I did check on cheyenne, everything was bit-for-bit between the hillslope_hydrology branch and this one (with this one rebased onto dev090 and the cdeps external change backed out on the hillslope branch).

I will post final results once cheyenne is back up, but it looks like this branch will be ready to come to master once cheyenne is back up.

billsacks:
I checked the rest of the tests of comparing the hillslope_hydrology branch against this one; all are bit-for-bit other than fieldlist diffs (other than the mct tests, which currently fail at runtime on the hillslope_hydrology branch).

So I think this verifies that the answer changes from the hillslope hydrology branch have been successfully migrated to this branch. However, the one thing I want to check is this comment from my hillslope hydrology review:

https://github.com/ESCOMP/CTSM/pull/1715#discussion_r870899923

- [ ] There was a sign change in accumulating rsub_top in the hillslope_hydrology branch that I don't see here. Apparently this doesn't impact answers in any of our tests, but I wanted to check if that is a (theoretically) answer-changing thing that should be here, too, or if that change is tied in with other hillslope hydrology changes.

Issue #1739, opened 2022-05-05
Some python system tests are failing (lilac build python tests)


negin513:
### Brief summary of bug

I was just testing python unit tests for PR #1727 and noticed some LILAC related unit tests are failing with a . I am not sure what is causing this issue. This is the error that I am getting from running `make` in `python/` directory:
```
EERROR: Current machine cheyenne does not match case machine ctsm-build.
```
The complete error is attached below.

### General bug information

**CTSM version you are using:** ctsm5.1.dev091

### Details of bug
```
python3 ./run_ctsm_py_tests  --unit
.....................................................................................................................
----------------------------------------------------------------------
Ran 117 tests in 0.210s

OK
python3 ./run_ctsm_py_tests  --sys
...ERROR: Current machine cheyenne does not match case machine ctsm-build.
EERROR: Current machine cheyenne does not match case machine ctsm-build.
E
======================================================================
ERROR: test_buildSetup_userDefinedMachine_allInfo (test.test_sys_lilac_build_ctsm.TestSysBuildCtsm)
Get through the case.setup phase with a user-defined machine
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/glade/scratch/negins/ctsm_1707/python/ctsm/test/test_sys_lilac_build_ctsm.py", line 90, in test_buildSetup_userDefinedMachine_allInfo
    inputdata_path=os.path.join(self._tempdir, 'my_inputdata'))
  File "/glade/scratch/negins/ctsm_1707/python/ctsm/lilac_build_ctsm.py", line 180, in build_ctsm
    inputdata_path=inputdata_path)
  File "/glade/scratch/negins/ctsm_1707/python/ctsm/lilac_build_ctsm.py", line 619, in _create_case
    subprocess.check_call([xmlchange, 'LILAC_MODE=on'], cwd=case_dir)
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/glade/scratch/negins/tmpm9i4wv5c/ctsm_build/case/xmlchange', 'LILAC_MODE=on']' returned non-zero exit status 1.

======================================================================
ERROR: test_buildSetup_userDefinedMachine_minimalInfo (test.test_sys_lilac_build_ctsm.TestSysBuildCtsm)
Get through the case.setup phase with a user-defined machine
[...]

PR #1723, opened 2022-04-26
First pass at getting fates-sp mode to run without soil BGC

ekluzek:
### Description of changes

Get FATES-SP mode to run without soil biogeochemistry

### Specific notes

  Set it up so that fates-SP mode with run with soil_decomp_method == None. I make some progress towards allowing soil decomposition to be off in general in FATES, but it's still hardcoded to assume this for FATES-SP mode and only allow it there.
  Also add a user-mod directory for fates_sp

Contributors other than yourself, if any:

CTSM Issues Fixed (include github issue #):
  Fixes: #1720
  Fixes: #1726
  Fixes: #1729 
  Fixes: #1610
  Initial add of FatesSP compset: #1697
  Starts adding a test mod for: #1722
  Some part of #1736

Are answers expected to change (and if so in what way)? Yes

Any User Interface Changes (namelist or namelist defaults changes)? No

Testing performed, if any: Ran a couple cases with the testmod in place
   SMS_Lm4.f45_f45_mg37.I2000Clm51Fates.cheyenne_intel.clm-fates_sp.20220425_151428_o9pw1i

which PASSes

SMS_D.f45_f45_mg37.I2000Clm51Fates.cheyenne_intel.clm-fates_sp

is currently dying with a subscript overflow error
(it looks like we don't run SatPhen with DEBUG on in the fates test list so this probably exists beforehand)

PR #1740, opened 2022-05-05
Workaround for some python system tests

billsacks:
### Description of changes

Resolves ESCOMP/CTSM#1739

### Specific notes

CTSM Issues Fixed (include github issue #): Resolves ESCOMP/CTSM#1739

Testing performed, if any: python testing


ekluzek (approved):
@billsacks and I talked about this. It does require a later version of cime (cime6.0.18), but I am updating cime in #1723 so I'll merge this in there and verify that this fixes the issue.

It may be a "hack", but I think it's acceptable for LILAC for right now.

PR #1741, opened 2022-05-06
Fixes to run_neon 

briandobbins:
### Description of changes

This change explicitly sets the NetCDF PIO mode and the PIO_REARRANGER_LND mode. 

### Specific notes

No

No

Testing performed, if any:

Tested simple ABBY runs on Cheyenne and AWS - both work.  (Previously, it would use the wrong PIO_REARRANGER_LND value on AWS, resulting in very slow writes or crashes.)

### DIFF
--- a/tools/site_and_regional/run_neon.py
+++ b/tools/site_and_regional/run_neon.py
@@ -416,6 +416,8 @@ class NeonSite:
 
                 # --change any config for base_case:
                 # case.set_value("RUN_TYPE","startup")
+                case.set_value("PIO_TYPENAME", "netcdf")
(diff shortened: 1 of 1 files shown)


ekluzek (changes requested):
Can you give a little more description to these settings? I'd also like to see your name or initials and a date, so we can think about removing it sometime later.

Something like the following:

# The following needs to be set because of a PIO bug (see PIO issue #...) BMD 05/06/2022

It's also OK to be unsure, but it's best if you can point to an issue.

briandobbins:
Added clearer comments.  Does that help?  I feel maybe I should create a PIO issue too, and will do that in the coming days, though @jedwards4b certainly has a better understanding of this issue, and it might be a known thing and not a bug.

ekluzek (approved):

Issue #742, opened 2019-06-04
Automatically download required raw data for making a new surface dataset


billsacks:
As part of making the toolchain more usable on other systems, we want to provide a mechanism for automatically downloading required raw data for making a new surface dataset.

ekluzek:
Here's what I would in slightly older versions of CLM...

http://www.cesm.ucar.edu/models/cesm1.2/clm/models/lnd/clm/doc/UsersGuide/x11868.html#getmksurfdata_map_datasets

I think the above will work with some modifications to paths. Obviously long term we need a tool that can just do this.

billsacks:
Great, thanks @ekluzek . Yes, something like this needs to be folded into the toolchain so it happens without any extra work from the user.

slevisconsulting:
Relates to #645

slevisconsulting:
> Here's what I would in slightly older versions of CLM...
> 
> http://www.cesm.ucar.edu/models/cesm1.2/clm/models/lnd/clm/doc/UsersGuide/x11868.html#getmksurfdata_map_datasets
> 
> I think the above will work with some modifications to paths. Obviously long term we need a tool that can just do this.

From discussion with @ekluzek 
we don't want user to have to modify a namelist when running check_inputdata
but otherwise check_inputdata seems like the way to go for easy downloads because it's supported by the cime community.

billsacks:
I agree that using check_inputdata is probably the way to go here. However, if I remember correctly from when I was leveraging check_inputdata for lilac, I think you'll need to create a fake case to do this and then put together an input data list in the correct location in the case directory. (Here's the wrapper script for downloading input data for lilac, though this assumes that the fake case directory has already been set up: https://github.com/escomp/CTSM/blob/master/python/ctsm/lilac_download_input_data.py .)

slevisconsulting:
Thank you @billsacks

slevisconsulting:
Works for mksurfdata_esmf (see #1748). Ok to close this issue?

billsacks:
> Works for mksurfdata_esmf (see #1748). Ok to close this issue?

Yes - thank you for getting this working!

PR #1748, opened 2022-05-12
Enable mksurfdata esmf to build and run on casper and izumi

slevisconsulting:
### Description of changes
Small changes to `gen_mksurfdata_build.sh` and `gen_mksurfdata_jobscript_single.py` to recognize casper.

### Specific notes

@ekluzek 

CTSM Issues Fixed (include github issue #): #742

I assume that generating fsurdat files on different machines doesn't produce identical results.

Added optional --machine option to `gen_mksurfdata_jobscript_single.py`.
Need to add it to `gen_mksurfdata_jobscript_multi.py`, as well.

Testing performed, if any:
Built the executable on casper.
Running it right now to generate an fsurdat file.


slevisconsulting:
This PR relates to the portability issue #645 

Two things:
- I mentioned making the same changes to `gen_mksurfdata_jobscript_multi.py` but now I wonder whether we want to encourage generating many or "all" the surface datasets anywhere other than cheyenne. @ekluzek interested in your feedback.
- While I'm thinking about it, I wanted to point out that the build works on **izumi**, but I haven't tried running the executable on izumi. This would need modifications to `gen_mksurfdata_jobscript_single.py` and `gen_mksurfdata_jobscript_multi.py` similar to the ones done here for casper.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
```diff
@@ -43,6 +43,14 @@ def get_parser():
         dest="tasks_per_node",
         required=True,
     )
+    parser.add_argument(
+        "--machine",
+        help="""currently this recognizes cheyenne and casper (default cheyenne)""",
```
ekluzek, line 48: Add a note to this that this needs to be the cime machine. So it's a machine that has been ported to cime and you can use it to build a cime model. You might also show the link to list cime machines and how to port a cime model.

ekluzek (changes requested):
I ask for one simple change here. But, I like this capability to make sure it works on Casper. And in principle this is great and a nice addition. I also like the addition of an argument, which would allow it to work on any general system without having to change the script.

I feel better about this being robust by being able to build/run it on more machines and compilers. So this is a nice addition. One thing we want to do is for users who don't have access to cheyenne, to be able to create surface datasets on their systems that they are going to run the model on. The more systems that we have access to that we can demonstrate that with -- the more confidence we have that someone else will be able to adapt it to their system.

ekluzek:
@slevisconsulting my view that I say above as well, is that we want to know that this tool can work on any system that someone is going to run the model on. At least the build anyway. Setting up the batch script for running is probably a bigger left to be done in a general way, and it's reasonable that the user can setup their own batch script for those purposes. So I think we should expect that a user will setup their own batch scripts for non-Cheyenne systems. By non-Cheyenne I mean Cheyenne or Derecho when it becomes our workhorse system.

For users that do have access to Cheyenne, I think we want to recommend using it, because that will be the system that we create surface datasets on. And we won't be creating them on other systems. We do want to know that you can create and run on other systems though. So doing some customized testing for that now, seems reasonable to me. You may have to setup some machine specific batch scripts to test this, for the machines we want to test on. But, doing one resolution on the non-Cheyenne systems seems reasonable to me. You don't really know if it's successful until you both build and run on the system, so you probably do have to have specific batch scripts for these systems to test on. But, these test batch scripts could just test one resolution and one hardcoded number of tasks. So they could be very hardcoded to test one thing.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
```diff
@@ -45,7 +45,10 @@ def get_parser():
     )
     parser.add_argument(
         "--machine",
-        help="""currently this recognizes cheyenne and casper (default cheyenne)""",
+        help="""currently this recognizes cheyenne and casper (default
+                cheyenne); this needs to be a cime machine, i.e. a machine
+                that has been ported to cime where you can build a cime model;
+                for details see the README in this directory""",
```
slevisconsulting, line 51: @ekluzek I added some more info here but didn't want to repeat too much since it's in the README, so I refer the user to the README for the details.

ekluzek: Perfect.

ekluzek (approved):

slevisconsulting:
I haven't, yet, successfully generated a surface dataset on casper, because I keep running out of wall-clock time. I bumped up the requested time to 4 hours, and I hope it's enough.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
```diff
@@ -81,14 +93,21 @@ def main ():
...
+        elif machine == 'casper':
+            runfile.write('#PBS -q casper \n')
+            runfile.write(f"#PBS -l select={number_of_nodes}:ncpus=12:mpiprocs={tasks_per_node} \n")
         runfile.write("\n")
 
         np = int(tasks_per_node) * int(number_of_nodes)
 
-        output = f"mpiexec_mpt -p \"%g:\" -np {np} ./bld/mksurfdata < {namelist_file}"
+        if machine == 'cheyenne':
+            output = f"mpiexec_mpt -p \"%g:\" -np {np} ./bld/mksurfdata < {namelist_file}"
```
slevisconsulting, line 108: @jedwards4b 
does this -p option seem useful to you here for running the mksurfdata_esmf executable? Or should we remove it?

jedwards4b: The -p option prepends the task id to the stdout file,  I find it useful and would leave it in for cheyenne.   But we should not need to have scripts which are specific to any machine, using cime you should be able to generate a exe command suitable for any platform.

slevisconsulting: Thank you @jedwards4b

slevisconsulting:
First successful generation of an fsurdat on casper:
`surfdata_10x15_hist_78pfts_CMIP6_1850_c220518.nc`
using this task geometry:
`#PBS -l select=12:ncpus=12:mpiprocs=12:mem=20GB`
took 51 wallclock minutes.

slevisconsulting:
Merge this PR before #1756 because some of the machine-related changes have continued in #1756

jedwards4b:
@slevisconsulting I don't think that you can do it - do you have a list of the files you need to rimport?

slevisconsulting:
> @slevisconsulting I don't think that you can do it - do you have a list of the files you need to rimport?

@jedwards4b are you concerned about permissions? I've done a few rimports in the last few yrs.

In other news:
First successful generation of a 2-deg fsurdat on casper: `surfdata_1.9x2.5_hist_78pfts_CMIP6_1850_c220512.nc` using this task geometry: `#PBS -l select=12:ncpus=12:mpiprocs=12:mem=80GB` took almost 2 wallclock hours.

slevisconsulting:
This far I have `./rimport`ed files that appear in two lists:
`/glade/work/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf/list1_to_rimport
`/glade/work/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf/list2_to_rimport

I had to `nccopy -k cdf5` several files to complete the ./rimport
For sanity, I have confirmed bit-for-bit same fsurdat results with the cdf5 files in the 4x5 resolution.

python/ctsm/mksurfdata_download_input_data.py
```diff
@@ -0,0 +1,86 @@
+"""Functions implementing the download_input_data command for mksurfdata"""
```
slevisconsulting, line 1: This code and its wrapper `download_input_data` worked!

slevisconsulting:
In the last commit, the comment "Jobscript now runs but fails right away" only applies to izumi.

slevisconsulting:
> In the last commit, the comment "Jobscript now runs but fails right away" only applies to izumi.

@ekluzek made some [recommendations here](https://github.com/ESCOMP/CTSM/pull/1756#issuecomment-1138146016).
@ekluzek I have now tried this:
`mpiexec --machinefile ../nodefile -n 6 --prepend-rank ./mksurfdata < ../surfdata_4x5_hist_78pfts_CMIP6_1850_c220524.namelist`
with the following nodefile contents based on a recommendation online for a `nodes=2:ppn=3` case:
```
n020101
n020101
n020101
n030102
n030102
n030102
```
The error I get is
`user specified host not in the PBS allocated list`
and I haven't figured out, yet, valid host names for izumi. Do you have a suggestion?

Alternatively, should/could we switch to openmpi since it doesn't require a nodefile? If so, what's required to make the switch?

ekluzek:
On the first issue, I think the problem is interpreting the $ENV{PBS_NODEFILE} environment variable for the --machinefile option. PBS should set the env variable PBS_NODEFILE, and you likely need to change the script so that the env variable is used correctly in the final script, and not interpreted too early.

ekluzek:
On the second issue of using openmpi, set the environment variable MPILIB to openmpi before you run the build script. You also need to explicitly set COMPILER as well when you do that.

jedwards4b:
PBS_NODEFILE variable should be set by PBS when you submit the job.  If you are not submitting the job and are only running on the login node, you don't need that argument.

slevisconsulting:
> PBS_NODEFILE variable should be set by PBS when you submit the job. If you are not submitting the job and are only running on the login node, you don't need that argument.

@jedwards4b I'm submitting the job on izumi from this directory:
`/fs/cgd/data0/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf`
with this jobscript:
`mksurfdata_jobscript_single`
Your insights are welcome. I'm happy to meet if that helps.

jedwards4b:
I think that maybe you need to source the .env_mach_specific.sh before you submit the script instead of inside the script.

ekluzek:
@slevisconsulting I think it's the syntax you are using in the script for the ENV variable. The $EVN{} syntax is used for python, for the bash script you need to just use the env variable name $PBS_NODEFILE.

slevisconsulting:
Thank you both.

@jedwards4b I tried sourcing like this:  `source bld/.env_mach_specific.sh` (is this how you meant?)
but my terminal froze with this message "badly placed ()"

@ekluzek your correction helped, though it got me to the same error I was getting before with `mpirun`: 
The job begins to write a surfdata...log file and stops in this line:
`rcode = pio_openfile(pio_iosystem, pioid, pio_iotype, trim(datafname), pio_nowrite)`
in subroutine mkurbanInit. Thoughts?

ekluzek:
@slevisconsulting when you are using a C-shell (like tcsh) you source the .env*.csh file, and for a Bourne shell (like Bash) you dot the .env*.sh file.

jedwards4b:
@slevisconsulting my guess is that you are using a csh based shell and not an sh one?
in that case source bld/.env_mach_specific.csh

I tried to poke around in your directory for the other issue but had permission problems.
What is file datafname?

ekluzek:
@slevisconsulting so it's now actually running, so that's a good thing! It's stopping on opening a dataset for Urban. I'd make sure that dataset is available, and readable, and see if you can find anything else on the error. This is where getting it to build in DEBUG mode would be really helpful. I'm not sure that merely getting DEBUG==TRUE is sufficient for the build to add extra debug flags.

slevisconsulting:
The file is present. I can look at it with ncview. I had to load a module for ncdump to work, which gives:
```
> ncdump -k /fs/cgd/csm/inputdata/lnd/clm2/rawdata/mksrf_urban_0.05x0.05_simyr2000.c220127.nc

cdf5
```

Here's the stderr file (nothing helpful to me in there):
```
[0] Obtained 10 stack frames.
[0] ./bld/mksurfdata() [0x1acd7e3]
[0] ./bld/mksurfdata() [0x1acde7b]
[0] ./bld/mksurfdata() [0x1ad6056]
[0] ./bld/mksurfdata() [0x1acc057]
[0] ./bld/mksurfdata() [0x1a5cc07]
[0] ./bld/mksurfdata() [0x4ce534]
[0] ./bld/mksurfdata() [0x4f101c]
[0] ./bld/mksurfdata() [0x420de2]
[0] /lib64/libc.so.6(__libc_start_main+0xf5) [0x7f59fb112505]
[0] ./bld/mksurfdata() [0x420ce9]
[0] [cli_0]: aborting job:
[0] application called MPI_Abort(MPI_COMM_WORLD, -1) - process 0
[240] Obtained 10 stack frames.
[240] ./bld/mksurfdata() [0x1acd7e3]
[240] ./bld/mksurfdata() [0x1acde7b]
[240] ./bld/mksurfdata() [0x1ad6056]
[240] ./bld/mksurfdata() [0x1acc057]
[240] ./bld/mksurfdata() [0x1a5cc07]
[240] ./bld/mksurfdata() [0x4ce534]
[240] ./bld/mksurfdata() [0x4f101c]
[240] ./bld/mksurfdata() [0x420de2]
[240] /lib64/libc.so.6(__libc_start_main+0xf5) [0x7fcb4c427505]
[240] ./bld/mksurfdata() [0x420ce9]
[mpiexec@i024.cgd.ucar.edu] HYDT_bscd_pbs_wait_for_completion (tools/bootstrap/external/pbs_wait.c:67): tm_poll(obit_event) failed with TM error 17002
[mpiexec@i024.cgd.ucar.edu] HYDT_bsci_wait_for_completion (tools/bootstrap/src/bsci_wait.c:23): launcher returned error waiting for completion
[proxy:0:0@i024.cgd.ucar.edu] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:911): assert (!closed) failed
[proxy:0:0@i024.cgd.ucar.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:0@i024.cgd.uca[proxy:0:1@i023.cgd.ucar.edu] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:911): assert (!closed) failed
[proxy:0:1@i023.cgd.ucar.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:1@i023.cgd.uca[proxy:0:2@i022.cgd.ucar.edu] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:911): assert (!closed) failed
[proxy:0:2@i022.cgd.ucar.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:2@i022.cgd.uca[proxy:0:3@i021.cgd.ucar.edu] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:911): assert (!closed) failed
[proxy:0:3@i021.cgd.ucar.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:3@i021.cgd.uca[proxy:0:4@i020.cgd.ucar.edu] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:911): assert (!closed) failed
[proxy:0:4@i020.cgd.ucar.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:4@i020.cgd.uca[proxy:0:5@i019.cgd.ucar.edu] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:911): assert (!closed) failed
[proxy:0:5@i019.cgd.ucar.edu] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:5@i019.cgd.ucar.edu] main (pm/pmiserv/pmip.c:202): demux engine error waiting for event
r.edu] main (pm/pmiserv/pmip.c:202): demux engine error waiting for event
r.edu] main (pm/pmiserv/pmip.c:202): demux engine error waiting for event
r.edu] main (pm/pmiserv/pmip.c:202): demux engine error waiting for event
r.edu] main (pm/pmiserv/pmip.c:202): demux engine error waiting for event
r.edu] main (pm/pmiserv/pmip.c:202): demux engine error waiting for event
Killed by signal 15.
```

slevisconsulting:
I now think the problem on izumi (I tried thorodin, too) is that it doesn't recognize cdf5, while mksurfdat_esmf works only with cdf5.

ekluzek:
Hmmm. That sounds like to me that the NetCDF library was enabled without CDF5 support. Which is possible, but there's two things that make me question that. The first is that when I do a build of mksurfdata_esmf on izumi, and source the .env_mach_specific.csh file the NetCDF module loaded does include NetCDF4/CDF5 according to the nf-config binary for the build.

``` csh
$NETCDF_PATH/bin/nc-config --has-nc4
yes
```

But, the other reason is that we are running the model with NUOPC which does have CDF5 files that it reads and that is working. So if it's failing on reading the CDF5 files, there must be some difference in the mksurfdata_esmf build/run configuration from the CESM configuration.

The differences between the two could be in the Makefile for CESM and/or in the mpirun setup to run.

jedwards4b:
@ekluzek cdf5 is not an hdf5 format,  --has-nc4, does not indicate cdf5 support.   ' --has-cdf5 ' is the correct flag.   @slevisconsulting it is not true that cdf5 is the only format that mksurfdata_esmf can support.   Have you tried netcdf3 classic format?   It's not at all clear to me why the files produced by this tool on izumi cannot be read - does it use pio for output?   If so try setting the pio iotype == 'netcdf'  instead of 'pnetcdf'.

slevisconsulting:
@ekluzek and I tried hardwiring
`pio_iotype = 2` in src/mkurbanparMod.F90
and the code continued to run until 
`Attempting to make Topography statistics`
where it crashes presumably due to insufficient memory.

So we need to decide where to set pio_iotype.
@ekluzek is recommending setting during the build phase and outputting a file that is then read in the namelist generation phase.

slevisconsulting:
> @slevisconsulting can start by updating the build generation phase by setting pio_iotype in a file depending on hostname and updating the namelist generation phase to read that file and put the information in the namelist. The fortran needs to get this additional info from the namelist.

@ekluzek I have implemented step 1, i.e. setting pio_iotype in a file depending on hostname. I'm now questioning whether we need to next read this file into the namelist. Seems simpler to read the file directly into the fortran. Would you be opposed to that?

slevisconsulting:
> @ekluzek and I tried hardwiring `pio_iotype = 2` in src/mkurbanparMod.F90 and the code continued to run until `Attempting to make Topography statistics` where it crashes presumably due to insufficient memory.

I have confirmed that the crash was due to insufficient memory by testing `--res 10x15`. This works because it uses `mksrf_ftopostats_override = '/fs/cgd/csm/inputdata/lnd/clm2/rawdata/surfdata_topo_10x15_c220303.nc`

tools/mksurfdata_esmf/gen_mksurfdata_build.sh
```diff
@@ -7,10 +7,17 @@ case $hostname in
...
+      ;;
+  ##casper
+  casper* )
+      export MACH="casper"
+      pio_iotype=1
       ;;
   ## Other machines
   *)
       export MACH="$hostname"
+      pio_iotype=2
```
slevisconsulting, line 20: @ekluzek recommendation:
Make izumi and hobart explicit (under separate checks).
Keep "other machines" as is and add note that says "we assume that pnetcdf is off; therefore, pio_iotype = 2."

slevisconsulting: clm_pymod testing IN PROGRESS
python testing OK
2 lilac tests fail because I'm working with ...dev090

Also confirm that mksurfdata_esmf still works on casper & izumi before merging PR to mksurfdata branch.

Issue #1717, opened 2022-04-24
Create create_test system test for running mksurfdata_esmf


ekluzek:
The new MPI program mksurfdata_esmf doesn't lend itself well to use the standard serial testing setup for the other tools. Since, all of the other tools are serial, it wouldn't be efficient to have some tests serial and some using multi-processors. So a better way to do it is to take advantage of the batch capability in cime create_test SystemTests and create a python test that builds and runs mksurfdata_esmf. Having at least system tests for mksurfdata_esmf will be critical in being able to bring it to CTSM main-dev. We also should assess adding unit testing for the new mksurfdata_esmf (see https://github.com/ESCOMP/CTSM/issues/842#issuecomment-1098697255).

mvertens:
@ekluzek - I do not think that we want to use any serial testing with the new create_test system. The whole reason for this new capability was to enable memory and performance scalability.  I don't think that the unit testing that is in place for other tools applies here. That said a system test would be helpful at some point. I am happy to discuss this in more detail.

billsacks:
@ekluzek thanks for opening this issue. I think having a system test that runs mksurfdata_esmf and then runs the model with the created dataset would be great – that would show us both that mksurfdata_esmf works and also that you can run the model with the created surface dataset. I think this could follow a pattern similar to what @slevisconsulting and I put together recently to test the fsurdat_modifier tool: https://github.com/ESCOMP/CTSM/blob/master/cime_config/SystemTests/fsurdatmodifyctsm.py – although there may be some extra complexities here due to wanting to run mksurfdata_esmf on multiple nodes (as opposed to the single-processor run of fsurdat_modifier).

mvertens:
@billsacks @ekluzek - I think all of this will become much easier once @jedwards4b and I have had a chance to update the build system for mksurfdata_esmf to be entirely CMake based. I would wait until then to have this implemented.

billsacks:
I think the two pieces (system test and more general build) are likely orthogonal, as long as we only want to run the test on cheyenne (which I think is fine): I think it would be fine to put in place a system test with a fragile build system for now. That said, I'm not personally trying to push one way or the other in terms of prioritization of putting in place this system test.

mvertens:
@billsacks - the reason I suggest waiting for the new build is because the commands to build the component will be different and not use a Makefile in the tools/mksurfdata_esmf/src/  directory. So the test would have to be rewritten with the new build.

slevisconsulting:
[...]

PR #1756, opened 2022-05-17
Add mksurfdata_esmf system test to test-suite

slevisconsulting:
### Description of changes
Confirm that mksurfdata_esmf builds and generates an fsurdat (CTSM surface dataset).
Confirm that the CTSM completes when pointing to the generated fsurdat.

Remove mksurfdata_map test from tools testing (if not done already).

### Specific notes

@billsacks @ekluzek 

#1717 

No, this is a new test.

New test will be added to test-suite.

Testing performed, if any:
Successfully ran a very early draft of the script using the test
`MKSURFDATAESMF_P144x1.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_intel.clm-default`
Generating the f10 surface dataset points to a lower resolution topography raw dataset, thereby bypassing the 1-km topography raw dataset. See corresponding failures below.

**Failures:**
`MKSURFDATAESMF_P144x1.f19_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-default` with mpiexec_mpt -np 144. Stops while working with the 1-km topography raw dataset, so I assume it needs more memory per node.
`MKSURFDATAESMF_P48x3.f19_g17.I1850Clm50BgcCrop.cheyenne_intel.clm-default` with mpiexec_mpt -np 48 (144 returned error). Stops while working with the 1-km topography raw dataset.


tools/mksurfdata_esmf/gen_mksurfdata_build.sh
```diff
@@ -40,7 +44,7 @@ if [ $? != 0 ]; then
   echo "Error doing configure for machine name: $MACH"
   exit 1
 fi
-. .env_mach_specific.sh
+. ./.env_mach_specific.sh
```
ekluzek, line 47: Either way does work here, so this isn't really required. Maybe the ./ form makes it more explicit though.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
```diff
@@ -81,15 +90,26 @@ def main ():
...
         runfile.write('#PBS -j oe \n')
-        runfile.write('#PBS -q regular \n')
         runfile.write('#PBS -l walltime=30:00 \n')
-        runfile.write(f"#PBS -l select={number_of_nodes}:ncpus=36:mpiprocs={tasks_per_node} \n")
+        if machine == 'cheyenne':
+            runfile.write('#PBS -q regular \n')
+            runfile.write(f"#PBS -l select={number_of_nodes}:ncpus=36:mpiprocs={tasks_per_node} \n")
+        elif machine == 'casper':
+            runfile.write('#PBS -q casper \n')
+            runfile.write(f"#PBS -l select={number_of_nodes}:ncpus=12:mpiprocs={tasks_per_node} \n")
```
ekluzek, line 99: You should add a check that only the allowed machine names can be set.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
```diff
@@ -43,6 +43,14 @@ def get_parser():
...
         required=True,
     )
+    parser.add_argument(
+        "--machine",
+        help="""currently this recognizes cheyenne and casper (default cheyenne)""",
+        action="store",
+        dest="machine",
+        required=False,
+        default='cheyenne'
+    )
```
ekluzek, line 53: There must be a way to list the valid options, rather than leaving it open ended.

ekluzek (changes requested):
This is really great to see. I have a few simple comments for small changes.

The other thing I wonder about is adding a "generic" machine option that would add some comments about how you need to change the script to work on a generic machine. So it would add something like this in front of the # PBS commands...

> # Edit the batch directives for your batch system
> # Below are the examples for the batch directives on cheyenne:
> # Edit the following to work on your batch system

And then for the mpiexec line have it setup to use mpirun, and add a comment like this

># Edit the mpirun command to use the MPI executable on your system and the arguments it's requires.

An expansion that we will want for this is to add the ability to also run on izumi. It would be good to know we can easily run on NCAR machines cheyenne and izumi. We could make running on izumi require a custom script as another solution. But, it also doesn't look that hard to extend this to work on izumi as well.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
slevisconsulting: Doing this takes care of your request for a check that only the allowed machine names can be set.

slevisconsulting:
> And then for the mpiexec line have it setup to use mpirun, and add a comment like this
> `# Edit the mpirun command to use the MPI executable on your system and the arguments it's requires.`

@ekluzek 
Are you suggesting that I set it up to use mpirun even for all machines, including cheyenne and casper? Or set it up with mpirun for other machines?

tools/mksurfdata_esmf/gen_mksurfdata_build.sh
billsacks: @ekluzek I asked for this change. It really is needed in some weird cases that have caused me significant time loss in the past. The problem is that you can have an `.env_mach_specific.sh` file somewhere in your $PATH, and if so, that one is used instead of the one in the current directory. That has happened for me because I have `cprnc` in my path, and at least at times, there has been an `.env_mach_specific.sh` file in the same directory as the `cprnc` executable. With a command like `. .env_mach_specific.sh` it picks up the one from the `cprnc` directory instead of the one in the current directory.

Ever since burning a few hours trying to figure out some weird behavior I was getting due to this issue, I have been vigilant about this.

ekluzek: Ahhh, good to know, and that does make sense.

tools/mksurfdata_esmf/gen_mksurfdata_jobscript_single.py
ekluzek: As noted above since you can only pick a valid choice, the else isn't needed here.

slevisconsulting:
@billsacks and @ekluzek 
I think this work is at a good stopping point, so I'm requesting your reviews.

cime_config/SystemTests/mksurfdataesmf.py
```diff
@@ -0,0 +1,96 @@
...
+        Build executable that will generate fsurdat
+        Generate namelist for generating fsurdat
+        SKIP Generate jobscript that runs executable
+        Modify user_nl_clm to point to the generated fsurdat
+        """
+        # These steps need only happen once
+        if not os.path.exists(os.path.join(self._get_caseroot(),
+            'done_MKSURFDATAESMF_setup.txt')):
+            # Paths and strings
+            self._rm_bld_dir = f"rm -rf {self._tool_path}/bld"
```
billsacks, line 50: It's not ideal that this needs to mess with directories in the CTSM sandbox. Some issues I see with that are:
- Trying to run multiple tests of this test type could cause failures due to race conditions: the tests could stomp on each other's build directory
- You could have problems if you're running the testing from a place where you also want to actually use the mksurfdata build for some real work at the same time
- You can't run tests from a directory where you don't have write permission (may not be an issue in practice)

I think the solution to this is to run the cmake and make commands from some other location that is inside the test directory rather than being under the ctsm root. If I remember correctly, the way cmake works is that you run cmake from your desired build location, pointing to the source directory. So you could create a build directory inside the test directory then run `cmake /path/to/ctsmroot/tools/mksurfdata_esmf/src`. For this to work, the paths in gen_mksurfdata_build would need to be generalized so that it could be run from anywhere. Or maybe there's some other way to accomplish this goal.

I don't feel this absolutely needs to be done immediately, but it should be done at some point.

billsacks (commented):
I haven't done a super-careful review of this, but I feel like I have reviewed it enough. It looks great! Thanks a lot for your work on this, Sam. I do have two inline comments, which I don't feel absolutely need to be addressed, but would be good to do – especially the more substantive one about the location of the build directory, though I realize that could involve some significant changes to the current build script.

One other thing that I try to address whenever I'm adding a new test: Did you test some failures at different points (e.g., build failure of mksurfdata_esmf, runtime failure of mksurfdata_esmf, runtime failure of the model) to ensure errors are reported correctly in those failure cases?

cime_config/SystemTests/mksurfdataesmf.py
```diff
@@ -0,0 +1,96 @@
...
+        self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
+
+    def run_phase(self):
+        """
+        Run executable to generate fsurdat
+        Submit CTSM run that uses fsurdat just generated
+        """
+        # Paths and command strings
+        self._executable_path = os.path.join(self._tool_path, 'bld/mksurfdata')
+        _machine = self._case.get_value("MACH")
```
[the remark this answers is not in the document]
slevisconsulting: I think I got all of them (and thank you for the reminder).

cime_config/SystemTests/mksurfdataesmf.py
slevisconsulting: As a temporary solution, the test now deletes the /tools/mksurfdata_esmf/bld directory when it's done needing it. This reduces the likelihood of the above risks without eliminating it.

To resolve this, I would like to discuss it in a meeting (me and @ekluzek ?)

slevisconsulting:
> Did you test some failures at different points (e.g., build failure of mksurfdata_esmf, runtime failure of mksurfdata_esmf, runtime failure of the model) to ensure errors are reported correctly in those failure cases?

Good point. While developing the test, I had to look at TestStatus.log for useful error messages. So I'm adding guidance along those lines when any of the subprocess commands fails. However, I'm leaving error messages unchanged for runtime failure of the model, with the assumption that such messages are standard across these tests.

cime_config/SystemTests/mksurfdataesmf.py
billsacks: In my view, the temporary solution could make things worse in a number of situations: where some other process is expecting the build to be there, and then it gets removed out from underneath. So I'd say what you had before this was probably a bit better... but really it's hard to avoid problems in certain situations until you move to a build in a test-specific location.

billsacks:
> Good point. While developing the test, I had to look at TestStatus.log for useful error messages. So I'm adding guidance along those lines when any of the subprocess commands fails. However, I'm leaving error messages unchanged for runtime failure of the model, with the assumption that such messages are standard across these tests.

Sounds good. Needing to look in TestStatus.log is reasonable. The most important thing is that a failure at any stage results in a FAIL result somewhere in the TestStatus file. This should be true both for the initial run of the test and a rerun. Ideally, a failure in the build or run of mksurfdata_esmf would cause the test to abort with a FAIL before it gets into the model run. Similarly, if you run the test once and it passes, and then you make a change to mksurfdata_esmf and rerun the existing test, but now there is a build or runtime failure in mksurfdata_esmf, the test should also report a FAIL result in the appropriate phase. We want to **avoid**, for example, the possibility that in a rerun, mksurfdata_esmf fails but there was already an output file in place so the test happily continues with the previously-generated output file.

slevisconsulting:
> Ideally, a failure in the build or run of mksurfdata_esmf would cause the test to abort with a FAIL before it gets into the model run.

As long as `subprocess` returns a non-zero when it fails, then we're good. I hope that nothing would cause it to return a false zero. However, I see how your next scenario could happen:

> Similarly, if you run the test once and it passes, and then you make a change to mksurfdata_esmf and rerun the existing test, but now there is a build or runtime failure in mksurfdata_esmf, the test should also report a FAIL result in the appropriate phase. We want to **avoid**, for example, the possibility that in a rerun, mksurfdata_esmf fails but there was already an output file in place so the test happily continues with the previously-generated output file.

I think this could happen in a rerun of the same test due to the presence of `done_MKSURFDATAESMF_setup.txt`, which ensures that most of the build_phase function occurs once. So I could remove `done_MKSURFDATAESMF_setup.txt` at the beginning of the run_phase to help avoid such scenario.

cime_config/SystemTests/mksurfdataesmf.py
slevisconsulting: Ok, I reverted to what I had before this.

billsacks:
> As long as `subprocess` returns a non-zero when it fails, then we're good.

Right, that's the main thing I wanted to make sure you have verified: I think there are ways that the subprocess call could be written wrong that would let things continue despite a failure. In addition, I think there are ways that a runtime abort could be done in the wrong way from the mksurfdata code, making it look to the subprocess call that it finished successfully when it really did not. As long as you have done some basic checking that a build or runtime failure in mksurfdata_map actually causes the test to stop at the appropriate point and put a FAIL in the TestStatus, then we should be good.

Mainly I'm thinking about the general rule: Testing that a test works correctly is kind of the opposite from testing that the production code works correctly: To test production code, you run it and make sure everything correctly runs to completion. But test code is all about picking up failures. So to test that test code is working correctly, you should inject artificial failures into the code under test and make sure that the test code actually flags those correctly. (Sorry to spend so much time belaboring this point, maybe unnecessarily: it comes from having seen a number of instances in the past with CESM testing where I discovered that the tests were not actually detecting failures: people had seen that the tests were working in the happy path so assumed that things were good, but actually the tests weren't doing their job of picking up problems.)

> I think this could happen in a rerun of the same test due to the presence of `done_MKSURFDATAESMF_setup.txt`, which ensures that most of the build_phase function occurs once. So I could remove `done_MKSURFDATAESMF_setup.txt` at the beginning of the run_phase to help avoid such scenario.

That's not what I was thinking about, but it is a very good point. The scenario I was thinking about was connected to the first point: if a failure condition from the mksurfdata run wasn't being sent / caught correctly, then it could fail upon a rerun but the test would happily continue, using the surface dataset that was previously generated. As long as failures in various stages of the mksurfdata build and run are caught correctly, then this shouldn't be a problem.

But your point is a very good one. I think there is too much done within the conditional on `done_MKSURFDATAESMF_setup.txt` here. Really, I think we want the bare minimum to be within that conditional. The only things there should be things that are problematic to do multiple times – or things that really don't need to be done upon rebuild / rerun and result in clutter if you do them multiple times: an example would be appending something to a user_nl file. I don't think you want to remove the `done` file, because that will cause these should-only-be-done-once things to be redone if you rerun the test. Instead, I think what you should do is:
- Take some things outside of this conditional, so they are done upon a rerun. At a glance, I think the only thing that should be in that conditional is the call to _modify_user_nl. If that's the case, it might end up being cleaner if you move that call – and the management of the `done` file – into the `__init__` method, though it doesn't really matter.
- For things taken outside of this conditional that are still in the build call: add a conditional so that these are only done when sharedlib_only=True – or only done when model_only=True; I don't think it matters which one you pick, but you only want it to be done for one of them. This is because I think build_phase is called twice: once with sharedlib_only=True and once with model_only=True. We don't want to build mksurfdata_esmf twice, just once, so just build it during one of those phases.

Together, what these changes will accomplish – I *think* – is that mksurfdata_esmf will get rebuilt when you rerun ./case.build, but still will only get built once for each run of ./case.build.

slevisconsulting:
> The scenario I was thinking about was connected to the first point: if a failure condition from the mksurfdata run wasn't being sent / caught correctly, then it could fail upon a rerun but the test would happily continue, using the surface dataset that was previously generated. As long as failures in various stages of the mksurfdata build and run are caught correctly, then this shouldn't be a problem.

I added code that preemptively removes preexisting .namelist and .nc files.

> [...] what these changes will accomplish – I _think_ – is that mksurfdata_esmf will get rebuilt when you rerun ./case.build, but still will only get built once for each run of ./case.build.

Made those changes, as well.

billsacks:
Thanks @slevisconsulting !

billsacks:
I skimmed back through my earlier comments. Everything is addressed except for the comments on trying to do the build within the test directory if possible.

slevisconsulting:
> The other thing I wonder about is adding a "generic" machine option that would add some comments about how you need to change the script to work on a generic machine. So it would add something like this in front of the # PBS commands...

@ekluzek per our conversation:
I included the guidance comments that you recommended and will not pursue implementation of a "generic" machine option for now.

slevisconsulting:
Testing performed:
`MKSURFDATAESMF_P144x1.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_intel.clm-default`  PASS
`make utest`  PASS
`make stest`  two lilac tests fail with "Current machine cheyenne does not match case machine ctsm-build"
...but this is expected for `git describe` --> `alpha-ctsm5.2.mksrf.03_ctsm5.1.dev090-26-g8b720987b` as explained in #1739 which is resolved in later versions.

@ekluzek pls let me know if you would like to review this PR in a meeting.

ekluzek:
I was looking this over and I think the best way to keep this updated and get this to work on izumi, will be to use the env_mach_specific.xml file to get what the mpirun command needs to look like. It looks like the mpirun command (from looking at it for a CTSM case is fairly complex and it depends on the MPI library being used.

So for mvapich2

`mpiexec --machinefile $ENV{PBS_NODEFILE} -n <ntasks> --prepend-rank`

and for openmpi

`mpiexec -n <ntasks> --tag-output`

The mpirun can be complex and changing with CESM, so using CESM to manage it might be the best way to go. That also provides a general solution that will work for any machine, compiler, and mpilib combination that cime is ported to.

ekluzek:
@slevisconsulting and I looked this over a bit and tried to get it working and run into some roadblocks. One thing that I think will be important to get working properly is the ability to turn DEBUG compiler options on in the build. We think that just requires the env variable DEBUG to be set before the configure line, but I'm suspicious that because this is a CMAKE build that there's more to it than that.

ekluzek:
Oh, I think I put my comments in the wrong PR...

slevisconsulting:
> @ekluzek pls let me know if you would like to review this PR in a meeting.

@ekluzek I wasn't pushing for wrapping up this PR and #1748 because of large diffs that I was finding in these files generated by mksurfdata_esmf:
`surfdata_0.9x1.25_SSP5-8.5_78pfts_CMIP6_1850-2100_c220525.nc`
`landuse.timeseries_0.9x1.25_SSP5-8.5_78_CMIP6_1850-2100_c220525.nc`
**The good news is that I have now tracked down the diffs, and they are fully expected.** In particular, the large diffs were against `release-clm5.0.18`, while the diffs are very small when I compare against the files in `/glade/work/oleson/ctsm_dynurb_mksurf_ctsm5.2.mksurfdata_I1564/tools/mksurfdata_map` because Keith generated these files with the same code mods that I used from #1586 and #1587 

So I would like to spend a few minutes going over this PR and #1748 so that we may finalize them.

cime_config/SystemTests/mksurfdataesmf.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def run_phase(self):
+        """
+        Run executable to generate fsurdat
+        Submit CTSM run that uses fsurdat just generated
+        """
+        # Paths and command strings
+        executable_path = os.path.join(self._tool_bld, 'mksurfdata')
+        machine = self._case.get_value("MACH")
+        if machine == 'cheyenne':
```
ekluzek, line 101: You should be able to refactor this a bit to rely on the gen_mksurfdata_jobscript_single.py script to produce the commands that need to be run rather than hardcoding the machines that can be used here. That would be a nice improvement, but this could also be a future change. The advantage is that it removes embedding of machine specific things into this script, and puts all the machine specific run into one place.

ekluzek (commented):
There's a couple things that I see that could be done. But, it's nice to get a working system test in place, so they could be done later. So let's chat about this and decide what should be done.

cime_config/SystemTests/mksurfdataesmf.py
slevisconsulting: ...so run `gen_mksurfdata_jobscript_single.py` and then run the jobscript without qsub in front.
...and most general approach is to pick up the mpi command from `env_mach_specific.xml`

slevisconsulting:
python testing OK
Two lilac sys tests fail due to this:
```
> git describe
alpha-ctsm5.2.mksrf.03_ctsm5.1.dev090-33-gde3fb92b6
```

pymod testing PASS

I plan on merging this PR to the ctsm5.2.mksurfdata branch tomorrow.

PR #1765, opened 2022-05-20
Fix missing hostname/CSMDATA in mkmapdata.sh

samsrabin:
### Description of changes
Update `mkmapdata.sh` to handle unset `hostname` and `CSMDATA`.

### Specific notes
I had not done `export CSMDATA=...` before calling `mkmapdata.sh`. The default `CSMDATA` for Cheyenne failed to be used, so it was empty, causing this script to fail with lots and lots of "how to use this script" output. Setting `set -e` at the top of the script helped narrow down the source of the problem. It turns out `hostname` was defined too late in the script.

Also added some helpful error messages for unrecognized `hostname` and not-found CSMDATA.

----

Testing performed, if any: This fixed the call of `mkmapdata.sh` for me.


ekluzek:
@samsrabin since you have this we might as well bring it in with another CTSM tag that we make. However, I did want to make sure you knew that mkmapdata.sh is going to go away when ctsm5.2 tags start to come in. So in a few months mkmapdata will be gone. I just want to make sure you are aware that this is a tool that's near it's end of life.

ekluzek (commented):
These changes look good! Thanks for sharing the fix you found.

samsrabin:
Sure, makes sense. Is there something else you recommend I go ahead and switch to, or should I wait for it to be officially replaced?

ekluzek:
Good question. You probably should stay with what's on the main dev (master branch) for now until ctsm5.2 comes out.

The ctsm5.2.mksurfdata branch is where the new tools are being developed. All new datasets to create surface datasets are going to go into that branch. So if you have new datasets that need to be part of the mksurfdata process we should work with you to get it into that branch. So let me know if that's the case. We are making tags along the branch, so it is easy get specific versions along the branch.

However, if you are just creating surface datasets that you want to work in the current model and you want them to be consistent with the existing datasets -- use main dev. Since, ctsm5.2.mksurfdata is going to be changing all the time until it's completed, you won't get anything consistent until the end, and it's not guaranteed to work with the model outside of the branch either. So for work in ctsm5.1 you want to use main dev.

samsrabin:
Okay, I'll email you.

ekluzek (approved):

PR #1770, opened 2022-05-27
add checkimport to clm for scam case

jedwards4b:
### Description of changes
Provide CheckImport function from nuopc library so that when in single column mode with no land points at play, this check can be bypassed. 

### Specific notes

Contributors other than yourself, if any: Rocky Dunlap

Testing performed, if any: will do standard regular testing


ekluzek:
@jedwards4b I tried our test cases for both a Land and Ocean point to make sure this works, and it does. But, I think for the new code to be exercised it needs a CMEPS update to it correct? Do we have a CMEPS branch where this would get invoked?

ekluzek (approved):

jedwards4b:
@ekluzek this is a standalone PR - no dependency on cmeps.

ekluzek:
Thanks @jedwards4b. 

By the way do you have any idea why this works fine for a CLM I case that runs for an ocean point, but it fails when you run with SCAM which uses CAM? I have a test that runs single-column mode (with or without this change) for an ocean point, that I was hoping would catch such problems. It's annoying that test didn't catch the problem in SCAM. It still might catch other problems so I'll keep it around, but I had hoped it would have seen this problem.

jedwards4b:
I think that it has to do with the flux fields sent from lnd to atm - with datm you aren't sending any.  When you run with scam atm is asking for fields from lnd but since you are on an ocn point ctsm isn't running.   With this change the esmf code that checks to make sure the fields are there is bypassed.   Another alternative would be to allow the lnd to run by removing the return at line 757 but that would introduce other issues to consider.

ekluzek:
I had hoped that adding FORCE_PROGNOSTIC_FOR_DATAMODELS=TRUE, might allow out pftsROA test to catch the problem, and unfortunately it doesn't. But, including it might catch other issues that would be unseen without it, so I'm adding that in.

PR #1787, opened 2022-06-24
Adding excess ground ice to CTSM 

### LINKED ISSUE #1229: adding a new feature to CTSM: excess ground ice physics
**Purpose with this issue:** 
We propose to add excess ground ice physics into the ctsm5.1.dev014 for better solving soil physics and hydrological impacts associated with the thawing process in the permafrost regions. This will imply changes in the following 8 files: 

- bld/namelist_files/namelist_definition_ctsm.xml
- src/biogeophys/SoilHydrologyMod.F90
- src/biogeophys/SoilTemperatureMod.F90
- src/biogeophys/TemperatureType.F90
- src/biogeophys/TotalWaterAndHeatMod.F90
- src/biogeophys/WaterStateType.F90
- src/main/clm_varctl.F90
- src/main/controlMod.F90

**Brief science information:**
The current CTSM does not include excess ground ice for the permafrost region, which inevitably leads to biases in calculating the soil physics and hydrological change along with the permafrost degradataion process (Burn and Nelson, 2006). We propose to add a feature of excess ice physics, which was initially developed based on CLM4.5 (Lee et al., 2014). Compared to the original development, we suggest to add a namelist entry and switches in the code so ctsm users could to turn on and off the excess ice physics easily. 

**Pseudocode for all file changes:**
- _bld/namelist_files/namelist_definition_ctsm.xml:_  adding an namelist entry in order to turn on and off excess ice physics upon users need. 
- _src/biogeophys/SoilHydrologyMod.F90:_ involving the added excess ice into the calculation of total volume of soil ice. 
- _src/biogeophys/SoilTemperatureMod.F90_: calculate the melt of excess ice and its impact to soil temperature and extra soil water. 
- _src/biogeophys/TemperatureType.F90:_ resetting the initial soil temperature to a freezing temperature.
- _src/biogeophys/TotalWaterAndHeatMod.F90:_ involving the added excess ice into the calculation of total mass of soil ice.
- _src/biogeophys/WaterStateType.F90:_ setting the initial condition of excess ice by reading variables from surface data (surface data file is revised to include excess ice initial condition variables). 
- _src/main/clm_varctl.F90:_ adding a switch to turn on and off excess ice, the default value of which is “false”. 
- _src/main/controlMod.F90:_ adding a flag to read the excess ice physics namelist entry. 

**Long term goal:**
 To add the suggested feature to the master branch of CTSM. 

**People involved:** 
@lca041, Hanna Lee (hale@norceresearch.no), @kjetilaas, @sunnivin, @dlawrenncar, @swensosc. ​

### LINKED ISSUE #1818: Code is repeated both inside if and in its else statement
The calculation of xmf in SoilTemperatureMod has an if statement where the if and else both contain the same expression. Thus the if statement should just be removed.

@mvdebolskiy noticed this.

```
                     if (j >= 1) then
                        xmf(c) = xmf(c) + hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime
                     else
                        xmf(c) = xmf(c) + hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime
                     endif

```

It looks like this goes back to this commit:

```
commit 248307d5bd096a23ff6f85efaf255c0254e0e348
Author: Keith Oleson <oleson@ucar.edu>
Date:   Tue Feb 27 09:25:33 2018 -0700

    Bug fixes for energy imbalance associated with surface water and lakes

```

Previous to that there was a difference and the else statement included a bit about frac_sno_eff
and looked like this...

`xmf(c) = xmf(c) + frac_sno_eff(c)*hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime`

From the commit log this difference was causing an energy imbalance.

mvdebolskiy:
**Description of changes**

Add parameterization to allow excess ice in soil and subsidence

Specific notes

Parameterization for excess ice is described in [Lee et al., 2014](http://dx.doi.org/10.1088/1748-9326/9/12/124006).

This code is a modified version of [code provided by Lei Cai](https://github.com/lca041/ctsm/tree/clm5.0.dev92_exice).

Works only for the `nuopc` driver.

Files changed:
_bld/CLMBuildNamelist.pm_, _bld/namelist_files/namelist_defaults_ctsm.xml_, _bld/namelist_files/namelist_definitionss_ctsm.xml_ -- added namelist options;
_src/main/clm_varctl.F90, _src/main/controlMod.F90_ -- added option to switch excess ice physics and read namelist option;
_src/biogeophys/WaterStateType.F90_ -- added prognostic excess ice variable and a history field;
_src/biogeophys/WaterStateBulkType.F90_, src/main/clm_instMod.F90_,  -- added arguments to soubrutiens for proper initialization;
_src/biogeophys/TemperatureType.F90_ -- initial soil temperature set to 268.15 K at the cold start (might be redundant if https://github.com/ESCOMP/CTSM/issues/1460 is added)
_src/biogeophys/WaterDiagnosticBulkType.F90_ -- added two diagnostic excess ice variables and two history fields;
_src/biogeophys/SoilTemperatureMod.F90_ -- added main excess ice calculations;
_src/biogeophys/TotalWaterAndHeatMod.F90_ -- added excess ice to total water for balance checks;
_src/biogeophys/SoilHydrologyMod.F90_ -- added excess ice for ice fraction calculation;

New files:
_src/cpl/share_esmf/ExcessIceStreamType.F90_ --  routines to read dataset with initial excess ice concentration

**CTSM Issues Fixed** (include github issue #):
   Fixes #1229 
   Fixes #1818

**Are answers expected to change (and if so in what way)?**

Yes, when excess ice parameterization is used.

New namelist options added:  
`use_excess_ice` (logical, in `clm_inparm`) default  - `.false.`; turns on excess ice physics
`stream_meshfile_exice`, `stream_fldfilename_exice` and `stream_mapalgo_exice` (char, in `exice_streams`) defaults - `lnd/clm2/paramdata/exice_init_0.125x0.125_ESMFmesh_c20220516.nc`, `lnd/clm2/paramdata/exice_init_0.125x0.125_c20220516.nc` and`bilinear`; meshfile, stream file and spatial interpolation algorithm for initial values of excess ice. Dataset is interpolated to 0.125x0.125 degrees grid from [Brown et al., 1997](https://pubs.er.usgs.gov/publication/cp45), can be found [here](https://drive.google.com/file/d/1mA457Oa52zG_MtLGB7KHuUYQvsS2-P5o/view?usp=sharing). Dataset is only used during the cold start or for hybrid runs (or start from finidat) that do not have excess ice. 

On sigma2 machines: comparisons with the _upstream/master_ version and switched-off excess ice physics give the same results for soil state variables.  More information and figures on changes and testing can be found [here](https://hackmd.io/@mvdebolskiy/BJj4M88Oc)

@ka7eh


bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -2586,4 +2586,14 @@ lnd/clm2/surfdata_map/release-clm5.0.30/surfdata_ne0np4.CONUS.ne30x8_hist_78pfts
 
 <init_interp_method>general</init_interp_method>
 
+<!-- =========================================  -->
+<!-- Defaults for excess_ice                    -->
+<!-- =========================================  -->
+
+<use_excess_ice>.false.</use_excess_ice>
+
+<stream_fldfilename_exice use_excess_ice=".true.">lnd/clm2/paramdata/exice_init_0.125x0.125_c20220516.nc</stream_fldfilename_exice>
```
ekluzek, line 2595: This isn't on cheyenne yet. We should get a copy to look at. Could you copy it to somewhere on cheyenne and let us know where it's at so we can look at it?

You also may not have permission to put under /glade/p/cesmdata/cseg/inputdata on cheyenne. We should discuss if you given permission to do so. If you won't be doing this often we won't need to. If you are going to be doing so, we'll need to do some training with you before we give you the permissions.

ekluzek: I'm also not sure if the paramdata subdirectory is the right location for this, but we need to look at it to make that assessment.

bld/CLMBuildNamelist.pm
```diff
@@ -4137,6 +4143,24 @@ sub setup_logic_fates {
...
+  #
+  # excess ice streams
+  #
+  my ($opts, $nl_flags, $definition, $defaults, $nl) = @_;
+  my $use_exice = $nl->get_value( 'use_excess_ice'               );
+  if (defined($use_exice) && $opts->{'driver'} eq "nuopc" && value_is_true($use_exice)) {
+    add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_meshfile_exice');
+    add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_fldfilename_exice');
+    add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_mapalgo_exice');
+  } elsif (defined($use_exice) && (! $opts->{'driver'} eq "nuopc") && value_is_true($use_exice)) {
```
ekluzek, line 4157: I like the test for nuopc driver here. We should add this check to the namelist testing. I can help with that.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1152,9 +1205,14 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
                imelt(c,j) = 0
                hm(c,j) = 0._r8
                xm(c,j) = 0._r8
+               xm2(c,j)=0._r8
                wice0(c,j) = h2osoi_ice(c,j)
                wliq0(c,j) = h2osoi_liq(c,j)
-               wmass0(c,j) = h2osoi_ice(c,j) + h2osoi_liq(c,j)
+               wexice0(c,j)=excess_ice(c,j)
+               wmass0(c,j) = h2osoi_ice(c,j) + h2osoi_liq(c,j) + wexice0(c,j)
+              if (j>=1) then
```
ekluzek, line 1213: Indentation issue.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1211,6 +1269,13 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
                   t_soisno(c,j) = tfrz
                endif
 
+               ! melt excess ice after normal ice
+               if (excess_ice(c,j) > 0._r8 .AND. t_soisno(c,j) > tfrz) then
+                imelt(c,j) = 1
```
ekluzek, line 1274: Indent at least 2 spaces to make it more clear. We are inconsistent with our indentation, but 2 or 3 spaces is typical, and sufficient to make the indentation noticeable.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1325,23 +1390,31 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
...
+                     if (xm(c,j) > 0._r8) then !if there is excess heat to melt the ice
+                       h2osoi_ice(c,j) = max(0._r8, wice0(c,j)-xm(c,j))
+                       heatr = hm(c,j) - hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime
+                       xm2(c,j) = xm(c,j) - h2osoi_ice(c,j) !excess ice melting
+                         if (h2osoi_ice(c,j) == 0._r8) then ! this might be redundant 
+                           if (excess_ice(c,j) >= 0._r8 .and. xm2(c,j)>0._r8 .and. j>=2) then ! if there is excess ice to melt
+                             excess_ice(c,j) = max(0._r8,wexice0(c,j) - xm2(c,j))
+                             heatr = hm(c,j) - hfus * (wexice0(c,j)-excess_ice(c,j)+wice0(c,j)-h2osoi_ice(c,j)) / dtime
+                           endif
+                         endif !end of excess ice block
```
ekluzek, line 1402: indentation issue on the close endif

src/biogeophys/TemperatureType.F90
```diff
@@ -647,11 +647,11 @@ subroutine InitCold(this, bounds, &
     ! !USES:
     use shr_kind_mod   , only : r8 => shr_kind_r8
     use shr_const_mod  , only : SHR_CONST_TKFRZ
-    use clm_varcon     , only : denice, denh2o, sb
-    use landunit_varcon, only : istwet, istsoil, istdlak, istice
+    use clm_varcon     , only : denice, denh2o, sb !what's sb? how it's never used here
```
ekluzek, line 650: If sb isn't used here, we should remove it. If it is used, it should fail in the build so you should see quickly whether you can do that or not. Thanks for pointing this out, by the way.

src/biogeophys/WaterDiagnosticBulkType.F90
```diff
@@ -277,8 +286,28 @@ subroutine InitBulkHistory(this, bounds)
          fname=this%info%fname('TOTSOILICE'),  &
          units='kg/m2', &
          avgflag='A', &
-         long_name=this%info%lname('vertically summed soil cie (veg landunits only)'), &
+         long_name=this%info%lname('vertically summed soil ice (veg landunits only)'), &
          ptr_col=this%h2osoi_ice_tot_col, l2g_scale_type='veg')
+    ! excess ice vars
```
ekluzek, line 291: Indentation looks wrong, but maybe was wrong previously?

src/biogeophys/WaterDiagnosticBulkType.F90
```diff
@@ -277,8 +286,28 @@ subroutine InitBulkHistory(this, bounds)
...
+    if (use_excess_ice) then
+      this%exice_vol_tot_col(begc:endc) = 0.0_r8
+      call hist_addfld1d ( &
+           fname=this%info%fname('TOTEXICE_VOL'),  &
+           units='m3/m3',  &
+           avgflag='A', &
+           l2g_scale_type='veg', &
+           long_name=this%info%lname('vertically averaged volumetric excess ice concentration (veg landunits only)'), &
+           ptr_col=this%exice_vol_tot_col)
```
ekluzek, line 300: Should both of these output by default anytime excess ice is turned on? They could be set to "inactive" and then turned on in the namelist when they are desired. We output so much that it's good to be careful on the default output we add.

src/biogeophys/WaterStateType.F90
```diff
@@ -150,6 +157,14 @@ subroutine InitAllocate(this, bounds, tracer_vars)
     call AllocateVar1d(var = this%dynbal_baseline_ice_col, name = 'dynbal_baseline_ice_col', &
          container = tracer_vars, &
          bounds = bounds, subgrid_level = subgrid_level_column)
+    !excess ice vars
```
ekluzek, line 160: Indent comment to same level as code.

src/biogeophys/WaterStateType.F90
```diff
@@ -268,6 +283,15 @@ subroutine InitHistory(this, bounds, use_aquifer_layer)
...
     end if
 
+      ! Add excess ice fields to history
+
+      if (use_excess_ice) then
+        data2dptr => this%excess_ice_col(begc:endc,1:nlevsoi)
+        call hist_addfld2d (fname='EXCESS_ICE',  units='kg/m2', type2d='levsoi', &
+             avgflag='A', long_name='excess soil ice (vegetated landunits only)', &
+             ptr_col=this%excess_ice_col, l2g_scale_type='veg')
+       end if
```
ekluzek, line 293: Looks like an indentation problem at the "end if"

src/biogeophys/WaterType.F90
```diff
@@ -300,14 +301,22 @@ subroutine DoInit(this, bounds, &
     real(r8)         , intent(in) :: watsat_col(bounds%begc:, 1:)            ! volumetric soil water at saturation (porosity)
     real(r8)         , intent(in) :: t_soisno_col(bounds%begc:, -nlevsno+1:) ! col soil temperature (Kelvin)
     logical          , intent(in) :: use_aquifer_layer ! whether an aquifer layer is used in this run
+    character(len=*) , intent(in) , optional    :: NLFilename ! Namelist filename
```
ekluzek, line 304: Make this a required argument, it'll simplify the code a bit.

src/biogeophys/WaterType.F90
```diff
@@ -300,14 +301,22 @@ subroutine DoInit(this, bounds, &
...
     !
     ! !LOCAL VARIABLES:
     integer :: begc, endc
     integer :: i
+    character(len=256) :: l_NLFilename ! local for namelist filename
 
     character(len=*), parameter :: subname = 'DoInit'
     !-----------------------------------------------------------------------
 
+    !check if NLFilename is passed
```
ekluzek, line 314: This can go away when you make NLFilename a required argument.

src/biogeophys/WaterType.F90
ekluzek: To do this the call to DoInit in the unit-test subroutine, just needs to send NLFilename=" " and pass that down.

src/cpl/share_esmf/ExcessIceStreamType.F90
```diff
@@ -0,0 +1,295 @@
+module ExcessIceStreamType
+
+#include "shr_assert.h"
+
+  !-----------------------------------------------------------------------
+  ! !DESCRIPTION:
+  ! Contains methods for reading in excess ice initial bulk values from data stream
```
ekluzek, line 7: Could you add a little more explanation of what "excess ice" refers to here? It would also be good to give the references to the papers.

src/cpl/share_esmf/ExcessIceStreamType.F90
```diff
@@ -0,0 +1,295 @@
...
+      stream_taxmode      = 'extend',                                           &
+      stream_dtlimit      = 1.0e30_r8,                                          &
+      stream_tintalgo     = 'nearest',                                          &
+      stream_name         = 'excess ice ',                                      &
+      rc                  = rc)
+      if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) then
+        call ESMF_Finalize(endflag=ESMF_END_ABORT)
+      end if
+      
+      ! Explicitly set current date to a hardcoded constant value. Otherwise
```
ekluzek, line 117: This is taken from my example of ch4FinundationStreamMod. We should assess if this problem still exists, if not it would be good to remove the hardcoding of the date. We'll talk about this more. I should file an issue on this problem. I'm glad you point it out here, I'm just hoping it's no longer a problem.

src/cpl/share_esmf/ExcessIceStreamType.F90
```diff
@@ -0,0 +1,295 @@
...
+  implicit none
+  private
+
+  type, public :: excessicestream_type
+     real(r8), pointer, private :: exice_bulk  (:)         ! excess ice bulk value (-)
+  contains
+
+      ! !PUBLIC MEMBER FUNCTIONS:
+      procedure, public :: Init            ! Initialize and read data in
+      procedure, public :: UseStreams      ! If streams will be used
```
ekluzek, line 29: It looks like UseStreams is actually private, so mark it as such.

src/cpl/share_esmf/ExcessIceStreamType.F90
```diff
@@ -0,0 +1,295 @@
...
+      mcdate = year*10000 + mon*100 + day
+      
+      call shr_strdata_advance(sdat_exice, ymd=mcdate, tod=sec, logunit=iulog, istr='exice', rc=rc) 
+         if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) then
+            call ESMF_Finalize(endflag=ESMF_END_ABORT)
+         end if
+        
+         ! Get pointer for stream data that is time and spatially interpolate to model time and grid
+         do n = 1,size(stream_varnames)
+          call dshr_fldbun_getFldPtr(sdat_exice%pstrm(1)%fldbun_model, stream_varnames(n), fldptr1=dataptr1d, rc=rc)
```
ekluzek, line 134: Indent more than one space.

src/main/clm_varctl.F90
```diff
@@ -298,6 +298,11 @@ module clm_varctl
   real(r8), public :: soil_layerstruct_userdefined(99) = rundef
   integer, public :: soil_layerstruct_userdefined_nlevsoi = iundef
 
+  !----------------------------------------------------------
+  !excess ice physics switch
+  !----------------------------------------------------------
+  logical, public :: use_excess_ice = .false. ! true. => use excess ice physics !MVD
```
ekluzek, line 304: Remove the !MVD comments.

src/main/controlMod.F90
```diff
@@ -242,6 +242,8 @@ subroutine control_init(dtime)
 
     namelist /clm_inparm/ use_soil_moisture_streams
 
+    namelist /clm_inparm/ use_excess_ice !MVD
```
ekluzek, line 245: Remove !MVD

src/main/controlMod.F90
```diff
@@ -733,6 +735,8 @@ subroutine control_spmd()
 
     call mpi_bcast (use_soil_moisture_streams, 1, MPI_LOGICAL, 0, mpicom, ier)
 
+    call mpi_bcast (use_excess_ice, 1, MPI_LOGICAL, 0, mpicom,ier) !MVD
```
ekluzek, line 738: Remove !MVD

src/main/controlMod.F90
```diff
@@ -866,6 +870,7 @@ subroutine control_print ()
     write(iulog,*) '    use_nitrif_denitrif = ', use_nitrif_denitrif
     write(iulog,*) '    use_extralakelayers = ', use_extralakelayers
     write(iulog,*) '    use_vichydro = ', use_vichydro
+    write(iulog,*) '    use_excess_ice = ', use_excess_ice !MVD to check if the value is read from namelists
```
ekluzek, line 873: Remove !MVD

src/cpl/share_esmf/ExcessIceStreamType.F90
```diff
@@ -0,0 +1,295 @@
...
+   integer  :: begc, endc
+   integer  :: begg, endg
+   integer  :: c, l, g  !counters
+   
+   exice_bulk_init(bounds%begc:bounds%endc)=0.0_r8
+
+   do c = bounds%begc,bounds%endc
+      g = col%gridcell(c)
+      l = col%landunit(c)
+      if ((.not. lun%lakpoi(l)) .and. (.not. lun%urbpoi(l)) .and. (.not. lun%itype(l) == istwet) .and. (.not. lun%itype(l) == istice)) then  !not lake
```
ekluzek, line 200: Rather than use an explicit list, we should possibly use a filter here.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1372,12 +1445,14 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
                         end if
                      endif  ! end of heatr > 0 if-block
 
-                     if (j >= 1) then
-                        xmf(c) = xmf(c) + hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime
+                     if (j >= 1) then !why before it was two same statements?
```
ekluzek, line 1448: I'm not sure what your question is here? Can you elaborate? I want to make sure we address any issues that you find the code confusing.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -654,7 +700,8 @@ subroutine SoilThermProp (bounds, num_urbanc, filter_urbanc, num_nolakec, filter
                   col%itype(c) /= icol_roof .and. col%itype(c) /= icol_road_imperv) .or. &
                   (col%itype(c) == icol_road_imperv .and. j > nlev_improad(l))) then
 
-                  satw = (h2osoi_liq(c,j)/denh2o + h2osoi_ice(c,j)/denice)/(dz(c,j)*watsat(c,j))
+                    ! TODO recalculate watsat and satw to have excess ice included
```
ekluzek, line 703: We should make sure we track this TODO item.

ekluzek (changes requested):
This looks really good and it's exciting to see this advancement come into the model.

I have some detailed requests for changes, most are pretty minor. There's some issues with indentation that would be good to straighten out, as it really helps with readability. I suggest indenting at least two spaces. Here's our style guide on indentation:

https://wiki.ucar.edu/display/ccsm/CLM+Coding+Conventions#CLMCodingConventions-Indentationindent

There's a few places where you have a !MVD comment to show your changes in the code. These need to be removed in final versions, as having these things scattered in the final code make it confusing and hard to read.

There's a hard coded date from my example in ch4FinundatedStreams, we should assess if this is still needed. We'll talk more about that.

I also ask you to put a copy of the stream files on cheyenne so we can look at them. We'll copy the files to the standard location and get it checked into inputdata. But, we should assess if you'll be doing that more often, and if so give you some training on doing it, and then give you the needed permissions.

The last more detailed thing is how this will be used in practice in terms of restarts and such. The stream file gives the level of excess ice at some point in time. If you want to run for a different date (especially a future date) the ice levels should have melted. And as such you should startup from levels from a restart file. Does that mean if you do a startup type run that it'll use values from the stream file, but if you do a hybrid or branch it'll be from the restart file and the stream is ignored? If so I wonder if the stream should only be read on startup to make this more clear? I'd like to discuss this more and figure out the best way to make this clear to CTSM users.

Thanks so much for your work on this. Very exciting to bring it in.

ekluzek:
It looks like this is up to date with ctsm5.1.dev071 and it should be update to the very latest. I tried doing that and was able to do so without conflicts.

src/biogeophys/SoilTemperatureMod.F90
mvdebolskiy: This block looks like this in master:
```
                     if (j >= 1) then
                        xmf(c) = xmf(c) + hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime
                     else
                        xmf(c) = xmf(c) + hfus*(wice0(c,j)-h2osoi_ice(c,j))/dtime
                     endif
```
I was unsure if the condition was redundant or not.

bld/namelist_files/namelist_defaults_ctsm.xml
mvdebolskiy: In the initial comment, there is [a google drive](https://drive.google.com/file/d/1mA457Oa52zG_MtLGB7KHuUYQvsS2-P5o/view?usp=sharing) link to the zipped file that contains both the stream file and the mesh file. I do not have access to cheyenne or glade. I don't think I will need one. I will ask Kaveh to put it there or if you can put it there, it would be awesome.

Regarding paramdata subdirectory -- I just followed the way ch4finundated is setup i.e. namelist defaults for _finundated_inversiondata*_ files are located.

> The last more detailed thing is how this will be used in practice regarding restarts and such. The stream file gives the level of excess ice at some point in time. 

Since there is only one dataset with initial concentration for excess ice concentration (unitless, for the part of the soil column, which is allowed to have excess ice in it) -- I followed the way _finundated inversion data_ is set up.

> If you want to run for a different date (especially a future date) the ice levels should have melted. And as such you should startup from levels from a restart file. Does that mean if you do a startup-type run, it'll use values from the stream file, but if you do a hybrid or branch it'll be from the restart file and the stream is ignored? 

Yes. This is correct, in the _WaterStateType%Restart_ if the variable is not found on the restart file or file containing initial conditions datasets -- the stream data will be used. There is no way this can work for branch-type runs since they do not allow to have missing variables in the restart file.  

This is done mainly for backward compatibility, so runs with `use_excess_ice==.true.` can be started from initial conditions that already exist (even if they don't have excess ice). In practice, it means that users should run a small spin-up run (I am looking into how long that should typically take) when they want to run excess ice option, for the soil hydro-thermal/bgc state to equilibrate.

For initial conditions and restart files that will be created with `use_excess_ice==.true.`, since the variable exists in the files it will be read and the stream value won't be used. However, the stream data is still read in any run with `use_excess_ice==.true.`. 
I was looking into the possibility of adding a logical variable that will look at if the `use_excess_ice==.true. has been already run, so there is no need to read stream data at all, but I am not sure that a small initialization cap/file download is worth overcomplicating things. Maybe @dlawrenncar and @swensosc  have better judgment for that.

src/biogeophys/SoilTemperatureMod.F90
mvdebolskiy: I've changed indents to 3 spaces, like in surrounding blocks

src/biogeophys/SoilTemperatureMod.F90
mvdebolskiy: Fixed.

src/biogeophys/SoilTemperatureMod.F90
mvdebolskiy: Fixed

src/biogeophys/SoilTemperatureMod.F90
mvdebolskiy: Fixed

src/biogeophys/TemperatureType.F90
mvdebolskiy: I was able to build and run after removing `sb`.

src/biogeophys/WaterDiagnosticBulkType.F90
mvdebolskiy: Do you mean the whole block or just the comment? Seems like indents around are the same.

src/biogeophys/WaterDiagnosticBulkType.F90
mvdebolskiy: At some point, I had them inactive, but then I assumed that maybe it is better to just have an if block there, so users when they run with `use_excess_ice=.true.` don't need to worry to switch the history variables on. I don't know if people run CTSM mostly with standard active output or change it most of the time. The diagnostic variable of course can be calculated in post-processing from the prognostic one but that requires some knowledge of how the parameterization works.

src/biogeophys/WaterStateType.F90
mvdebolskiy: Indented code to the same level as code above.

src/biogeophys/WaterStateType.F90
mvdebolskiy: Fixed
As well as the indents in the InitCold and Restart

src/biogeophys/WaterType.F90
mvdebolskiy: Done

src/biogeophys/WaterType.F90
mvdebolskiy: Deleted.

src/cpl/share_esmf/ExcessIceStreamType.F90
mvdebolskiy: Added

src/cpl/share_esmf/ExcessIceStreamType.F90
mvdebolskiy: Should `!TODO` be put here?

src/cpl/share_esmf/ExcessIceStreamType.F90
mvdebolskiy: Fixed

src/cpl/share_esmf/ExcessIceStreamType.F90
mvdebolskiy: Fixed

src/main/clm_varctl.F90
mvdebolskiy: Fixed

src/main/controlMod.F90
mvdebolskiy: Fixed

src/main/controlMod.F90
mvdebolskiy: Fixed

src/main/controlMod.F90
mvdebolskiy: Fixed

src/cpl/share_esmf/ExcessIceStreamType.F90
mvdebolskiy: It's a bit complicated but my logic was to have it here the same way it is `WaterStateType` where it is called from.

mvdebolskiy:
@ekluzek I was able to build and run the model with the commits above (both with `use_excess_ice=.true.` and `.false.`, cold start and from finidat, resubmits also worked). I haven't resolved anything above, only commented since I am not sure if I am supposed to resolve it. Not sure if unit_test code is built when normal cases are built though (regarding `NLFilename` in _WaterType_ initialization routines.

bld/CLMBuildNamelist.pm
ekluzek: Since, we are scheduling to end support for MCT by Sep/30th there really isn't need to add a check for this, since it won't matter after that point. So I won't add the check that I thought I should add earlier.

src/biogeophys/WaterDiagnosticBulkType.F90
ekluzek: Looking at it again, I don't see what I was seeing before. So I'll just resolve this comment.

src/biogeophys/WaterDiagnosticBulkType.F90
ekluzek: This is probably something that would be good to discuss on a call. Eventually excess ice is likely to be a pretty standard way to run the model, so some people will be using it, but not looking into the details of what it's outputting. So I think these probably should be within the use_excess_ice if, but unless they are really standard things that everyone should be outputting, they should also have default='inactive'

src/biogeophys/WaterStateType.F90
ekluzek: Perfect.

src/cpl/share_esmf/ExcessIceStreamType.F90
ekluzek: Yes, we should look into this more, and add a TODO item in the code here.

src/cpl/share_esmf/ExcessIceStreamType.F90
ekluzek: Cool.

src/cpl/share_esmf/ExcessIceStreamType.F90
ekluzek: This would be good to discuss. The nots really mean everything but natural-PFT or crop. So it would probably be more clear to use the soilc filter. With something like:

``` fortran
do fc = 1,num_soilc
   c = filter_soilc(fc)
   g = col%gridcell(c)
   exice_bulk_init(c)=this%exice_bulk(g)
```

Which is much simpler and more easily understood.

You don't really need the else in here, because exice_bulk_init is already being explicitly set to 0.0 everywhere.

ekluzek: @mvdebolskiy what part of WaterStateType did you mean? The InitCold part? That is using if statements on all the different types, but it's doing that to make sure it sets something for each type. I'd say this is a little different than that.

src/main/controlMod.F90
ekluzek: Awesome.

ekluzek: Actually I still see it there, so it would be good to remove the comment..

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: OK, I grabbed those files and have put them on cheyenne and also on our inputdata repository.  I had to change the format of the Mesh file from NetCDF-4 to CDF5. I'm still not sure if paramdata is the right subdirectory for this, but since the methane data is also there, it's at least consistent.

ekluzek: @mvdebolskiy can you give me push access to your fork, so I can push the change about the file? You do that by going to settings on github in your fork and then collaborators.

src/biogeophys/SoilTemperatureMod.F90
ekluzek: Cool.

src/biogeophys/SoilTemperatureMod.F90
ekluzek: Made an issue to look into this #1818. Since, you do need the if statement we might resolve this just by removing your comment.

src/biogeophys/SoilTemperatureMod.F90
ekluzek: Made this an issue in #1819

src/biogeophys/WaterStateType.F90
ekluzek: Awesome. Looking at the diff to master, I now only see one small indentation issue:

``` fortran
+            g = col%gridcell(c)
+            l = col%landunit(c)
+           if (.not. lun%lakpoi(l)) then  !not lake
+              if (lun%itype(l) == istsoil .or. lun%itype(l) == istcrop) then
```

src/cpl/share_esmf/ExcessIceStreamType.F90
ekluzek: I've filed an issue for this #1820, and am looking into seeing if this is still a problem.

bld/namelist_files/namelist_defaults_ctsm.xml
mvdebolskiy: Done, check for an invite

src/cpl/share_esmf/ExcessIceStreamType.F90
mvdebolskiy: I meant InitCold. I've put those redundant conditions everywhere, just to ensure that `excess_ice_col` will get 0.0 everywhere except for vegetated and crop units. The filter should be fine, but I was just too cautious.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: OK I pushed my changes.

src/biogeophys/WaterStateType.F90
ekluzek: OK, this is fixed.

ekluzek:
- [ ] Make sure we have at least one test in the aux_clm testlist that runs with excess ice
- [ ] Add an excess ice test to the science list
- [ ] Talk about #1819 
- [ ] Talk about #1820 
- [ ] Use filter in excess_iceStreams
- [ ] Decide about output of history variables
- [ ] Talk about general flow of use of excess ice

src/biogeophys/WaterDiagnosticBulkType.F90
ekluzek: Let's make this top one inactive by default, and the other two active by default, when excess_ice is on.

src/biogeophys/WaterStateType.F90
```diff
@@ -268,6 +283,15 @@ subroutine InitHistory(this, bounds, use_aquifer_layer)
             ptr_col=this%wa_col, l2g_scale_type='veg')
     end if
 
+    ! Add excess ice fields to history
+
+    if (use_excess_ice) then
+      data2dptr => this%excess_ice_col(begc:endc,1:nlevsoi)
+      call hist_addfld2d (fname='EXCESS_ICE',  units='kg/m2', type2d='levsoi', &
```
ekluzek, line 290: Because this is 3D, and not changing rapidly it should normally be on the annual file.

ekluzek: We probably need to add a user-mod for CMIP output for excess ice that adds this to the annual file. I think we can do that sort of thing.

mvdebolskiy:
I've checked TWS variable. It gets changed with the amount of excess ice, as expected. So, this one passes.

mvdebolskiy:
@ekluzek added log and fixed history fields. Also, maybe #1819 should be closed.

When the initial data interpolation is used you automatically get logs like:
 `Skipping     : VARNAME is NOT on input file`
 
 So I just added a reminder in the stream output.
 The restart without the variable gets an additional log line.

ekluzek:
I found a simple problem in DEBUG mode that I'll fix:

```
1:forrtl: severe (194): Run-Time Check Failure. The variable 'waterstatetype_mp_restart_$NBEDROCK' is being used in '/glade/work/erik/ctsm_worktrees/excess_ice/src/biogeophys/WaterStateType.F90(740,33)' without being defined
1:Image              PC                Routine            Line        Source             
1:cesm.exe           0000000002FECE06  waterstatetype_mp         740  WaterStateType.F90
1:cesm.exe           0000000002FBE99B  waterstatebulktyp         218  WaterStateBulkType.F90
1:cesm.exe           000000000303685F  watertype_mp_rest         752  WaterType.F90
1:cesm.exe           000000000097EA90  clm_instmod_mp_cl         533  clm_instMod.F90
1:cesm.exe
```

mvdebolskiy:
@ekluzek Is this because excess ice is read before `nbedrock` is read from a restart? I do not remember encountering this when I was testing running from `finidat` or hybrid run.

mvdebolskiy:
> @mvdebolskiy the fix is straightforward, see what I did in [212e104](https://github.com/ESCOMP/CTSM/commit/212e10476a64911da1ca9dff9f1a172515c4707a) which I just pushed. Let me know if you see any issues with it...

No issues. Nice catch, Thanks.

ekluzek:
@mvdebolskiy I've ran most of the testing and the only issue I'm seeing is a difference in answers for some cases without excess_ice. This might be something simple where an order of operations was changed. We do like to preserve the answers when a new feature is added and it's turned off, because it makes evaluation much simpler. So if we can find it and fix it, we'd like to. Let me know if you know what this might be.

For example, this list shows changes in answers

ERI_D_Ld9.f10_f10_mg37.I1850Clm45Bgc.cheyenne_gnu.clm-default
ERI_D_Ld9.ne30_g17.I2000Clm50BgcCru.cheyenne_intel.clm-vrtlay
ERP_D_Ld5.f09_g17.I2000Clm50Vic.cheyenne_intel.clm-vrtlay
ERP_D_Ld5.f10_f10_mg37.I2000Clm50Vic.cheyenne_intel.clm-vrtlay
ERP_D_Ld5.f10_f10_mg37.IHistClm45Sp.cheyenne_intel.clm-decStart
ERP_D_P36x2_Ld3.f10_f10_mg37.I2000Clm45BgcCrop.cheyenne_gnu.clm-no_subgrid_fluxes
ERP_Ld5.f09_g17.I2000Clm50Vic.cheyenne_intel.clm-vrtlay
ERP_P36x2_D_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.cheyenne_intel.clm-crop
ERP_P36x2_D_Ld5.f10_f10_mg37.I1850Clm45BgcCru.cheyenne_intel.clm-default
ERP_P36x2_D_Ld5.f10_f10_mg37.I2000Clm45Sp.cheyenne_intel.clm-default
ERP_P36x2_D_Ld5.f10_f10_mg37.IHistClm45BgcCru.cheyenne_intel.clm-decStart
ERS_D_Ld6.f10_f10_mg37.I1850Clm45BgcCrop.cheyenne_gnu.clm-clm50CMIP6frc
ERS_D_Ld6.f10_f10_mg37.I1850Clm45BgcCrop.cheyenne_intel.clm-clm50CMIP6frc
ERS_Ly5_P72x1.f10_f10_mg37.IHistClm45BgcCrop.cheyenne_intel.clm-cropMonthOutput
SMS_D_Ld5.f10_f10_mg37.I2000Clm45Fates.cheyenne_intel.clm-FatesColdDef
SMS_D_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm45BgcCropQianRs.cheyenne_intel.clm-cropMonthOutput
SMS_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.cheyenne_gnu.clm-crop
SMS_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.cheyenne_intel.clm-crop
SMS_Ld5.f10_f10_mg37.I2000Clm45Fates.cheyenne_intel.clm-FatesColdDef

ekluzek:
Another thing I'm seeing that izumi helps us with is that some arrays need to be dimensioned by nlevmaxurbgrnd rathn than nlevgrnd. I think that will be easy to fix as well.

ekluzek:
@mvdebolskiy it looks like the answer changing thing was the change in the soil depth for excess ice in SoilTemperature. It shouldn't matter, but turning it off when excess ice isn't on, seems to resolve the change in answers. Look at 782fc1923e8e050358588937d112cd20f6ed7712. I think we probably want it this way anyway, although it is also good to know that in most cases this change didn't matter in most instances.

mvdebolskiy:
@ekluzek  I have a suspicion that the fails were caused because some compilers did not like when you add anything to the `spval`. In `initVerticalMod`  we have loops like [this one](https://github.com/mvdebolskiy/CTSM/blob/9bb110f9ae81fb872a064597407dc4edfef0c3c5/src/main/initVerticalMod.F90#L425-L467) where depth between `nlevgrnd` and `nlevurb` are assigned `spval` and  in `SoilTemperature` we adjust depths for `1:nlevmaxurbgrnd`.

ekluzek:
@mvdebolskiy I added an issue for adding this to the tech note in #1923

ekluzek (approved):
There's only a couple more things to do with this now. So I'm adding approve to it.

ekluzek:
@mvdebolskiy this is looking almost ready right now. Do you want to try running the latest on Fram and make sure results look as you expect? Also please assign yourself to #1923

mvdebolskiy:
@ekluzek I will run the latest on fram during the weekend. 
Is it even possible for me to assign myself to the issue on this repository? I don't see a settings button on the issue or anything.

src/biogeophys/WaterDiagnosticBulkType.F90
```diff
@@ -194,6 +199,10 @@ subroutine InitBulkAllocate(this, bounds)
     allocate(this%h2osoi_ice_tot_col     (begc:endc))                     ; this%h2osoi_ice_tot_col     (:)   = nan
     allocate(this%h2osoi_liq_tot_col     (begc:endc))                     ; this%h2osoi_liq_tot_col     (:)   = nan
     allocate(this%swe_old_col            (begc:endc,-nlevsno+1:0))        ; this%swe_old_col            (:,:) = nan   
+    allocate(this%exice_subs_tot_col     (begc:endc))                     ; this%exice_subs_tot_col     (:)   = 0.0_r8
+    allocate(this%exice_subs_col         (begc:endc, 1:nlevmaxurbgrnd))   ; this%exice_subs_col         (:,:) = 0.0_r8
+    allocate(this%exice_vol_col          (begc:endc, 1:nlevgrnd))         ; this%exice_vol_col          (:,:) = 0.0_r8
```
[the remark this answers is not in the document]
mvdebolskiy: It can be nlevsoi.  This one is only used in this WaterDiagnostics and not involved in any loops that go all the way to the `nlevmaxurbgrnd`.

ekluzek (changes requested):
I met with @billsacks and we went over the test list and had a few comments.

- [ ] Add a note to the issue handling that this is a permanent feature as opposed to others that might be removed.
- [ ] Add issue as a parameter to restUtils and use in restFileMod so it's defined in one place.
- [ ] Shorten the 12 month test, but add a note that we aren't really testing excess ice unless there's the rare chance that a gridcell will have excess ice melt. Since, this isn't a standard feature that's always on, we think this might be OK.
- [ ] Only have one test for this since this isn't currently a standard feature right now.

ekluzek:
@mvdebolskiy make sure you have set a watch level to the CTSM repository, and then I should be able to assign you to the issue. Watch level is at the very top on the right side with other settings. You are right you might not have permissions to assign people, I thought you could assign yourself, but looks like not.

Also be sure to "star" CTSM, just so we can say we are winning the repository popularity contest. :-)

src/biogeophys/WaterDiagnosticBulkType.F90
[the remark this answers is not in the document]
ekluzek: - [ ] Change the dimension to nlevsoi

mvdebolskiy:
@ekluzek I've just commented on the issue and you should be able to assign me. I already starred and watched the repository long ago.

mvdebolskiy:
> * [ ]  Matvey will run on Fram and make sure it looks as he expects with the latest version

Just ran two 2 year global cases form with `CLM_FORCE_COLDSTART=off`  and `RESUBMIT=1` on fram with and without excess ice.  Both worked.
Running a couple of single point ones made with `subset_data`, just to make sure.

mvdebolskiy:
@ekluzek even if you run 2100 all of the excess ice won't melt, since there is a lot of cold cells left with permafrost. To melt everything, one can adjust any forcing f.e. 1901-1951 with a rediculous uniform temperature  change (like 20 degrees per  decade or more).  Or just run a small spinup  1901-1921 and then put a line in the atm2lnd to adjust air temperature by 30 degrees or so and run another 20-30 year case. 

And then just look for if the model falls apart or not what difference in timing of getting rid of the permafrost inclusion of excess ice makes.

I usually try to test some ridiculous extreme cases to see if my model can withstand it.

ekluzek:
To test the code we don't need ALL the excess ice to melt, but we want to know that at least one point melted so that code was exercised. You have a good point that it can be useful to stress test a parameterization with a ridiculous case, so that you know it functions under unusual conditions. For now I'm just musing about what starting conditions or length of simulation would guarantee that at least one point melts. You are right we could do a crazy case using anomaly forcing to apply a ridiculous warming to a short simualtion and see if that would work.

mvdebolskiy:
Oh, in that case -- we need to run a 20-50 years case from `finidat`  and just look for cells that had `TOTEXICE_VOL > 0` at the first time step and `TOTEXICE_VOL = 0`. Then just pick one cell that melts the fastest. I can't run it on `fram` -- too little allocated hours left on the project I am on right now. I can try on `betzy` but that will take some time to set up.

slevis-lmwg:
TODOs
1) Update to latest tag
2) Run test-suites
3) Go through this [check list](https://github.com/ESCOMP/CTSM/pull/1787#pullrequestreview-1211991831)

slevis-lmwg:
@mvdebolskiy @ekluzek 
I'm ready to push the update to escomp/master and need permission to do so to this PR. If I remember right, I need to be made collaborator on this PR by one of you.

ekluzek:
That's awesome @slevis-lmwg. @mvdebolskiy can you make @slevis-lmwg a collaborator on your fork of CTSM? On your fork go to the "Settings" tab and then "Collaborators" and they'll be a way to add a collaborator and you can invite Sam and give him permission to work on your fork.

mvdebolskiy:
@slevis-lmwg Added you to my fork. [This branch is linked to this PR](https://github.com/mvdebolskiy/CTSM/tree/exice_nuopc_pr).

slevis-lmwg:
@swensosc @mvdebolskiy @ekluzek 
I made a branch tag as follows:
```
git tag -a branch_tags/exice_wrpmip.n01_ctsm5.1.dev129
git push escomp branch_tags/exice_wrpmip.n01_ctsm5.1.dev129
```
Test suites OK and generated baselines using the branch tag's name `exice_wrpmip.n01_ctsm5.1.dev129`

slevis-lmwg:
> * [ ]  Add a note to the issue handling that this is a permanent feature as opposed to others that might be removed.
> * [ ]  Add issue as a parameter to restUtils and use in restFileMod so it's defined in one place.
> * [ ]  Shorten the 12 month test, but add a note that we aren't really testing excess ice unless there's the rare chance that a gridcell will have excess ice melt. Since, this isn't a standard feature that's always on, we think this might be OK.
> * [ ]  Only have one test for this since this isn't currently a standard feature right now.

To address these pending TODOs, I likely need a follow-up with @ekluzek :
1) I don't think that I am familiar with issue handling. I also want to clarify the meaning of "issue" in restUtils.
2) How short were you planning on making this test? If we are shortening this test, then I assume that we are removing the other ExcessIce test.

slevis-lmwg:
PASS: `SMS_Lm12.f09_f09_mg17.I1850Clm51Sp.cheyenne_intel.clm-ExcessIceStartup_output_sp_exice`

I think this PR is ready. When it's next in line to merge, I will need to:
- Rerun the test-suites
- Update ChangeLog/ChangeSum

bld/unit_testers/build-namelist_test.pl
```diff
@@ -163,9 +163,21 @@ sub cat_and_create_namelistinfile {
...
 #
+<<<<<<< HEAD
+my $ntests = 1863;
+||||||| 17e2acb6a
+my $ntests = 1846;
+=======
 my $ntests = 1975;
+>>>>>>> escomp/master
 if ( defined($opts{'compare'}) ) {
+<<<<<<< HEAD
```
slevis-lmwg, line 174: I somehow missed a couple of conflicts here. For now I will use the value that I see in escomp/master; however, I wonder whether the exice branch added tests that I need to account for in the final numbers...

src/biogeophys/WaterDiagnosticBulkType.F90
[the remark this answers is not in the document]
slevis-lmwg: Looks as though this suggestion wasn't tested. Making a note to try it out.

bld/unit_testers/build-namelist_test.pl
slevis-lmwg: @ekluzek suggested the following approach:
1) Add the difference between HEAD and starting point to MAIN.
2) Run the tests and look at the output to confirm how many tests ran and how many should have run.

src/biogeophys/WaterDiagnosticBulkType.F90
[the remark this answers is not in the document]
slevis-lmwg: PASS `SMS_Lm12.f09_f09_mg17.I1850Clm51Sp.cheyenne_intel.clm-ExcessIceStartup_output_sp_exice`
but cheyenne test-suite 2 unexp. FAILs to investigate
Once resolved, run also the izumi test-suite.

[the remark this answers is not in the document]
slevis-lmwg: Reran the 3 unexpected fails on cheyenne, and they passed.

bld/unit_testers/build-namelist_test.pl
slevis-lmwg: I tried ./build-namelist_test.pl and it finished with this message:
```
Successfully ran all testing for build-namelist

Cleanup files created
/bin/rm: cannot remove 'env_run.xml': No such file or directory
/bin/rm: cannot remove 'user_nl_ctsm_real_parameters': No such file or directory
/bin/rm: cannot remove 'temp_file.txt': No such file or directory
```
I did not keep track of other output that scrolled past me on the screen. I think that this result is fine based on our conversation this morning @ekluzek but pls let me know if I should investigate further. Thanks!

src/biogeophys/WaterDiagnosticBulkType.F90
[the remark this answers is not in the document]
slevis-lmwg: Summary
1) PASS SMS_Lm12.f09_f09_mg17.I1850Clm51Sp.cheyenne_intel.clm-ExcessIceStartup_output_sp_exice
2) Test-suites
OK cheyenne
OK izumi

slevis-lmwg:
Izumi test-suite IN PROGRESS

src/cpl/share_esmf/ExcessIceStreamType.F90
ekluzek: We realized this is just how we do things for InitCold, so this should stay the same. So I'm resolving this conversation.

src/biogeophys/WaterDiagnosticBulkType.F90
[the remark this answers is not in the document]
ekluzek: This was done so resolving.

bld/unit_testers/build-namelist_test.pl
ekluzek: Yep, this is all good, so resolving.

src/biogeophys/SoilTemperatureMod.F90
ekluzek: In the discussion in the issue we decided to remove the comment and leave the code as it is. I also searched for other places like this and didn't find any. And Matvey did previously as well, so we are good here.

src/biogeophys/SoilTemperatureMod.F90
ekluzek: We removed the comment. The issue doesn't need to be resolved anymore since in excess ice there is a difference. With excess ice there is an additional term added in for the j>=1 case. It would be better to have the equation the same, but then add the extra term for j>=1 case. But, we'll forget about that for now.

src/cpl/share_esmf/ExcessIceStreamType.F90
ekluzek: This is an issue we'll leave in place, until more work is done on it. I looked at this last year, and I'm pretty sure I remember it still being an issue, so we'll leave it like it is.

ekluzek (approved):

PR #1793, opened 2022-06-30
Improvements for CTSM initialization

mvertens:
### Description of changes
Improvements for CTSM initialization

### Specific notes
- A new directory is now in your run directory `init_generated_files`. These two run time generated files  – `finidat_interp_dest.nc` and the land fraction file `ctsm_landfrac.nc `(which is new in this PR) – are put in that directory when they are created. Then, if they already exist, they are reused rather than being regenerated. (Technically, it looks for a status file that flags successful creation of this file.)
- For tests, the`init_generated_files `directory is removed before running, so that rerunning a test will do the same thing the second time. **Note t**hat for this to be enabled for testing, the CIME tag `cime6.0.42 `needs to be included.
- There are occasional, rare cases when a user would need to manually remove the init_generated_files directory – e.g., when changing things about N dominant landunits or PFTs. (we may want a test that exercises this and ensures things are bit-for-bit).
- In addition, NDEP has been added to CTSM history output for future use.
- Output was cleaned up in a few cases so that it did not span multiple lines 

CTSM Issues Fixed: None

Testing performed, if any: Ran the ctsm test suite and compared to ctsm5.1.dev098 (this PR however is up to date with ctsm5.1.dev099 - but baselines were not available for that tag when the testing was done).


olyson:
I believe we use NDEP_TO_SMINN as the history variable that indicates atmospheric nitrogen deposition.
In subroutine CNNDeposition in biogeochem/CNNDynamicsMod.F90, we have:

    associate(                                                                &
         forc_ndep     =>  atm2lnd_inst%forc_ndep_grc ,                       & ! Input:  [real(r8) (:)]  nitrogen deposition rate (gN/m2/s)
         ndep_to_sminn =>  soilbiogeochem_nitrogenflux_inst%ndep_to_sminn_col & ! Output: [real(r8) (:)]  atmospheric N deposition to soil mineral N (gN/m2/s)
         )

      ! Loop through columns
      do c = bounds%begc, bounds%endc
         g = col%gridcell(c)
         ndep_to_sminn(c) = forc_ndep(g)

      end do

    end associate

And in soilbiogeochem/SoilBiogeochemNitrogenFluxType.F90:

    this%ndep_to_sminn_col(begc:endc) = spval
    call hist_addfld1d (fname='NDEP_TO_SMINN', units='gN/m^2/s', &
         avgflag='A', long_name='atmospheric N deposition to soil mineral N', &
         ptr_col=this%ndep_to_sminn_col)


Unless there is some distinction between the new NDEP and NDEP_TO_SMINN that I'm not aware of.

mvertens:
@olyson thanks. Do we only need the nitrogen deposition from the atmosphere if the biogeochemistry is on? I've added it to the general fields send from the atmosphere - so I'm wondering if we should use that instead. I'm validating the new fields using SP mode since its cheaper. The idea is that you will always get ndep from the atmosphere right now - regardless of the CTSM mode.

olyson:
Right, as far as I know we only need nitrogen deposition in BGC mode.  Although maybe there are new developments that might require it in SP mode?  Or possibly maybe FATES?  @ekluzek

olyson:
I think that is fine with me, otherwise we would end up with both NDEP_TO_SMINN and ATM_NDEP in the history file.  Is there an option to remove the current NDEP_TO_SMINN, but rename ATM_NDEP to NDEP_TO__SMINN?  I'm thinking the land diagnostics package would have to be changed otherwise (easily done however) and possibly other people's scripts.  On the other hand, perhaps ATM_NDEP is more appropriate.  Wishy-washy Friday....

mvertens:
@olyson , @ekluzek - I have removed NDEP_TO_SMINN from the history output in this PR now.

mvertens:
@erik, @billsacks - I seem to be encountering problems with this PR in another use case I am using. Please hold off until I can resolve it.

mvertens:
@erik, @billsacks - I fixed the problem and pushed the fix back.

billsacks (changes requested):
Thank you very much for this enhancement, @mvertens! This is going to help a lot of users by significantly speeding up their initialization times when rerunning a case for debugging or development purposes!

I do have a few requests to make this more robust. I don't feel like we need to cover every edge case, but I have been trying to think through scenarios that feel reasonably likely to occur in practice and for which the current logic could silently do the wrong thing.

- [x] **Handling changing finidat:** I feel like a common scenario would be that a user does an initial test run, then changes `finidat` (with `use_init_interp` explicitly set to true) and reruns. My understanding is that the code will do the wrong thing in this situation: it will detect that the `finidat_interp_dest.nc` file already exists and will reuse it without regard for the fact that `finidat` has changed and so the interpolation needs to be redone. I think the best way to prevent this would be to add metadata to the `finidat_interp_dest.nc` file saying what the source file was (if such metadata doesn't already exist... this would be useful metadata to have around anyway). Then, when you check to see if the `finidat_interp_dest.nc` file already exists, you can expand this logic to both check the existence of the file and also to read this piece of metadata from the file; if the metadata gives a different source file from the one currently being pointed to, then act as if the `finidat_interp_dest.nc` file doesn't yet exist (i.e., rerun the interpolation). (There may be other similar namelist flags we should check, but I can't think of any off-hand that would be nearly as important as `finidat`.) (Another way to implement this would be to write the source file to the `.status` file, but I feel like using netcdf metadata will be more robust and flexible, particularly if we want to check multiple items in the future.)

- [x] **Handling a switch to cold start:** A special case of the above is where the user does an initial run by interpolating an initial conditions file, but then changes the model to use cold start (via `./xmlchange CLM_FORCE_COLDSTART=on`). I haven't traced the logic for this scenario, but can you confirm that the code does the right thing in this scenario – i.e., uses a cold start rather than reusing the `finidat_interp_dest.nc` file?

- [x] **Handling a custom finidat_interp_dest file name:** The ability for the user to set their own `finidat_interp_dest` file name adds some complexity here; I haven't thought through how this will work in all possible scenarios. What will happen in the following cases? (This may already work right. If not, and if it's hard to get it working right, then maybe the logic can be changed so your new logic only applies if using the default `finidat_interp_dest` rather than a user-specified file name??? It seems okay to me if the result of these things is that interpolation is unnecessarily redone, but I want to make sure the model won't crash or silently do the wrong thing in these scenarios.) (Update: based on our discussion today (2022-07-04), I would be fine with removing `finidat_interp_dest` as a namelist flag and just having it hard-coded: I doubt people use this namelist flag often, if at all.)
  - User sets finidat_interp_dest to something else before first run, keeps it set like that for a rerun
  - User uses the default finidat_interp_dest for the initial run, but then sets finidat_interp_dest and reruns
  - User sets finidat_interp_dest before first run, unsets it for a rerun
I have removed this functionality from the namelist. I think this solves the problem.

- [x] **Removing the status file before writing:** (I don't think this is an issue currently, but I think it's something we want to make sure of with other changes requested above – specifically, the possible need to recreate `finidat_interp_dest.nc` because the user has changed `finidat`.) Can you please add code that removes each status file before starting to write the new file? e.g., for the `finidat_interp_dest` file: if we find that we need to recreate this file (e.g., because `finidat` has changed), then the status file should be removed. Probably a good time to do this removal is as soon as we discover that we need to recreate this file (e.g., as soon as we see that `finidat` doesn't match the metadata on the pre-existing `finidat_interp_dest` file). (I wasn't sure how to delete a file in Fortran; here is a pattern for doing so: https://programming-idioms.org/idiom/155/delete-file/3521/fortran ).

In addition, here are a few optional requests to clean up the code. If you don't have time to do these, then I'm okay with bringing in this PR without these clean-ups:

- [ ] [OPTIONAL] Given that we are now using the status file in two places, and it seems like there's a good chance that we may expand this usage throughout CTSM / CESM, it feels worth introducing some generic functions that take care of this status file functionality, removing some code duplication. I can envision a small module (which for now could be in the CTSM code but perhaps eventually should be moved to the share code) with the following functions:
  - `logical function status_file_exists(nc_filename)` (returns true if the status file associated with `nc_filename` exists; false otherwise)
  - `subroutine write_status_file(nc_filename)` (writes the status file associated with `nc_filename`)
  - `subroutine delete_status_file(nc_filename)` (deletes the status file associated with `nc_filename`)
  - `function status_filename_from_nc_filename(nc_filename)` (private helper function: given the netcdf filename, returns the corresponding status filename)

- [ ] [OPTIONAL] (I recognize that this mainly relates to pre-existing code, which is the main reason I'm listing it as optional) `lnd_set_read_write_landmask` could be simplified by splitting it into two separate subroutines: one to do the read and one to do the write: currently there is no overlap between those two operations, and the current interface is a bit confusing, in that you need to pass two logical flags (`read_file` / `write_file`), exactly one of which should be true. If you don't split this, then (a) ideally these two arguments would be collapsed down to one, or at least there would be an error check that exactly one is true; (b) ideally the calls to this function would use keyword arguments, particularly to document what the logical arguments represent.

src/cpl/share_esmf/lnd_set_decomp_and_domain.F90
```diff
@@ -743,24 +775,19 @@ logical function chkerr(rc, line, file)
...
   !===============================================================================
-  subroutine lnd_set_read_write_landmask(write_file, read_file, lndmask_glob, lndfrac_glob, gsize)
+  subroutine lnd_set_read_write_landmask(flandfrac, flandfrac_status, write_file, read_file, &
+       lndmask_glob, lndfrac_glob, gsize)
 
-    ! This subroutine is currently unused (as of 2021-03-17), but it may be needed in the
-    ! future. Its purpose is: Now that we get landmask and landfrac at runtime, from
+    ! Write or read landfrac and landmask to file so that mapping does not have to be done each time
     ! mapping the ocean mask to the land grid, it's possible that landfrac will be
-    ! roundoff-level different with different processor counts. Mariana Vertenstein
```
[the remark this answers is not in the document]
mvertens: This comment is no longer there.

billsacks:
I tried a multi-instance run (a test with `_C2`) and noticed that there is only one `ctsm_landfrac.nc` file produced in the `init_generated_files` directory. I'm thinking there should be one for each instance. It's hard to imagine a situation where they would differ, but I think that having a single file could lead to the different instances stepping on each other. (Alternatively, the write could happen just from the first instance, but it seems simpler to just have a general rule that all of these files should have an instance suffix.)

- [ ] If you agree, can you please change the landfrac file name – and its associated status file – to have an instance suffix?

billsacks:
I'm done my updates to this branch. Note that I just pushed a change that includes a new system test. This requires the cime changes in https://github.com/ESMCI/cime/pull/4279

mvertens:
@billsacks  - I don't see how this case can occur for different land fractions - but I'm happy to change the name for the sake of consistency.

billsacks:
> I don't see how this case can occur for different land fractions - but I'm happy to change the name for the sake of consistency.

I don't see how either. I'm not concerned about that but rather am concerned that two instances would be trying to write the same file at the same time, leading to problems.

mvertens:
@billsacks - that makes sense. Thanks for clarifying.

src/main/clm_initializeMod.F90
```diff
@@ -478,40 +482,66 @@ subroutine initialize2(ni,nj)
...
-          write(iulog,*)'Reading restart file ',trim(fnamer)
+          write(iulog,'(a)')'Reading restart file '//trim(fnamer)
        end if
        call restFile_read(bounds_proc, fnamer, glc_behavior, &
             reset_dynbal_baselines_lake_columns = reset_dynbal_baselines_lake_columns)
     end if
 
     ! If appropriate, create interpolated initial conditions
     if (nsrest == nsrStartup .and. finidat_interp_source /= ' ') then
-       is_interpolated_start = .true.
```
billsacks, line 507: - [ ] It looks like this variable is no longer used in the code (it looks like I removed its one use at some point but didn't remove the variable). Can you please go ahead and remove it from clm_varctl? I think that, with the changes you have made, its value is no longer correct; rather than fixing it, I feel like we might as well just remove it.

src/main/clm_initializeMod.F90
```diff
@@ -478,40 +482,66 @@ subroutine initialize2(ni,nj)
...
        end if
 
+       ! Determine name if finidat_interp_dest status file
+       klen = len_trim(finidat_interp_dest) - 3 ! remove the .nc
+       locfn = finidat_interp_dest(1:klen)//'.status'
+
+       ! Remove file if it already exists
+       inquire(file=trim(locfn), exist=lexists)
+       if (lexists) then
+          open(unit=1234, file=locfn, status='old', iostat=ioe)
```
billsacks, line 539: - [ ] Here and two lines down: this should use newunit rather than a hard-coded unit number that could theoretically already be in use.

src/main/controlMod.F90
```diff
@@ -1113,11 +1116,50 @@ subroutine apply_use_init_interp(finidat_interp_dest, finidat, finidat_interp_so
...
+       if (status /= nf90_noerr) call handle_err(status)
+       status = nf90_close(ncid)
+       if (status /= nf90_noerr) call handle_err(status)
+
+       ! If the input source file in finidat_interp_dest does not match the namelist input for mapping
+       ! then use the namelist input (i.e. finidat) - otherwise just use the generated file
+       if (trim(initial_source_file) /= trim(finidat)) then
+          finidat_interp_source = finidat
+          finidat = ' '
+          if (masterproc) then
```
billsacks, line 1134: - [ ] [OPTIONAL] Here and a little below: this function is only called by the masterproc, so these masterproc conditionals aren't needed. They don't hurt, though they threw me off because it made me think that the routine was called by all procs – which I think would be somewhat problematic in that it would mean that all procs are opening the netcdf file to read the global attribute at once.

billsacks (changes requested):
Thank you for all of your changes here, and for addressing my requests!

Reviewing the latest changes, I see a few possible issues with the status files:

- [ ] I see you remove the finidat_interp_dest status file before writing finidat_interp_dest - thank you. However, I don't see a similar removal of the landfrac status file before writing that file. Can you please apply similar logic there?

I believe there is a possible race condition in the code for removing a pre-existing status file: it looks like you execute this removal code on all processors. This creates the possibility of:
- proc 1 checks and finds the file exists
- proc 2 checks and finds the file exists
- proc 1 removes the file
- proc 2 tries to remove the file and fails because it has already been removed by proc 1

- [ ] I think the solution to this race condition is to just have the masterproc do the checking and removal of the status file; other procs should probably wait until it's done (e.g., via an mpi barrier) (because we don't want other procs to go on to start writing the new file until the status file has been removed)

- [ ] I think there is a similar possible race condition in the code to write the status file - both for the finidat_interp_dest file and for the landfrac file: it looks like you do this write from all procs. But I think it could be a problem (at least on some systems) if multiple procs are trying to write the same file at the same time. I would again suggest just doing this from the masterproc. In this case, I think we want an mpi barrier *before* the writing of the status file – so that we ensure that all processors have successfully completed writing the real file before the masterproc writes the status file signaling successful completion.

src/main/clm_initializeMod.F90
mvertens: @billsacks - this does not work with newunit. I tried that. 

https://github.com/ESCOMP/CTSM/blob/2671dfc54aa91cc5c3ba5e3ca984046dfe1216a7/src/main/clm_initializeMod.F90#L539-L542

When you use newunit - the actual unit number is negative. And trying to use the delete flag on a negative unit number was problematic. @jedwards4b suggested this code and I think this is why he used 1234. I am happy to put a comment in here if that would be useful.

src/main/controlMod.F90
mvertens: @billsacks - thanks for pointing this out. I did not appreciate that this function was only called on masterproc. That said - I think there is a real problem in having an endrun call in a function that is only called on masterproc.
If you did encounter an issue - I think that the model would hang - since only the master processor would actually call endrun. So I think parts of controlMod need to be rewritten so that an error is broadcast to all processors and then the model can exit gracefully. I'm happy to implement that - but its a bit outside the scope of this PR. Thoughts?

src/main/clm_initializeMod.F90
billsacks: Interesting, good to know. I guess you could use the mechanism we used before the existence of newunit, which is to use shr_file_getUnit and shr_file_freeUnit, but I'm okay with this being left as is. (I'd support adding a comment, but don't feel it's critical.)

src/main/controlMod.F90
billsacks: (We discussed this. It's fine to call endrun from one proc.)

mvertens:
@billsacks - I have addressed the requests that you made.

src/main/controlMod.F90
mvertens: @billsacks - thanks for pointing this out. You are right!

src/cpl/share_esmf/lnd_set_decomp_and_domain.F90
```diff
@@ -770,15 +802,32 @@ subroutine lnd_set_read_write_landmask(write_file, read_file, lndmask_glob, lndf
...
     type(file_desc_t) :: pioid ! netcdf file id
     integer           :: dimid
-    character(len=CL) :: flandfrac = 'landfrac.nc'
+    integer           :: iun
+    integer           :: ioe
+    integer           :: ier  
+    logical           :: lexists
     !-------------------------------------------------------------------------------
 
+    ! TODO: remove status file before writing (unit 9876)
```
billsacks, line 811: - [ ] It looks to me like you have done what you wanted to do here. Do you agree? (i.e., should this TODO be removed?)

billsacks (changes requested):
Okay, this all looks great now - thank you!!! The only remaining thing is that I want to double-check that it's okay to remove a TODO comment (see inline comment that I just made). If so, I'll do that, update CIME, update to the latest master, then run final testing.

src/cpl/share_esmf/lnd_set_decomp_and_domain.F90
mvertens: @billsacks - thank you for checking. Yes - this is no longer needed. Thanks so much for reviewing the code.

billsacks (approved):

Issue #1791, opened 2022-06-30
mksurfdata_esmf should use site/grid name rather than "NX x NY" name


ekluzek:
In order to be more compatible with single point site names and also all of the different types of global grid names such as the 360x720cru name and the NLDAS2 name, we need mksurfdata_esmf to output names using the users input site name, rather than a name with "NX"x"NY".

slevisconsulting:
Looking again at `gen_mksurfdata_namelist.py` I find that the code does what this issue requests. For example:
- `./gen_mksurfdata_namelist.py --res 5x5_amazon --start-year 1850 --end-year 1850` returns
`surfdata_5x5_amazon_hist_78pfts_CMIP6_1850_c220706.namelist`
- `./gen_mksurfdata_namelist.py --res 360x720cru --start-year 1850 --end-year 1850` returns
`surfdata_360x720cru_hist_78pfts_CMIP6_1850_c220706.namelist`
It uses NX x NY for custom grids such as:
- `./gen_mksurfdata_namelist.py --start-year 2000 --end-year 2000 --model-mesh /glade/work/slevis/data_UCB/ca_fates_input_data/wrfinput_to_meshes/wrfinput_d01_unstruct.nc` returns
`surfdata_109x104_hist_78pfts_CMIP6_1850_c220705.namelist`

Keep in mind that the code obtains mesh_file paths from .../ccs_config/component_grids_nuopc.xml which does not include paths to the 1x1 mesh_files. In a meeting with @ekluzek and @mvertens towards the beginning of the year we agreed that 
- I would test mksurfdata_esmf for 1x1 to confirm that it works. I did that in late March. I had to generate an UNSTRUCT 1x1_brazil destination mesh and add the path to ccs_config/component_grids_nuopc.xml for this to work.
- Also that for the most part we expected to generate fsurdat files for 1x1 using the subset_data tool.

If there's nothing additional here, I think this issue may be closed.

ekluzek:
@slevisconsulting it's those custom cases that need to change. For the WRF case I'd make "res" still required and "res" would be the string that would be used in the surface dataset filename. When people make custom cases they need to be able to have appropriate names for them.

So it would be something like:

```
./gen_mksurfdata_namelist.py --res wrfd01 --start-year 2000 --end-year 2000 --model-mesh /glade/work/slevis/data_UCB/ca_fates_input_data/wrfinput_to_meshes/wrfinput_d01_unstruct.nc returns
[...]

Issue #1790, opened 2022-06-30
mksurfdata_esmf produce wrong output when using a mesh file and does not check for mesh compatibility with nx and ny


negin513:
### Brief summary of bug

I was trying to use mksurfdata_esmf for south Africa domain. At first, I've noticed all the variables on the output were zeros and was not correct. I've noticed a few other issues while working with this that I summarized below.
These issues are also relevant for non-WRF regional cases that require mesh files.

1. The user can specify mesh file with ./gen_mksurfdata_namelist.py through `--model-mesh` option. The code by default assume values of `nx=-999` and `ny=-999` which produce files that are have only zeros as their output.
It was not clear that the code needs `--model-mesh-nx` and `--model-mesh-ny`, and by default it was assigning a wrong value (-999) which produced wrong results (only zeros) without any errors or warnings. 

@ekluzek and I was not sure what is causing this issue, so I highly suggest improving this as it might cause confusion for other users.

Solution: 
* These `--model-mesh-nx` and `--model-mesh-ny` flags should be required when using `--mesh-files` option instead of assigning a default value of -999.  There are different ways to do this, but at the very least the code should check via ifs and print out some errors instead of proceeding with default values.

2.  Since mesh files are in 1-d format, I am not sure how to figure out values of nx and ny from the mesh files. How can the user find nx and ny from the mesh files to provide inputs of ./gen_mksurfdata_namelist.py?

3. For a mesh file, the workflow works without any complains with any random values of nx and ny but the output file is not correct.  For example to figure out nx and ny I looked at my wrfout files (2d files), but it is not clear which direction is nx and which dimension is ny. 
The help does not provide any information about this (directions of nx and ny) and using the wrong nx and ny (for example flipped nx and ny) still produces a surface dataset files that is wrong. We need to add some testing to make sure nx and ny is compatible with the mesh file provided.

Solution: 
- We either need to remove the requirements for nx and ny and require the code to detect nx and ny from the mesh file on the fly. 
- At the very least we need to put in some error checks for wrong nx and ny, so the code does not proceed without any errors or warning. 

I have talked with @slevisconsulting about this today, and he mentioned he noticed similar issues. 

* It seems like the best solution to address all the above would be for the workflow to automatically detect nx and ny,  instead of relying on the user input which might not be compatible with each other.
*  At the very least we need to add some error checking to make sure the values provided by the user are compatible with the mesh file. 

### General bug information

[...]

Issue #1776, opened 2022-06-08
Have mksurfdata_esmf pick up the mpi command from env_mach_specific.xml


slevisconsulting:
Motivated by this [comment](https://github.com/ESCOMP/CTSM/pull/1756#issuecomment-1138146016):
The best way of keeping the mpi command up-to-date and working on various machine/compiler/mpilib combinations is to query the env_mach_specific.xml file for the mpirun command. The mpirun command is fairly complex and depends on the MPI library being used.

E.g. for mvapich2
`mpiexec --machinefile $ENV{PBS_NODEFILE} -n <ntasks> --prepend-rank`
and for openmpi
`mpiexec -n <ntasks> --tag-output`

@ekluzek I thought I hadn't resolved my issues with PBS_NODEFILE, but this [comment](https://github.com/ESCOMP/CTSM/pull/1748#issuecomment-1139101946) and the one after it suggest that I got past it.

slevisconsulting:
@jedwards4b I will start by asking you and let me know if I should be asking somebody else.

I'm reading the `env_mach_specific.xml` file to piece together mpirun commands for the new mksurfdata_esmf tool on different machines. In the next lines I show what I the .xml file provides versus what I need for my python code to work:

On cheyenne I get the first but want the second:
```
mpiexec_mpt -p "%g:" -np {{ total_tasks }}
mpiexec_mpt -p \"%g:\" -np {total_tasks}
```

On casper:
```
mpirun  -np {{ total_tasks }}
mpirun  -np {total_tasks}
```

On izumi:
```
mpiexec --machinefile $ENV{$PBS_NODEFILE} -n {{ total_tasks }}
mpiexec --machinefile $PBS_NODEFILE -n {total_tasks}
```

The question is about changing the strings in the .xml to look the way I need them. Would that break things elsewhere? If not, then that seems like the ideal solution. If things elsewhere will break, then I will need to manipulate the strings in my code.

jedwards4b:
At a higher level there is a case object [get_mpirun_cmd ](https://github.com/ESMCI/cime/blob/master/CIME/case/case.py#L1960).   In either you use the overrides argument to set the values for stuff in {{ }}
For example overrides = {"total_tasks" : 8,} will replace {{ total_tasks }} with the value 8.

slevisconsulting:
That's probably what I want, thank you! (I was putting together my own code to read the .xml, not using cime.)

jedwards4b:
[...]

PR #1796, opened 2022-07-06
Add consistency checks for mesh_nx mesh_ny relative to mesh_file in gen_mksurfdata_namelist.py

slevisconsulting:
### Description of changes
If mesh_file contains mesh_nx mesh_ny info, read it from the file.
Else abort and require user to enter it at the command line.
Confirm that the product of mesh_nx x mesh_ny (user-entered or from mesh_file) = elementCount (from mesh_file).

Address #1791 here, so as to name fsurdat files by their site/grid name rather than by their NX x NY.

Also address #1776 here, so as to stop hardwiring the mpirun command according to --machine in gen_mksurfdata_jobscript_single.py. Now get the mpirun command from env_mach_specific.xml.

### Specific notes

CTSM Issues Fixed (include github issue #): #1790 #1791 #1776 

Testing performed, if any:
Running gen_mksurfdata_namelist.py for cases with and without --model_mesh in the user-entered arguments and confirming expected behavior.

### DIFF
--- a/tools/mksurfdata_esmf/gen_mksurfdata_namelist.py
+++ b/tools/mksurfdata_esmf/gen_mksurfdata_namelist.py
@@ -285,16 +285,18 @@ def main ():
         element_count = mesh_file.dimensions['elementCount'].size
         if 'origGridDims' in mesh_file.variables:
             orig_grid_dims = mesh_file.variables['origGridDims']
-            force_model_mesh_nx = orig_grid_dims[0]
-            force_model_mesh_ny = orig_grid_dims[1]
-            mesh_file.close()
-            important_msg = 'Found data for force_model_mesh_nx and ' \
-                            'force_model_mesh_ny in the mesh file so ' \
-                            'IGNORING ANY CORRESPONDING USER-ENTERED VALUES.'
-            logger.info(important_msg)
+            if int(force_model_mesh_nx) == orig_grid_dims[0] and \
+               int(force_model_mesh_ny) == orig_grid_dims[1]:
+                mesh_file.close()
+            else:
+                error_msg = 'ERROR: Found variable origGridDims in ' \
+                            f'{force_model_mesh_file} with values ' \
+                            f'{orig_grid_dims[:]} that do not agree with the ' \
+                            'user-entered mesh_nx and mesh_ny values of ' \
+                            f'{[force_model_mesh_nx, force_model_mesh_ny]}.'
+                sys.exit(error_msg)
         elif force_model_mesh_nx is None or force_model_mesh_ny is None:
-            error_msg = 'ERROR: You set --model-mesh but the file does not ' \
-                        'contain the variable origGridDims, so you MUST ALSO ' \
+            error_msg = 'ERROR: You set --model-mesh so you MUST ALSO ' \
                         'SET --model-mesh-nx AND --model-mesh-ny'
             sys.exit(error_msg)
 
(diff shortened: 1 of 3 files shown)


slevisconsulting:
@ekluzek this PR could be ready for review and merging UNLESS you think I should fold #1776 here, too. I wouldn't mind doing that rather than opening a separate PR for that. Whatever you think makes more sense.

ekluzek:
@slevisconsulting this likely means you haven't pulled all the tags in both of your clones. If you fetch everything in both clones they will likely be consistent. The tags it's pointing to are the ESCOMP tags, so fetch from ESCOMP, and you might need to use the "-t" option to make sure you pull all the tags.

slevisconsulting:
> @slevisconsulting this likely means you haven't pulled all the tags in both of your clones. If you fetch everything in both clones they will likely be consistent. The tags it's pointing to are the ESCOMP tags, so fetch from ESCOMP, and you might need to use the "-t" option to make sure you pull all the tags.

```
Interesting. The -t option returned this:

> git fetch -t escomp
From https://github.com/ESCOMP/ctsm
 ! [rejected]            alpha-ctsm5.2.mksrf.04_ctsm5.1.dev090 -> alpha-ctsm5.2.mksrf.04_ctsm5.1.dev090  (would clobber existing tag)
 ! [rejected]            alpha-ctsm5.2.mksrf.05_ctsm5.1.dev090 -> alpha-ctsm5.2.mksrf.05_ctsm5.1.dev090  (would clobber existing tag)
```

If it's not concerning, I can leave as is and make the next tag (after we go over my mods) from izumi in case it makes a difference.

ekluzek:
That sounds like your local tag in your clone is in disagreement with the tag on ESCOMP. I would trust the tag on ESCOMP and be suspicious of the tag on your local clone. You could look into it in more detail to see which is right, and if the ESCOMP ones are actually correct. But, even if the tags are wrong we are unlikely to do anything about it for these branch tags. But, you probably don't want to have a tag on your local clone that disagrees with ESCOMP.

tools/mksurfdata_esmf/gen_mksurfdata_namelist.py
```diff
@@ -274,8 +279,32 @@ def main ():
...
+        element_count = mesh_file.dimensions['elementCount'].size
+        if 'origGridDims' in mesh_file.variables:
+            orig_grid_dims = mesh_file.variables['origGridDims']
+            force_model_mesh_nx = orig_grid_dims[0]
+            force_model_mesh_ny = orig_grid_dims[1]
+            mesh_file.close()
+            important_msg = 'Found data for force_model_mesh_nx and ' \
+                            'force_model_mesh_ny in the mesh file so ' \
+                            'IGNORING ANY CORRESPONDING USER-ENTERED VALUES.'
+            logger.info(important_msg)
```
slevisconsulting, line 294: @ekluzek request:
if orig_grid_dims disagrees with user-entered values, then abort with error message
if agrees, go on

tools/mksurfdata_esmf/gen_mksurfdata_namelist.py
```diff
@@ -433,14 +467,24 @@ def main ():
...
-                    f'valid grid values are {valid_grids}'
-        sys.exit(error_msg)
+        if res in valid_grids:
+            error_msg = 'ERROR: You have requested a valid grid for which ' \
+        '../../ccs_config/component_grids_nuopc.xml does not include a mesh ' \
+        'file. For a regular regional or 1x1 grid, you may generate the ' \
+        'fsurdat file using the subset_data tool instead. Alternatively ' \
+        'and definitely for curvilinear grids, you may generate ' \
+        'a mesh file using the workflow currently (2022/7/6) described in ' \
+        'https://github.com/ESCOMP/CTSM/issues/1773#issuecomment-1163432584'
```
slevisconsulting, line 482: @ekluzek comment:
add TODO that this workflow should be included in the User's Guide.
add similar comment in the issue.

Issue #1580, opened 2021-12-17
Add an action to test python code through different python versions


ekluzek:
We talked about this at the CTSM software meeting this morning. Something that wouldn't be hard to do would be to run the python unit tests in the python directory through different versions of python that we want to support. I would envision this would try with a list of python versions like:

3.7
3.8
3.9
3.10

As new versions are available we can add them to the list. If there's a reason to try micro versions within a major version that could be done as well. If the minimum version required is advanced, older versions could be taken off the list.

ekluzek:
There's some discussion here

https://github.com/ESCOMP/CTSM/pull/1419#issuecomment-1018975634

in PR #1419 about sensitivity that likely has to do with python versions (or python environment or user environment). This makes me think that possibly this issue should be elevated, so we get more information on these type of issues.

ekluzek:
Note, one interesting thing I noticed with this is that when I test with different versions of python on cheyenne (using conda) pylint can show errors in one version, but not others.

ekluzek:
The bit longer term solution for the pylint issues will be that @negin513 will setup some conda environments that we can use. We should have a few different python versions for those test environments as well.

In the short term I think the standard should be to do this on cheyenne:

module load python.   # Which gives you the default python version, currently 3.7.9)

@negin513 is going to teach us about python environments and also get the environments setup in a future PR. This part of this issue should be a separate issue.
ncar_pylib

negin513:
Similar issue also came up in the discussion of PR#1419:
@slevisconsulting comment about this : https://github.com/ESCOMP/CTSM/pull/1419#issuecomment-1013772176
My response on what possibly causing this issue: https://github.com/ESCOMP/CTSM/pull/1419#issuecomment-1018975634

ekluzek:
I have a conda environment file that makes some progress here for ctsm5.1.dev100. It looks like merely moving the python version ahead doesn't cause it to break. And numpy and xarray can go to the current latest version for example. Using a 3.10 version of python is a problem though, and pylint must be carefully controlled or else it shows errors.

But, for example the following requirements file does work:

```
python>=3.9.13,<3.10    # Moving to 3.10 runs into conflicts
scipy
netcdf4
requests
numpy>=1.23.0
xarray>=2022.3.0
pylint>=2.8.3,<2.9.0  # Once, you move off of 2.8.3, make lint shows 2 errors, at 2.11.1 there are 2 more errors beyond that
[...]

PR #1798, opened 2022-07-06
Have fsurdat_modifier system test use conda environment rather than ncar_pylib

### LINKED ISSUE #1786: ncar_pylib
looks like CISL is changing how they handle ncar_pylib.  Not sure how this needs to change our workflow for python tools, but I wanted to raise the issue here to consider.  Below is the error I received when activating ncar_pylib
```
**************************
*
*   DEPRECATION NOTICE!
*
*   The "ncar_pylib"-based NCAR Package Library has been deprecated in favor of
*   the newer conda-based NPL. Please migrate your workflows to the conda solution
*   as soon as possible. This older method and associated virtual environments
*   will be removed from the system in the coming months. A specific date will be
*   announced once it is known via the Daily Bulletin.
*
*   See https://arc.ucar.edu/knowledge_base/83853599 for documentation on using the
*   conda-based NCAR Python Library environments.
*
**************************
```

ekluzek:
### Description of changes

Use the manage_python_env script to setup the conda environment and use it rather than ncar_pylib

### Specific notes

  Fixes #1786
  Fixes #1925 

Testing performed, if any:
FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel runs and PASSes on cheyenne


billsacks:
Thanks for working on this. It's not as pretty or simple as I hoped, but it's still an improvement over what we had before - and  I really hope any such mechanism can go away before too long.

Can you confirm two things:
- Does this work both for a user's shell being csh and bash? (given the shell-specific logic you have here)
- Does it work to run FSURDATMODIFYCTSM as one of a suite of tests? I'm both wanting to confirm that earlier tests don't mess anything up about the environment that would interfere with this one (which I think I've seen before) and that this one doesn't mess anything up with the environment that could impact following tests.

ekluzek:
I think I can improve on what I have here, so I'll try to do that. But, I also really like your suggestions on tests as well, so I'll do that.

To simplify this further we'll need conda to be loaded on machines in ccs_config/cime rather than python. That should be OK as when you activate an environment with conda you get a python version with it. But, we've got to convince CESM people that doing that would be OK.

billsacks:
I'm not sure it's worth spending much more time on this. (Sorry: my last comment probably suggested that it would be good to improve this, but I don't really think that should be a priority.) The vision we have discussed in CSEG is that users will load an appropriate python environment before doing anything with CESM / CIME. We don't want CIME to be responsible for controlling the python environment. Once we are comfortable with that (maybe the switch to derecho will be a good time???), I believe we can remove any 'module load python' from CIME. (I don't think we want a conda load in ccs_config/cime: I think we want to leave this up to the user - at least based on previous discussions.) So I feel like a less-than-elegant solution is okay for now as long as it works, and then we can hopefully remove this soon.

billsacks:
Also: I appreciate that this new solution works on multiple machines, not just cheyenne.

billsacks (approved):
Approving with the testing caveats noted in my earlier comments. Thanks for your work on changing this so it will work on other systems as well as not relying on the deprecated ncar_pylib!

slevisconsulting:
> Thanks for working on this. It's not as pretty or simple as I hoped, but it's still an improvement over what we had before

Thanks, @ekluzek 
It's not a lot of code, but I doubt that I would have figured this out without help :-)
Let me know if you'd like me to try any or all of the testing that @billsacks recommended.

ekluzek:
Oh, I had forgotten about this one too. Thanks @billsacks

slevisconsulting:
> @slevisconsulting I'm not going to be able to get back to working on this one for awhile, I wonder if you shouldn't merge this into your PR and shepherd bringing it in to CTSM main?

@ekluzek looks as though I can push directly to your PR, so I will work here for now, and we will likely close the redundant PR that I opened (#1925)

ekluzek:
@slevisconsulting I think some of the documentation changes you have in #1925, might still apply. So you could merge this into that, or the other way around, whichever you prefer. I'm glad you'll be able to work on it though, it would be nice to see this come in.

slevisconsulting:
Out of the box
`./create_test FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev115`
fails with
```
Lmod has detected the following error: The following module(s) are unkno                                                                                                                                  wn:
        "lang/python"

        Please check the spelling or version number. Also try "module spider ...                                                                                                                                  "
        It is also possible your cache file is out-of-date; it may help to try:
          $ module --ignore-cache load "lang/python"

        Also make sure that all modulefiles written in TCL start with the string
        #%Module



        python3: can't open file '/glade/work/slevis/git/mksurfdata_toolchain/to                                                                                                                                  ols/modify_fsurdat/fsurdat_modifier': [Errno 2] No such file or directory
          File "./case.build", line 256, in <module>
            _main_func(__doc__)
          File "./case.build", line 216, in _main_func
          File "/glade/work/slevis/git/mksurfdata_toolchain/cime_config/SystemTe                                                                                                                                  sts/fsurdatmodifyctsm.py", line 44, in __init__
            self._run_modify_fsurdat()
          File "/glade/work/slevis/git/mksurfdata_toolchain/cime_config/SystemTe                                                                                                                                  sts/fsurdatmodifyctsm.py", line 74, in _run_modify_fsurdat
            subprocess.run( conda_env+"python3 "+tool_path+" "+self._cfg_file_pa                                                                                                                                  th, shell=True, check=True)
          File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/lib/python3.7/subpro                                                                                                                                  cess.py", line 512, in run
            output=stdout, stderr=stderr)
        subprocess.CalledProcessError: Command '. /glade/scratch/slevis/FSURDATM                                                                                                                                  ODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.C.20221215_                                                                                                                                  143118_58x138/.env_mach_specific.sh; module unload python; module load conda; mo                                                                                                                                  dule load lang/python; /glade/work/slevis/git/mksurfdata_toolchain/py_env_create                                                                                                                                   && conda activate ctsm_pylib && python3 /glade/work/slevis/git/mksurfdata_toolc                                                                                                                                  hain/tools/modify_fsurdat/fsurdat_modifier /glade/scratch/slevis/FSURDATMODIFYCT                                                                                                                                  SM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.C.20221215_143118_                                                                                                                                  58x138/modify_fsurdat.cfg' returned non-zero exit status 2.
```
In the last line, I think I need to remove the `module load lang/python` based on how I got fsurdat_modifier to work in interactive mode.

slevisconsulting:
Previous test passed. Next testing the full clm_pymods test-suite
`./run_sys_tests -s clm_pymods -c ctsm5.1.dev115 --testroot-base /glade/scratch/slevis --skip-generate`

slevisconsulting:
In the test-suite, the same test stopped with:
`FAIL FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel SHAREDLIB_BUILD failed to initialize`
The TestStatus.log suggested a retry, but I got the same error a second time:
```
CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://conda.anaconda.org/ncar/linux-64/current_repodata.json>
Elapsed: -

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
'https//conda.anaconda.org/ncar/linux-64'


  File "./case.build", line 256, in <module>
    _main_func(__doc__)
  File "./case.build", line 216, in _main_func
  File "/glade/work/slevis/git/mksurfdata_toolchain/cime_config/SystemTests/fsurdatmodifyctsm.py", line 44, in __init__
    self._run_modify_fsurdat()
  File "/glade/work/slevis/git/mksurfdata_toolchain/cime_config/SystemTests/fsurdatmodifyctsm.py", line 74, in _run_modify_fsurdat
    subprocess.run( conda_env+"python3 "+tool_path+" "+self._cfg_file_path, shell=True, check=True)
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/lib/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '. /glade/scratch/slevis/tests_1215-151410ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.C.1215-151410ch_int/.env_mach_specific.sh; module unload python; module load conda; /glade/work/slevis/git/mksurfdata_toolchain/py_env_create && conda activate ctsm_pylib && python3 /glade/work/slevis/git/mksurfdata_toolchain/tools/modify_input_files/fsurdat_modifier /glade/scratch/slevis/tests_1215-151410ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.C.1215-151410ch_int/modify_fsurdat.cfg' returned non-zero exit status 254.
```
@billsacks you alluded to the possibility of encountering problems in the test-suite, though this test is one of two in clm_pymods and it runs first. I have no ideas how to solve this and would welcome suggestions.

slevisconsulting:
Just in case, I reran the whole clm_pymods test-suite, because earlier I just reran the failed test like this:
`./run_sys_tests -s clm_pymods --rerun-existing-failures --testid-base 1215-151410ch -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev115 --skip-generate`
However, I get the same error...

billsacks:
This sounds like a different issue than what I was concerned about (in terms of running in a test suite). My guess is that you're encountering this when running via run_sys_tests because run_sys_tests does the test builds on the compute nodes, whereas I'm guessing that, with your first test, you did the build on the login node. You could verify this by getting an interactive session on a compute node and trying to run the relevant conda commands from there. I'm guessing that the compute nodes don't have the same access permissions to the outside world as the login nodes do.

As for how to solve this, my initial thoughts are: You could first email CISL help (help@ucar.edu), asking them if they can change some permissions to allow this (or if they have thoughts about a way to get this to work from the compute nodes). 

If that won't work, then I wonder if we could do something like have run_sys_tests set up the appropriate conda environment for you, and then just have the tests load the already-set-up conda environment. My main concern with that approach is that, if this setting up of the conda environment takes more than a few seconds, I would ideally want this done in a way that doesn't significantly delay the starting of the system tests – especially in cases where you're just running one or a few tests that don't need this conda environment, so this delay would just be a pain.

If setting up the conda environment takes a while and we can't come up with a way to avoid having this setup delay the start of the system tests, then maybe we need to consider requiring anyone running these tests to have the appropriate conda environment already created. I don't like that that makes running the tests harder, but maybe that's our only viable option.

billsacks:
> If that won't work, then I wonder if we could do something like have run_sys_tests set up the appropriate conda environment for you, and then just have the tests load the already-set-up conda environment. My main concern with that approach is that, if this setting up of the conda environment takes more than a few seconds, I would ideally want this done in a way that doesn't significantly delay the starting of the system tests – especially in cases where you're just running one or a few tests that don't need this conda environment, so this delay would just be a pain.
> 
> If setting up the conda environment takes a while and we can't come up with a way to avoid having this setup delay the start of the system tests, then maybe we need to consider requiring anyone running these tests to have the appropriate conda environment already created. I don't like that that makes running the tests harder, but maybe that's our only viable option.

Or maybe we could do a combination of these two options: If run_sys_tests sees that you already have an appropriately-named conda environment, then it won't do anything; but if you don't yet have one, then it creates a conda environment for you. That way first time users of run_sys_tests will get a CTSM conda environment created for them automatically, and after that it will reuse the existing environment. Major caveat that this is coming from someone with almost no experience with conda environments, so it may not actually work the way I'm imagining it could....

slevisconsulting:
I will start with an email to help@ucar.edu and we'll see how it goes.

ekluzek:
@billsacks the one problem with just checking that a given conda environment exists is that we do expect the conda environment to be updated (hopefully infrequently). So the only way to be sure is to construct the environment each time. You might be right that maybe run_sys_test should create the environment and then system tests just use it though. It does take an annoying amount of time to create the environment, so you wouldn't want to unnecessarily run conda all the time. Maybe one way to do it would be to have run_sys_tests to create the conda environment by default, but have an option to skip it with a command line option.  Since, run_sys_tests runs on the login nodes this also bypasses needing to get this to work on the compute nodes. If it's a problem for cheyenne, I imagine it could be a problem on other machines as well.

slevisconsulting:
I'm ready to email help@ucar.edu but I will hold off in case we have an acceptable solution in the above posts.
@ekluzek how about we take a few minutes at the end of the SLIM stand-up tomorrow to discuss?

billsacks:
> the one problem with just checking that a given conda environment exists is that we do expect the conda environment to be updated (hopefully infrequently). So the only way to be sure is to construct the environment each time. You might be right that maybe run_sys_test should create the environment and then system tests just use it though. It does take an annoying amount of time to create the environment, so you wouldn't want to unnecessarily run conda all the time. Maybe one way to do it would be to have run_sys_tests to create the conda environment by default, but have an option to skip it with a command line option. Since, run_sys_tests runs on the login nodes this also bypasses needing to get this to work on the compute nodes. If it's a problem for cheyenne, I imagine it could be a problem on other machines as well.

My hope would be that we aren't so tied to specific versions of python packages that we often *need* to update the ctsm python environment. Can we start with something simple like using the environment if it exists and creating it if it doesn't, and leave it up to the user to update this environment if it's ever needed?

But this also feels like it's getting messy. Something I seem to remember from discussions in CSEG meetings is the suggestion that, at least on our standard platforms (cheyenne & izumi), we should try to get the appropriate python environments supported at the system level. Have you looked into whether the environment that you get from "conda activate npl" contains everything we need? If it doesn't, would it be feasible to ask CISL to install any missing packages – or try to remove those dependencies from our python code if they are non-standard enough to not be included in npl? Then for testing we could use that system-level environment? I don't know, maybe that creates as many problems as it solves, but this management of python environments is just feeling like such a headache.

ekluzek:
I think checking for the environment and using it does make sense. That means they've at least run py_env_create once. And the environment isn't changing that often, so it's probably OK most of the time. And if there is a missing package it will abort, which would be the most common thing. That would just mean that they need to rerun py_env_create again on the login node. Just activating the environment is also fast, so that would have reasonable performance.

On using npl -- I actually think we should NOT do that as I view it as a requirement to get our tools to work on all systems and not just NCAR systems. The advantage of using conda is that we can provide an environment that someone can use for ANY machine. We will only test on cheyenne and izumi, but in principle the conda environment will allow you to work on any machine. Since, we do want to support the university community as well as people using any of the supported CESM machines in ccs_config I think this makes sense as a requirement.

slevisconsulting:
@ekluzek and I met and agreed on the following:
- This test should stop using py_env_create.
- I will add graceful aborts for when conda doesn't exist and when ctsm_pylib doesn't exist telling the user that they need to first manually install conda, unload python, load conda, and run py_env_create for this test to work.
- I will add text when the tool itself fails, suggesting that the user's ctsm_pylib environment may be out-of-date, and that running py_env_create may resolve the issue.

ekluzek:
@slevisconsulting note I did update the name of the conda environment to ctsm_pylib in this PR, so be sure to use it.

slevisconsulting:
This works now, though I haven't found a way to separate the single subprocess that
- loads conda and activates the ctsm_pylib environment AND
- runs the fsurdat_modifier tool

into two or more subprocesses without losing access to the ctsm_pylib environment. So instead I have written a more elaborate error message with the various possible scenarios leading to an error in the subprocess. I have tested it by commenting out the "conda activate" line and confirming that I see my error message in TestStatus.log.

If that's acceptable because we're still only looking at an interim solution, then I think that this PR is ready for pre-merge preparations:
- I need to make sure pylint passes
- Should I run aux_clm? I don't know if any of the modified files affect anything outside of clm_pymods...

billsacks:
Okay, all sounds good - thank you @slevisconsulting and @ekluzek . It might be good to confirm that this works in the context of a larger test suite, if you haven't already. This wouldn't necessarily need the full aux_clm, but you could create a text-based testlist file containing this test and a few others – with one or more tests before this one and one or more after it – to make sure this doesn't interfere with the python environment loaded by the test system and vice versa.

test/tools/test_driver.sh
```diff
@@ -54,6 +54,8 @@ module load nco
 module load ncl
```
ekluzek, line 54: @slevisconsulting you should remove most of these changes. When you update to the latest version the py_env_create line is actually only done in one place. So the new conda name will also only be in one place.

cime_config/SystemTests/fsurdatmodifyctsm.py
```diff
@@ -65,54 +65,49 @@ def _create_config_file(self):
...
-            return match_line.group(1)
-
-    return None
+    def _get_conda_env(self):
+        #
+        # Add specific commands needed on different machines to get conda available
+        # Use semicolon here since it's OK to fail
+        #
+        # Remove python and add conda to environment for cheyennne
+        conda_env = "module unload python; module load conda"
```
ekluzek, line 104: @slevisconsulting what happens when you try running on izumi? That's one advantage of using a conda environment we should be able to get it work on izumi as well.

.github/workflows/python_testing.yml
```diff
@@ -0,0 +1,31 @@
...
+            manage_externals/checkout_externals
+        - name: chdirectory python
+          run: |
+              cd python
+        - name: Run python tests
+          run: |
+              make test
+        - name: Run pylint
+          run: |
+              make lint
```
ekluzek, line 31: @slevisconsulting this is a new action for running python testing on pushes and PR's. It looks like it's failing right now, so it needs more work. Is this something you'd like to dig into to get working? Maybe we should meet to discuss it next week....

ekluzek: This is working on #1580

cime_config/SystemTests/fsurdatmodifyctsm.py
slevisconsulting: I checked out the branch on izumi and submitted the same test by changing cheyenne to izumi in the suffix:
`./create_test FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.izumi_intel`

First attempt gave this error:
```
        ModuleCmd_Load.c(213):ERROR:105: Unable to locate a modulefile for 'conda'

        CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
        If your shell is Bash or a Bourne variant, enable conda for the current user with

            $ echo ". /usr/local/anaconda-3.7.0/etc/profile.d/conda.sh" >> ~/.bashrc

        or, for all users, enable conda with

            $ sudo ln -s /usr/local/anaconda-3.7.0/etc/profile.d/conda.sh /etc/profile.d/conda.sh

        The options above will permanently enable the 'conda' command, but they do NOT
        put conda's base (root) environment on PATH.  To do so, run

            $ conda activate

        in your terminal, or to put the base environment on PATH permanently, run

            $ echo "conda activate" >> ~/.bashrc

        Previous to conda 4.4, the recommended way to activate conda was to modify PATH in
        your ~/.bashrc file.  You should manually remove the line that looks like

            export PATH="/usr/local/anaconda-3.7.0/bin:$PATH"

        ^^^ The above line should NO LONGER be in your ~/.bashrc file! ^^^
```
For the second attempt, I tried my usual approach:
```
module unload lang/python/3.7.0
conda activate ctsm_py
```
and got the same error, which makes sense in retrospect, since `module avail` shows no available conda modules.

@ekluzek I'm not confident about following the instructions in the above error msg, but I'm happy to try if you recommend it.

slevisconsulting:
> Okay, all sounds good - thank you @slevisconsulting and @ekluzek . It might be good to confirm that this works in the context of a larger test suite, if you haven't already. This wouldn't necessarily need the full aux_clm, but you could create a text-based testlist file containing this test and a few others – with one or more tests before this one and one or more after it – to make sure this doesn't interfere with the python environment loaded by the test system and vice versa.

I ran aux_clm on cheyenne and it passed other than expected failures. I did not get motivated to try a custom list of tests :-)

billsacks:
> I ran aux_clm on cheyenne and it passed other than expected failures. I did not get motivated to try a custom list of tests :-)

Awesome, thanks @slevisconsulting

test/tools/test_driver.sh
ekluzek: Ahhh, OK. Yes, actually this is done on line 384 for all machines, so there is no need for each machine. So yes, remove the lines you added such as 57 and 58 to run py_env_create and conda activate.

.github/workflows/python_testing.yml
slevisconsulting: @ekluzek and I agreed that this is not within the scope of this PR, so we're leaving unresolved.

cime_config/SystemTests/fsurdatmodifyctsm.py
slevisconsulting: @ekluzek and I confirmed that we both get the same behavior on izumi:
`./py_env_create` works
`conda activate ctsm_pylib` works
BUT we both need to `module unload` python for the fsurdat_modifier tool to work correctly.

Regardless, I will implement code in the test that will check for conda ("which conda") and when this fails (as expected when running the test-suite on cheyenne) we will execute module commands and when it passes we won't.

slevisconsulting:
Tested the latest commit in clm_pymods, as well as with ./create_test and both passed.
@ekluzek please let me know if you agree with my implementation.

.github/workflows/python_testing.yml
slevisconsulting: Remove python_testing.yml from this PR

slevisconsulting:
@ekluzek and @slevisconsulting agreed this morning (2023/1/24) that this PR will likely get merged/tagged as dev118, pending successful testing once I update to dev117.

slevisconsulting:
@ekluzek I think something has changed in my permissions here. I updated to dev117, but the push to the PR has been failing. If nothing has changed on your side, we may need to troubleshoot tomorrow morning after Stand-up.

slevisconsulting:
Meanwhile
IN PROGRESS `./run_sys_tests -s clm_pymods -c ctsm5.1.dev117 --skip-generate`
and
PASS `make all` from the /python directory

slevisconsulting:
I have tried once more doing everything from scratch and the FSURDAT system test still fails.

This is how I resolved the conflicts:

1) In `test/tools/test_driver.sh` I'm keeping escomp/master:
```
# Setup conda environement
<<<<<<< HEAD
\$CLM_ROOT/py_env_create
conda activate ctsm_pylib
||||||| 17e2acb6a
\$CLM_ROOT/py_env_create
conda activate ctsm_py
=======
conda activate ctsm_py
if [ \$? -ne 0 ]; then
   echo "ERROR: Trouble activating the ctsm_py conda environment, be sure it's setup with \$CLM_ROOT/py_env_create, then rerun"
   exit 4
fi
>>>>>>> escomp/master
```
but changing ctsm_py to ctsm_pylib as I did elsewhere in this PR.

2) In `cime_config/SystemTests/fsurdatmodifyctsm.py` I'm keeping HEAD:
```
<<<<<<< HEAD

        self._case.load_env(reset=True)
        conda_env = ". "+self._get_caseroot()+"/.env_mach_specific.sh; "
        # Preprend the commands to get the conda environment for python first
        conda_env += self._get_conda_env()
        # Source the env
        try:
            subprocess.run( conda_env+"python3 "+tool_path+" "+self._cfg_file_path, shell=True, check=True)
        except subprocess.CalledProcessError as error:
            print("ERROR while getting the conda environment and/or ")
            print("running the fsurdat_modifier tool: ")
            print("(1) If your ctsm_pylib environment is out of date or you ")
            print("have not created the ctsm_pylib environment, yet, you may ")
            print("get past this error by running ./py_env_create ")
            print("in your ctsm directory and trying this test again. ")
            print("(2) If conda is not available, install and load conda, ")
            print("run ./py_env_create, and then try this test again. ")
            print("(3) If (1) and (2) are not the issue, then you may be ")
            print("getting an error within the fsurdat_modifier tool itself. ")
            print("Default error message: ")
            print(error.output)
        except:
            print("ERROR trying to run fsurdat_modifier tool.")
            raise
||||||| 17e2acb6a
        # Need to specify a specific python version that has the required
        # dependencies
        python_path = _get_python_path()
        subprocess.check_call([python_path, tool_path, self._cfg_file_path])
=======
        # Need to specify a specific python version that has the required
        # dependencies
        python_path = _get_python_path()
        subprocess.check_call([python_path, tool_path, self._cfg_file_path, "--verbose", "--overwrite", "--allow_ideal_and_include_non_veg"])
>>>>>>> escomp/master
```
I now suspect that one or more new things in this PR fail for some reason. New items include:
- `load_env`
- `.env_mach_specific.sh`
- `which conda`
- `module unload python; module load conda`
- `conda activate ctsm_pylib`

Also, I still get an error when I try to push to the PR:
```
ERROR: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'https://github.com/ekluzek/CTSM.git'
```

billsacks:
Regarding the permissions: I have noticed that there is a bug with git-lfs that can prevent you from pushing to other people's forks. The workaround is to do a `git lfs uninstall`, then `git push`, then restore with `git lfs install`.

ekluzek:
I think the permission is just the problem with making @slevisconsulting a collaborator on my fork, which we've resolved now.

In test driver you should change the use of ctsm_py to ctsm_pylib as follows...

``` bash
conda activate ctsm_pylib
if [ \$? -ne 0 ]; then
   echo "ERROR: Trouble activating the ctsm_py conda environment, be sure it's setup with \$CLM_ROOT/py_env_create, then rerun"
   exit 4
fi
```

in fsurdatmodifyctsm.py you do need to change the line in

``` python
        try:
            subprocess.run( conda_env+"python3 "+tool_path+" "+self._cfg_file_path, shell=True, check=True)
        except subprocess.CalledProcessError as error:
```

to

``` python
        try:
            subprocess.check_call(conda_env+"python3 "+tool_path+" "+self._cfg_file_path+"--verbose --overwrite --allow_ideal_and_include_non_veg", shell=True, check=True)
        except subprocess.CalledProcessError as error:
```

Do you see why those need to change? Just want to make sure this makes sense to you...

slevisconsulting:
@ekluzek I think I have found the source of the problem. For some reason it eluded me until now. I have discovered this message, and wanted to ask whether you had a specific reason to add this requirement or if simply it made sense to you:
`ERROR: idealized AND include_nonveg can NOT both be on, pick one or the other`

slevisconsulting:
If you had a specific reason, then I will keep it. If not, then I would rather remove this requirement.

slevisconsulting:
I now see (probably saw in the review, as well, but forgot) that you had added this option:
--allow_ideal_and_include_non_veg
I will try to resolve using this option.

ekluzek:
@slevisconsulting the reason I thought that idealized and include_nonveg should be on together is because idealized explicitly sets the dominant PFT to zero, while include_nonveg (actually false) zeros' out the PFT's. So they contradict each other. It looks like I got the logic wrong though -- and that combination should be allowed. First I got it wrong about it being about include_nonveg True, when it should be False. But, also I got it wrong that include_nonveg=False isn't zeroing vegetation -- just all the other landunit types -- so that's actually fine.

So you should remove that restriction and option I added.

The one that still makes sense is idealized and dom_pft, because idealized explicitly sets the dominant PFT to bare-soil, while dom_pft sets it to the value the user asks for. What behavior you get is then dependent on how it's implemented in the code -- but since it's ambiguous that means it's something that should be checked for.

Sorry I added a non useful restriction there. It is good to think these things through and make sure options make sense together and restrict ones that don't.

slevisconsulting:
@ekluzek I thought about it some more and decided that I also disagree about keeping the second restriction. The comments in the configure file explain why:
```
# PFT/CFT to be set to 100% according to user-defined mask.
# If idealized = True and dom_pft = UNSET, the latter defaults to 0
# (bare soil). Valid values range from 0 to a max value (int) that one can
# obtain from the fsurdat_in file using ncdump (or method preferred by user).
# The max valid value will equal (lsmpft - 1) and will also equal the last
# value of cft(cft).
dom_pft = UNSET
```
In other words, dom_pft is meant to be zero if the user leaves it UNSET or if the user explicitly sets it to zero.

slevisconsulting:
> Regarding the permissions: I have noticed that there is a bug with git-lfs that can prevent you from pushing to other people's forks. The workaround is to do a `git lfs uninstall`, then `git push`, then restore with `git lfs install`.

Thanks @billsacks .
@ekluzek made me collaborator, so I am able to push again. 
My only confusion is that I had pushed to this PR previously (e.g. Dec 15-19). Do collaborator permissions expire?

billsacks:
> @ekluzek made me collaborator, so I am able to push again.
> My only confusion is that I had pushed to this PR previously (e.g. Dec 15-19). Do collaborator permissions expire?

I really think this is due to the lfs issue. There are two ways you can have permission to push to someone else's fork:

(1) If they make you a collaborator. This works even if you have git lfs installed.

(2) If you are a "maintainer" of CTSM (which I think means you have write permission), then by default you are given permission to push to any branch for which there is a PR. (Authors can choose not to give this permission, but the default is to give it. For this pr, it is allowed: I see the message "Maintainers are allowed to edit this pull request" in the metadata on the top right of the PR.) This feature is broken by git lfs for some reason.

ekluzek:
@slevisconsulting that makes sense, and nicely documented in the file. Thanks for pointing that out. 

It might be good to add a comment in the code that the dom_pft section HAS to be after the the idealized part for those reasons above. You could also add a comment to the idealized section that this should be done first, so that other options could override the behavior there. The idea that the idealized section is done and other things override its behavior then becomes an expectation of how the script is supposed to function. It's good to make all the expectations on how the script is intended to behavior to be as clear as possible.

slevisconsulting:
- Tests now PASS:
clm_pymods test-suite and `make all` in /python directory
I linked the dev117 baselines to dev118: `ln -s ctsm5.1.dev117 ctsm5.1.dev118`
- I have now also updated the ChangeLog
- @ekluzek I have added comments in the code as you suggested. Pls let me know if you have further concerns. Otherwise I think this is ready for the merge.

Issue #1789, opened 2022-06-29
A bug in calculating accumulated fields (24/240 hours averaged) when using a smaller timestep


Duseong:
### Brief summary of bug

It looks like there's a bug in calculating accumulated fields (e.g., TV24, TV240, PAR24_sun, PAR240_sun, etc.) when using a smaller timestep. 

### General bug information

**CTSM version you are using:**
ctsm5.1.dev019 (cam6_3_018)

**Does this bug cause significantly incorrect results in the model's science?**
Yes

**Configurations affected:**
Regional refinement simulations or any configurations that use a much smaller timestep than the global 1-degree model.

### Details of bug

Please see the attached file to see the time series of variables (TV, TV24, TV240, PAR_sun, PAR24_sun, PAR240_sun) at Manaus point over Amazon. Black dots are from the ne30 simulation, and red dots are from the same ne30 simulation but with a different timestep (3.75 minutes -> ATM_NCPL = LND_NCPL of 384). 
There are unexpected big fluctuations in 24/240 hours averaged fields in the smaller timestep run, so I guess there is a bug when calculating those fields. 
Since regional refinement simulations use smaller timestep, results from those models are heavily affected by these fields. I found that global biogenic emissions, OH, and other chemical fields like CO were changed substantially. It may also affect other atmospheric fields. 
[Accumulated_fields_time_series.pdf](https://github.com/ESCOMP/CTSM/files/9014924/Accumulated_fields_time_series.pdf)

### Important details of your setup / configuration so we can reproduce the bug

I changed ATM_NCPL (=LND_NCPL) from 48 to 384 to see the effects of the timestep on those fields. The spatial resolution was the same for both (ne30).

billsacks:
Thank you for opening this issue. I agree that this appears to be a significant bug!

From looking through the code, I think I see what's happening here: the PERIOD of accumulated fields is read from the initial conditions file, but I don't think it needs to be, and doing so is problematic if the run you're doing uses a different time step than the time step used in creating the initial conditions file originally.

Can you try redoing your test after rebuilding the code with this block of code deleted:

https://github.com/escomp/CTSM/blob/master/src/main/accumulMod.F90#L767-L776

If that gives other problems, then a simpler but safer experiment would be to redo your test after setting `./xmlchange CLM_FORCE_COLDSTART=on` so that the model doesn't use a restart file at all. (That won't be good for science, but if my hypothesis is right, then it should get around this issue. I'd like to see if that's true.)

[...]

PR #1802, opened 2022-07-11
Remove accumulation fields' PERIOD from the restart file

billsacks:
### Description of changes

This doesn't need to be on the restart file, and having it read from the
restart file causes wrong behavior when changing the model time step.

See https://github.com/ESCOMP/CTSM/issues/1789 for details

### Specific notes

Resolves ESCOMP/CTSM#1789

Are answers expected to change (and if so in what way)? YES - for any case where the time step differs from the one originally used to create the initial conditions file. This typically means changing answers for any case with a time step other than 1/2 hour.

Testing performed, if any: None yet; about to run test suite

### DIFF
--- a/src/main/accumulMod.F90
+++ b/src/main/accumulMod.F90
@@ -764,17 +764,6 @@ contains
                data=accum(nf)%nsteps, readvar=readvar)
        end if
 
-       if (accum(nf)%old_name /= "") then
-          varname = trim(accum(nf)%name) // '_PERIOD:' // trim(accum(nf)%old_name) // '_PERIOD'
-       else
-          varname = trim(accum(nf)%name) // '_PERIOD'
-       end if
-       call restartvar(ncid=ncid, flag=flag, varname=varname, xtype=ncd_int, &
-            long_name='', units='time steps', &
-            imissing_value=ispval, ifill_value=huge(1), &
-            interpinic_flag='copy', &
-            data=accum(nf)%period, readvar=readvar)
-
     end do
 
   end subroutine accumulRest


billsacks:
I have spent more time looking at the relevant code, with a particular focus on whether new problems / inconsistencies will arise from my proposed fix of removing the accumulation fields' `period` from the restart file (e.g., will there be problems because now we're using a newly-initialized `period` value, but still reading each accumulation field's `nsteps` from the restart file?). **My overall conclusion is that there could be some weirdness at the start of a run, but at least for a startup or hybrid run, that weirdness should work itself out within about the first averaging period. A branch or restart run could have some longer-term potential weirdness, so for now I think we should recommend that people NOT change the time step on a branch or restart run. With (significant?) additional work, we could probably avoid this additional weirdness, but my feeling is that it isn't worth the effort right now. In any case, I feel like my proposed fix will bring things much closer to being correct than they currently are when changing the time step.** More details below:

In `extract_accum_field_timeavg`, there is:

``` fortran
    if (mod(nstep,this%period) == 0) then
```

In a branch run, where nstep continues to accumulate, that is going to be weird when changing time steps, but I don't think it will be hugely problematic: I think things will work themselves out after one averaging period. The weirdest thing that might happen is that the averaging periods may no longer be aligned with day boundaries. **I guess this argues for not changing the time step in a branch run, but using a hybrid or startup run instead.**

There is similar potential weirdness in `update_accum_field_timeavg`. That includes some additional potential weirdness due to the use of `this%nsteps`:

``` fortran
    if (mod(nstep,this%period) == 0) then
       do k = begi,endi
          if (this%active(k)) then
             this%val(k,level) = this%val(k,level) / this%nsteps(k,level)
          end if
       end do
    end if
```

But it looks like `this%nsteps` should be reset to 0 (earlier in that subroutine) when `nstep` is 1, so as long as we're in a startup or hybrid case, I *think* this will be okay. If we're in a branch case (where `nstep` has continued from before), then there will be weirdness until we finish out the current accumulation period.

There is also weirdness of a different kind for runmean accumulation (in `update_accum_field_runmean`):

``` fortran
    do k = begi,endi
       if (this%active(k)) then
          kf = k - begi + 1
          this%nsteps(k,level) = this%nsteps(k,level) + 1
          ! Cap nsteps at 'period' - partly to avoid overflow, but also because it doesn't
          ! help us to accumulate nsteps beyond a value of 'period' (because nsteps is
          ! just used to determine accper, and accper needs to be capped at 'period'). A
          ! side-benefit of this capping of nsteps is that accper (the accumulation
          ! period) is always equal to nsteps.
          this%nsteps(k,level) = min(this%nsteps(k,level), this%period)
          accper = this%nsteps(k,level)
          this%val(k,level) = &
               ((accper-1)*this%val(k,level) + field(kf)) / accper
       end if
    end do
```

I think this will be okay when going to *longer* time steps (i.e., decreasing `this%period`), but when moving to *shorter* time steps (increasing `this%period`), the first part of the run will see too-small `this%nsteps`. This isn't a huge problem, but until we have gotten up to `this%nsteps = this%period`, each new time step's value will have a bigger influence on the running mean than it really should. It looks like this problem will exist in startup and hybrid runs as well as branch runs.

-   We could probably fix this problem by allowing `this%nsteps` to get really big rather than capping it at `this%period` (and then capping the local `accper` instead), but I'm not sure it's worth making that change.

`%nsteps` will also be weird at first for a `runaccum` field. However, it looks like `%nsteps` is irrelevant for a `runaccum` field (i.e., it's never used for anything), so that shouldn't be an issue.

PR #1803, opened 2022-07-12
Make zetamaxstable consistently 2.0 when BHS on

### LINKED ISSUE #1689: Set zetamaxstable to 2 consistently for BHS
### Brief summary of bug

The zetamaxstable used when Biomass Heat Storage is on is hardcoded to 100 in FrictionVelocityMod/CanopyFluxesMod, but the zetamaxstable value is used for BareGround, Lake, and Urban. The value of 100 should be used in all of these cases to be consistent.

### General bug information

**CTSM version you are using:** ctsm5.1.dev085 (but goes back to introduction of BHS ctsm5.1.dev021)

**Configurations affected:** When use_biomass_heat_storage is TRUE, including non-FATES clm5_1

### Details of bug

It's hard to say how often this will affect results. It only applies under very stable conditions which might not happen that often in our testing, especially our shorter tests.

One reason this issue came up is because of specific code for BHS rather than using zetamaxstable which is on the namelist.

### LINKED ISSUE #1690: Set and use zetamaxstable for BHS cases
There is special code in CanopyFluxes and FrictionVelocity to essentially hardcode zetamaxstable when Biomass Heat Storage (BHS) is on to 100. What should be done is the namelist value should always be used, but set to 100 when BHS is on. That value of zetamaxstable should be used by all instances including: bare-ground, veg, lake, and urban.

This is blocked by #1689

Some discussion on this is in: #1596

ekluzek:
### Description of changes

Set zetamaxstable to 2.0 when BHS is on for all subgrid types. This also means to remove the hardcoded setting to 100 for vegetation when BHS is on and use zetamaxstable from the namelist.

### Specific notes

  Fixes #1690
  Fixes #1689

Testing performed, if any: namelist testing, will do regular testing


ekluzek:
@olyson can you look this over and make sure this is correct? This uses zetamaxstable from the namelist for Canopy fluxes, so that the same value is consistent for all subgrid types: lake, bareground, veg, and urban. It also sets zetamaxstable to 2.0 whenever BHS is turned on. So this means zetamaxstable will be 2.0 for ctsm5_1 cases.

olyson:
This is already done on Ronny Meier's branch and in such a way as to maintain bfb when z0param_method is not Meier2022:

https://github.com/RonnyMeier/ctsm/commits/main

Is there a reason to do this separately?

ekluzek:
I wanted to bring in the answer changing parts in a small tag, so that the surface roughness tag could be non answer changing.

olyson:
There is another answer-changing aspect of the branch that updates the forcing heights when roughness/displacement height is updated.  This wasn't being done consistently before.  Although I guess you could do that as another answer-changing tag.

ekluzek:
@olyson and I chatted about this a bit so that we understood what each other was doing. He suggested that I ask @wwieder to decide how to move forward with this. I then chatted with @wwieder and we brought in @swensosc and we agreed on going this way so that a consistent zetamaxstable of 2.0 is used when BHS is on for all surface types. @swensosc didn't see a scientific reason to preserve the current behavior of having a different zetamaxstable value over vegetated land types. 100 was actually used in BHS just as a way to say "there is no cap", it was kind of a way of setting it to infinity.

Having a consistent value for all land surfaces makes the code easier to understand, from both a software and scientific perspective.

bld/CLMBuildNamelist.pm
```diff
@@ -3899,10 +3899,11 @@ sub setup_logic_rooting_profile {
 #-------------------------------------------------------------------------------
 
 sub setup_logic_friction_vel {
-  #
+  # Must be after canopyfluxes so that use_biomass_heat_storagge will be set
```
olyson, line 3902: use_biomass_heat_storagge -> use_biomass_heat_storage

olyson (approved):
In that case, this looks good, other than a misspelling.

bld/CLMBuildNamelist.pm
ekluzek: Got it. Thanks.

ekluzek:
I ran some of the longer aux_test suite tests and found that two gave the same answers with only one diverging. Since, the change is to a max value you have to reach that value before answers will change.

This test showed answers that diverge:

And these tests were identical:
SMS_Lm37.f10_f10_mg37.I1850Clm50SpG.cheyenne_intel.clm-glcMEC_long			
SMS_Ly1_Mmpi-serial.1x1_brazil.IHistClm50BgcQianRs.cheyenne_intel.clm-output_bgc_highfreq

Issue #1813, opened 2022-07-18
1D lon/lat in clm history need to be 2D arrays to correctly represent irregular grids


slevis-lmwg:
### Brief summary of bug
In connection with work/simulations documented in #1773:
@ckoven points out that 1D lon/lat variables in clm history do not represent the grid correctly when running on a curvilinear wrf grid.

The correct 2D variables appear in the fsurdat file (LATIXY, LONGXY) and could be used in post-processing for now, until we resolve this issue by outputting the 2D variables to clm history.

### General bug information

**CTSM version you are using:** ctsm5.1.dev103

**Does this bug cause significantly incorrect results in the model's science?** No, but causes trouble in post-processessing unless user takes LONGXY LATIXY from the fsurdat file.

**Configurations affected:** Simulations running on grids with latitudes/longitudes that change along the grid's rows/columns. To be clear, along a row, the latitude does not remain constant, and along a column, the longitude does not remain constant.

### Details of bug

@mvertens points out the area of code where we can address this issue:
`src/cpl/share_esmf/lnd_set_decomp_and_domain.F90` (for example, search for lon1d)
We will discuss code modifications in this issue.

### Important details of your setup / configuration so we can reproduce the bug

The case is here:
`/glade/u/home/slevis/cases_FATES/wrf_CA_01`
and the history output is:
`/glade/scratch/slevis/wrf_CA_01/run/wrf_CA_01.clm2.h0.2000-01.nc`
until I delete it, move it, or it gets scrubbed.

billsacks:
Looking briefly at the code, I noticed that CNDVDriverMod.F90 appears to output longxy and latixy... I haven't looked closely at whether that's what we want, but maybe that just needs to be moved to histfilemod??? Though maybe part of the issue is that this is coming from ldomain and that isn't currently set with nuopc?

slevisconsulting:
@billsacks thank you for this suggestion. Presumably these longxy and latixy are the same as in fsurdat. I will confirm. If so, we could write them to history, checking first that their values are unique in every column and row. And otherwise we still output the 1D variables.

wwieder:
As with #1828, I'd suggest we don't fix this because we have the ability to plot unstructured grids with uxarray and we can also regrid to regular grids with tools I've added to ADF that will be available in CUPiD.

I'll close this PR and the associated issue for now, but others can reopen if I'm being to hasty.

billsacks:
[...]

PR #1828, opened 2022-08-12
Add 2D lat/lon vars to history

slevis-lmwg:
### Description of changes

Add 2D lat/lon vars to src/main/histFileMod.F90
Useful when running on irregular grids.

### Specific notes

@mvertens @ekluzek @billsacks 

#1813 

No but 2D lat/lon variables will now appear in history.

No.

Testing performed, if any:
Ran a case on an irregular grid. Confirmed that 2D lat/lon variables now appear in history. Their x, y coordinates are still wrong.

For the concern of filling up disk space, I could write the 2D lat/lon variables in the first time step only. It's more complicated to figure out whether the 2D variables are needed at all. We'd need to loop through lat values in a row (or lon values in a column) and confirm that there are multiple values (other than the missing value).


ekluzek:
@slevisconsulting there is a convention that these variables (LATIXY and LONGXY) should be in caps. Most of our files that have them are in caps.  From this UNIDATA report I found on a Google search it seems to be a standard as well...

https://rda.ucar.edu/thredds/uddc/files/g/ds314.3/1901_1909/clmforc.cruncep.V7.c2016.0.5d.Prec.1908-08.nc?catalog=https://rda.ucar.edu/thredds/catalog/files/g/ds314.3/1901_1909/catalog.html&dataset=files/g/ds314.3/1901_1909/clmforc.cruncep.V7.c2016.0.5d.Prec.1908-08.nc

slevisconsulting:
This far my code mods are simple and they work BUT I have eliminated the lat/lon coordinates because they were 1D and I haven't, yet, found a way to make them 2D. So I have kept lat/lon as dimensions and LATIXY/LONGXY as variables.

I will need to research this more, and I'm open to suggestions.

ekluzek:
Here's the CF convention which gets into how these conventions work: https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html

billsacks:
For what it's worth, CISM uses the "coordinates" attribute on variables like you mention @slevisconsulting - e.g., see /glade/p/cesmdata/cseg/inputdata/glc/cism/Greenland/greenland_4km_epsg3413_c171126.nc (though that appears to just be on input files, not on the history output files, for whatever reason).

ekluzek:
@slevis-lmwg this seems like a worthwhile thing to still do. What do you think? And should this be in for ctsm6.0.0?

wwieder:
Not sure I really understand this question, as it seems to be focused on irregular grids (see also #1813)

If this is true, I'd suggest we don't fix this because we have the ability to plot unstructured grids with uxarray and we can also regrid to regular grids with tools I've added to ADF that will be available in CUPiD.  

I'll close this PR and the associated issue for now, but others can reopen if I'm being to hasty

PR #1866, opened 2022-10-06
Remove pftmask from history and PFTDATA_MASK from fsurdat files

### LINKED ISSUE #1703: Reduce confusion about fsurdat vars LANDFRAC_PFT, PFTDATA_MASK vs. history vars landfrac, landmask, pftmask

slevisconsulting:
### Description of changes
Simple mods to reduce confusion as explained in #1703 

### Specific notes

CTSM Issues Fixed:
Fixes #1703 

No but pftmask and PFTDATA_MASK will not appear in history and fsurdat files, respectively.

Testing:
Confirmed that fsurdat files do not include PFTDATA_MASK and have no other changes.

### DIFF
--- a/src/main/histFileMod.F90
+++ b/src/main/histFileMod.F90
@@ -3347,17 +3347,6 @@ contains
               imissing_value=ispval, ifill_value=ispval)
        end if
        if (ldomain%isgrid2d) then
-          call ncd_defvar(varname='pftmask' , xtype=ncd_int, &
-              dim1name='lon', dim2name='lat', &
-              long_name='pft real/fake mask (0.=fake and 1.=real)', ncid=nfid(t), &
-              imissing_value=ispval, ifill_value=ispval)
-       else
-          call ncd_defvar(varname='pftmask' , xtype=ncd_int, &
-              dim1name=grlnd, &
-              long_name='pft real/fake mask (0.=fake and 1.=real)', ncid=nfid(t), &
-              imissing_value=ispval, ifill_value=ispval)
-       end if
-       if (ldomain%isgrid2d) then
           call ncd_defvar(varname='nbedrock' , xtype=ncd_int, &
               dim1name='lon', dim2name='lat', &
               long_name='index of shallowest bedrock layer', ncid=nfid(t), &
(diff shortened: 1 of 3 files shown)


slevisconsulting:
@ekluzek this PR is ready for review. Let me know if you would like to discuss it in a meeting.

ekluzek (approved):
This looks correct to me.

PR #1735, opened 2022-05-03
Subset mesh files for regional cases and add user-mods for regional cases. 

negin513:
### Specific notes
This PR adds capability to subset mesh files for a regional case based on the region of choice. 

Contributors other than yourself, if any: @swensosc and @adrifoster

CTSM Issues Fixed (include github issue #): #1513 

Are answers expected to change (and if so in what way)? No

Any User Interface Changes (namelist or namelist defaults changes)? No, option `--subset-mesh` is added to `subset_data`. 

Testing performed, if any:
Plotting the mesh files to make sure the extracted mesh file makes sense.

PR #1884, opened 2022-10-27
Update subset-mesh capability of the subset_data tool with nco/esmf two-line approach

slevis-lmwg:
### Description of changes
Starting from #1735 I will introduce to the subset_data tool a two-line subset-mesh approach mentioned elsewhere, [e.g. here](https://github.com/ESCOMP/CTSM/issues/1513#issuecomment-1195667006).

Until this PR is merged, one can use the "two-line approach" manually, as shown in a couple of detailed examples [here](https://docs.google.com/document/d/1kyd0AFNi3YcuzWbcHVtxAV3DHVM7O3amRl1agKcsf3Q/edit?usp=sharing).

### Specific notes

@jkshuman 

CTSM Issues Fixed (include github issue #): #1513


wwieder:
This is awesome, Sam.  Is there information in the attached google doc that
would be helpful as a discussion post before we document this properly?

slevisconsulting:
@glemieux had the same suggestion, and I think it's a good idea.

slevisconsulting:
In this morning's SE meeting (2022/10/27), after some back and forth, we concluded that I should wait and that
- this PR may be wasted effort, because ultimately we may return to Negin's native Python approach for subset-mesh, mainly due to the user-facing complexities of requiring nco and esmf to run with the "two-line approach"
- @ekluzek will follow up with @negin513 to get an understanding of the status of her unpushed version

jkshuman:
@billsacks given that this subsetting capability is to obtain regions for use with CTSM and NUOPC, can we assume that the user will have both nco and esmf?
in the meantime I will test the workflow provided by @slevisconsulting

billsacks:
> can we assume that the user will have both nco and esmf?

We can assume that the user has esmf, but not necessarily nco. However, even for esmf, the challenge is ensuring that the user's path is set up correctly to have the esmf tools in their path. I imagine for most users, they have no idea how to load the appropriate esmf module to ensure this, instead relying on CIME to do the correct module loads behind the scenes – but that mechanism wouldn't be invoked for this standalone tool.

My feeling is that @slevisconsulting 's approach is an excellent stop-gap measure, but I began to feel that it doesn't seem like a great long-term solution because of these portability (or user education) issues, so it felt to me that spending a day to implement something temporary like this might not be a good use of time.

jkshuman:
Thanks for the explanation @billsacks I understand your concerns on this with regard to relying on CIME, and that these CIME loads are not used for this stand alone tool. Given your explanation of this, it is likely to eventually fail due to the disconnect between modules.  

Should @negin513 PR https://github.com/ESCOMP/CTSM/pull/1735 go back onto the NUOPC project board?

billsacks:
> Should @negin513 PR #1735 go back onto the NUOPC project board?

I put it back on the upcoming tags board. The relevant issue is still on the NUOPC (dropping MCT) project board: https://github.com/orgs/ESCOMP/projects/2/views/12

negin513:
I am back working on #1735 again today and I will try to get #1735 done by next week.

ekluzek:
@negin513 that's awesome to hear. We do want to be respectful of your time in your new position. I'm wondering if you and I could meet next week to go over where you are at in the process? Even if you finish it out, it would be good for you to explain it to me. Thanks so much...

negin513:
@ekluzek Thanks for your message. I would like to give a quick presentation at this CTSM SE meeting (Nov 3 2022) about how this works and what additional capabilities it has.

ekluzek:
@negin513 awesome. I'll add your presentation as one of the first things in the agenda. It'll be good to hear from you on Thursday. Take care.

slevis-lmwg:
@ekluzek I think you're correct. I see nothing new (by me or anyone else) in this PR relative to its starting point #1735

ekluzek:
Closing as we think other work already in covers this.

PR #1892, opened 2022-11-07
Regional CTSM Simulations and Capability of Creating Mesh Files

### LINKED ISSUE #1397: Generating ESMFMesh files for WRF-CTSM
Following the documentation on using CTSM with WRF, in section [3.4.1.7 Set ctsm runtime options](https://escomp.github.io/ctsm-docs/versions/master/html/lilac/specific-atm-models/wrf.html#set-ctsm-runtime-options), we are told to modify `lilac_in` in the following way: 
`atm_mesh_filename = '/glade/scratch/negins/wrf_ctsm_files/wrf2ctsm_land_conus_ESMFMesh_c20201110.nc'`
`lnd_mesh_filename = '/glade/scratch/negins/wrf_ctsm_files/wrf2ctsm_land_conus_ESMFMesh_c20201110.nc'`

Could you please let us know how these files are generated?

### LINKED ISSUE #1513: Need a process to subset ESMF mesh files from global ones for regional grids
With the update to NUOPC we'll need a way for users to create ESMF mesh files from domain files (or from their grid) for any regional grids they create. This needs to be done for regional cases for CTSM, but also when using LILAC to run WRF with CTSM.

The current subset_data process creates domain files as well as surface datasets for regional areas. We'll need some steps to convert the domain file into an ESMF mesh. Or possibly we should directly create an ESMF mesh file, or maybe the global ESMF mesh file should also be subset (but because ESMF mesh files are 1D vector files and not 2D, this may be difficult).

This isn't a problem with single point simulations as there we provide the latitude and longitude with PTS_LAT and PTS_LON so a mesh file isn't used. But, it is a problem for regional grids of greater than a single point.

### LINKED ISSUE #1773: High resolution regional simulations
We need a clearer workflow for how users can run high resolution regional cases.

Briefly, Xiulin is wanting to run a high resolution CTSM-FATES case over regions of California with DATM inputs that are archived from a WRF simulation.  @lmkueppers and @ckoven is this something you've done previously?  Moving forward, encouraged use of nuopc, and thought the following would need to be done: 
1. Subset surface data (from NLDAS) to the region of interest, given #1408, I wonder if this will be very straightforward?
2. Create a mesh file, #1735 (the mesh file for nuopc is similar in function to the old domain file for mct, but is created differently).  
3. Create a regional case that points to the surface data, mesh, and datm files.  You can point to the correct DATM files using the datm.streams.xml file in a nuopc case.  This may be easier to set up with a usermods_dir.

Here are the questions that may trip us up.
- Are the existing surface data and datm files at the same resolution as the NLDAS grid?
- How does CLM handle information on soil properties coming from NLDAS (presumably they aren't the same)?  
- @negin513 and @adrifoster, can the ./subsetdata script be manually changed to create a mesh and surface dataset from higher resolution base maps (3 or 9 km NLDAS).  If so,
- Do we have an appropriate resolution mesh file to subset for this regional case? 
- @ckoven , masking out regions within the region may be a little trickier.  I'd assume this is done with the surface dataset and mesh files?  @ekluzek  may have ideas based on how we're creating the sparse grid for nuopc.

I'm sure there are additional considerations I'm not thinking about here, but wanted to move this discussion onto github to avoid confusion

negin513:
#### Quick Summary:
This PR includes: 
1 - Script that enables our users to easily set up and run regional CTSM simulations.
2- Python class and scripts for creating a mesh file from any netCDF file and visualize it. 

* **Detailed instructions for running CTSM regional cases are available in this [link](https://gist.github.com/negin513/d71268419f808b91a82cde5530d390b1#file-regional_ctsm-md).** 

* **Instructions on how to use `./mesh_maker.py` for different domains including one example WRF domain are available [here](https://gist.github.com/negin513/dde385a28145928786e2bc498e2b5137).** 

#### Additional Details:
In this PR, we first introduced `mesh_type` python class and its new capabilities. 
This new mesh object enables our users to create a mesh file from any netCDF files with either 1D or 2D latitude and longitude coordinates.

* `mesh_type` is implemented in `./subset_data` to create mesh file for surface dataset for easily running regional cases. Next, we added additional capabilities to automatically set up usermods necessary to run a regional case. This enable our users to create and run a regional case with 2 commands. 
* For other purposes, `mesh_type` is implemented in a script called `mesh_maker.py` to create a mesh file from any netCDF file.

Please note that the method that is used for creating a mesh file is different from #1735, where we subset a global mesh file for a region. 

### Specific notes

CTSM Issues Fixed :
 Fixes #1513: Create ESMF mesh file for regional grids
 Fixes #1773: High Resolution Regional Simulations
 Fixes #1397: Generating ESMFMesh files for WRF-CTSM cases --> discussions #1832

Are answers expected to change? No

Any User Interface Changes? 
1- Changes in `subset_data` user interface with addition of `--create-mesh`
2- Adding `mesh_maker.py` with the following user interface:
```
usage: mesh_maker.py [-h] --input INPUT [--output OUTPUT] [--outdir OUT_DIR]
                     --lat LAT_NAME --lon LON_NAME [--mask MASK_NAME]
                     [--area AREA_NAME] [--overwrite] [-v]

|------------------------------------------------------------------|
|---------------------  Instructions  -----------------------------|
|------------------------------------------------------------------|
This script creates ESMF unstructured GRID (mesh file) from a netcdf
file with valid lats and lons. Provided lats and lons can be 1D or 2D.

For example for running WRF-CTSM cases, the user can create a mesh
file for their domain :
    ./mesh_maker.py --input wrfinput_d01 --output my_region
        --lat XLAT --lon XLONG --verbose

optional arguments:
  -h, --help        show this help message and exit
  --input INPUT     Netcdf input file for creating ESMF mesh.
  --output OUTPUT   Name of the ESMF mesh created.
  --outdir OUT_DIR  Output directory (only if name of output mesh is not
                    defined)
  --lat LAT_NAME    Name of latitude varibale on netcdf input file. If none
                    given, looks to find variables that include 'lat'.
  --lon LON_NAME    Name of latitude varibale on netcdf input file. If none
                    given, looks to find variables that include 'lon'.
  --mask MASK_NAME  Name of mask varibale on netcdf input file. If none given,
                    create a fake mask with values of 1.
  --area AREA_NAME  Name of area variable on netcdf input file. If none given,
                    ESMF calculates element areas automatically.
  --overwrite       If meshfile exists, overwrite the meshfile.
  -v, --verbose     Increase output verbosity
```

Testing performed, if any: regional simulations for different regions.


negin513:
I am adding [my instructions](https://gist.github.com/negin513/d71268419f808b91a82cde5530d390b1#file-regional_ctsm-md) for running regional cases here for future reference.

# Regional CTSM simulations:

This comment includes steps for running a regional CTSM simulations. In this comment we used CONUS domain as an example. 

1- First clone my fork and branch that includes these capabilities:

``` Shell
git clone https://github.com/negin513/ctsm.git ctsm_mesh
cd ctsm_mesh

git checkout subset_mesh_dask
./manage_externals/checkout_externals -vv

```


2- Run `./subset_data` for the region of interest to create surface dataset, mesh files, and all the usermods necessary to run the regional case. In this example, CONUS region is used:

```
cd tools/site_and_regional

./subset_data region --create-surface --create-mesh --create-user-mods --lat1 20 --lat2 55 --lon1 230 --lon2 300 --overwrite --verbose --outdir /glade/scratch/negins/conus_example

```

This takes ~20 seconds to complete on Cheyenne. `--verbose` helps with additional outputs. 
Example output of this script shows the commands the user needs to run next. :tada: :sparkles:

Sample output for this region:
```
For running this regional case with the created user_mods : 

./create_newcase --case case --res CLM_USRDAT --compset I2000Clm51BgcCrop --run-unsupported --user-mods-dirs  /glade/scratch/negins/conus_example/user_mods 

INFO: Successfully ran script for a regional case.
```

This script creates both regional and global plots of the mesh file for region of interest. Next, the user should check the plots to confirm the region:


For example for this region, the plots are:

image
image





3- Next, run the `./create_newcase` command to create a new case from the region of interest. 
Hint: the command for running `./create_newcase` is printed out as output of last step (so just copy and paste that.)

```
cd ../../cime/script/

./create_newcase --case conus_example --res CLM_USRDAT --compset I2000Clm51BgcCrop --run-unsupported --user-mods-dirs /glade/scratch/negins/conus_example/user_mods
```

Now go to the created case, build it and run it. 

Running the following will build your case and run it for 5 days (default):
```
cd conus_example
./case.setup
./case.build #on cheyenne qcmd -- ./case.build
./case.submit
```
-----
Now, if we want to run this case for 2 years with 8 nodes and 4 thread per node, instead of the commands above:

```
cd conus_example

./xmlchange STOP_OPTION=nyears
./xmlchange STOP_N=2

#-- change the number of cores to run on:
./xmlchange NTASKS=64
./xmlchange NTHRDS=4
./case.setup
./preview_run

./case.build #on cheyenne qcmd -- ./case.build
./case.submit

```

Voila: Now you successfully run CTSM for the CONUS region. :clap:

Next: Try running CTSM for another region.

negin513:
@jkshuman, @swensosc, and @swensosc : Please feel free to use this PR to test regional CTSM capabilities.

chrystali2002:
@negin513, you are the best! :) Thanks for sharing this step by step procedure. I am gonna try it ASAP

negin513:
@jkshuman mentioned an issue about `CLM_USRDAT_DIR` not inlining correctly in user_nl_clm and it was working in the previous versions ([fcb3405](https://github.com/ESCOMP/CTSM/pull/1892/commits/fcb3405b3092e4aaa6a35771472f43f5174b478f)). It seems like one of pylint cleanups created this issue. This issue is fixed with my latest commit [543e424](https://github.com/ESCOMP/CTSM/pull/1892/commits/543e4243a7bb981c3c3b00bffd19228dd8fbd4c8). Thanks to @jkshuman for noticing this.

jkshuman:
Updating this to remove my last post. That was a misunderstanding on my part. Things are working.

jkshuman:
This script is functional and runs as expected. Huge thanks to @negin513 for getting this through and helping me with trouble-shooting.

future development options that are outside of this PR
  - option to change input surface dataset (not edit cfg file)
  - remove ocean tiles

Testing the regional script shows performance slowdowns with nuopc compared to previous mct regional runs. @jkshuman will do a serious of test runs that iterate through setup changes from discussion with @slevisconsulting Mariana @ekluzek @billsacks 
Tests may prompt recommendation of case setup changes to improve timing performance, and will highlight this issue, but addressed in a separate issue thread on potential changes.

negin513:
Hello @jkshuman You can give the name of the landmask to this script and use `--mask` flag to remove ocean points from the mesh or modify mesh to use non 1 mask everywhere. If `--mask` is NOT given then it assumes all the points are land points.

wwieder:
Thanks for this note @negin513 can you provide this in your CONUS example for `./subset_data region` above, as it will help us with documentation?

In the SE meeting today, we also discussed using an ocean mask required for regional cases, but this may not need to come in with this PR.

billsacks:
> In the SE meeting today, we also discussed using an ocean mask required for regional cases, but this may not need to come in with this PR.

To be a bit more specific: What we'd like is that, for regional cases, the user should be required to *either* give the `--mask` argument or to give an argument like `--no-mask`. That way the user needs to explicitly think about whether they want a mask, as opposed to what's in place now where if you don't realize that you need to give a mask you end up with no mask, which can cause problems.

ekluzek:
Since, @negin513 has transitioned to CISL, I'll go ahead and take care of bringing this tag to CTSM main. This tag adds important capability for our work. And with being able to deprecate MCT, and fully move to NUOPC. I checked with her to make sure that is OK. We will fully credit her in the Change files and documentation.

ekluzek:
Working on this now. I updated to the latest, and there are a bunch of pylint errors to resolve, and the python unit tests are failing (the sys tests pass though).

ekluzek:
I've taken a look at the new code here, and there is some complexity that would be worth adding testing for. We talked about this concern in a CTSM-Software meeting and we questioned if this would be worthwhile, since it's the most efficient to add tests while something is being developed. Which is certainly true. We also figure that the mesh capability that is added here, may get replaced with a CESM wide effort on handling ESMF mesh files. 

However, to fix some of the pylint issues I would be more confident if at least some tests were added. But, I totally do not want to take too much time in adding testing. So I plan to add some limited testing, but be careful that I don't take too much time to add too much. Adding some testing now though would make it easier to add testing later for refactoring that's done. subset_data is an important tool that we need to continue to maintain long term.

ekluzek:
I added the new mesh option to the tools test regional mesh test case, and the plots enable me to see that this is all working correctly. It's nice to have that validation.

tools/site_and_regional/mesh_maker.py
```diff
@@ -0,0 +1,255 @@
...
+|------------------------------------------------------------------|
+This script creates ESMF unstructured GRID (mesh file) from a netcdf
+file with valid lats and lons. Provided lats and lons can be 1D or 2D.
+
+For example for running WRF-CTSM cases, the user can create a mesh
+file for their domain :
+    ./mesh_maker.py --input wrfinput_d01 --output my_region
+        --lat XLAT --lon XLONG --verbose
+
+"""
```
ekluzek, line 14: I'm not sure in what context this would be used. I suppose the idea is that this could be used in place of the two step process that relies on NCO and ESMF. This would enable you to create a mesh anywhere with a valid python environment rather than requiring the use (and build) of NCO and ESMF. Using those are also very machine specific.

A test for that would be to compare the results between the two methods.

test/tools/nl_files/subset_data_region1
```diff
@@ -1 +1 @@
-region --lat1 -40 --lat2 15 --lon1 275 --lon2 330 --create-domain --create-surface --create-landuse --verbose --overwrite --reg test1 --inputdata-dir CSMDATA
```
ekluzek, line 1: Since mesh_masker.py was added as a new script.  We should at least add a tools test for it.

python/ctsm/site_and_regional/regional_case.py
```diff
@@ -228,3 +322,220 @@ def create_landuse_at_reg(self, indir, file, user_mods_dir):
...
+
+        self.mesh = mesh_out
+
+        node_coords, subset_element, subset_node, conn_dict = self.subset_mesh_at_reg(mesh_in)
+
+        f_in = xr.open_dataset(mesh_in)
+        self.write_mesh(f_in, node_coords, subset_element, subset_node, conn_dict, mesh_out)
+
+    def subset_mesh_at_reg(self, mesh_in):
+        """
```
ekluzek, line 353: A complex method that could use testing.

python/ctsm/site_and_regional/mesh_type.py
```diff
@@ -0,0 +1,560 @@
+"""
```
ekluzek, line 1: Since, this is a type that's not a top level program it should be moved to the python directory, so that the python building tools can be used around it.

ekluzek (changes requested):
I really like the mesh plots that this produces, visualizing these mesh files has been problematic and this has a nice clear method to display them which is really good. This regional capability is an important functionality that we need to bring in. This is also the last critical thing needed so that we can deprecate MCT and fully commit to NUOPC.

In general the code is well broken up into sensible subroutines and easy to follow. It doesn't have testing implemented for it, and I think it needs some. 

The mesh type should be moved over to the python directory. We should add a simple test for the mesh_maker script. I think it could also be useful to have a mesh plotting script and with this base code that should be easy to do.

I'll be the one working on this. So I'll add limited testing, and do the other changes I suggest. I'm going to limit the time I spend on this so we'll see how much I can do.

It's nice to see this work there and will be good to bring it into main CTSM.

ekluzek:
OK I ran into an annoying problem with conda when I updated to dev118. The creation of the conda environment was taking hours without completing (on multiple machines). It would work when I just gradually added a few new requirements at a time adding to the ctsm_py -- but was hanging when creating from scratch.

This made me look into conda environments more and I finally found a helpful article.

https://pythonspeed.com/articles/conda-dependency-management/

This answers a question I've been wondering about. How do you support both ease of upgrading environments, as well as supporting getting environments that work? And since you really want both at the same time how can you do both at once? The article gives a solution which makes sense to me and will just require a bit of a change in how py_env_create works.

jkshuman:
That's great @ekluzek ! I look forward to hearing about this. I am in fear of updating my conda environment for exactly these reasons.

wwieder:
+1 this seems like helpful information @ekluzek.  I wonder if folks in CISL or ESDS would be interested in this, since it seems like a broader challenge as we all are using more conda environments to manage python workflows.

negin513:
Hi everyone, I hope you are doing fine. I just wanted to follow up on this. Thanks @ekluzek for developing tests for this. 

Can I ask what is causing the conda environment hangup? 
A quick tip is that mamba is usually faster than conda. But creating an environment from the beginning is usually time-consuming. 

One suggestion I have is not creating a new conda environment for every ctsm_py clone.
A lot of time spent on conda is spent on resolving compatibilities issues.
It always make sense to directly specify the versions (especially if they don't have compatibility issues, it saves up on time) but that requires sometime on the maintenance side to keep testing newer versions and updating the versions accordingly. 
 
Overall: I suggest specifying the versions, but this require some personnel time to update these packages on a regular cadence. 
 
Another option is for machines such as Cheyenne having the environment saved somewhere (for example on Erik's path) and ask the users to clone that environment, this creates an environment for them from Erik's environment but it does not take as long as starting from scratch. 


In this particular PR I used `Dask.Arrays` instead of `Numpy` to speed the mesh creation part up. This significantly helped with the speed (x5-x10) but added `Dask` as a dependency.  This is especially important for creating high-resolution mesh files.

ekluzek:
@negin513 thanks so much for checking back in here! And thanks for the suggestions.

Yes, you lay out the conflicting goal of adding exact versions in, as opposed to putting minimum requirements for updating and flexibility. I've been working on the approach I found about using both minimum requirements and then use conda-lock to save the exact specification. In principle that seems like the right solution to go for and it's used by other projects -- but in practice for our machines it's had issues.

I'm loathe to have a user-based path to use a shared environment, as that goes against the goal of being able to run anywhere on any systems. That could be a temporary solution though.

Thanks for pointing out that using dask is important for high resolution meshes.

I'm thinking of two things here. One is to separate out the dask requirement into a separate python environment. That way it would work for everything, but mesh creation. Another would be to either back out the addition of dask, or make it an optional addition. I'm not sure if it's possible to make a module optional or not though.

negin513:
Hello @ekluzek, I was wondering if you have confirmed that the 'dask' package is causing issues with the ctsm-py environment? I strongly advise against removing 'dask' as it would require extensive work and lead to significant inefficiencies.

Additionally, even if the environment building process takes longer due to the issue, I believe it's important to keep 'dask' in the environment for its benefits in distributed computing. It is especially important for high-resolution mesh creation and even if we don't use `dask` with this code, it will eventually come back again with other codes. Numpy-backed Xarray Arrays are not efficient (or as effiecient) as Dask-backed ones, so eventually all will be migrated to `Dask` especially for high-resolution datasets. 

I suggest you first pinpoint which package is causing the long build time on the system. 

Additionally, I have to add using NPL on NCAR machines will fix the issues as it already have all these dependencies and 100s more from our community. 

For example, the workflow on Cheyenne/casper (and what I used while developing this script):
```
module load conda/latest
conda activate npl-2022b (or any other version that is maintained by us). 
```

negin513:
Moreover, this tool could be useful for a broader ESM user community. Although I haven't tested it with MPAS and it may not work with certain unstructured datasets, other CESM components and WRF users have expressed interest in a mesh creation code. IMO it might be worth considering making it less closely linked to the CTSM Python framework.

negin513:
Hello @ekluzek and @jkshuman 
I hope this message finds you well. I wanted to follow up on this pull request. I was wondering if there is anything that needs to be addressed before it can be merged. 

While I am not sure if this comment is still related, I have done some testing with the environments to figure out what might has caused the issue you mentioned (long conda environment build time).

> OK I ran into an annoying problem with conda when I updated to dev118. The creation of the conda environment was taking hours without completing (on multiple machines). It would work when I just gradually added a few new requirements at a time adding to the ctsm_py -- but was hanging when creating from scratch.
> 
> This made me look into conda environments more and I finally found a helpful article.
> 
> https://pythonspeed.com/articles/conda-dependency-management/
> 
> This answers a question I've been wondering about. How do you support both ease of upgrading environments, as well as supporting getting environments that work? And since you really want both at the same time how can you do both at once? The article gives a solution which makes sense to me and will just require a bit of a change in how py_env_create works.

I was curious about this and did some testing and found the culprit is not the packages introduced in this PR but the `black` version. So when I removed the black version from the [dependencies here](https://github.com/ESCOMP/CTSM/blob/a6c1c63f1eeb479164a7fb3d5b7b6449af1856fa/python/conda_env_ctsm_py.txt), creating the environment took only <~90 seconds on Cheyenne .

You can reproduce my results on Cheyenne and Casper using the following bash script:

```bash
#!/bin/bash
# -- for CTSM group
# following https://arc.ucar.edu/knowledge_base/83853599
module reset
module load conda/latest
module list

conda env list

# -- create a yml file for ctsm python environment
echo "name: ctsm_python
dependencies:
    - python==3.7.9
    - pandas
    - tqdm
    - scipy
    - netcdf4
    - dask
    - cartopy
    - matplotlib
    - requests
    - packaging
    - numpy=1.18.5
    - xarray=0.16.2
    - pylint=2.8.3
    - black" > environment.yml

start=$(date +%s.%N)
echo "Starting to build conda environment using mamba:"

mamba env create -f environment2.yml --force

end=$(date +%s.%N)
runtime=$(echo "$end - $start" | bc)
echo "Execution time (mamba): $runtime seconds"
```

You might also notice here I used `mamba` which I recommend for better performance, but even with conda we get ~2minutes build time for this conda environment (depending on the system resource of course). 


----------------------------------------

FYI: Usually, when building a Conda environment with some dependencies take a long time, it is due to compatibility issues,  complex dependencies, or limited resources on the machine. 
Since the requirements are minimal and you are experiencing this issue on all the system, it is most likely due to compatibility issues. The above experiment confirms this too. As I mentioned earlier, it is always good to specify the packages explicitly; however, this lengthen the build time if there are any conflicts and it requires resources (a.k.a. personnel time) to assure it is properly maintained.

IF what is causing in merging this PR is the issue with building the conda environment, we can proceed removing the black version. 

Please let me know if there is any additional information that I can provide or if there is a better time for us to discuss the pull request.

Thank you for your time and attention, and I look forward to hearing back from you soon.

Best regards,
Negin

ekluzek:
@negin513 hmm interesting. Thanks for trying that experiment out. I did get a robust way to create the environment by changing how its setup. So it's interesting to hear how your approach worked as well. We did want to ensure the version of black was specified to ensure different people wouldn't experience differences in how black operates. That makes me wonder if we need multiple python environments.

There are some requirements that I'll give in what we need here, that I'll give in a bit....

ekluzek:
We talked more as a group about this and came to some group decisions regarding it. And some of those group decisions outline the requirements we have for this work, and some of this work in general. We need our tool conda environments to work on any system that we can run the model on, which is part of why we need a system for specifying the environment so that it can be done anywhere. Since, github is forever even work that is backed out can be easily added back in at a later date. I added a bunch of complexity for working with conda-lock, which looked like it was a great solution, but turned out it didn't work that well on either cheyenne or izumi. So I'm going to back all of that out. But, with git we can add it back in later if it starts working better and we decide it will be useful in the future. A good Agile Programming Methodology concept is that you only add the capabilities that you have a demonstrated need for, and you only add them once you need them. This keeps the code as simple as possible until the extra complexity is proven to be needed.

Requirements for CTSM python tools:
- Only include packages that are absolutely required
- Tools that require extra packages should be done in a "contrib" type area out of the main part of tools (this would apply for advanced plotting capability for example)
- We need to be able to reproduce working conda environments minimally on our test machines (currently cheyenne and izumi), but also on any machine that we run CTSM on. If there is a machine that we can run CTSM on that we can't build the conda environments or run the tools on -- that needs to be fixed.
-  We need to tell the user how long to expect the conda environment to load, and give them options if the conda load is taking too long
- Conda environments need to build robustly even for users who don't have ctsm_pylib loaded in their conda environment
-  Currently dask will NOT be something we require for any of the main CTSM tools
-  Currently we won't use conda-lock
- We specify the black version exactly so that black will function identically for all users
- We specify the pylint version exactly because pylint is finicky with version and we need it to work identically for all developers
- We might remove the need for python packages that  aren't providing enough utility
- Packages where specific versions seem to be required will have the version requirements in a least the >= form if not an exact version

Some things to try?:
An experiment we will try is that I also have a "specific" file that gives specific versions that work on cheyenne. This makes the conda install quicker, but it might require maintenance. We'll try this out and see if it works. Another thing I'm trying is a "latest" file that we can experiment with the latest versions of packages that seems to work. The best way both of these could work is if we have automated testing for both of them. Which I think we can add eventually

python/ctsm/site_and_regional/mesh_type.py
```diff
@@ -0,0 +1,560 @@
...
+        ds_out.attrs["conventions"] = "ESMFMESH"
+        ds_out.attrs["date_created"] = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+        # -- write Xarray dataset to file
+        if mesh_fname is not None:
+            logger.info("Writing ESMF Mesh file to : %s", mesh_fname)
+            ds_out.to_netcdf(mesh_fname)
+            logger.info("Successfully created ESMF Mesh file : %s", mesh_fname)
+
+        self.make_plot = True
```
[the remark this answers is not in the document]
ekluzek: This is solved by spinning off the plotting into a different script. As a first step I had made this an argument.

python/ctsm/site_and_regional/regional_case.py
```diff
@@ -112,6 +137,49 @@ def create_tag(self):
...
+            lon1 points to the westernmost longitude of the region. {}
+            lon2 points to the easternmost longitude of the region. {}
+            Please make sure lon1 is smaller than lon2.
+
+            Please note that if longitude in -180-0, the code automatically
+            convert it to 0-360.
+            """.format(
+                self.lon1, self.lon2
+            )
+            raise argparse.ArgumentTypeError(err_msg)
```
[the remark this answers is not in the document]
ekluzek: There's now some unit and sys tests in place.

tools/site_and_regional/mesh_maker.py
ekluzek: Comparing results to the two step process is probably a worthwhile thing to do. We had decided a long time ago that a python script with a conda environment would be an easier way to support this.

test/tools/nl_files/subset_data_region1
ekluzek: I added both tools test and python unit and sys tests.

python/ctsm/site_and_regional/mesh_type.py
ekluzek: This is something that I must have misread the directory name. So resolving.

python/ctsm/site_and_regional/mesh_type.py
```diff
@@ -0,0 +1,560 @@
...
+
+        if self.make_plot:
+            plot_regional = os.path.splitext(mesh_fname)[0] + "_regional" + ".png"
+
+            plot_global = os.path.splitext(mesh_fname)[0] + "_global" + ".png"
+
+            self.make_mesh_plot(plot_regional, plot_global)
+
+    def make_mesh_plot(self, plot_regional, plot_global):
```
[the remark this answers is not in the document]
ekluzek: I'm working on spinning off a separate script to do the plots.

negin513:
Hi everyone! I want to follow up on this PR! 

A few different people reached out to me and mentioned they are interested in such a tool enabled with Dask. I was thinking about creating a small package out of the dask versions of this and just wanted to let you know about that!

ekluzek:
Hey @negin513 thanks for that update. I assume this is something you can do with your work in CISL?

We do continue to have concerns with the complexity of dask environments, but as an alternative for a different environment that is something that we could consider. So I'm not sure about a path forward for that. So I can't guarantee that we could take it in, but I'm always interested in seeing alternative's. I think it can also be useful for us to learn about dask for workflows that do need it. So you should be aware of that.

Note, you should branch from this PR. I want to bring this in pretty much as it is. Actually there's some github action work that I added here that I should have put into a different PR, and I'm going to remove to put in a different PR so they aren't tied together. I've started making "merge PR's" that merge other PR's together. This makes it easier to back out one of the PR's if I run into trouble with it. It also makes the contents of PR's easier to evaluate if they are small and separate.

Anyway, as you can see I haven't been able to work on this in quite some time. But, am planning on getting back to it and finishing it off.

If you branch from here, this will be close to what goes to the main branch, so a good way to move forward.

wwieder:
Thanks for reaching out, @negin513.

It would likely be good to merge this PR, @ekluzek.  I can't remember what's left to be done here?

Subsequently, we can explore the requirements for implementing the dask features for regional subsetting.  It's been a while since we discussed this, but I thought we decided the non-dask approach was _good enough_ for most applications of regional simulations?  Were there particular applications were users were requesting a dask-enabled tool, @negin513?

wwieder:
Hi @negin513 thanks for following up on this.

I know @ekluzek had been working on this as part of another tag and was having issues wrapping around the prime meridian.  I thought we we're going to address the issue, but I honestly don't recall why the tag got stalled out?  I'd assume we need some testing to go along with this PR, but maybe that's already been created?

Regarding your dask comment, I'm not sure where we as a CTSM-SE group stand on the complexity of our python environments...?  

For timing of merging this PR I'd argue this is a feature we want supported for the CESM3 release, but right now other code changes to CTSM are higher priority at the moment.  I appreciate this isn't a very satisfying response, but maybe this is something that can come on on a b4b_dev tag more easily?

ekluzek:
Yes, @negin513 thanks for pinging us about this. It's good to know there are others interested in this PR. And I want to point out that @adrifoster used the branch as is for some work fairly recently where this was critical. And thank you for your work on this it is an important piece for LMWG.

@wwieder and everyone the main thing here is that we have to be working on the most critical LMWG priorities for the release which is Dust-emissions and CTSM5.2 for me right now. After that is done I'll probably be able to work on this again. I do agree that this should be part of the CESM3 release. I'm not sure if we'll get this in before the planned June science freeze, but we should be able to get it in before the planned CESM3 release next Winter (ish). And since this is more of a software thing it doesn't have to get done by June. And unfortunately this PR is not alone in taking a long time to come in, we do have 50 PR's that need to come in at this point. Some are many years old. Some are drafts and not ready to come in. But, we do have a big backlog so we have to triage what needs to come in first. We also do seem to be PR's faster than they can come in as well -- which obviously is a concern.

@wwieder the prime meridian wrapping (talked about for example here #1897) should just be something we worry about later for this PR. It would be good to have though, but shouldn't slow this one down. Again the main reason it's stalled is because I've needed to work on other things.

@wwieder on dask, that could be a future update that we bring in. But, we shouldn't slow this PR down by doing that here. And this isn't a clear win to bring in dask for this work. So that's something that we'll continue to talk about and possibly bring in later. The most important thing is to bring it in as it is.

The other thing that I'll confess to, is that I added some github actions to this PR, which I then ran into problems with which stalled the progress on the main effort here (you see those actions as not passing here, and they need to pass to bring the PR in). What I will actually do is to remove those actions into a separate PR and bring this on largely as is. I've had a habit of bringing in things into whatever PR I'm working on (making them too big), but we are realizing as a group that separating things out is beneficial. I added it because I thought it was going to be simple, easy and quick, but ran into trouble. There are four conversations to resolve as well. So once I get a chance to do that this should be good to come in.

@wwieder yes we will likely rebase this to b4b-dev when I can work on it. Which will help to bring it in. But, I still need to be able to spend time on it before it goes there either. @negin513 something we have started to help PR's come in quicker and be smaller -- is a bit-for-bit branch to bring in developments that don't change answers. We've just started it, but I already see where it's been helping.

So @negin513 hopefully that helps. This is important work that you did and it does need to come up. But, the current triage priority for CESM3 for LMWG is this is work I'll do after dust-emissions and CTSM5.2. It is listed among the many priorities for CESM3 in our planning page https://github.com/ESCOMP/CTSM/projects/40 (currently under slow roast).

I know this is long, but here's the synopsis:

TLDR:
 -- Yes this is important! 
-- Thank you for the ping and your work on this!
-- We will get to it among the triage of CESM3 priorities.
-- We can't guarantee *when* that will be -- but it is a CESM3 priority

slevis-lmwg:
@negin513 and everyone, does this PR address the "sparse grid" use-case, which is in high demand these days? If so, I recommend the consolidation/clean-up of efforts (issue #1731 and discussion #1919) under this PR.

Regardless though, getting this PR wrapped up sooner rather than later, yet (as always) within the context of overall priorities, as pointed out by @ekluzek, makes sense to me.

ekluzek:
@slevis-lmwg -- nope this does NOT have to do with the sparse grid. Happy to clear that up.

adrifoster:
@ekluzek from our discussion it sounds like all that is needed to bring this in is to remove the github actions-related changes? 

If so, I can pull those into a separate branch/PR, and delete them here?

Another update that is required is setting nx/ny, which I *think* we can get the user_mods to do?

py_env_create
```diff
@@ -20,6 +20,7 @@ if [ $error != 0 ]; then
    echo "For notes on installing on a user system see: https://docs.conda.io/projects/conda/en/latest/user-guide/install/index.html"
```
ekluzek, line 20: These changes get moved over the the github action branch, and removed from here.

python/Makefile
```diff
@@ -42,6 +42,17 @@ utest: FORCE
 stest: FORCE
 	$(PYTHON) ./run_ctsm_py_tests $(TEST_ARGS) --sys
 
+adv_test: adv_utest adv_stest
```
ekluzek, line 45: Move to the branch.

python/conda_env_requirements_ctsm_pylib.txt
```diff
@@ -0,0 +1,16 @@
+#
```
ekluzek, line 1: The above text files that are deleted should be put back and the files below added should be removed.
And those changes go on the branch.

python/ctsm/.pylintrc
```diff
@@ -433,6 +433,12 @@ good-names=i,
            k,
            ex,
            Run,
+           ds,
```
ekluzek, line 436: This should stay in.

python/ctsm/run_ctsm_py_tests.py
```diff
@@ -28,6 +28,10 @@ def main(description):
         pattern = "test_unit*.py"
     elif args.sys:
         pattern = "test_sys*.py"
+    elif args.adv_unit:
```
ekluzek, line 31: These changes go to the other branch and removed here.

test/tools/README.testnames
```diff
@@ -20,15 +20,15 @@ n is the configuration type:
 6 -- unused
 7 -- unused
 8 -- unused
-9 -- unused
+9 -- mesh_maker
```
ekluzek, line 23: I add some test/tools testing here, which is in the process of being deprecated. So maybe doesn't need to come in. But, it also doesn't seem like it hurts to have it either...

ekluzek:
Yes, move everything doing with github actions and almost everything that changes py_env_create and related files to a different branch, and remove them from here.

And yes we should add the needed update about the NX/NY size.

We also need to resolve the last four conversations. They should be turned into an issue or two.

python/conda_env_requirements_ctsm_pylib_dask.txt
```diff
@@ -0,0 +1,9 @@
+#
+# To install this on with conda loaded in modules
+# use the top level bash script:
+# ../py_env_create                                           # Do this each time you update your CTSM Version
+# conda activate ctsm_pylib                                  # Do this anytime you want to run a CTSM python script
+#
+dask=2021.10.0
```
ekluzek, line 7: Oh, this is important to have so that the plots can be done.

py_env_create
```diff
@@ -69,8 +91,19 @@ while [ $# -gt 0 ]; do
...
+           custom="Yes"
            shift
            ;;
+       --read-spec )
+           readspec="Yes"
+           ;;
+       --read-latest )
+           latest="Yes"
+           ;;
+       --dask )
```
ekluzek, line 103: The dask option should stay.

py_env_create
```diff
@@ -48,15 +54,31 @@ usage() {
...
+  echo "     So if it fails with this option try building without it"
   echo "[--option <option>]  "
   echo "     Option(s) to pass to 'conda install' step"
   echo "***********************************************************************"
 }
 
 verbose="No"
+readspec="No"
+custom="No"
+includedask="No"
```
ekluzek, line 80: The dask option needs to stay, but others go to the branch.

py_env_create
```diff
@@ -48,15 +54,31 @@ usage() {
...
   echo "     Default is: $condafile"
+  echo "[--read-spec] "
+  echo "     Also read the requirements file with specific versions"
+  echo "     This will likely be faster"
+  echo "     Will use: $condadir/$specfile"
+  echo "[--read-latest] "
+  echo "     Read the experimental conda environment file with the latest working versions"
+  echo "     Can NOT use with either the --file, or --read-spec options"
+  echo "     Will use: $condadir/$latestfile"
+  echo "[--dask] "
```
ekluzek, line 68: dask option stays.

py_env_create
```diff
@@ -29,11 +30,16 @@ ctsm_python=ctsm_pylib
...
-   condafile="conda_env_ctsm_py_cgd.txt"
+   condaplatformfile="conda_env_requirements_ctsm_pylib_cgd.txt"
 else
-   condafile="conda_env_ctsm_py.txt"
+   condaplatformfile="conda_env_requirements_ctsm_pylib_notcgd.txt"
 fi
+condafile="conda_env_requirements_ctsm_pylib.txt"
+specfile="conda_env_requirements_ctsm_pylib_specific.txt"
+latestfile="conda_env_requirements_ctsm_pylib_latest.txt"
+daskfile="conda_env_requirements_ctsm_pylib_dask.txt"
```
ekluzek, line 42: dask file stays...

py_env_create
```diff
@@ -84,20 +117,42 @@ while [ $# -gt 0 ]; do
...
+   condafilepath="$condadir/$latestfile"
+else
+   condafilepath="$condadir/$condafile"
+fi
+if [ ! -f $condafilepath ]; then
+   echo "$condafilepath does NOT exist"
    echo "Use the --file option with a valid filename"
    exit -1
 fi
+if [ "$includedask" == "Yes" ] && [ "$latest" == "Yes" ]; then
```
ekluzek, line 133: dask part stays.

py_env_create
```diff
@@ -130,6 +207,26 @@ if [ $? != 0 ]; then
    fi
    exit -2
 fi
+if [ "$includedask" == "Yes" ]; then
```
ekluzek, line 210: This dask part stays.

samsrabin:
When testing this, in addition to the instructions [here](https://gist.github.com/negin513/d71268419f808b91a82cde5530d390b1), I had to specify `--create-domain`. I also had to specify `--inputdata-dir` when trying on Izumi.

wwieder:
should we remove the `--create-domain` feature of this work since mct is no longer being supported for CESM3?

adrifoster:
> should we remove the `--create-domain` feature of this work since mct is no longer being supported for CESM3?

yes I think that is a good idea

ekluzek:
@wwieder as I recall we still need the domain files. I think they are to create the mesh files, but don't quite remember.

@negin513 those instructions @samsrabin pointed to are really helpful. I suggest moving them somewhere in ESCOMP/CTSM if you agree. I added a longer comment to your gist. This would make them more visible to the CTSM community.

python/ctsm/subset_data.py
```diff
@@ -403,6 +413,43 @@ def check_args(args):
...
+        )
+        raise argparse.ArgumentError(None, err_msg)
+
+    if args.surf_year != 1850 and args.create_landuse:
+        err_msg = textwrap.dedent(
+            """\
+                \n ------------------------------------
+                \n --surf-year option is NOT set to 1850 and the --create-landuse option
+                \n is selected which requires it to be 1850
+                """
```
samsrabin, line 440: - [ ] Why is this required to be 1850? Some additional text in the error message would be helpful, explaining what happens when you do set it to 1850 (you still get a landuse timeseries file with 1850-2015).
- [ ] When I did set this to 1850, I still got an fsurdat file for 2000.

adrifoster:
> @wwieder as I recall we still need the domain files. I think they are to create the mesh files, but don't quite remember.
> 
> @negin513 those instructions @samsrabin pointed to are really helpful. I suggest moving them somewhere in ESCOMP/CTSM if you agree. I added a longer comment to your gist. This would make them more visible to the CTSM community.

oh okay good to know!

python/ctsm/subset_data.py
ekluzek: On the first. When you have a landuse timeseries file for a year range of XXXX-YYYY, the surface dataset HAS to be the beginning of the range XXXX. This enables the model to do internal consistency checking to make sure the two files go together.

ekluzek: On the second, there is a config file that stores the location for the files that is separate from build-namelist XML and it's currently setup to be hardcoded to either 1850 or 2000. We do have a different file on ctsm5.2 for 1850, with the default being 2000. It would be good to handle this differently. But we ought to brainstorm on what it should look like, as well as needing to spend time to work on it.

samsrabin: > On the first. When you have a landuse timeseries file for a year range of XXXX-YYYY, the surface dataset HAS to be the beginning of the range XXXX. This enables the model to do internal consistency checking to make sure the two files go together.

I guess my thought is that `--surf-year` should be allowed to be set to 2000 with `--create-landuse`, but that the resulting landuse timeseries file in that case would be 2000-2015.

> On the second, there is a config file that stores the location for the files that is separate from build-namelist XML and it's currently setup to be hardcoded to either 1850 or 2000. We do have a different file on ctsm5.2 for 1850, with the default being 2000. It would be good to handle this differently. But we ought to brainstorm on what it should look like, as well as needing to spend time to work on it.

That doesn't really explain why, when I specified `--surf-year=1850` in the call of `subset_data`, I got an fsurdat file for 2000. Not only is `simyr2000` in the filename, the files' data are identical.

ekluzek: Hmmm. We might need to go over some of this together to see exactly what you are seeing.

OK, interesting idea on 2000-2015, although I don't think it truncates the years for the landuse timeseries file. But, I could be wrong. And it could be added also. That actually might be a useful feature for single-point cases because you could have it be over just the period that you have your tower site data for.

This is definitely a pain point that could definitely use some work to get it to work more intuitively...

mvertens:
@ekluzek @wwieder - In general with the new nuopc capability you do not need domain files - even for regional cases. Where you do still need then is for PTS_MODE - and if you look in component_grids_nuopc.xml you will see a few entries like the following:

  <domain name="T42">
    <nx>128</nx> <ny>64</ny>
    
    <file>$DIN_LOC_ROOT/share/domains/domain.lnd.T42_gx1v7.180727.nc</file>
    <mesh>$DIN_LOC_ROOT/share/meshes/T42_ESMFmesh_c20200629.nc</mesh>
    <desc>T42 is Gaussian grid:</desc>
  </domain>

You can also see the following in CTSM's lnd_comp_nuopc.F90
! If single_column is true - used single_column_domainfile.....

You also still need a domain file for LILAC.

adrifoster:
@negin513 I am working on bringing this PR in, after removing some of the conda and git workflow-related things @ekluzek was working on. This branch is [here](https://github.com/adrifoster/CTSM/tree/subset_mesh). 

How would you like me to work from here? Shall I just push to your branch/PR? If so, I think I need collaborator access on your CTSM fork. Otherwise I can just make another PR to CTSM, which would include all of the code mods you made (and still be linked to you).

I'm happy to do whatever you'd like!

python/ctsm/subset_data.py
samsrabin: Right, it probably doesn't truncate the years, but that behavior could be added.

For the result I'm describing, I did the following:
```bash
conda run -n ctsm_pylib \
tools/site_and_regional/subset_data region \
    --create-landuse --create-surface \
    --surf-year=1850 \
    --create-domain \
    --lat1 -15 --lat2 -11 \
    --lon1 -60 --lon2 -52 \
    --outdir /glade/work/samrabin/arc_of_deforestation/mato_grosso \
    --verbose --overwrite
```
and got the `*_c240313.nc* files in `/glade/work/samrabin/arc_of_deforestation/mato_grosso/`.

negin513:
Hello @adrifoster , Thanks! I added you to the collaborators on the fork.

negin513:
> should we remove the `--create-domain` feature of this work since mct is no longer being supported for CESM3?

No, for WRF-CTSM runs domain files are needed. So I would suggest keeping `--create-domain`

negin513:
> @wwieder as I recall we still need the domain files. I think they are to create the mesh files, but don't quite remember.

Yes @ekluzek , domain files are needed for WRF-CTSM (if the procedure is still the same). ;-)

negin513:
Hello. Everyone, 
Comparing the code with my last commit ([3588342](https://github.com/ESCOMP/CTSM/pull/1892/commits/358834295a2377bcd8edf03200f10a91ebff63fd)) , I noticed the dask arrays are converted to numpy arrays, which means it does not benefit from the lazy loading offered by dask. 




I suggest reverting these back to use dask array instead, which should make subsetting for high resolution mesh faster. 

For example:
```
git diff 543e424 python/ctsm/site_and_regional/mesh_type.py

-            self.center_lat2d = da.from_array(self.center_lats)
-            self.center_lon2d = da.from_array(self.center_lons)
+            # self.center_lat2d = da.from_array(np.array(self.center_lats))
+            # self.center_lon2d = da.from_array(np.array(self.center_lons))
+            self.center_lat2d = np.array(self.center_lats)
+            self.center_lon2d = np.array(self.center_lons)
....

I can start from [3588342](https://github.com/ESCOMP/CTSM/pull/1892/commits/358834295a2377bcd8edf03200f10a91ebff63fd) and check the changes made here.

python/ctsm/run_ctsm_py_tests.py
ekluzek: Looks like this should still be removed. It adds some complexity that we don't need.

ekluzek (approved):
Awesome, this is great to see. I resolved the conversations I had after checking.

I do have one change to do from the last conversation. Go ahead fix that and resolve and merge this in.

samsrabin:
> I think it would be best to get these instructions (once updated, if necessary) moved into the CTSM User's Guide as part of this PR. I can help if needed.
> 
> > * **Detailed instructions for running CTSM regional cases are available in this [link](https://gist.github.com/negin513/d71268419f808b91a82cde5530d390b1#file-regional_ctsm-md).**
> > * **Instructions on how to use `./mesh_maker.py` for different domains including one example WRF domain are available [here](https://gist.github.com/negin513/dde385a28145928786e2bc498e2b5137).**

Just wanted to re-raise this. It'd probably be best to get these brought in to the documentation with this PR, rather than letting it wait until users ask us questions about it. I can do this and submit a PR to [negin513:subset_mesh_dask](https://github.com/negin513/ctsm/tree/subset_mesh_dask) if you prefer, @adrifoster.

negin513:
Thanks Sam. I am going to add them later today. Do you parse md in you docs
build? If not i can change them to rst.

negin513:
Okay reading through conf.py, it seems like you don’t current parse
markdown file.  If interested, I can add that capability. LMK!

samsrabin:
@negin513 If it's not too much trouble, that'd be great! As long as it can do things like linking to other notes/sections in notes, being included in chapter lists, etc.

PR #1920, opened 2022-12-02
mksurfdata: Rework mapping of landunits to handle coastal areas more rigorously

### LINKED ISSUE #1716: Wetland area and areas of landunits in coastal grid cells should be determined more rigorously
### Brief summary of bug

I believe that there are some problems with how we calculate wetland area, both in mksurfdata_map and at runtime. These problems lead to a situation where I believe there is currently no correct answer to the question, "Should PCT cover raw datasets specify the percent of the total grid cell or just the percent of the land area in which the given landunit appears?"

@olyson @lawrencepj1 @dlawrenncar  @wwieder @ekluzek @slevisconsulting @mvertens 

### General bug information

**Does this bug cause significantly incorrect results in the model's science?** Yes, at least to some degree – though errors are probably relatively small

### Background / motivation

@olyson recently asked:

> Should the raw input dataset fields (e.g., PCT_URBAN in 0.05deg raw input urban dataset) represent the fields in terms of a percentage of total gricell area, or in terms of a percentage of the total land area in the gridcell?

After giving this some thought, I have come to the conclusion that either choice will do some things correctly and some things incorrectly. I feel that it would be helpful if we could provide a clear answer to this question and have the resulting landunit fractions in the model be consistent with the raw datasets.

I should note that I have *not* carefully examined all of the relevant code or tried running various scenarios: My thoughts below are based mainly on my recollections of how mksurfdata_map works. Also note that, although I refer to mksurfdata_map in my comments, really I don't imagine any changes being made to that deprecated code: all changes can be applied to the new mksurfdata_esmf.

#### Problems with specifying raw dataset PCT cover as percent of the entire grid cell

(A) If a raw dataset (e.g., for glacier) specifies PCT cover as percent of the entire grid cell, for a grid cell that is partially ocean, then we can end up with too much vegetated area and too little area of this special landunit in the final surface dataset.

For example, consider the idealized scenario where all raw data sets are at the same resolution, all raw data sets agree about the land/ocean mask/fraction, the model resolution matches this raw data set resolution, and the ocean model grid also agrees exactly with this land/ocean fraction. Let's consider a 0.05 deg grid cell in this scenario where the raw data sees 98% ocean, 1% glacier and 1% vegetated. If the glacier raw data set encoded this as 1% glacier, then I think that mksurfdata_map would create a model grid cell that is 1% glacier and 99% vegetated. When run in the model, with a landfrac of 2%, the final grid cell cover would be 98% ocean (i.e., owned by the ocean model), 0.02% glacier and 1.98% vegetated; this would be wrong. If, instead, the glacier raw data set encoded this as 50% glacier, then I think mksurfdata_map would create a model grid cell that is 50% glacier and 50% vegetated. When run in the model, with a landfrac of 2%, the final grid cell cover would be 98% ocean, 1% glacier and 1% vegetated, as desired. So for this idealized scenario where all data sets agree in their resolution and land fraction and no mapping is needed, I think we would want raw data sets to give percent cover in terms of percentage of just the land area in the grid cell, at least given the current treatments of wetland area.

#### Problems with specifying raw dataset PCT cover as percent of just the land area

However, there are also problems with doing the reverse – specifying raw dataset PCT cover as percent of just the land area. I can think of two off-hand:

(B) Averaging multiple source grid cells to form a destination cell: Consider a model (destination) grid cell that exactly contains two raw data (source) grid cells. One of the source grid cells is as above – with 98% ocean, 1% glacier and 1% vegetated; if we encode glacier area as percent of just the land area, the raw dataset would prescribe 50% glacier area. The other source grid cell is 100% land, but has 0% glacier. As above, let's assume that the ocean grid agrees exactly with the land fractions on the raw data sets. Then the resulting land grid cell would be 51% land and 49% ocean. The glacier landunit in this grid cell would cover 25% of the land area – since it is determined by averaging the 50% in grid cell 1 with 0% in grid cell 2. This is much more glacier area than there should really be: since grid cell 1 is mostly ocean, its percent glacier cover should have gotten much less weight in forming the average, and the actual glacier percent over the land portion of this grid cell should be a little less than 1%. (We could handle this by having a land fraction on each raw data set and including that in the mapping, but this could be hard to get right if different raw data sets have different land masks, and I believe that there is a simpler solution, as I explain below.)

(C) Filling missing areas with wetlands: Another scenario is one where the ocean model grid doesn't resolve the coast line exactly, so CTSM needs to fill in some missing areas with wetland. Our logic for how much wetland area to apply is currently inconsistent: see https://github.com/ESCOMP/CTSM/issues/1001#issuecomment-678518813 , where I state:

> I think that the faulty logic is in what we do with wetland % in grid cells with pftdata_mask > 1e-6: I'm starting to think that our handling of wetlands with respect to lake and glacier is correct (in places with pftdata_mask = 0, fill the remainder of the grid cell with wetlands where it wasn't filled with glacier or lake), and that the real problem is that, anywhere where we have pftdata_mask > 1e-6, we fill the whole grid cell with vegetation.

(see that issue comment for more elaboration). However, I believe that comment is only correct if the glacier and lake raw data sets give percent cover in terms of percentage of the entire grid cell.

Consider a scenario like that in (A), but now the ocean model grid does not fully resolve the coast line, so the land model grid cell in question ends up with 100% landfrac. In this case, we would want a wetland area of 98%. I think there is currently no way to achieve this 98% wetland cover in this scenario, but we certainly cannot achieve it with a glacier raw data set that prescribes 50% glacier cover, as suggested in (A).

(This was my motivation for this comment last year: https://github.com/ESCOMP/CTSM/issues/1406#issuecomment-871555350 )

### Suggested path forward

I believe that we can handle these scenarios and others correctly if we put in place the following conventions and changes:

(1) **Raw data sets of PCT cover should represent the fields in terms of percentage of the total grid cell area.** This keeps the averaging of raw data grid cells (scenario (B)) simple, without requiring use of the land fraction associated with each raw data set. This may already be how things are done; I'm not sure.

(2) **Change the setting of wetland cover in mksurfdata_map so that it fills all area outside the current raw data set masks, rather than the current logic where wetland area is only added if** `pctlnd_pft(n) < 1.e-6_r8`. More exactly, I think that it would make sense to consider two areas: (a) the mapped land fraction given by `pctlnd_pft`, and (b) the total special landunit area. It is important to consider (b) so that we don't overwrite some special landunits with wetland where these special landunits extend beyond the PFT data's land fraction. (This is particularly important for glaciers, where we can have floating ice shelves: see https://github.com/ESCOMP/CTSM/issues/545 .) We could probably do even better by considering each raw data set's notion of the land fraction, but thinking about these multiple, disagreeing land fractions makes my head hurt. So I propose treating the PFT data's land fraction as the definitive land fraction for the sake of setting wetland area, but imposing a limit that ensures that it doesn't overwrite any special landunits: something like:

``` fortran
pctwet(n) = min(100._r8 - pctlnd_pft(n), 100._r8 - pctgla(n) - pctlak(n) - pcturb(n))
pctwet(n) = max(pctwet(n), 0._r8)
```

In addition to setting wetland area more correctly, this formulation will also deal with scenario (A): The problem with the current code that leads to the issue in scenario (A) is that, for a grid cell with `pctlnd_pft(n) > 1e-6`, there is no wetland area and the natural vegetated landunit is given all of the remaining area in the grid cell. With the proposal here, most of the grid cell in scenario (A) would be filled with wetland, and natural vegetation would only fill the remainder: 1% of the grid cell, as desired.

Notes about the above:
- I am assuming that `pctlnd_pft` is formed by mapping a field that specifies the land *fraction* – not just land *mask* – in each source grid cell. That is, I'm assuming that the source data for this field can be any value between 0 and 1 on the raw data grid. I'm not sure whether that is currently the case.
- Note that all special land units are now included here, in contrast to the current code for setting wetland area, which ignores urban area. (Maybe that should have been changed in the context of https://github.com/ESCOMP/CTSM/pull/1587 but I didn't think of it then.) I am *not* including crop area... I think it's correct not to include crop area in this because I think it's effectively handled via `pctlnd_pft` in this respect (though I'm not positive about this).
  - Update (2022-11-17): I think I *will* include crop area: see https://github.com/ESCOMP/CTSM/issues/1716#issuecomment-1318748012
- This isn't completely correct in some cases – e.g., consider the case where `pctlnd_pft(n) = 50` and `pctgla(n) = 50`, but all of the glacier cover is floating ice shelves that are complementary to the land fraction. Then we would really want `pctwet(n)` to be 0, but we would end up with it being 50. I think that handling this entirely correctly could be done with some extra mappings, but I think we can live with this error.

(See also https://github.com/ESCOMP/CTSM/issues/1001#issuecomment-678518813 for a similar suggestion, though I think I got the exact solution wrong there.)

Update (2022-04-29): If we wanted to accommodate inland wetlands (which are no longer used by default but may still be allowed), then we would need to add the above expression to any existing wetland areas read from the wetland raw dataset.

(3) **Change CTSM's runtime code so that, in grid cells that have landfrac < 1, wetland area is preferentially decreased and then the remaining landunit areas (including any remaining wetland areas) are renormalized appropriately.** This is needed because, within CTSM, we think of each landunit's weight as being the fractional area of the land-owned portion of the grid cell, *not* the fractional area of the entire grid cell. I'm not aware of this correction being done currently, but I'm pretty sure that it should be done – at least if we do the other things suggested above. This is particularly important with the change proposed in (2), since now we will more often have fractional wetland areas in a grid cell. (Without the change in (2), I think it has often – though not always – been the case that wetland area is either 0% or 100% of a grid cell, so this change (3) hasn't been as important.)

**Care will need to be taken here to ensure that this same reweighting is also applied to transient land cover – including possibly the dynamic glacier areas that come from CISM (I want to give this glacier issue a little more thought).**
- Update (2022-04-25): see https://github.com/ESCOMP/CTSM/issues/1716#issuecomment-1108959121 for thoughts on the dynamic glacier areas from CISM. The summary is that I think these are already expressed as a fraction of the land-owned area, so they do NOT need to be reweighted – i.e., this coupling can stay as it currently is.
- Update (2022-04-29): My initial thought here is that, in initialization, we would compute a scaling factor for each grid cell based on a combination of landfrac and the wetland area on the surface dataset. We would store a gridcell-level array with those scaling factors. Then, when reading landunit areas from the landuse timeseries file, we would apply the same saved scaling factors.

For example, in scenario (A), the surface dataset would specify 98% wetland, 1% glacier and 1% natural vegetation. The runtime land fraction is 2%. Wetland area would be preferentially decreased – to exactly 0% in this case – and then the remaining areas would be renormalized so that the resulting grid cell has 50% glacier and 50% natural vegetation over the small land area – as desired.

In scenario (C) we again have 98% wetland, 1% glacier and 1% natural vegetation on the surface dataset, but now the runtime land fraction is 100%. There would be no decrease of wetland area in this scenario, and the resulting grid cell would remain with 98% wetland, 1% glacier and 1% natural vegetation – again, as desired.

Also consider an intermediate scenario with the same specification on the surface dataset but where the runtime land fraction is 50%. Wetland area would be decreased to 48%, then the remaining landunit areas would be renormalized, resulting in a grid cell where the land-owned areas are covered by 96% wetland, 2% glacier and 2% natural vegetation. (So, in terms of the total grid cell area, we would have 50% ocean, `0.5*96%` = 48% wetland `0.5*2%` = 1% glacier, and `0.5*2%` = 1% natural vegetation – once again, as desired.)

Finally, consider a scenario where the runtime land fraction is greater than the wetland area: again, the surface dataset specifies 98% wetland, 1% glacier and 1% natural vegetation, but now the runtime land fraction is 1%. This would work out the same as the first example: wetland area would be completely decreased, and other landunit areas would be renormalized. No explicit changes are needed to account for the land fraction being smaller than the specified wetland area.

(4) Update (2022-05-03): I realized that a fourth piece is needed, too: **Change the remapping of PCT landunit areas in mksurfdata to NOT normalize by the mask.** See https://github.com/escomp/ctsm/issues/1716#issuecomment-1116375444 for details.

### Final thoughts / summary about how this changes the treatment of wetland areas

In the past, we have *sort of* had a rule that wetlands covered either 0% or 100% of a grid cell. Before the fix for https://github.com/ESCOMP/CTSM/issues/545 the exception was lakes, and after that fix the exception was lake and glacier areas. But I am coming to see that – in addition to the inconsistency in how different landunits are treated with respect to determining wetland area – this rule makes it hard/impossible to treat other aspects of landunit areas correctly.

I am therefore proposing a more rigorous determination of wetland areas, both in mksurfdata_map and at runtime, which I believe will allow us to have a more correct and more conceptually consistent treatment of landunit areas in coastal grid cells.

### LINKED ISSUE #1930: North and South pole set to glacier on surface datasets regardless of raw data specification
### Brief summary of bug

mksurfdata has been forcing the North and South Poles to glacier on the surface datasets regardless of the raw data specification. In general this won't cause problems – the South Pole truly should be glacier, and the North Pole is typically owned by the ocean – but in unusual cases (e.g., different continental arrangements or very warm paleo climates) this could cause incorrect land cover at the poles.

### General bug information

**CTSM version you are using:** Latest master, but has existed essentially forever.

**Does this bug cause significantly incorrect results in the model's science?** Yes, but only very small and only in rare setups

**Configurations affected:** Rare cases (see above)

billsacks:
### Description of changes

This PR reworks the mapping of landunit areas and the determination of wetland areas according to the plan laid out in #1716. (Note that the final implementation differs from the initial comment in #1716, as discussed in follow-up comments in that issue.) See the discussion there as well as the design document in this PR for details.

### Specific notes

Contributors other than yourself, if any: Various discussions with others, especially @ekluzek and @slevisconsulting 

- Resolves #1716 
- Resolves #1930 

Are answers expected to change (and if so in what way)? Areas will change on surface datasets

Testing performed, if any:

(1) Compared 1850 f09 surface dataset before and after these changes (and after each individual change from the individual commits on this branch)

(2) By hard-coding landunit areas before the call to `normalize_and_check_landuse`, I performed various "manual unit tests" of the normalization code, using diffs like this:

``` diff
diff --git a/tools/mksurfdata_esmf/src/mksurfdata.F90 b/tools/mksurfdata_esmf/src/mksurfdata.F90
index eb91be30c..0f347f8c0 100644
--- a/tools/mksurfdata_esmf/src/mksurfdata.F90
+++ b/tools/mksurfdata_esmf/src/mksurfdata.F90
@@ -593,6 +593,17 @@ program mksurfdata
   ! Perform other normalizations
   ! -----------------------------------
 
+  do n = 1, lsize_o
+     pctlnd_pft(n) = 0._r8
+
+     call pctnatpft(n)%set_pct_l2g(0._r8)
+     call pctcft(n)%set_pct_l2g(0._r8)
+     pctlak(n) = 0._r8
+     pcturb(n) = 0._r8
+     pctgla(n) = 0._r8
+     pctwet(n) = 0._r8
+  end do
+
   ! Normalize land use and make sure things add up to 100% as well as
   ! checking that things are as they should be.
   call normalize_and_check_landuse(lsize_o)
```

In the following, any landunit not explicitly mentioned had 0% area:

(a) With pctlnd_pft = 0% and all initial landunit areas 0%: wetland area = 100%

(b) With pctlnd_pft = 0% and initial glacier area 1%: glacier area = 100%

(c) With pctlnd_pft > 0% and all initial landunit areas 0%: natural vegetated area = 100%

(d) With pctlnd_pft = 40%, initial crop area 20%, natural vegetated area 10%: crop area = 50%, natural vegetated area = 50%

(e) With pctlnd_pft = 40%, initial crop area 20%, natural vegetated area 10%, glacier area 10%: crop area = 50%, natural vegetated area = 25%, glacier area = 25%

(f) With pctlnd_pft = 40%, initial crop area 20%, natural vegetated area 10%, glacier area 15%: crop area = 50%, natural vegetated area = 12.5%, glacier area = 37.5%

(g) With pctlnd_pft = 40%, initial crop area 20%, natural vegetated area 10%, glacier area 20%: crop area = 50%, glacier area = 50%

(h) With pctlnd_pft = 40%, initial crop area 20%, natural vegetated area 10%, glacier area 30%: crop area = 40%, glacier area = 60%

(i) With pctlnd_pft = 40%, initial crop area 0%, natural vegetated area 40%, glacier area 40%: glacier area = 100%

(j) With pctlnd_pft = 2%, initial natural vegetated area 1%, glacier area 1%: natural vegetated area = 50%, glacier area = 50%

(k) With pctlnd_pft = 2%, initial natural vegetated area 0%, glacier area 1%: natural vegetated area = 50%, glacier area = 50%

(l) With pctlnd_pft = 2%, initial natural vegetated area 2%, glacier area 1%: natural vegetated area = 50%, glacier area = 50%

### DIFF
--- a/tools/mksurfdata_esmf/src/mklanwatMod.F90
+++ b/tools/mksurfdata_esmf/src/mklanwatMod.F90
@@ -25,8 +25,8 @@ module mklanwatMod
   public :: mkwetlnd    ! make %wetland
   public :: update_max_array_lake  ! Update the maximum lake percent
 
-  real(r8), allocatable  :: frac_o_mklak(:)
-  type(ESMF_RouteHandle) :: routehandle_mklak
+  real(r8), allocatable  :: frac_o_mklak_nonorm(:)
+  type(ESMF_RouteHandle) :: routehandle_mklak_nonorm
 
   character(len=*) , parameter :: u_FILE_u = &
        __FILE__
@@ -114,10 +114,12 @@ contains
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ! Create a route handle between the input and output mesh
-    if (.not. ESMF_RouteHandleIsCreated(routehandle_mklak)) then
-       allocate(frac_o_mklak(ns_o))
-       call create_routehandle_r8(mesh_i=mesh_i, mesh_o=mesh_o, norm_by_fracs=.true., &
-            routehandle=routehandle_mklak, frac_o=frac_o_mklak, rc=rc)
+    if (.not. ESMF_RouteHandleIsCreated(routehandle_mklak_nonorm)) then
+       allocate(frac_o_mklak_nonorm(ns_o))
+       ! Note that norm_by_fracs is false in the following because this routehandle is
+       ! used to map fields that are expressed in terms of % of the grid cell.
+       call create_routehandle_r8(mesh_i=mesh_i, mesh_o=mesh_o, norm_by_fracs=.false., &
+            routehandle=routehandle_mklak_nonorm, frac_o=frac_o_mklak_nonorm, rc=rc)
        if (ChkErr(rc,__LINE__,u_FILE_u)) return
        call ESMF_VMLogMemInfo("After create routehandle in "//trim(subname))
     end if
@@ -138,14 +140,14 @@ contains
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
     ! Regrid lake_i to lake_o
-    call regrid_rawdata(mesh_i, mesh_o, routehandle_mklak, lake_i, lake_o, rc)
+    call regrid_rawdata(mesh_i, mesh_o, routehandle_mklak_nonorm, lake_i, lake_o, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
     do no = 1,ns_o
        if (lake_o(no) < 1.) lake_o(no) = 0.
     enddo
 
     ! Check global areas
-    call output_diagnostics_area(mesh_i, mesh_o, mask_i, frac_o_mklak, &
+    call output_diagnostics_area(mesh_i, mesh_o, mask_i, frac_o_mklak_nonorm, &
          lake_i, lake_o, "pct lake", percent=.true., ndiag=ndiag, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
@@ -159,8 +161,8 @@ contains
 
     ! Release memory
     if (mksrf_fdynuse == ' ') then  ! ...else we will reuse it
-       deallocate(frac_o_mklak)
-       call ESMF_RouteHandleDestroy(routehandle_mklak, nogarbage = .true., rc=rc)
+       deallocate(frac_o_mklak_nonorm)
+       call ESMF_RouteHandleDestroy(routehandle_mklak_nonorm, nogarbage = .true., rc=rc)
        if (chkerr(rc,__LINE__,u_FILE_u)) call shr_sys_abort()
     end if
     call ESMF_MeshDestroy(mesh_i, nogarbage = .true., rc=rc)
@@ -193,9 +195,11 @@ contains
 
     ! local variables
     type(ESMF_Mesh)        :: mesh_i
+    type(ESMF_RouteHandle) :: routehandle
     type(file_desc_t)      :: pioid_i
     integer , allocatable  :: mask_i(:)
     real(r8), allocatable  :: rmask_i(:)
+    real(r8), allocatable  :: frac_o(:)
     real(r8), allocatable  :: lakedepth_i(:) ! iput grid: lake depth (m)
     real(r8), allocatable  :: lakedepth_o(:) ! output grid: lake depth (m)
     integer                :: ni,no,k        ! indices
@@ -254,13 +258,11 @@ contains
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ! Create a route handle between the input and output mesh
-    if (.not. ESMF_RouteHandleIsCreated(routehandle_mklak)) then
-       allocate(frac_o_mklak(ns_o))
-       call create_routehandle_r8(mesh_i=mesh_i, mesh_o=mesh_o, norm_by_fracs=.true., &
-            routehandle=routehandle_mklak, frac_o=frac_o_mklak, rc=rc)
-       if (ChkErr(rc,__LINE__,u_FILE_u)) return
-       call ESMF_VMLogMemInfo("After create routehandle in "//trim(subname))
-    end if
+    allocate(frac_o(ns_o))
+    call create_routehandle_r8(mesh_i=mesh_i, mesh_o=mesh_o, norm_by_fracs=.true., &
+         routehandle=routehandle, frac_o=frac_o, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    call ESMF_VMLogMemInfo("After create routehandle in "//trim(subname))
 
     ! ----------------------------------------
     ! Create lake parameter (lakedepth)
@@ -278,11 +280,11 @@ contains
 
     ! regrid lakedepth_i to lakedepth_o - this also returns lakedepth_i to be used in the global sums below
     allocate (lakedepth_o(ns_o)); lakedepth_o(:) = spval
-    call regrid_rawdata(mesh_i, mesh_o, routehandle_mklak, lakedepth_i, lakedepth_o, rc)
+    call regrid_rawdata(mesh_i, mesh_o, routehandle, lakedepth_i, lakedepth_o, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
     call ESMF_VMLogMemInfo("After regrid_rawdata for lakedepth in "//trim(subname))
     do no = 1,ns_o
-       if (frac_o_mklak(no) == 0._r8) then
+       if (frac_o(no) == 0._r8) then
           lakedepth_o(no) = 10._r8
        end if
     enddo
@@ -300,7 +302,7 @@ contains
     ! Check global areas for lake depth
     call output_diagnostics_continuous(mesh_i, mesh_o, &
          lakedepth_i, lakedepth_o, "lake depth", "m", &
-         ndiag=ndiag, rc=rc, mask_i=mask_i, frac_o=frac_o_mklak)
+         ndiag=ndiag, rc=rc, mask_i=mask_i, frac_o=frac_o)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
     ! ----------------------------------------
@@ -310,8 +312,8 @@ contains
     call pio_closefile(pioid_i)
 
     ! Release memory
-    deallocate(frac_o_mklak)
-    call ESMF_RouteHandleDestroy(routehandle_mklak, nogarbage = .true., rc=rc)
+    deallocate(frac_o)
+    call ESMF_RouteHandleDestroy(routehandle, nogarbage = .true., rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) call shr_sys_abort()
     call ESMF_MeshDestroy(mesh_i, nogarbage = .true., rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) call shr_sys_abort()
@@ -339,12 +341,12 @@ contains
     integer           , intent(out)   :: rc
 
     ! local variables
-    type(ESMF_RouteHandle) :: routehandle
+    type(ESMF_RouteHandle) :: routehandle_nonorm
     type(ESMF_Mesh)        :: mesh_i
     type(file_desc_t)      :: pioid_i
     integer , allocatable  :: mask_i(:) 
     real(r8), allocatable  :: rmask_i(:)
-    real(r8), allocatable  :: frac_o(:)
+    real(r8), allocatable  :: frac_o_nonorm(:)
     real(r8), allocatable  :: swmp_i(:)      ! input grid: percent wetland
     integer                :: ni,no,k        ! indices
     integer                :: ns_i,ns_o      ! local sizes
@@ -403,9 +405,11 @@ contains
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ! Create a route handle between the input and output mesh
-    allocate(frac_o(ns_o))
-    call create_routehandle_r8(mesh_i=mesh_i, mesh_o=mesh_o, norm_by_fracs=.true., &
-         routehandle=routehandle, frac_o=frac_o, rc=rc)
+    allocate(frac_o_nonorm(ns_o))
+    ! Note that norm_by_fracs is false in the following because this routehandle is
+    ! used to map fields that are expressed in terms of % of the grid cell.
+    call create_routehandle_r8(mesh_i=mesh_i, mesh_o=mesh_o, norm_by_fracs=.false., &
+         routehandle=routehandle_nonorm, frac_o=frac_o_nonorm, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
     call ESMF_VMLogMemInfo("After create routehandle in "//trim(subname))
 
@@ -416,7 +420,7 @@ contains
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
     ! regrid swmp_i to swmp_o - this also returns swmp_i to be used in the global sums below
-    call regrid_rawdata(mesh_i, mesh_o, routehandle, swmp_i, swmp_o, rc)
+    call regrid_rawdata(mesh_i, mesh_o, routehandle_nonorm, swmp_i, swmp_o, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
     call ESMF_VMLogMemInfo("After regrid_data for wetland")
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
@@ -425,7 +429,7 @@ contains
     enddo
 
     ! Check global areas
-    call output_diagnostics_area(mesh_i, mesh_o, mask_i, frac_o, &
+    call output_diagnostics_area(mesh_i, mesh_o, mask_i, frac_o_nonorm, &
          swmp_i, swmp_o, "pct wetland", percent=.true., ndiag=ndiag, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
@@ -433,7 +437,7 @@ contains
     call pio_closefile(pioid_i)
 
     ! Release memory
-    call ESMF_RouteHandleDestroy(routehandle, nogarbage = .true., rc=rc)
+    call ESMF_RouteHandleDestroy(routehandle_nonorm, nogarbage = .true., rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) call shr_sys_abort()
     call ESMF_MeshDestroy(mesh_i, nogarbage = .true., rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) call shr_sys_abort()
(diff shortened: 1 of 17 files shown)


billsacks:
@ekluzek @slevisconsulting - I just pushed a somewhat-related change to this branch, where I have removed the code that hard-codes the pole points to glacier. I never liked that we had this code - e.g., this could trip someone up if they were trying to create a surface dataset for a very different climate / continental configuration - and it was leading to weirdness in the landfrac at the north pole (setting that landfrac to 1, which wasn't a problem per se, but was weird). I looked at diffs in an f09 surface dataset from removing it, and the only diff is as expected: the north pole is now classified as wetland rather than glacier; there is no change at the south pole (which is still classified as glacier).

Please let me know if you can think of any reason why this hard-coding at the poles should be retained.

billsacks:
I have looked at the difference in LANDFRAC_MKSURFDATA for an f09 transient dataset, between 1850 and 2005. There are only 8 grid cells that have any difference in LANDFRAC_MKSURFDATA between years 1850 and 2005. Of these, only 1 has a difference exceeding 1e-4: one grid cell has a difference of 1.45e-3; this is a grid cell at the mouth of the Mediterranean Sea. This is good: there are NOT big differences in our estimated land fraction in a transient case.

@ekluzek and @slevisconsulting - at this point I feel like this is ready to be merged if you're satisfied with it.

slevisconsulting (approved):
@billsacks thank you for your careful work in this PR! Looks good to me. I'm clicking "approve" and will let you and @ekluzek decide whether you also need his approval.

ekluzek:
@billsacks the change you made on the poles in general makes sense to me. Although it does seem a bit weird if the north pole is now wetland rather than glacier. I'm just wondering if you found some information in git logs (or elsewhere) about why this was done way back in the day? That context can be really helpful in figuring out if we should retain a feature we now question.

billsacks:
@ekluzek I just checked the history and this adjustment of the pole points dates all the way back to the initial version in the CLM svn repository. If I had to guess, my guess would be that it was needed at one time due to problems with mapping to the poles.

Why do you think it's weird if the north pole is now wetland? That is consistent with the raw datasets, where the north pole looks like ocean and isn't claimed by any landunit. This makes the north pole end up with the same land cover as the latitudes slightly south of the north pole, which seems more correct and consistent to me.

slevisconsulting:
> [...] That is consistent with the raw datasets, where the north pole looks like ocean and isn't claimed by any landunit. This makes the north pole end up with the same land cover as the latitudes slightly south of the north pole, which seems more correct and consistent to me.

Am I right to also conclude that, if a user put a continent over the North Pole with evergreen trees, then the NP would be land with evergreen trees?

ekluzek:
Ahh, OK, thanks for that clarification. I wasn't clear what this really looked like. It sounds like you used to have a completely isolated glacier point around the North pole that was disconnected to anything, the new method does make more sense. We did used to have more issues with remapping, so it makes sense it had something to do with that.

billsacks:
> It sounds like you used to have a completely isolated glacier point around the North pole that was disconnected to anything

Yes, exactly. Here's what PCT_GLACIER looked like before my recent change: notice the red band (100% glacier) in the far northern latitude band:

Screen Shot 2022-12-05 at 11 43 09 AM

slevisconsulting:
I talked with @ekluzek this morning and we agreed:
- He will look over this PR before it gets merged.
- Once ready, do you @billsacks have a preference for who does the merging/tagging? I think any of the three of us could do it.

billsacks:
> * Once ready, do you have a preference for who does the merging/tagging? I think any of the three of us could do it.

Thanks. No preference.

tools/mksurfdata_esmf/src/mksurfdata.F90
```diff
@@ -1142,13 +1104,116 @@ subroutine normalize_and_check_landuse(ns_o)
             call shr_sys_abort()
          end if
 
-         ! Natural vegetated cover is 100% minus the sum of all other landunits
+         ! Determine the percent of each grid cell considered to be land. (See comments in
+         ! https://github.com/ESCOMP/CTSM/issues/1716 for details.)
+         !
+         ! Start by using the land fraction field from the PFT raw data set:
+         pct_land = pctlnd_pft(n)
+         !
```
ekluzek, line 1112: I appreciate you being very explicit in your thinking here. I think that is good, and will help when we look at this in the future. But, lots of text in comments is also problematic in code. I'm wondering if you could also have a TLDR section that lists the underlying assumptions in a really short way with bullet points?

tools/mksurfdata_esmf/src/mksurfdata.F90
```diff
@@ -1242,15 +1307,6 @@ subroutine normalize_and_check_landuse(ns_o)
 
       end do
 
-      ! Make sure that there is no vegetation outside the pft mask
```
ekluzek, line 1245: Why was this removed?

ekluzek (approved):
OK, I have a few questions and comments. And mostly I wanted to make sure I understood this a little better by looking through it.

The general idea is that the routehandles now normally are normalized by fractional area, and if they are NOT they have "_nonorm" at the end to say they aren't.

There's also a bug that you caught that made the North pole glacier, which is really good to catch. I'm thinking that would be good to make into an issue for visibility across the project.

tools/mksurfdata_esmf/src/mklanwatMod.F90
```diff
@@ -253,12 +258,11 @@ subroutine mklakdep(file_mesh_i, file_data_i, mesh_o, pioid_o, fsurdat, rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ! Create a route handle between the input and output mesh
-    if (.not. ESMF_RouteHandleIsCreated(routehandle_mklak)) then
```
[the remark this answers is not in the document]
billsacks: Right. This may have been needed when the same routehandle was used for both lake areas and the lake depth parameter: then, for transient datasets, we wanted to make sure that the routehandle was only created once even though it was used many times. Now I have separated the routehandle used for lake areas from the one used for lake depth, because the mappings are now done differently. So you'll notice that I introduced a new local variable for routehandle (line 198) and, even in a transient run, this routine is just called once, so this local routehandle is created when the routine is called and just used the one time.

billsacks:
> The general idea is that the routehandles now normally are normalized by fractional area, and if they are NOT they have "_nonorm" at the end to say they aren't.
> 
> There's also a bug that you caught that made the North pole glacier, which is really good to catch. I'm thinking that would be good to make into an issue for visibility across the project.

Yes, thanks for the summary. I just opened #1930 and marked it as resolved in the initial comment in this PR. In general, raw datasets on fractional area are mapped without normalization, whereas most parameters are mapped with normalization.

tools/mksurfdata_esmf/src/mklanwatMod.F90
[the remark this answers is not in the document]
ekluzek: Thanks for that explanation. That makes perfect sense to me.

tools/mksurfdata_esmf/src/mksurfdata.F90
billsacks: Good question. This error check felt like it wasn't serving much purpose anymore, because the new code in the above block - starting at line 1168 - explicitly zeroes out the vegetation in this case. If I remember correctly, this might have served more of a purpose before my changes, because the way this was done might have been less direct at that point.

tools/mksurfdata_esmf/src/mksurfdata.F90
billsacks: Good point. I think the tldr was really the first sentence ("But we don't want to..."). I have reorganized the comment a bit to highlight that. See the commit I just pushed.

tools/mksurfdata_esmf/src/mksurfdata.F90
```diff
@@ -386,17 +389,7 @@ program mksurfdata
        pctlnd_o=pctlnd_pft, pctnatpft_o=pctnatpft, pctcft_o=pctcft, rc=rc)
   if (ChkErr(rc,__LINE__,u_FILE_u)) call shr_sys_abort('error in calling mkdomain')
 
-  ! If have pole points on grid - set south pole to glacier
-  ! north pole is assumed as non-land
   do n = 1,lsize_o
-     if (abs((lat(n) - 90._r8)) < 1.e-6_r8) then
```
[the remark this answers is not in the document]
ekluzek: This was done, as #1930

tools/mksurfdata_esmf/src/mksurfdata.F90
ekluzek: OK

tools/mksurfdata_esmf/src/mksurfdata.F90
ekluzek: Perfect

ekluzek:
I'm going to go ahead and merge this in.

PR #1925, opened 2022-12-13
Replace ncar_pylib because it is deprecated and will disappear soon

slevisconsulting:
### Description of changes
In `fsurdatmodifyctsm.py` prepare conda environment before running the fsurdat_modifier tool instead of using ncar_pylib.

### Specific notes

@ekluzek 

addresses #1786 (fixes?)

No, and we'll see...

Not visible to user.

Testing performed, if any:
Goal is for test
`FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel`
to continue working without relying on ncar_pylib.

### DIFF
--- a/cime_config/SystemTests/fsurdatmodifyctsm.py
+++ b/cime_config/SystemTests/fsurdatmodifyctsm.py
@@ -65,54 +65,24 @@ class FSURDATMODIFYCTSM(SystemTestsCommon):
     def _run_modify_fsurdat(self):
         tool_path = os.path.join(self._ctsm_root,
                                  'tools/modify_input_files/fsurdat_modifier')
-        # Need to specify a specific python version that has the required
-        # dependencies
-        python_path = _get_python_path()
-        subprocess.check_call([python_path, tool_path, self._cfg_file_path])
+        # Prepare conda environment before running the tool
+        # TODO slevis (delete these lines when all steps are done):
+        # 1) DONE Confirm that
+        # ./create_test FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev115
+        # works: PASS
+        # 2) DONE Confirm that the README instructions work in interactive mode
+        # >>> module unload python
+        # >>> module load conda
+        # >>> ./py_env_create
+        # >>> conda activate ctsm_py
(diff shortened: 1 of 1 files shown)


cime_config/SystemTests/fsurdatmodifyctsm.py
```diff
@@ -65,54 +65,24 @@ def _create_config_file(self):
...
+        # ./create_test FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev115
+        # works: PASS
+        # 2) DONE Confirm that the README instructions work in interactive mode
+        # >>> module unload python
+        # >>> module load conda
+        # >>> ./py_env_create
+        # >>> conda activate ctsm_py
+        # >>> cd tools/modify_input_files
+        # >>> ./fsurdat_modifier islas_examples/modify_fsurdat/fill_indian_ocean/modify_slevis_16pft.cfg
+        # 3) Execute those here
```
slevisconsulting, line 80: @ekluzek @billsacks 
Running the plan by you before I spend time on it because I think I remember Bill having hesitations about implementing the new commands inside the test:
- I have removed references to python_path.
- I will have the test execute the module load, module unload, py_env_create, and conda activate before running the tool.

ekluzek: Another way to do it would be to just ensure that python and conda are in the users path. And that the ctsm_py environment is active for conda. This would expect the user to execute those things outside of the system test itself.

billsacks: I forget the details, but I remember having issues in the past with loading and unloading python modules from within a system test, since CIME does its own module load of python. It may have been that it worked when running a single system test, but messed things up when running a whole test suite... but again, I forget the details.

There is some additional discussion in https://github.com/ESCOMP/CTSM/pull/1798.

I guess I'd say go ahead and try it, and as long as it works even when running a whole test suite, then it's fine by me.

slevisconsulting: I need to remind myself what @ekluzek had already accomplished in #1798.

@billsacks thanks for the reminder of the existence of #1798!

slevisconsulting:
#1798 is close to getting merged to main.
The present PR #1925 is obsolete, so I'm closing it.

PR #1933, opened 2023-01-11
Update neon_sites_dompft.csv

wwieder:
### Description of changes
incorrect lat-lon being used for ONAQ

### Specific notes
Likely also need to update surface dataset and generate new initial conditions files.

@TeaganKing 

- Yes, for NEON ONAQ sites

- Also need to check usermods


TeaganKing:
Thanks for making this change, @wwieder !

wwieder:
This PR also requires creating new surface data for ONAQ. For simplicity sake, I've made new surface datasets for all sites here `/glade/scratch/wwieder/single_point_neon_updated/surfdata_1x1_NEON_*_hist_78pfts_CMIP6_simyr2000_c230111.nc`

@ekluzek I'm assuming we're better off changing the default surface datasets in `cime_config/usermods_dirs/NEON/defaults/user_nl_clm` to point to these new surface data files, even though it's only ONAQ that's changed?

wwieder:
While we're mucking around I went ahead and looked at lat-lon coordinates for other NEON sites.  I'm not sure of the history of the `neon_sites_dompft.csv` file, but comparing to the [NEON_Field_Site_Metadata_20220412.csv file that's available from NEON](https://www.neonscience.org/field-sites/explore-field-sites). There are a few other differences at ORNL, UNDE, and BLAN that I'll go ahead and change here now.

PR #1946, opened 2023-01-31
Newest raw datasets for pft, lai, soilcolor

slevis-lmwg:
### Description of changes

Mainly I'm updating file paths/names in `gen_mksurfdata_namelist.xml`.
While testing, I'm also updating codes and README with necessary corrections or updates.

### Specific notes

@ekluzek @lawrencepj1 

Fixes #1873 

Yes: new raw datasets; eventually all new fsurdat files.

New raw datasets. Eventually all new fsurdat files.

Testing performed, if any:
This far I have used mksurfdata_esmf to generate 2-degree fsurdat (and landuse when applicable) files for
1) 2000
2) 1850-2015
3) 1850-2100
4) 2016-2100

@samsrabin is using (2) with ctsm5.1 code. He had to address the following ctsm5.1 vs. ctsm5.2 inconsistencies:
- Put PFTDATA_MASK back in the new fsurdat file (relates to #1703 )
- Hardwire nlevurb = 10 for consistency with new fsurdat file
- Possible problem with TK_IMPROAD and CV_IMPROAD?


samsrabin:
One other thing I noticed in the 2-degree landuse.timeseries files: The ctsm5.2 file's latitude values differ very slightly from those in the ctsm5.0 file. We're talking about floating-point error magnitudes (largest discrepancy < 5e-14), but it did initially mess me up when trying to compare the two files in Python.

olyson:
I saw here a reference to a couple of urban-related ctsm5.1 vs ctsm5.2 inconsistencies.  Let me know if you need help.

samsrabin:
The 2-degree 5.2 dataset is also missing the AREA variable, which might not be needed in the model but is really helpful for postprocessing.

slevisconsulting:
> The 2-degree 5.2 dataset is also missing the AREA variable, which [...] is really helpful for postprocessing.

Agreed, though `area` still appears in CTSM history. I don't know (or remember) if there was discussion that led to removing AREA from fsurdat files or if @mvertens made an executive decision. I'm open to putting AREA back in fsurdat files. I'd like feedback from the group.

wwieder:
It seems like area and landfrac are already on history files, which I assume means they aren't needed on surface data too (unless the history files get this information from the surface datasets)?

mvertens:
landfrac doesn't belong on the surface dataset since it is computed at run time from an ocean mask file and many different landfrac profiles can be associated with a single surface dataset that has land properties even over ocean (before this was wetland - I'm not sure what it is now).
I don't remember why I removed area. I think that @wwieder is correct - this is already on the history files.

samsrabin:
> It seems like area and landfrac are already on history files, which I assume means they aren't needed on surface data too

The reason I thought area and landfrac might be good to include on the surface and landuse timeseries files is that it facilitates working with those files directly, as I've done to compare between different versions. It also seems like it wouldn't be much extra data, since they're just lat x lon. 

However, @mvertens, good point about landfrac potentially being in conflict with what is actually used in a run. Could it be named something different? If the datasets being used to generate the files have their own landfrac equivalent, perhaps it could be called `LANDFRAC_DATASOURCE` or something.

All that said, I'm not sure my use case is common enough that this is worth addressing!

billsacks:
LANDFRAC is no longer on the surface dataset for the reasons mentioned by @mvertens . However, there are (or should be) two new variables:
- LANDFRAC_PFT: the land fraction mapped from the PFT raw data files
- LANDFRAC_MKSURFDATA: the actual land fraction used internally by mksurfdata to renormalize landunit areas; this will often match LANDFRAC_PFT, but can differ depending on the areas of various landunits

samsrabin:
Ahh, okay, I didn't know "landfrac" referred to something different from `LANDFRAC_PFT`. Thanks for the explanation. Both `LANDFRAC_PFT` and `LANDFRAC_MKSURFDATA` are indeed on the ctsm5.2 file I'm looking at.

slevisconsulting:
Note to self that I had opened #1873 in October:
Make sure that mods in #1873 are present in (or get merged with) this PR.

slevis-lmwg:
To generate a 2000 fsurdat file for @Ivanderkelen on the 360x720cru grid, I had to update two things:
1) The .xml file (see [last commit](https://github.com/ESCOMP/CTSM/pull/1946/commits/2f84b2608d6bcf17845ff81eed0ba0304f1811ec)). This problem was unrelated to the specific grid requested, and I'm surprised that I had not encountered it making timeslice files before.
2) `./rimport -file share/meshes/360x720_120830_ESMFmesh_c20210507_cdf5.nc`  because the mesh file that we were pointing to before was netcdf-4.

slevis-lmwg:
I completed this test to confirm that the Gross Unrepresented Land Use (GULU) code addition works:
`/glade/scratch/slevis/SMS_Ly3_PS.f09_g17.IHistClm50BgcCrop.cheyenne_intel.clm-f09_dec1990Start_GU_LULCC.20230516_140916_1q2hsw`

@lawrencepj1 (in a meeting yesterday) suggested what to look for in the new variables prefixed with GRU. As far as I can tell the simulation worked correctly, so I will go ahead with next steps:
- Merge and tag #1946

Issue #1979, opened 2023-04-02
Need some changes to avoid negative h2osoi_ice in UpdateState_TopLayerFluxes


billsacks:
### Brief summary of bug

In UpdateState_TopLayerFluxes, there is a somewhat arbitrary epsilon used to determine if h2osoi_ice and h2osoi_liq are close enough to zero that they should be truncated to zero. If the state remains negative after this truncation, we deem this to be a problem. It seems that this tolerance is occasionally exceeded, leading runs to abort. Since this tolerance is somewhat arbitrary, we will loosen it by an order of magnitude.

### General bug information

**CTSM version you are using:** Recent versions

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** Unknown

### Details of bug

For details, see comments in #1253 starting with https://github.com/ESCOMP/CTSM/issues/1253#issuecomment-1490812859

See also #988

billsacks:
From discussion with @olyson and @swensosc - We're going to loosen the tolerance further to 1e-10 in the hopes that we'll avoid this problem moving forward. If this problem crops up again, see code in https://github.com/ESCOMP/CTSM/blob/4e126c75a2eaebc6c1c32da4f728bd0891599ff5/src/biogeophys/SoilFluxesMod.F90#L307-L359 and see https://github.com/ESCOMP/CTSM/pull/1282 for some guidance on what might be needed. We do see some potential for issues if there are big differences between h2osoi_liq and h2osoi_ice (in which case the liquid or ice flux might have relatively large errors relative to the liquid or ice state – since relative errors depend on the largest term in the equation), so we may need to do some corrections to ensure that fluxes aren't leading us to exceed either the liq or ice state values, probably when these fluxes are computed in SoilFluxesMod.

billsacks:
@olyson and @swensosc - I woke up this morning wondering if I could find a relatively easy solution that would get more to the heart of the problem, and spent a bit of time looking into this - though didn't want to spend too long on it (analyzing, developing & testing), so where I currently stand is about where we were yesterday – that, for now, we should just loosen the tolerance. **However, I'm inclined to loosen it even more than we suggested - to 1e-9 rather than 1e-10. Are you comfortable with that?** I figure that a reduction in state by a factor of 1 billion is effectively reducing the state to 0, so we might as well set it to exactly 0.

That said, I'll document here the thoughts I had this morning in case we want to come back to this - or in case either of you think this is worth pursuing further here.

[...]

PR #1980, opened 2023-04-02
Loosen tolerance on near-zero truncation of h2osoi_ice and h2osoi_liq

billsacks:
### Description of changes

In UpdateState_TopLayerFluxes, there is a somewhat arbitrary epsilon used to determine if h2osoi_ice and h2osoi_liq are close enough to zero that they should be truncated to zero. If the state remains negative after this truncation, we deem this to be a problem. It seems that this tolerance is occasionally exceeded, leading runs to abort. Since this tolerance is somewhat arbitrary, we will loosen it by an order of magnitude.

Note that, although the issue was only reported for h2osoi_ice, I am changing the tolerance for both liq and ice so they remain consistent.

### Specific notes

Resolves ESCOMP/CTSM#1979

Yes, possible small answer changes due to the possibility that slightly non-zero states are now truncated to exactly zero.


billsacks:
Closing this because I'm going to go with a more robust fix.

PR #1987, opened 2023-04-10
Rework handling of evaporation constraint in SoilFluxes

### LINKED ISSUE #1979: Need some changes to avoid negative h2osoi_ice in UpdateState_TopLayerFluxes

billsacks:
### Description of changes

Occasionally, h2osoi_ice was going significantly negative in UpdateState_TopLayerFluxes - see
https://github.com/ESCOMP/CTSM/issues/1979. As noted in that issue, this seems to be due to h2osoi_ice having a very different magnitude from h2osoi_liq, leading to greater-than-roundoff-level differences from zero final state in a relative sense (i.e., relative to the magnitude of h2osoi_ice) - I think because of the appearance of the sum (h2osoi_ice + h2osoi_liq) in the equations that limit fluxes.

To try to deal with this, I have reworked the handling of the evaporation constraint to directly limit both the liqevap and solidevap, so that both of them should result in the equivalent liq or ice states going to 0 within roundoff.

To do that, I needed to move the partitioning of the total flux into liquid and solid to earlier in the subroutine and then recalculate those partitioning fluxes in conditions where we're applying an evaporation constraint.

Note that the deleted lines in this commit have just been moved as-is to earlier in this subroutine.

### Specific notes

Contributors other than yourself, if any: Discussions with @olyson and @swensosc 

CTSM Issues Fixed (include github issue)
- Resolves #1979 

Are answers expected to change (and if so in what way)? YES - occasional roundoff-level changes

Testing performed, if any:
Ran full aux_clm test suite. Most tests were bit-for-bit; a few had roundoff-level changes. One (SMS_D_Ln9_P36x3.f19_g17.IHistClm50Sp.cheyenne_intel.clm-waccmx_offline) had greater-than-roundoff-level changes; I am going to investigate that further to confirm that the fundamental changes are just roundoff-level and we're just seeing error growth.

### DIFF
--- a/src/biogeophys/SoilFluxesMod.F90
+++ b/src/biogeophys/SoilFluxesMod.F90
@@ -205,6 +205,74 @@ contains
          endif
       end do
 
+      ! Partition evaporation into liquid and solid
+      do fp = 1, num_nolakep
+         p = filter_nolakep(fp)
+         c = patch%column(p)
+         l = patch%landunit(p)
+         j = col%snl(c)+1
+
+         qflx_liqevap_from_top_layer(p)   = 0._r8
+         qflx_solidevap_from_top_layer(p) = 0._r8
+         qflx_soliddew_to_top_layer(p)    = 0._r8
+         qflx_liqdew_to_top_layer(p)      = 0._r8
+
+         ! Partition the evaporation from snow/soil surface into liquid evaporation,
+         ! solid evaporation (sublimation), liquid dew, or solid dew.  Note that the variables
+         ! affected here are all related to the snow subgrid patch only because of the use of qflx_ev_snow.
+         ! In the situations where there are snow layers or there is snow without an explicit snow layer,
+         ! the partitioned variables will represent the components of snow evaporation
+         ! (qflx_ev_snow = qflx_liqevap_from_top_layer + qflx_solidevap_from_top_layer
+         ! - qflx_liqdew_to_top_layer - qflx_soliddew_to_top_layer).
+         ! In the case of no snow, qflx_ev_snow has already been set equal to qflx_ev_soil (the evaporation
+         ! from the subgrid soil patch) and the partitioned variables will then represent evaporation from the
+         ! subgrid soil patch.
+         ! In the case of urban columns (and lake columns - see LakeHydrologyMod), there are no subgrid
+         ! patches and qflx_evap_soi is used. qflx_evap_soi = qflx_liqevap_from_top_layer + qflx_solidevap_from_top_layer
+         ! - qflx_liqdew_to_top_layer - qflx_soliddew_to_top_layer.
+         if (.not. lun%urbpoi(l)) then
+            if (qflx_ev_snow(p) >= 0._r8) then
+               ! for evaporation partitioning between liquid evap and ice sublimation,
+               ! use the ratio of liquid to (liquid+ice) in the top layer to determine split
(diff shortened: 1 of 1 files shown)


billsacks:
@olyson - can you please test this in your failing test (without changing the relative epsilon) to confirm that it solves the problem?

@swensosc - and @olyson if you want - I would welcome a review of the code diffs to make sure this makes sense to you.

src/biogeophys/SoilFluxesMod.F90
```diff
@@ -205,6 +205,74 @@ subroutine SoilFluxes (bounds, num_urbanl, filter_urbanl, &
...
+               if (t_grnd(c) < tfrz) then
+                  qflx_soliddew_to_top_layer(p) = abs(qflx_evap_soi(p))
+               else
+                  qflx_liqdew_to_top_layer(p) = abs(qflx_evap_soi(p))
+               end if
+            end if
+
+         end if
+
+      end do
```
billsacks, line 274: This block of code was moved as-is from below: it looks safe to do this partitioning earlier in the subroutine, and doing so lets us directly adjust the partitioning later.

src/biogeophys/SoilFluxesMod.F90
```diff
@@ -235,11 +305,27 @@ subroutine SoilFluxes (bounds, num_urbanl, filter_urbanl, &
                evaporation_demand = qflx_evap_soi(p)
                qflx_evap_soi(p)   = evaporation_limit
                qflx_ev_snow(p)    = qflx_evap_soi(p)
+               qflx_liqevap_from_top_layer(p)   = h2osoi_liq(c,j)/dtime
+               qflx_solidevap_from_top_layer(p) = h2osoi_ice(c,j)/dtime
```
billsacks, line 309: Key piece 2 of 2: As above, now we modify the partitioning directly in a way that is less sensitive to roundoff issues when h2osoi_liq and h2osoi_ice have significantly different magnitudes.

billsacks:
@olyson - I realized that you may be using an older code base in your test that illustrates the problem. If so, you can bring in my change my cherry-picking my commit into your code base: `git cherry-pick c8b80d84d`.

olyson:
git cherry-pick c8b80d84d gives me:

fatal: bad revision 'c8b80d84d'

Is that because my code/case is on a branch, not on main?  I.e.,

"On branch dynurb_mksurf_ctsm5.2.mksurfdata_I1564"

olyson:
Hmm, still getting that error after doing git fetch.

olyson:
Ok, I got the cherry-pick to work by adding the remote, then fetching.

olyson (approved):
I reviewed the code, it makes sense to me.
I used the revised code in my case and it solved the error we were getting.

billsacks:
@swensosc and @olyson - great, thanks a lot! I want to investigate the larger diffs in the one test case a little further, then will do final testing on this branch.

PR #1991, opened 2023-04-21
remove deprecated shr_mpi from this module

jedwards4b:
### Description of changes
Remove the depricated shr_mpi_mod from this module and replace with ESMF_VMBroadcast. 
### Specific notes

Testing performed, if any:

### DIFF
--- a/src/cpl/nuopc/lnd_import_export.F90
+++ b/src/cpl/nuopc/lnd_import_export.F90
@@ -65,7 +65,7 @@ module lnd_import_export
   logical                :: flds_co2a        ! use case
   logical                :: flds_co2b        ! use case
   logical                :: flds_co2c        ! use case
-  logical                :: force_send_to_atm = .true.  ! Force sending export data to atmosphere even if ATM is not prognostic
+  logical                :: force_send_to_atm   ! Force sending export data to atmosphere even if ATM is not prognostic
   integer                :: glc_nec          ! number of glc elevation classes
   integer, parameter     :: debug = 0        ! internal debug level
 
@@ -1294,9 +1294,8 @@ contains
     ! ----------------------------------------------------
     ! Read in tne namelist for CTSM nuopc cap level items
     ! ----------------------------------------------------
+    use ESMF             , only : ESMF_VMGetCurrent, ESMF_VMBroadcast, ESMF_VM
     use clm_nlUtilsMod   , only : find_nlgroup_name
-    use shr_mpi_mod      , only : shr_mpi_bcast
-    use spmdMod          , only : mpicom
     use abortutils       , only : endrun
     use shr_log_mod      , only : errMsg => shr_log_errMsg
     ! !ARGUMENTS:
@@ -1304,10 +1303,18 @@ contains
     ! !LOCAL VARIABLES:
     integer            :: nu_nml                           ! unit for namelist file
     integer            :: nml_error                        ! namelist i/o error flag
+    integer            :: rc                               ! ESMF return code
+    integer            :: tmp(1)
+    type(ESMF_VM)      :: vm
     character(*), parameter :: nml_name = "ctsm_nuopc_cap" ! MUST match with namelist name below
+    
+
     namelist  /ctsm_nuopc_cap/ force_send_to_atm
 
+    tmp = 0
+    rc = ESMF_SUCCESS
     ! Read namelist
+    force_send_to_atm = .true.
     if (masterproc) then
        open( newunit=nu_nml, file=trim(NLFilename), status='old', iostat=nml_error )
        call find_nlgroup_name(nu_nml, nml_name, status=nml_error)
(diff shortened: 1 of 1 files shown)


billsacks (commented):
I am fine with the changes here, but can you say why shr_mpi_mod is being deprecated? We use that in a lot of places. (Apologies if I'm forgetting something that has already been discussed.)

jedwards4b:
shr_mpi is being deprecated because ESMF_VMBroadcast does the same thing plus it anticipates communications paradigms beyond MPI.

src/cpl/nuopc/lnd_import_export.F90
```diff
@@ -1294,20 +1294,27 @@ subroutine ReadCapNamelist( NLFilename )
...
-    use shr_mpi_mod      , only : shr_mpi_bcast
-    use spmdMod          , only : mpicom
     use abortutils       , only : endrun
     use shr_log_mod      , only : errMsg => shr_log_errMsg
     ! !ARGUMENTS:
     character(len=*), intent(IN) :: NLFilename   ! Namelist filename
     ! !LOCAL VARIABLES:
     integer            :: nu_nml                           ! unit for namelist file
     integer            :: nml_error                        ! namelist i/o error flag
+    integer            :: rc                               ! ESMF return code
```
billsacks, line 1306: - [ ] I'm thinking that rc should be a subroutine output argument that is checked by the caller so that errors can propagate up the call chain; do you agree?

billsacks (changes requested):
Thanks for the explanation of the deprecation of shr_mpi_mod. I feel like the best way to approach this, though, would be to maintain something like shr_mpi_mod (named the same or different) that wraps the relevant ESMF routine: that way the implementation could be based on ESMF_VMBroadcast but the use from CTSM and elsewhere would be no more complex than is currently the case – and we wouldn't need to make this many changes throughout the CTSM code.

I would tentatively envision an interface similar to the current shr_mpi_bcast but without the mpicom. For broadcasting a scalar logical, you could call something like `shr_[something]_broadcast(force_send_to_atm)`. In my tentative vision, there would be two optional arguments:
1. `vm`: Input: If not provided, this is obtained through a call to `ESMF_VMGetCurrent`. (It may not be worth providing this optional argument; the only reason I can see for it is a slight efficiency if you are doing a lot of calls with the same vm, to avoid having to get the vm over and over for each call.)
2. `rc`: Output: If provided, then the caller should check this value; if not provided, then the broadcast routine should abort if there is an error. I know that aborting in place doesn't follow the standard of what's done for ESMF return codes, but I feel that adding and maintaining all of the necessary propagation of errors up the call chain could be awkward in some places, so having the convenience of allowing the code to abort in place in the (unlikely) event of an error seems helpful.

If we don't do that, then I have a couple of inline comments that should possibly be addressed for the current implementation.

src/cpl/nuopc/lnd_import_export.F90
jedwards4b: done

src/cpl/nuopc/lnd_import_export.F90
```diff
@@ -1318,10 +1325,16 @@ subroutine ReadCapNamelist( NLFilename )
...
        end if
        close(nu_nml)
+       if (force_send_to_atm) tmp(1) = 1
     endif
+    call ESMF_VMGetCurrent(vm, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
     ! Broadcast namelist to all processors
-    call shr_mpi_bcast(force_send_to_atm  , mpicom)
+    call ESMF_VMBroadcast(vm, tmp, 1, 0, rc=rc)
```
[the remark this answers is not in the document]
jedwards4b: I tried using the logical first, I couldn't make it work.

target attribute added

billsacks:
Thanks for the changes, @jedwards4b . This looks good now.

I discussed this with @ekluzek and others at this morning's CTSM software meeting, and our feelings are:
- If there is a need for this to come in soon – e.g., because the current code is causing some problems – then we're happy to bring this in.
- But if this isn't critical, then we'd prefer to put in place a consistent solution for all of CTSM's many broadcasts, and ideally one for which the interface is about as simple as the current shr_mpi_bcast interface (through use of a wrapper function like in my above comment).

Can you please let us know the importance of bringing this in soon? Was there a reason, for example, why you needed to change this particular broadcast call?

billsacks (approved):
Approving, but see the comment I just made.

jedwards4b:
I changed it because a user reported a hang in the shr_mpi routine that was replaced, the problem wasn't ultimately with shr_mpi but I had done the work so I thought I could get it on the trunk.

billsacks:
We definitely appreciate the work and pushing things towards a better method long-term. Given that this isn't needed now and that we'd like to wrap it in an easy-to-use interface that we can use consistently throughout CTSM, I think we'd like to close this current version and talk about what we want to see long-term.

billsacks:
I feel like I shouldn't have closed this: even if we end up wanting to do this differently, it's good to have this starting point to guide our future implementation. So I'm reopening it for that purpose and also to invite further discussion. @jedwards4b I'm sorry if my closing this earlier came across as unappreciative of your work: in fact, I truly appreciate the work you do like this to try to incrementally move us to a better future position, even if we end up wanting to do the final implementation a little differently.

jedwards4b:
I think that this is a great example of perfection being the enemy of progress.   The development of the "easy-to-use interface" that you propose is probably never going to happen.   Interactions like this one actively discourage contributions from the community.

samsrabin:
From CTSM software meeting this morning: I will go ahead and bring this in, and make issues for the remaining concerns.

samsrabin:
All aux_clm tests pass bit-for-bit (aside from expected failures). Baselines saved as `remove_shr_mpi.merge142`.

PR #1996, opened 2023-05-04
Multilayer Canopy

### LINKED ISSUE #1922: Multi-layer canopy code
Gordon, @olyson and @slevis-lmwg have been discussing putting multilayer canopy code into the main repository for future developments.  This code will have limited applicability (short runs in SP mode only), but provide the hooks for others to continuing developing the code.  

This is a placeholder issue for this development until more details are provided about the code

slevisconsulting:
### Description of changes

I am putting Gordon Bonan's code on a branch starting at ctsm5.1.dev123. I have not tested, yet. I'm opening this PR to keep a record of the work and to allow Gordon to review how I brought his code into ctsm5.1.dev123.

### Specific notes

@gbonan

Fixes #1922 

Yes. New science.

Not, yet.

Testing performed, if any:
I have not tested, yet.


slevis-lmwg:
I'm testing like this:
`./create_test SMS_Ld3_PS.f09_g17.IHistClm50BgcCrop.cheyenne_intel.clm-f09_dec1990Start_GU_LULCC`

For now the code builds and begins to run but stops with this error:

```
Attempting to read RSL look-up table .....
(GETFIL): attempting to find local file
psihat.nc
(GETFIL): failed getting file from full path: ../rsl_lookup_tables/psihat.nc

 ERROR: GETFIL: FAILED to get ../rsl_lookup_tables/psihat.nc
```

slevis-lmwg:
Obtained psihat.nc from @gbonan's repository and submitted new test.

slevis-lmwg:
I have found that
`zw(p,0) = -2.8e-14`
in line 128 of MLinitVerticalMod.F90

This seems like a computer precision error to me, so I proposed to Gordon that I change the if statement to:
```
if (zw(p,0) < 0._r8 and zw(p,0) >= -1.e-10_r8) then
   zw(p,0) = 0._r8
else if (zw(p,0) >= 1.e-10_r8 .or. zw(p,0) < -1.e-10_r8) then
   call endrun (msg=' ERROR: initVerticalStructure: zw(p,0) improperly defined')
end if
```

slevis-lmwg:
Switching to this simpler test:
`./create_test SMS_Ld3_D.f19_g17.I2000Clm50BgcCrop.cheyenne_intel --walltime 00:20:00`

slevis-lmwg:
Increasing the following error tolerances and submitting the f19 test above:
- `zw(p,0) improperly defined` to 1e-9
- pai_err to 1e-5
- sai_err to 1e-5

Now I get an error that seems bigger than the previous ones:
```
Attempting to initialize multilayer canopy vertical structure .....
 ncan, p, nlevmlcan =         304           5         100
 ENDRUN:
 ERROR:  ERROR: initVerticalStructure: ncan > nlevmlcan
```
I printed a few more variables for p = 5:
```
ntop, n_zref, dz_zref, zref, ztop =
11   293   29.776757   30.89297   1.116
```
This leads me to believe that ncan(p) > nlevmlcan due to small ztop. Does this make sense to you @gbonan? If so, how would you like to handle it differently?

slevis-lmwg:
Notes from discussing with Will and Gordon last week:
- [x] Will updated the #2155 branch to a recent dev tag (dev168)
- [x] Merge #2155 branch to this one
- [x] Try Will's Plumber example /glade/campaign/cgd/tss/people/wwieder/cheyenne_archive/PLUMBER2_tests/ZM-Mon_test2
- [x] Are single- and multi-layer history fields are working? There may be one example of each in the code already. Does anything come out to history?
- [x] Try for US-NR1 (Ameriflux and Fluxnet code for Niwot Ridge).
- [x] Gordon takes over from there. Gordon and I will have a "tutorial" on derecho and/or izumi.

slevis-lmwg:
Attempting the 3rd checkbox above:
`./create_newcase --case ~/cases_plumber2/ZM-Mon_test2_slevis --res CLM_USRDAT --compset HIST_DATM%1PT_CLM51%BGC_SICE_SOCN_SROF_SGLC_SWAV_SESP --run-unsupported --user-mods-dir PLUMBER2/ZM-Mon`

The create_newcase ends with `errput: ERROR: No variable PLUMBER2SITE found in case`

This may not be a fatal error, so I'm proceeding with
```
./case.setup
./case.build
./case.submit
```
The job is pending in the queue.

ekluzek:
@slevis-lmwg the setting of PLUMBER2SITE should have come from the update from #2155 so I'm doubtful this will work. But, you should also check that the update from #2155 came in correctly.

slevis-lmwg:
GPP_ML and LWP_ML appear in history, though I'm not sure that they contain valid data...

slevis-lmwg:
Gordon, I created a new case for US-NR1 and submitted a run. It fails with this error:
```
Attempting to initialize multilayer canopy vertical structure .....
ENDRUN:
ERROR:  ERROR: initVerticalStructure: ncan > nlevmlcan
```
The site that I tested last week, ZM-Mon, completed a year without error. Maybe Niwot Ridge has no trees and triggers this error?

gbonan:
Hi Sam,

Is this an SP run or a BGC run? The difference is that canopy height is
specified in SP but is calculated in BGC. If the canopy height is small
(e.g., zero in an initial BGC run) and the forcing height is large, the
maximum number of canopy layers can be exceeded. I cannot remember if we
discussed this last week, but the run should be SP.

Here is what should be happening:

The tower height (forcing height) at US-NR1 is 26 m, and the canopy height
is 12-13 m. The model should be using a layer thickness of 0.5 cm, so there
is a total of 52 layers up to the tower forcing height. This is less than
the maximum number of layer (nlevmlcan = 100).

Here are some things to check in subroutine
initVerticalStructure (MLinitVerticalMod.F90):

What is the value for zref(p)? This should  be the tower height (about 26 m)
What is the value for ztop(p)? This should be the canopy height (about
12-13 m)
What is the value for dz_within? This should be 0.5 m for a tall canopy (>
2m), but will be set to 0.1 m for a short canopy (< 2m)
What is the value for ncan(p)? This should be 52 (i.e., 26 m / 0.5 m)

wwieder:
Maybe one other question, were you able to run a 'normal' SP or BGC case at US-NR1 before turning on the ML canopy?

I do know that the default surface dataset for this site has a very high LAI for evergreen needleleaf trees, but I don't know what is being assigned for HTOP?

slevis-lmwg:
Thank you for these comments, Gordon and Will.

I have switched the case from BGC to SP and got past the initialization error.
The model now completes 4 time steps and triggers this error:
```
(shr_strdata_readstrm) reading file ub: /glade/work/oleson/PLUMBER2/input_files/US-NR1/LAI_stream_US-NR1_1999-2014.nc       6
 
 hist_htapes_wrapup : Writing current time sample to local history file 
 ./US-NR1_sp.clm2.h1.1999-01-01-25200.nc at nstep =            4 
  for history time interval beginning at   6.250000000000000E-002  and ending at
    8.333333333333333E-002
 
(shr_orb_params) ------ Computed Orbital Parameters ------
[...]
(shr_orb_params) -----------------------------------------
(shr_strdata_readstrm) reading file ub: /glade/work/oleson/PLUMBER2/input_files/US-NR1/LAI_stream_US-NR1_1999-2014.nc       7
 ENDRUN:
 ERROR: ERROR: Norman: total longwave conservation error
```
I like @wwieder's suggestion that I confirm next whether this runs with the multilayer canopy OFF.

gbonan:
Sam,

The error message relates to longwave radiation in the multilayer canopy.
Check that CLM runs okay without the multilayer canopy. Then we will have
to figure out how to debug this error. I have not encountered it in my own
uncoupled simulations, so I suspect something is not getting initialized
correctly.

Gordon

slevis-lmwg:
I started two "control" simulations with the PLUMBERcsv branch. They appear in /glade/derecho/scratch/slevis:
- `US-NR1_sp_default` which used ctsm5.1.dev168
- `US-NR1_sp_def_dev170` which used ctsm5.1.dev170 (same as the multilayer canopy run `US-NR1_sp`)

They both completed 1999 to 2004. They both crashed in 2005 but on different months, so I suspect that they ran out of wallclock time.

Gordon and @wwieder, please let me know if you would like me to go ahead and debug the multilayer canopy case `US-NR1_sp` or if you would rather discuss next steps.

gbonan:
I am not sure of the best way to proceed with the debugging. The Norman
radiation error typically means (when it occurs) that the leaf temperatures
in the model have attained unrealistic values. This is an indication that
something else has gone wrong, which then gives the strange temperatures. I
suspect that something is not getting passed from CLM to the multilayer
canopy correctly (or it is some value that the multilayer canopy has never
encountered before). In this respect, I have always done simulations for
the warm season (e.g., June, July, August). CLM is probably starting up on
January 1. Maybe it is something to do with snow on the ground or with lai
and sai in the winter?

At the top of subroutine MLCanopyFluxes (MLCanopyFluxesMod.F90) there are
many variables that are copied from CLM to multilayer canopy variables.
These variables need to be printed out to make sure they have reasonable
values at every time step. Also, this code should be checked, because the
simulation starts in the non-growing season:

       ! Get values for current time step from CLM

       lai(p) = elai(p)
       sai(p) = esai(p)

       ! Vertical profiles

       do ic = 1, ncan(p)
          dlai(p,ic) = dlai_frac(p,ic) * lai(p)
          dsai(p,ic) = dsai_frac(p,ic) * sai(p)
          dpai(p,ic) = dlai(p,ic) + dsai(p,ic)
       end do

We probably should discuss how to do the debugging and also an estimate of
time. Since the model is crashing after 4 timesteps, I think it should be
fairly easy to find the problem. It will require lots of write statements,
though. Also we need to see if this is how Will wants you to spend time.

gbonan:
That's a good idea, and would be a good starting point for debugging if
needed.

ekluzek:
@wwieder the one thing about a July 1st start date is that the initial condition files we save are all from Jan/1st. So if you use a Jan/1st finidat file for a July/1st start date you are starting from a really weird conditions. If you save a restart file for July 1st, that would be better here...

gbonan:
Thanks, Erik. Yes, the simulation will need a restart file for July 1
(without the multilayer canopy) and then a restart on July 1 with the
multilayer canopy turned on. Since this is a multistep process, I was
hoping the multilayer canopy would work with a Jan 1 initialization. It
still might, if we can debug why it is blowing up.

Sam - we can talk next week if you want to discuss options

slevis-lmwg:
As a very quick test, I started on July 1st without a new finidat file. The model completed one more time step (nstep = 5) and stopped with the same error.

Next I will prepare a July 1 finidat to have it ready. I will look into setting up a meeting time.

slevis-lmwg:
Sequence of tests and their results, this far, so that I don't lose track:

1) Case `ZM-Mon_test2_slevis` with MLCanopy completed one year using this compset and this finidat:
`HIST_DATM%1PT_CLM51%BGC_SICE_SOCN_SROF_SGLC_SWAV_SESP`
`/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/initdata_map/clmi.I2000Clm50BgcCrop.2011-01-01.1.9x2.5_gx1v7_gl4_simyr2000_c190312.nc`
2) Case `US-NR1_slevis` with MLCanopy fails using the same compset and finidat:
`ERROR: initVerticalStructure: ncan > nlevmlcan`
3) Case `US-NR1_sp` with MLCanopy fails using this compset:
`HIST_DATM%1PT_CLM51%SP_SICE_SOCN_SROF_SGLC_SWAV_SESP`
Cold start in Jan, cold start in July, as well as July 1 finidat give this error:
`ERROR: Norman: total longwave conservation error`
Jan 1 finidat gives this error:
`ERROR: initVerticalStructure: ncan > nlevmlcan`
4) Case `US-NR1_sp_def_dev170` runs without MLCanopy for several years (same compset, cold start in Jan).
5) I thought I'd try an 1850 instead of Hist simulation. This runs into trouble in the build phase for now because PLUMBER2 cases are designed as Hist simulations.

src/multilayer_canopy/MLCanopyFluxesMod.F90
```diff
@@ -0,0 +1,1114 @@
...
+       if (masterproc) then
+          write (iulog,*) 'Successfully initialized multilayer canopy vertical structure'
+       end if
+    end if
+
+    ! Copy CLM variables to multilayer canopy variables. Note the
+    ! distinction between grid cell (g), column (c), and patch (p)
+    ! variables. All multilayer canopy variables are for patches.
+
+    do fp = 1, num_mlcan
```
slevis-lmwg, line 294: Debugging:
Focus on this loop and next ending with line 417.

slevis-lmwg: This far the "write statements" seem to point to tleaf(p=1, ic=9), which increases out of control from 285 to 1264K by nstep = 10. The trend in tleaf(1,9)  is already appearent from nstep > 3.

Likely related... I see lw_source < 0 for ic = 9 and 10. From the tiny fracsun I infer that ic = 9 is nearest to the ground.

What I don't know how to interpret is whether the tridiagonal matrix solver has anything to do with the problem.

Gordon, it may help if you look at the output that I'm looking at. 
- Open the last lnd.log file in:
/glade/derecho/scratch/slevis/US-NR1_sp/run
- The write statements are in
``` 
/glade/work/slevis/git/multilayer_canopy/src/multilayer_canopy/src/multilayer_canopy/MLCanopyFluxesMod.F90
/glade/work/slevis/git/multilayer_canopy/src/multilayer_canopy/src/multilayer_canopy/MLLongwaveRadiationMod.F90
```
I think you can type `git diff` in /glade/work/slevis/git/multilayer_canopy to see clearly what I did.

gbonan:
Sam,

Thanks. This is very good detective work! And the model output and print
statements are easy to follow. You are correct in identifying lw_source < 0
for layers ic=9 and ic=10 on the very first time step as the problem.
lw_source is the emitted longwave radiation from the leaf layer. It has two
components: lw_source_sun for sunlit leaves and lw_source_sha for shaded
leaves. These depend on leaf temperature (tleaf) for the sunlit and shaded
leaves. These temperatures are okay (283.600006103516 K), so the component
fluxes should be good. The sunlit and shaded fluxes are then weighted by
the sunlit fraction (fracsun) and shaded fraction (1-fracsun). The sunlit
fraction is very small (10^-39), but this small value is not directly
causing the negative lw_source. The weighted fluxes are also multiplied by
(1-td).  td > 1 for layers ic=9 and ic=10, which is causing the negative
lw_source for these layers.

td is the fractional transmittance of diffuse radiation for the canopy
layer. It should be < 1. td is calculated in MLSolarRadiationMod (lines
196-203). For each leaf layer, the code loops through 9 angle classes
(angle = 5, ..., 85 degrees) to calculate an extinction coefficient (gdirj)
that is then weighted and summed over the 9 classes. Within this loop print
out:

solar_zen(p), kb(p,ic), tb(p,ic)
gdirj, phi1(p,ic), phi2(p,ic), angle, chil(p,ic)
exp(-gdirj / cos(angle) * dpai(p,ic) * clump_fac_ic(p,ic)) * sin(angle) *
cos(angle)
dpai(p,ic), clump_fac_ic(p,ic)
td(p,ic)

You only have to do this for ic=9 and ic=10 and for the first time step.
The first line of output is for the direct beam transmittance (tb). I want
to see if something strange is going on with that too.

slevis-lmwg:
Ok, thank you, Gordon. You can go back to the same place to see my code changes.

The lnd.log files (including the latest one) now appear in `/glade/derecho/scratch/slevis/archive/US-NR1_sp/logs` because I ran for 3 timesteps and the simulation completed (i.e. didn't crash) and, therefore, copied files to the archive directory.

gbonan:
Thanks, Sam. I think I see what is going on. The total plant area index
(leaves + stem) for these two layers is very small:

ic = 10, dpai = 3.569665593485061E-003
ic = 9, dpai = 1.325427162346319E-003

These low values of dpai cause td > 1. In my tower simulations, dpai is
always greater than 0.01. If that value is used, I do not get td > 1. So
let's try a simple test of printing out some canopy structure information
(so I can see the default values) and then resetting dlai and dsai > 0.01.
Let's see if this gets beyond the error.

MLCanopyFluxesMod.F90:

line 404:

! Vertical profiles

write(iulog,*)
lai(p),sai(p),mlcanopy_inst%ztop_canopy(p),mlcanopy_inst%zbot_canopy
write(iulog,*) mlcanopy_inst%ntop_canopy,mlcanopy_inst%nbot_canopy

do ic = 1, ncan(p)
   dlai(p,ic) = ...
   dsai(p,ic) = ...
   dpai(p,ic) = ...
   write (iulog,*)
ic,dlai(p,ic),dsai(p,ic),dpai(p,ic),mlcanopy_inst%dz_profile(p,ic)
    ! Now reset values to minimum
    dlai(p,ic) = max(dlai(p,ic), 0.01_r8)
    dsai(p,ic) = max(dsai(p,ic), 0.01_r8)
    dpai(p,ic) = dlai(p,ic) + dsai(p,ic)
end do

What this does is print out canopy structure:
   lai, sai, top height, bottom height
   layer index for canopy top, layer index for canopy bottom
   for each layer: layer index,lai, sai, lai+sai, layer thickness
 Then it resets lai and sai for each layer to > 0.01

slevis-lmwg:
Gordon, I commented out the error check right after the lines of code that you suggested and now this works. The simulation is in progress in the same scratch/.../run directory that I mentioned above.

gbonan:
Sam,

It looks like the code ran for 20937 timesteps (when I last checked), which
is over a year. You can stop the run. There is no need to keep it going.
Check to see if there are valid multilayer canopy fields on the history
file (GPP_ML, LWP_ML). It looks like LWP_ML has some numbers, but GPP_ML is
empty.

slevis-lmwg:
Gordon, I see the same. Shall we debug that next?

gbonan:
Yes, debug the two multilayer fields that go onto the history file (GPP_ML,
LWP_ML). The header is correct for both, but the values seem to be wrong.
GPP_ML is a single-level field, but has no data. LWP_ML is a multi-level
field (nlevmlcan) and seems to have bad data. Data is written for 73
levels, but the leaf layers are levels 9 to 24. All other layers should
have missing data.

You can set up the run for 1 month (July, 31 days) and write history files
every time step.

slevis-lmwg:
Hi Gordon,
For now I ran 16 timesteps rather than a month, and the output got moved to the /archive directory:
```
/glade/derecho/scratch/slevis/archive/US-NR1_sp/lnd/hist
/glade/derecho/scratch/slevis/archive/US-NR1_sp/logs
```
I used ncdump to make ascii copies of the first timestep from this run and of the first month from the earlier run. My initial concern is that GPP_ML is NaN from the first timestep.

You'll see output from write statements in the latest lnd.log file showing that gpp = NaN because sunlit agross = NaN for ic = 1 to 8. (Sorry I had intended but didn't write out the ic values.) So for starters I would stop adding the non-existent layers to the sum, right?

I will put this aside tomorrow and hope to get back to it on Friday.

gbonan:
Hi Sam,

Try running again printing out values for ic (so we know what layers we are
looking at). Then we should discuss how to debug the error. It seems to
suggest that agross (which comes from LeafPhotosynthesis) is not calculated
correctly. It might be better for me to debug because I know the code. I
talked with Erik and Adrianna and they both suggested running on izumi
using a login node. They said the code will run without queuing, so jobs
can turn around quickly (i.e., add write statements, run the model, look at
output, add more write statements, ...). We can talk about this to see if
it is the best path forward.

slevis-lmwg:
Ok, Gordon, look for the new lnd.log file in the same /archive directory and let me know what you'd like to do next.

About izumi, I did not know that you could run outside the queues (and would be interested in seeing how), but even within the izumi queues there's rarely any wait-time.

gbonan:
This is very helpful, Sam. I think I see the problem (or at least one
problem). In our previous quick fix to prevent small values of dpai, we
messed up the indexing of what canopy layers have leaves and stems.

nbot = Bottom canopy layer with leaves and stems. My notes show nbot = 9
ntop = Top canopy layer with leaves and stems. My notes show ntop = 24

Our quick fix was to set dlai and dsai to a minimum of 0.01 for all layers
(1 to ncan) regardless of if they had leaves or stems. So there is a
mismatch now between the top and bottom of the canopy (ntop, nbot) and
layers with leaves and stems (dpai > 0). I think this is what is causing
NaN for agross.

So let's try this fix in CanopyFluxesMod.F90 (lines 411-413):

if (dlai(p,ic) > 0._r8) dlai(p,ic) = max(dlai(p,ic), 0.01_r8)
if (dsai(p,ic) > 0._r8) dsai(p,ic) = max(dsai(p,ic), 0.01_r8)
dpai(p,ic) = dlai(p,ic) + dsai(p,ic)

This way, dlai and dsai are only reset in layers that have leaves and stems.

ekluzek:
@gbonan and @slevis-lmwg FYI on running on the login node. You do this using:

`./case.submit --no-batch`

I remember having some weird subtleties the last time I did it. So if you see that -- let me know and I
can likely help. I don't remember well enough off hand what they were. So just try that and see if it
works.

As @slevis-lmwg says sending to the queue on Izumi is rarely a problem. While it often is on Derecho.

src/multilayer_canopy/MLCanopyFluxesMod.F90
```diff
@@ -406,10 +406,9 @@ subroutine MLCanopyFluxes (bounds, num_exposedvegp, filter_exposedvegp, &
        do ic = 1, ncan(p)
           dlai(p,ic) = dlai_frac(p,ic) * lai(p)
           dsai(p,ic) = dsai_frac(p,ic) * sai(p)
-          dpai(p,ic) = dlai(p,ic) + dsai(p,ic)
           ! Now reset values to minimum
-          dlai(p,ic) = max(dlai(p,ic), 0.01_r8)
-          dsai(p,ic) = max(dsai(p,ic), 0.01_r8)
+          if (dlai(p,ic) > 0._r8) dlai(p,ic) = max(dlai(p,ic), 0.01_r8)
+          if (dsai(p,ic) > 0._r8) dsai(p,ic) = max(dsai(p,ic), 0.01_r8)
           dpai(p,ic) = dlai(p,ic) + dsai(p,ic)
```
slevis-lmwg, line 412: Gordon, these are the lines that you recommended changing (410 - 411).

src/multilayer_canopy/MLCanopyNitrogenProfileMod.F90
```diff
@@ -186,12 +186,13 @@ subroutine CanopyNitrogenProfile (num_filter, filter, mlcanopy_inst)
...
+       ! NB slevis 2024/3/22: The next was a valid error check before introducing minimum dlai and dsai
 
        numerical = sum(vcmax25_profile(p,1:ncan(p)) * dpai(p,1:ncan(p)))
        analytical = vcmax25top * (1._r8 - exp(-kn*(lai(p) + sai(p)))) / kn
-       if (abs(numerical-analytical) > 1.e-06_r8) then
-          call endrun (msg='ERROR: CanopyNitrogenProfile: canopy integration error')
-       end if
+!      if (abs(numerical-analytical) > 1.e-06_r8) then
+!         call endrun (msg='ERROR: CanopyNitrogenProfile: canopy integration error')
+!      end if
```
slevis-lmwg, line 195: After changing lines 410-411 in MLCanopyFluxesMod.F90, this error got triggered, so I commented it out, because I do not think that we should expect it to pass anymore.

src/multilayer_canopy/MLCanopyFluxesMod.F90
```diff
@@ -407,12 +407,18 @@ subroutine MLCanopyFluxes (bounds, num_exposedvegp, filter_exposedvegp, &
...
 
        totpai = sum(dpai(p,1:ncan(p)))
-       if (abs(totpai - (lai(p)+sai(p))) > 1.e-06_r8) then
-          call endrun (msg=' ERROR: MLCanopyFluxes: plant area index not updated correctly')
-       end if
+       ! NB slevis 2024/3/19: The next was a valid error check before
+       ! introducing minimum dlai and dsai a few lines up from here
+!      if (abs(totpai - (lai(p)+sai(p))) > 1.e-06_r8) then
+!         call endrun (msg=' ERROR: MLCanopyFluxes: plant area index not updated correctly')
+!      end if
```
slevis-lmwg, line 420: I don't think I mentioned before, but this error was triggered by the previous change to lines 410-411. Again, I commented it out because I do not think that we should expect it to pass anymore.

src/multilayer_canopy/MLCanopyFluxesMod.F90
```diff
@@ -362,19 +362,19 @@ subroutine MLCanopyFluxes (bounds, num_exposedvegp, filter_exposedvegp, &
        p = filter_mlcan(fp)
        c = patch%column(p)
        g = patch%gridcell(p)
-       lat = grc%latdeg(g) * pi / 180._r8
-       lon = grc%londeg(g) * pi / 180._r8
+       lat = grc%lat(g)
+       lon = grc%lon(g)
```
slevis-lmwg, line 366: I did not bring up this one before, either. It's a simplification that does not change answers.

slevis-lmwg:
Gordon, I think that results look better now, but I will let you tell me what you think.

I used the same write statements in the latest lnd.log file, and I made new ncdumps in the corresponding /lnd/hist directory. Both are in the same /archive area as before.

gbonan:
Yea!!! Well done, Sam. It looks like the model is running correctly, based
on the history files. The output for GPP_ML and LWP_ML look good. As a
recap, the problem was thin canopy layers with very small dlai and/or dpai.
We put in a quick fix, but that caused additional problems. The fix we have
now works, but requires some warnings to be turned off because the sum of
dlai+dsai does not equal the expected total lai+sai. So I will need to
investigate the correct fix.

The best way to do that will be for me to run the code so that I can
investigate alternatives. It does sound like running on izumi is a good
option. So, Sam, maybe you could get the case moved over to izumi and try
it out. Then you can give be a tutorial on running the code.

Thanks for your help with this, and have a good weekend.

slevis-lmwg:
Ok, I have the same case running on izumi now.

gbonan:
I have a login on izumi. We can schedule a meeting at your convenience so I
can learn how to run the case.

slevis-lmwg:
@gbonan and I met today and completed a step-by-step tutorial on izumi on cloning the ctsm, checking out this PR's branch, creating a new case, and building and running the case.

Gordon, you should receive a github notification that I gave you collaborator permissions so that you may push commits to this PR. When you're ready we will go over instructions for that.

slevis-lmwg:
Though, best if you act on the "collaborator" invitation by the end of the week, because it has an expiration date, I'm pretty sure...

gbonan:
Sam,

I accepted the github invitation. Also, I have been running the code today
on izumi. All is going well. Thanks for the tutorial.

Gordon

ekluzek:
@slevis-lmwg is the plan to have this preliminary version as part of ctsm6.0.0? Or is it going to come in after that?

gbonan:
This branch is for ongoing model development. It is not part of any release.

Gordon

slevis-lmwg:
Testing the latest update of this branch with this sequence of cases (note that the multilayer case currently fails with error listed below):
```
./create_newcase --case /glade/u/home/slevis/cases_plumber2/US-NR1_ctsm54021_ml_HISTsp_def --res CLM_USRDAT --compset HIST_DATM%1PT_CLM60%SP_SICE_SOCN_SROF_SGLC_SWAV_SESP --run-unsupported --user-mods-dir clm/PLUMBER2/US-NR1
./create_newcase --case /glade/u/home/slevis/cases_plumber2/US-NR1_ctsm54021_ml_HISTsp --res CLM_USRDAT --compset HIST_DATM%1PT_CLM60%SP_SICE_SOCN_SROF_SGLC_SWAV_SESP --run-unsupported --user-mods-dir clm/PLUMBER2/US-NR1
```
Ran the first (default) case for 6 months to generate finidat for the second (multilayer) case.
Diffs between the second case and the first:
`US-NR1_ctsm54021_ml_HISTsp/user_nl_datm_streams US-NR1_ctsm54021_ml_HISTsp_def/user_nl_datm_streams`
```diff 
34,38d33
< !TODO, add $PLUMBER2SITE to this naming string + accurate years after moving files to inputdata
< CLM_USRDAT.PLUMBER2:datafiles = /glade/work/oleson/PLUMBER2/datm_files/US-NR1/CLM1PT_data/CTSM_DATM_US-NR1_1999-2014.nc
< CLM_USRDAT.PLUMBER2:taxmode=cycle
< CLM_USRDAT.PLUMBER2:dtlimit=30
<
45a41,42
> CLM_USRDAT.PLUMBER2:taxmode=extend
> CLM_USRDAT.PLUMBER2:dtlimit=1e30
```
`US-NR1_ctsm54021_ml_HISTsp/user_nl_clm US-NR1_ctsm54021_ml_HISTsp_def/user_nl_clm`
```diff 
20d19
< finidat = '/glade/derecho/scratch/slevis/archive/US-NR1_ctsm54021_ml_HISTsp_def/rest/1999-07-01-25200/US-NR1_ctsm54021_ml_HISTsp_def.clm2.r.1999-07-01-25200.nc'
```
`US-NR1_ctsm54021_ml_HISTsp/env_run.xml US-NR1_ctsm54021_ml_HISTsp_def/env_run.xml`
```diff 
267c267
<     <entry id="RUN_STARTDATE" value="1999-07-01">
---
>     <entry id="RUN_STARTDATE" value="1999-01-01">
279c279
<     <entry id="STOP_OPTION" value="nsteps">
---
>     <entry id="STOP_OPTION" value="nmonths">
286c286
<     <entry id="STOP_N" value="16">
---
>     <entry id="STOP_N" value="6">
```

Error in the second case:
```
 Successfully initialized multilayer canopy vertical structure
  
 nstep =            1
 htop, zref, ncan
   12.0000000000000        36.3593409135170               73
 lai, sai, nbot, ntop
   2.02601981163025       0.233227646437726               12          24
  
 ENDRUN:
 ERROR: CanopyNitrogenProfile: canopy integration error
```

src/multilayer_canopy/MLCanopyFluxesMod.F90
slevis-lmwg: Resolving this post as outdated.

src/multilayer_canopy/MLCanopyNitrogenProfileMod.F90
slevis-lmwg: Resolving this post as outdated.

slevis-lmwg:
> @gbonan and @slevis-lmwg FYI on running on the login node. You do this using:
> 
> `./case.submit --no-batch`
> 
> I remember having some weird subtleties the last time I did it. So if you see that -- let me know and I can likely help. I don't remember well enough off hand what they were. So just try that and see if it works.
> 
> As @slevis-lmwg says sending to the queue on Izumi is rarely a problem. While it often is on Derecho.

@gbonan I scrolled through the PR in case I found something useful, and I found this post by Erik. It suggests running the model on the login node if you find that the batch system becomes frustrating.

src/multilayer_canopy/MLCanopyFluxesMod.F90
slevis-lmwg: I will resolve as outdated.

PR #2007, opened 2023-05-17
black reformat python files for consistancy across cesm

jedwards4b:
### Description of changes
Black reformats buildlib and buildnml 
### Specific notes

Testing performed, if any:

### DIFF
--- a/cime_config/SystemTests/fsurdatmodifyctsm.py
+++ b/cime_config/SystemTests/fsurdatmodifyctsm.py
@@ -12,33 +12,33 @@ from CIME.SystemTests.test_utils.user_nl_utils import append_to_user_nl_files
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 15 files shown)


ekluzek:
I support using black for our python code. There's probably other code that we should run through black as well. We do run the python code under the python directory through black, but other python codes should as well.

ekluzek:
@jedwards4b what version of black did you use for this PR?

jedwards4b:
black --version
black, 22.3.0

ekluzek:
We talked about their briefly yesterday. We are supportive of this. The version of black we are using right now is the same as used here.

I do think we should regularly test buildlib and buildnml through black. That can be added in our python directory testing. Those tests don't run it through black, just test and it and show the user if it fails. We also have a black action and it could be extended to also check these two files.

I'll bring this in with some other simple bit-for-bit type changes and we'll schedule it to come in...

PR #2008, opened 2023-05-18
Get the Makefile working for CTSM5.2 surface dataset creation

### LINKED ISSUE #9: mksurfdata_map raw data grids should just be listed in hgrid, not res
**Bill Sacks < sacks@ucar.edu > - 2013-04-12 06:02:12 -0600**
**Bugzilla Id:** [1661](http://bugz.cgd.ucar.edu/show_bug.cgi?id=1661)
**Bugzilla CC:**  rfisher@ucar.edu, 

Currently, it looks like grids for mksurfdata_map raw data files are listed in two places in namelist_definition_clm4_5.xml: the valid values for 'res' and the valid values for 'hgrid'. The fact that they're listed in the valid values for 'res' can be an annoyance when making mapping files with mkmapdata/regridbatch, because this tool then tries to make mapping files TO these raw data resolutions (which is unnecessary). 

This annoyance is becoming magnified with my addition of a 1km raw data file, for which mapping files take a long time to generate. 

I think these raw data grids should just be listed in hgrid, not res. But I haven't looked closely at this, so I'm not positive about it. I think this will require changing some things in queryDefaultNamelist.

### LINKED ISSUE #1869: Bring in Makefile for ctsm5.2 branch to build all datasets with multiple batch submissions at the same time
I'm going to bring back the Makefile mechanism we have used to the mksurfdata_esmf ctsm5.2 branch so that the Makefile can send off multiple batch submission to create datasets at the same time. This also allows us to use different number of processors for different tests. This also allows us to integrate with #1674. This still relies on the new python batch scripts submission in mksurfdata_esmf, it just adds a thin layer on top of it that actually runs each of the batch script generation and then submits.

One change I'll need to make to get this to work is that right now the python batch submission creates a batch submission filename thats the same name no matter what scenario is chosen. This will need to change so that the filename depends on the scenario. That way the can all run at the same time and be submitted together without having to lock files for each.

### LINKED ISSUE #2073: Change names on ctsm5.2 datasets to include context for "_78_" to "_78pfts"
Names for landuse.timeseries on the ctsm5.2 branch currently should change.

We want to go to this format:

lnd/clm2/surfdata_esmf/ctsm5.2.0/landuse.timeseries_0.9x1.25_hist_1850-2015_78pfts_c230601.nc
lnd/clm2/surfdata_esmf/ctsm5.2.0/fsurdat_0.9x1.25_hist_1850_78pfts_c230601.nc
lnd/clm2/surfdata_esmf/ctsm5.2.0/fsurdat_0.9x1.25_hist_2000_78pfts_c230601.nc
lnd/clm2/surfdata_esmf/ctsm5.2.0/fsurdat_0.9x1.25_hist_2015_78pfts_c230601.nc
lnd/clm2/surfdata_esmf/ctsm5.2.0/landuse.timeseries_0.9x1.25_SSP2-4.5_1850-2100_78pfts_c230601.nc
lnd/clm2/surfdata_esmf/ctsm5.2.0/fsurdat_0.9x1.25_SSP2-4.5_2100_78pfts_c230601.nc

The _78_ part should become _78pfts_

### LINKED ISSUE #2178: Don't create/test most SSP landuse timeseries datasets for CTSM5.2
Since the future scenarios will be replaced with CMIP7 it doesn't seem important to make all of the CMIP6 scenarios with CTSM5.2. However, what is important is to make sure the process works, and to make datasets if needed for use by a group. The SSP scenarios are also important to have in order to have data to run transient simulations that go up to present-day, rather than being restricted to 2015, because of out historical landuse data currently ends there.

@lawrencepj1 and I talked about this a bit yesterday, and he also recommended that SSP2-4.5 be used as the one SSP scenario that we create datasets for, because it's a nice "middle of the road" scenario.

We currently test 4 of the SSP's at f10 resolution and one grid for CAM ne0ARCTICGRISne30x8_ne0ARCTICGRISne30x8_mt12, for SSP5-8.5. So for CTSM5.2 I suggest we only make SSP2-4.5 datasets and only test a couple resolutions for these. Although we'd also use the SSP2-4.5 landuse.timeseries for some resolutions so we don't make both a SSP scenario and a historical, with the SSP scenario covering 1850-2100.

@slevis-lmwg any thoughts on this?

### LINKED ISSUE #2179: Remove the list of valid resolutions in namelist_definition
For mksurfdata_map and mksurfdata we needed a list of valid resolutions. The need for this goes way back before config_grids and query_config --grids was a thing.

It also has been problematic in the recent past because developers had to edit namelist_definition as well as ccs_config to add new grids. Removing this makes it easier for developers. This has also been a problem in #2008 because the list is so long. The other use of this list is in the build-namelist tester, but right now what it does is remove a long list of resolutions, when the list of valid resolutions to use would be shorter. So good for past and current time.

This also resolves #9 As we should remove both hgrid and res from namelist_definition.

### LINKED ISSUE #2180: Remove scripts no longer needed with mksurfdata_esmf
Some scripts that were useful a long time ago before check_input_data, query_config and other's were created are:

listDefaultNamelist.pl
queryDefaultNamelist.pl  queryDefaultXML.pm 
checkmapfiles.ncl
createMapEntry.pl
createMkSrfEntry.py
getco2_historical.ncl

These all were required with mksurfdata_map, but with mksurfdata_esmf no longer are so should be removed.

Definition of done:
- [x] Remove all the files
- [x] Update README files for the change
- [x] Update any other documentation
- [x] Make sure both the code and tools work

### LINKED ISSUE #2189: Need to be able to change number of tasks on node/wallclock for gen_mksurfdata_jobscript_* scripts for mksurfdata_esmf
For some resolutions to be created with mksurfdata_esmf both the wallclock time and the number of tasks per node needs to change (they currently have fixed defaults). This appears to be the case especially for low and higher resolutions. 12 tasks/node and an hour of wallclock seems to work fine for the bulk of the mid-range resolutions. The number of nodes needed is already an input and it of course is the big knob for different sized grids. For some resolutions getting this right is important to getting the dataset to be created.

By storing the values in the Makefile (in #2008) we can keep the build of all model grids going as we move forward.

### LINKED ISSUE #2191: Move mksurfdata_esmf CTSM5.2 python support scripts gen_mksurfdata_*.py to python directory to get python development support
The gen_mksurfdata_*.py support scripts for mksurfdata_esmf in ctsm5.2 are all separate scripts under the tools/mksurfdata_esmf directory. As with other python scripts in order to use other python development tools, I'll be moving the core to the python subdirectory. Then for development we automatically get python linting as well as black formatting, and it facilitates adding unit testing for the scripts. This will improve the current development as well as future development. Since, these scripts are core to the support of surface dataset creation it's critical to have support for them.

Also as part of that the gen_mksurfdata_multi and gen_mksurfdata_single scripts have code duplication that makes them hard to support and requires separate testing for them. As part of this I'll reduce that duplication that's hard to support by having the multi script be based on top of the single script with additional functionality to support different standard resolution/scenarios to be built.

The other thing to fix is that there is a subprocess.run from the jobscript scripts to the namelist one. I've been having similar problems with it as we have in other contexts. By moving the guts of the python scripts to the python directory, I'll be able to avoid using a subprocess.run and fix that issue as well. Strangely the subprocess.run can sometimes work fine, and then unrelated changes seem to make it break. But in any case this will make the use of the scripts more robust.

I thought about continuing to fight with the scripts as they are, but this will improve current and future development. And by focussing on the minimum changes to make do this it will improve the current development.

@slevis-lmwg and I discussed this earlier today.

ekluzek:
### Description of changes

Get the Makefile working for CTSM5.2 surface dataset creation so all files can be created at the same time.
Also update to ctsm5.1.dev125

### Specific notes

   Fixes #1869
   Fixes #2073 
   Fixes #2178 
   Fixes #2179
   Fixes #2180
   Fixes #2189
   Fixes #2191
   Fixes #9
   Some work on #2149

  Error checking added to gen_mksurfdata_jobscript_multi.py as well as --walltime option

Testing performed, if any: So far have ran a few cases, will run the complete set before merging


ekluzek:
The PET files can interfere with each other so I need to set defaultDefaultLogFilename in ESMF_Initialize from the default of ESMF_LogFile to something that includes the resolution. This also requires moving the namelist read to earlier just before the ESMF_Initialize, but will also allow to change the setting of create_esmf_pet_files. The PET files are normally NOT written out, unless there is an error, but in that case more than one resolution erroring out would conflict with each other without different filenames.

ekluzek:
This also includes the merge to ctsm5.1.dev125, and we should bring that merge in before this, so that the changes for this PR are clearer.

ekluzek:
The subset data makefile is working for everything, but the urban datasets. This is because of the change of nlevurb from 5 to 10. We need to decide how this should be handled.

ekluzek:
@olyson is going to find nlevurb==10 versions of the datasets, and we'll use the settings from them.

olyson:
This was the Mexico City dataset with nlevurb=10 that was used in Oleson and Feddema (2019):

/glade/p/cgd/tss/people/oleson/modify_surfdata/surfdata_1x1_mexicocityMEX_simyr2000_c170927.nc

olyson:
I realized you may need Vancouver also:

/glade/p/cgd/tss/people/oleson/modify_surfdata/surfdata_1x1_vancouverCAN_simyr2000_c170927.nc

Anything else?

ekluzek:
There's the urban-C alpha site as well. As long as it's still useful.

olyson:
The paper used this dataset for urban-C:

/glade/p/cgd/tss/people/oleson/modify_surfdata/surfdata_1x1_urbanc_alpha_simyr2000_c170927.nc

This has nlevurb=4 as originally specified for a model intercomparison exercise.
I would prefer to leave it as nlevurb=4 since that is what is documented in the paper.
The new mksurfdata_esmf code should be able to handle that automatically as nlevurb is now read in from the surface dataset instead of being hard-coded in clm_varpar.F90.  I assume the latest branch version has this code in place....

ekluzek:
OK, I created all the single point datasets including urban ones, with the urbanc_alpha for nlevurb==4. In testing all of them the non-urban ones all pass, but the urban ones have trouble. It looks like I need to do a COLDSTART as part of this, as one test failed in the init_interp part. I assume we'll need to have initial conditions with the right number of nlevurb for it to work?

olyson:
The Mexico City and Vancouver sites are normally run with cold start since the initialization is set in the code.  I think it is fine to have the urbanc site as cold start as well.

ekluzek:
I've been able to cheat and run some urban single point tests on login nodes of cheyenne and I get issues with both Vancouver (with a coldstart), and the urbanc_alpha site. The urbanc_alpha site looks like it has problems in the datm streams for it's setup. This is a site we don't test in aux_clm, so it's not too surprising that an error cropped up. Fulfilling the maxim, "If you don't continue to test it, it will break". So I think this is in the baseline version as well, but I haven't proved it yet, but will.

For Vancouver I see the following crash in the cesm.log...

```
 /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/share/src/shr_file_mod.F90         912 This routine is depricated - use shr_log_setLogUnit instead         -12
 /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/share/src/shr_file_mod.F90         912 This routine is depricated - use shr_log_setLogUnit instead         -13

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7f1ad023631f in ???
	at /usr/src/debug/glibc-2.22/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
#1  0xc43633 in soilthermprop
	at /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/src/biogeophys/SoilTemperatureMod.F90:721
#2  0xc77209 in __soiltemperaturemod_MOD_soiltemperature
	at /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/src/biogeophys/SoilTemperatureMod.F90:285
#3  0x5bb64d in __clm_driver_MOD_clm_drv
	at /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/src/main/clm_driver.F90:1231
#4  0x575ab4 in modeladvance
	at /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/src/cpl/nuopc/lnd_comp_nuopc.F90:904
#5  0x7f1ad43630d9 in _ZNK5ESMCI13MethodElement7executeEPvPi

```

The line in the code it's pointing to is:

``` fortran
               if (j >= snl(c)+1 .AND. j <= nlevurb-1) then
                  tk(c,j) = thk(c,j)*thk(c,j+1)*(z(c,j+1)-z(c,j)) &
                       /(thk(c,j)*(z(c,j+1)-zi(c,j))+thk(c,j+1)*(zi(c,j)-z(c,j)))
```

I can probably check values and see what's failing here. Something is probably not set correctly at this point, either due to the cold-start or the surface dataset....

olyson:
@ekluzek , did you still need help with these tests?

ekluzek:
OK, so running this in the debugger, I see that for c==3, and j==7, the divisor ends up being zero, so
we have a divide by zero.

This is because the first term in the denominator is negative
`thk(c,j)*(z(c,j+1)-zi(c,j))`

and the second term is positive for the same value as above.
`thk(c,j+1)*(zi(c,j)-z(c,j)))`

olyson:
Ok, I realized that what's probably going on here is that the Vancouver and Mexico City sites have a "soil" layer structure that is hardcoded in initVerticalMod.F90 as five layers specifically for those two sites.  Looking back to the code I used in conjunction with the nlevurb=10 datasets, I removed this hardcoding.  So I propose we remove this hardcoding which is triggered by "use_vancouver" and "use_mexicocity".
Based on your comments, it looks like urbanc_alpha has a problem with datm files.  Happy to look at that if you point me to your setup.

ekluzek:
@olyson, I compared the new fsurdat file with the old one referenced above for the urban variables and they look identical to me. I tried running with the old file, but that doesn't work since there are a lot of fields NOT on the old file.

If I'm understanding this right all of the fields in the equation that fails come from either TK_* or THICK_* values from the fsurdat file. So if this worked before, likely the code changed slightly from when these 10 level were files were being used. So we might be able to compare the code from then to now to figure it out.

Anyway, I think it's time for me to sit down with you and see if we can make progress on this. Thanks for the offer, hopefully we could do that tomorrow, or schedule for another time...

ekluzek:
@olyson our messages crossed. Thanks for that insight! I'll try that, and get back to you!

ekluzek:
@olyson it looks like that's it! Thanks for the tip!

olyson:
Great!  Apologies, that should have occurred to me sooner.

python/ctsm/test/test_sys_mesh_modifier.py
```diff
@@ -96,8 +96,8 @@ def setUp(self):
         self._lon_dimname = fsurdat_in_data[self._lat_varname].dims[1]
 
         ncap2_cmd = (
-            "ncap2 -A -v -s 'mod_lnd_props=PFTDATA_MASK' "
-            + "-A -v -s 'landmask=PFTDATA_MASK' "
+            "ncap2 -A -v -s 'mod_lnd_props=LANDFRAC_PFT' "
```
ekluzek, line 99: @slevis-lmwg I had to change this because we decided to no longer put PFTDATA_MASK on the surface datasets in ctsm5.2. This works this way, but I could either use LANDFRAC_PFT or LANDFRAC_mksurfdata do you see any reason to prefer one field over the other?

slevis-lmwg: I feel more familiar with LANDFRAC_PFT, so I'm leaning in favor of that.

ekluzek:
@slevis-lmwg whenever you get a chance if you could do this...

`chmod g+w /glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_esmf/ctsm5.2.0/`

that would allow me to add the files to inputdata in that directory. Thanks in advance...

slevis-lmwg:
> @slevis-lmwg whenever you get a chance if you could do this...
> 
> `chmod g+w /glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_esmf/ctsm5.2.0/`
> 
> that would allow me to add the files to inputdata in that directory. Thanks in advance...

Done!

python/ctsm/test/test_sys_mesh_modifier.py
ekluzek: @slevis-lmwg so this isn't working for me now. So I'd like to have you and I look at it together. I also realized that this test doesn't effectively do anything because it's not changing the mask. I think we should have a test that does change the mask and make sure it's correct.

ekluzek: @slevis-lmwg and I went over this yesterday and worked on it together. It helped that we worked together, but we didn't find a solution. But, there were several things that we noticed looking at it together, and I'll work on it more, and then get back with Sam if I'm stuck.

ekluzek:
Above I thought I had to do COLD starts for these urban cases. 

https://github.com/ESCOMP/CTSM/pull/2008#issuecomment-1653015256

It turns out that must have been something else, because I am able to get the cases working now without a COLD start using the default global finidat file. So that's a win.

python/ctsm/test/test_sys_mesh_modifier.py
ekluzek: I got this working now, by both converting to integer earlier in the process. And finding that the error we saw was in a different place in the code. I'm close on that one now.

ekluzek: Working with the latest commit. But, we should still add a test that changes the mesh and verifies the change is correct.

slevis-lmwg:
Erik, I can check when I arrive in Colorado:  I wonder whether we have such
test(s) in the unit tests.

python/ctsm/test/test_sys_mesh_modifier.py
slevis-lmwg: > Erik, I can check when I arrive in Colorado:  I wonder whether we have such
test(s) in the unit tests.

From what I can tell, we do not...

ekluzek:
@slevis-lmwg I'm going to bring this one in as is. And then make a new PR for updating to Derecho ctsxm5.1dev159 to get started on the processing of running on Derecho. I think there will be external updates required to get the build to work on Derecho, so that may become a different PR, but we'll see.

Issue #1906, opened 2022-11-16
Remove the non-working options in run_neon


ekluzek:
There are a couple of non working options in the run_neon script that we should remove until they can be implemented and tested to work.

--start-date
--end-date
--run-type sasu (until SASU comes to CTSM main)

WRW is also adding: 
- [x] create a unit test for run_neon.py to the definition of done for this item, as such I'll close #1441, which is now duplicated.
- [ ] Remove the above argument options
- [x] Remove diff_month

ekluzek:
We should also make sure the --rerun, --run-from-postad,  and --no-batch options are functional and work correctly. Also does --end-date and --run-length provided together work? It probably should abort with an error if both are provided.

@TeaganKing and @wwieder and I talked about this some yesterday.

ekluzek:
Also diff_month can be removed as it's not used. And the end-date option can be removed as well.

ekluzek:
@TeaganKing took care of most of this in her PR. The one thing left to do is to remove SASU for run_type.

TeaganKing:
SASU actually has been removed already as a valid run type in `arg_parse`. See [here](https://github.com/ESCOMP/CTSM/blob/master/python/ctsm/site_and_regional/neon_arg_parse.py#L124). I left the option commented out so that we can easily add it back in once SASU comes into main (although I am unaware of the details/timing on this, so let me know if there's a better alternative).

So, I think we can close this issue, as long as #1926 addresses all of the other concerns raised in this issue?

ekluzek:
Awesome. Thanks @TeaganKing. Leaving it commented out is fine, since it will come back with CN-Matrix. That will still be a while though, likely after LMWG.

Yes, I will close this now! Which feels good to do!

PR #2015, opened 2023-06-02
Address NEON Bugs

### LINKED ISSUE #2013: Incorrect NEON PFT at TOOL
Toolik lake is an Arctic tundra site and should be run as an arctic shrub or arctic grass. 
 https://github.com/ESCOMP/CTSM/blame/394f9fd473731cc7d87e268a385cc45fd3f3daf6/tools/site_and_regional/neon_sites_dompft.csv#L43

Thanks to members of the NEON workshop who noticed this error!

### LINKED ISSUE #2014: NEON usermods_dirs not working as intended
Users reported failures for runs at the TEAK site, which seemed to be occurring because conditionals in the shell commands for sites with non-standard datm records are not being used as expected. 

Thanks to members of the NEON workshop who helped identify this error at for TEAK site!

### LINKED ISSUE #2029: fire_method is used before it's set and hence fire_res isn't set to none when fire_method="nofire"
### Brief summary of bug

light_res doesn't get properly set to none, when fire_method is set to nofire. This is because the variable fire_method is used before being set.

### General bug information

**CTSM version you are using:** ctsm5.1.dev128

**Does this bug cause significantly incorrect results in the model's science?** No, this adds unneeded datasets that won
t be used

**Configurations affected:** BGC with fire_method="nofire"

### LINKED ISSUE #2030: Logic in build-namelist not functioning correctly for FATES when light_res is explicitly set
### Brief summary of bug

If FATES is on, and light_res is explicitly set the logic for testing if the configuration is valid doesn't function correctly.
So if you set light_res to none, and fates_spitfire_mode is 0 (fire off) it doesn't properly identify that as valid, nor things
like light_res=360x720 and fates_spitfire_mode=2 which should be valid as well.

### General bug information

**CTSM version you are using:** ctsm5.1.dev128

**Configurations affected:** 

FATES on, and light_res set to something besides "default".

wwieder:
### Description of changes
NEON bug fixes identified at the NCAR-NEON workshop

### Specific notes
Corrects the dominant PFT at TOOL site & usermods_dirs

*CTSM Issues Fixed (include github issue #):*
 Fixes #2013 
 Fixes #2014 
 Fixes #2029 
 Fixes #2030

*Are answers expected to change (and if so in what way)?*
Results will change for single point TOOL simulations

*Any User Interface Changes (namelist or namelist defaults changes)?*
New surface dataset needs to be stored on input data).


wwieder:
I can't remember how we handle surface datasets for NEON cases?  I don't think we push these to NEON, which means they have to go onto input data (/glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/NEON/).  @ekluzek can you help with this?

Are other code changes needed to make sure we're pulling the latest surface datasets for NEON cases?  

Updated surface data for this PR are here `/glade/scratch/wwieder/single_point_neon_updated`.  Only TOOL has been updated, but surface data for sites were run for time-stamp consistency.  

I haven't tested a site that automatically grabs these data, but otherwise I think this PR is ready to go and can be merged with #2021 .  Doing this relatively soon would be helpful, as updates to the NEON tutorial depend on these usermod changes. I'm assuming that @danicalombardozzi and @briandobbins would like to use tutorial updates and code changes for the cloud instance for the Flux course workshop that will be held June 23.

Apologies, for the short turn around on this, which is compounded by the CESM workshop next week.  Hopefully changes are minor enough to not be a huge burden.

wwieder:
I don't think I'm jumping the line by moving this to the top of the upcoming tags list, but hopefully this is a relatively simple PR that can come in quickly?

TeaganKing (approved):
Hi @wwieder , thanks for making these updates. I left a minor comment, but this looks good to me. I don't know much about the SP simulations details, so perhaps someone else should check that bit.

cime_config/usermods_dirs/NEON/FATES/MLBS/shell_commands
```diff
@@ -2,9 +2,13 @@
 ./xmlchange NEONSITE=MLBS
 ./xmlchange PTS_LON=279.47575
 ./xmlchange PTS_LAT=37.37783
-./xmlchange DATM_YR_END=2019
-# Different default number of months to run for transient case
-if  [[ $compset =~ ^HIST ]]; then
-  ./xmlchange DATM_YR_END=2020
-  ./xmlchange STOP_N=24
+# NEON precipitation data for this site is bad so end early
```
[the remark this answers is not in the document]
ekluzek: @TeaganKing that's a nice insight. "Bad" is a value judgement which is maybe not what we want here. We should make this as accurate as we can.

It was my understanding that the end data for these sites exists, but has quality issues so you don't want to use it. But, maybe it's not even available because of those issues, so you have to truncate it? What's the most accurate way to describe the issues here?

ekluzek (changes requested):
I'm suggesting a couple changes. The changes to comments should go in right away. Other things could wait, but some of what I suggest is easy enough that I think it would be OK to do if I can do it quickly which I think I can. But, I can also back it out if I can't get it working quickly enough. I'll create a couple issues on some of this for discussion.

cime_config/usermods_dirs/NEON/defaults/shell_commands
```diff
@@ -23,5 +24,25 @@ else
   ./xmlchange CLM_NML_USE_CASE="2018_control"
 fi
 
+# If needed for SP simulations: turn off lightning & set history file variables
+if  [[ $compset == *'SP_SICE'* ]]; then
+  ./xmlchange STOP_OPTION=nyears
+  ./xmlchange CLM_FORCE_COLDSTART=on
+  # comment out lightning file
+  :> user_nl_clm.tmp
```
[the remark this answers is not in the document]
adrifoster: I agree I think moving the lightning stuff to the build-namelist would be helpful. Maybe not in the short-term but is a good long-term goal.

cime_config/usermods_dirs/NEON/defaults/shell_commands
```diff
@@ -23,5 +24,25 @@ else
...
+  while read line; do
+    if [[ $line != *"_lightng"* ]]; then
+        echo $line>>user_nl_clm.tmp
+    else
+  echo '!'$line>>user_nl_clm.tmp
+    fi
+  done<user_nl_clm
+  mv user_nl_clm.tmp user_nl_clm
+
+  echo "hist_mfilt(2) = 48" >> user_nl_clm
```
ekluzek, line 42: @wwieder these two strictly aren't required because it's already set this way. Did you want this set here in case it's desired to change it for SP cases? I just wanted to understand the reasoning.

The difference in hist_fincl2 is required, as the list of history fields does need to be different between SP and BGC (and FATES for that matter).

cime_config/usermods_dirs/NEON/FATES/MLBS/shell_commands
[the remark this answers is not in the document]
wwieder: I've replaced 'bad' with 'missing', which is more accurate

cime_config/usermods_dirs/NEON/defaults/shell_commands
```diff
@@ -23,5 +24,25 @@ else
   ./xmlchange CLM_NML_USE_CASE="2018_control"
 fi
 
+# If needed for SP simulations: turn off lightning & set history file variables
+if  [[ $compset == *'SP_SICE'* ]]; then
```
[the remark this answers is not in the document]
wwieder: It seems to me like this is tied in with the larger lightning discussion below, which seems like a longer term goal.  Maybe we don't fix this right now?

cime_config/usermods_dirs/NEON/defaults/shell_commands
wwieder: Good point, I've removed this

ekluzek (approved):

cime_config/usermods_dirs/NEON/defaults/shell_commands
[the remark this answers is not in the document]
ekluzek: I did work on this.

cime_config/usermods_dirs/NEON/defaults/shell_commands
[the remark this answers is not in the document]
ekluzek: I was able to fix this readily, and got it done.

ekluzek:
The tag for this will be #2031

ekluzek:
I just added the new NEON surface datasets to CESM inputdata, so they can be downloaded now.

ekluzek:
@wwieder I compared some of the surface datasets to make sure they are the same. In many of the files LATIXY and lsmlat and/or lsmlon is different by roundoff. For some ORGANIC is also different to roundoff, TOOL is only different in NAT_PFT as expected. This all sounds reasonable to me, but documenting so we know.

ekluzek:
@wwieder I ran into a glitch with my testing that I need to run by you. I also realize that we need this tag ASAP, so it might be best to live with the glitch. But, there is a potential conflict for users using run_neon.py to set the stop date. So I'm not sure if we should deal with it now...

cime_config/usermods_dirs/NEON/defaults/shell_commands
```diff
@@ -23,5 +24,13 @@ else
...
+  ./xmlchange STOP_OPTION=nyears
```
ekluzek, line 29: @wwieder the problem with explicitly setting STOP variables is that it conflicts with the ways either run_neon.py might set it (based on the --end-date option), or their own settings, or in my case my test cases.

cime_config/usermods_dirs/NEON/YELL/shell_commands
```diff
@@ -3,14 +3,12 @@
...
-if [ $CLM_USRDAT_NAME=='NEON' ]
-then
+if [[ $CLM_USRDAT_NAME == 'NEON' ]]; then
   # Different default start date and number of months to run for transient case
   ./xmlchange DATM_YR_START=2019
-  if  [[ $compset =~ ^HIST ]]
-  then
+  if  [[ $compset =~ ^HIST ]]; then
     ./xmlchange RUN_STARTDATE=2019-01-01
     ./xmlchange STOP_N=39
```
ekluzek, line 13: This applies to all the user-mods, but explicitly setting STOP_N may interfere with the users own setting, test-mods, or run_neon.py --end-date option.

TeaganKing: We also have STOP* changes in cime_config/usermods_dirs/NEON/defaults/shell_commands, which then get overwritten by these site-specific xmlchanges.

Is there a way to set the STOP variables to only do an xmlchange IF there is not already an xmlchange (or two) for STOP variables listed in the case status, perhaps?

ekluzek: It would be awkward to check if there already is an xmlchange for it. I think it's possible, but I'd also view it as really fragile code that could be problematic.

But, that also wouldn't know about if the --end-date option was used in run_neon.py or if the user wanted to set it to something explicit.

Since, we have a time crunch here, I'm also OK with documenting it as a problem, and then making the tag. But, I also wanted to hear from everyone about if they think it will cause a problem for the tutorial users.  And if so we probably should do something. Although I think it would require that they "go away from the instruction" and do something wild.

TeaganKing: That certainly makes sense. Good point. I personally don't recall folks adjusting STOP variables in the previous tutorial, and wouldn't really expect users to need to do that in the flux course, but perhaps others can chime in.

wwieder: as far as I know the `--end-date` (and `--start-date`) options used in run_neon.py are not functional, we should likely just take this out to avoid confusion (but so far it hasn't been an issue).

ekluzek: Thanks for the discussion. From #1906 it looks like --end-date (and --run-length) are functional. I think this warrants more discussion because a common thing for a user is to be the one that controls the length of the simulation. So putting this in the user-mod directory is problematic.

But, for right now to get a tag out -- I'm going to check on the XML variable "TEST" so I can ignore this for my tests. I'll file an issue about setting STOP options and we can also discuss it tomorrow.

cime_config/usermods_dirs/NEON/defaults/shell_commands
ekluzek: This will be discussed in #2036

cime_config/usermods_dirs/NEON/YELL/shell_commands
ekluzek: We'll discuss more in #2036

PR #2020, opened 2023-06-07
Add more documentation to history tape code

johnpaulalex:
### Description of changes
Add more documentation to history tape code
  * Add more breadcrumbs between related variables and methods
  * Put related namelist flags/methods together
  * Update some out of date comments

Doc-only change.

### Specific notes

CTSM Issues Fixed (include github issue #):  none fixed, but preparation for fixing issue 27.

Testing performed, if any:
Ran unit tests to confirm new comments didn't break compilation

### DIFF
--- a/src/main/histFileMod.F90
+++ b/src/main/histFileMod.F90
@@ -5,6 +5,7 @@ module histFileMod
   !-----------------------------------------------------------------------
   ! !DESCRIPTION:
   ! Module containing methods to for CLM history file handling.
+  ! See 'history_tape' type for more details.
   !
   ! !USES:
   use shr_kind_mod   , only : r8 => shr_kind_r8
@@ -84,8 +85,8 @@ module histFileMod
 
   logical, public :: &
        hist_empty_htapes  = .false.      ! namelist: disable default-active history fields (which
-                                         ! only exist on history tape 1). Overridden by hist_fincl1 
-                                         ! flag.
+                                         ! only exist on history tape 1). Use hist_fincl1 to enable
+                                         ! select fields on top of this.
 
   character(len=max_namlen+2), public :: &
        hist_fincl1(max_flds) = ' '       ! namelist: list of fields to include in history tape 1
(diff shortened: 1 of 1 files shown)


src/main/histFileMod.F90
```diff
@@ -84,75 +82,86 @@ module histFileMod
   character(len=max_namlen), public :: &
        hist_type1d_pertape(max_tapes)  = (/(' ',ni=1,max_tapes)/)   ! namelist: per tape type1d
 
-  character(len=max_namlen+2), public :: &
-       fincl(max_flds,max_tapes)         ! namelist-equivalence list of fields to add
+  logical, public :: &
+       hist_empty_htapes  = .false.      ! namelist: disable default-active history fields (which
+                                         ! only exist on history tape 1). Overridden by hist_fincl1 
+                                         ! flag.
```
ekluzek, line 88: Hmmm, my first reaction here is that it isn't "overridden" by the hist_fincl1 flag. But, in a way it is in that anything in hist_fincl1 -- will be output. The way I've thought about it is that this turns all the default fields off -- and you have to explicitly add the list that you want to hist_fincl1.

src/main/histFileMod.F90
```diff
@@ -84,75 +82,86 @@ module histFileMod
...
 
   character(len=max_namlen+2), public :: &
-       hist_fincl1(max_flds) = ' '       ! namelist: list of fields to add
+       hist_fincl1(max_flds) = ' '       ! namelist: list of fields to include in history tape 1
```
ekluzek, line 91: I wonder if you should also include something that says "the "h0" set of files" or something to that effect?

src/main/histFileMod.F90
```diff
@@ -289,7 +315,7 @@ end subroutine copy_entry_interface
   !
   character(len=max_length_filename) :: locfnh(max_tapes)  ! local history file names
   character(len=max_length_filename) :: locfnhr(max_tapes) ! local history restart file names
-  logical :: htapes_defined = .false.            ! flag indicates history contents have been defined
+  logical :: htapes_defined = .false.        ! flag indicates history output fields have been selected
```
ekluzek, line 318: The wording of "defined" is specific here, because it's a term used in setting up NetCDF files. I like your switch of contents to "output fields", but I would keep defined over selected.

src/main/histFileMod.F90
```diff
@@ -873,7 +899,7 @@ subroutine htapes_fieldlist()
           if (ff > 0) then
 
              ! if field is in include list, ff > 0 and htape_addfld
-             ! will not be called for field
```
ekluzek, line 876: Nice catch!

src/main/histFileMod.F90
```diff
@@ -1102,8 +1128,7 @@ end function is_mapping_upto_subgrid
   subroutine htape_addfld (t, f, avgflag)
     !
     ! !DESCRIPTION:
-    ! Add a field to the active list for a history tape. Copy the data from
-    ! the master field list to the active list for the tape.
+    ! Add a field to a history tape, copying metadata from the master field list
```
ekluzek, line 1131: OK, I think your wording is more clear here. Had to think about it...

ekluzek (approved):
This is some great improvements to the documentation, and naming of some variables as well as the ordering of some things. There's a few suggestions I have about changes, but largely this is good to go. There's not a lot of code that actually changes here, so this all seems quite safe.

src/main/histFileMod.F90
johnpaulalex: thanks!

src/main/histFileMod.F90
johnpaulalex: Makes sense, I reworded to " Use hist_fincl1 to enable select fields on top of this."

src/main/histFileMod.F90
ekluzek: Yes, exactly, we call it "history tape 1", but the filenames are actually labeled as "h0". This is kind of an awkward thing we've been preserving. I like what you added below for history_tape and the breadcrumbs you put for pointing to it. Maybe there could be another one on top of the hist_fincl and hist_fexcl definitions?

ekluzek (approved):

src/main/histFileMod.F90
johnpaulalex: SGTM. How do I add a change at this point - looks like this commit got pulled into PR #2031, so should I just send a new PR for this final thing?

ekluzek: You could either: create a new PR, add it to #2022 which is still outstanding, or add it to here, and this PR will stay open and I can pull it into another tag. I have a slight preference for adding to #2022, but any of these would work fine. So I don't care too much.

Because we run a bunch of testing by hand to create our tags, I am very likely to merge a few PR's together to create the tag (especially if they are simple bit-for-bit type things). So you should expect that. But, it's also easier to see things broken up into smallish PR's so you can see what's going on with a specific PR. So I do still like having things broken up into separate PR's.

johnpaulalex: Added PR #2034 (note its commit is on top of the commits in this PR)

PR #2021, opened 2023-06-08
small changes to fix warnings; add print statements for clarity to users

adrifoster:
### Description of changes

Some small changes for quality of life improvements for the `run_neon` script.

### Specific notes

1. fixed a couple lines so that python will stop complaining about deprecated things
2. added print statements about warning messages being not an issue, and that building/running may take a while
3. added a "success" print statement - **I'm not sure this will actually only print if "successful"**

@ekluzek is there a way to get a return value from case.submit that tells you if it actually submitted?

@ekluzek @TeaganKing @wwieder 

Are answers expected to change (and if so in what way)? No changes

Testing performed, if any:
only affects `run_neon`

### DIFF
--- a/tools/site_and_regional/run_neon.py
+++ b/tools/site_and_regional/run_neon.py
@@ -356,22 +356,6 @@ def parse_isoduration(s):
     dt = datetime.timedelta(days=int(days) + 365 * int(years) + 30 * int(months))
     return int(dt.total_seconds() / 86400)
 
-def get_queue_command(batch_system):
-    match batch_system.lower():
-        'pbs':
-            queue_command = 'qstat -u <user_name>'
-        'slurm':
-            queue_command = 'squeue -u <user_name>'
-        'cobalt':
-            queue_command = 'qstat -u <user_name>'
-        'lsf':
-            queue_command = 'bqueues -u <user_name>'
-        case _:
-        queue_command = ''
-
-    return queue_command
-
-
 
 class NeonSite:
     """
@@ -490,7 +474,22 @@ class NeonSite:
         d1 = datetime.datetime(self.end_year, self.end_month, 1)
         d2 = datetime.datetime(self.start_year, self.start_month, 1)
         return (d1.year - d2.year) * 12 + d1.month - d2.month
+    
+    def get_batch_query(self, case):
+        """
+        Function for querying the batch queue query command for a case, depending on the 
+        user's batch system. If no batch system, empty string is returned
 
+        Args:
+        case:
+            case object
+        """
+        
+        if case.get_value("BATCH_SYSTEM") == "none":
+          return ""
+        else:
+          return case.get_value("batch_query")
+          
     def run_case(
         self,
         base_case_root,
@@ -553,7 +552,7 @@ class NeonSite:
                         case.submit(no_batch=no_batch)
                         logger.info("-----------------------------------")
                         logger.info("Successfully submitted case!")
-                        logger.info("Use 'qstat -u <user-name>' to check its run status")
+                        logger.info(f"Use {self.get_batch_query(case)} to check its run status")
                     return
             else:
                 logger.warning(
@@ -633,7 +632,7 @@ class NeonSite:
                 case.submit(no_batch=no_batch)
                 logger.info("-----------------------------------")
                 logger.info("Successfully submitted case!")
-                logger.info("Use 'qstat -u <user-name>' to check its run status")
+                logger.info(f"Use {self.get_batch_query(case)} to check its run status")
 
     def set_ref_case(self, case):
         rundir = case.get_value("RUNDIR")
--- a/tools/site_and_regional/run_neon.py
+++ b/tools/site_and_regional/run_neon.py
@@ -5,17 +5,17 @@
 |---------------------  Instructions  -----------------------------|
 |------------------------------------------------------------------|
 This is a wrapper script for running CTSM simulation for one or more
-neon sites. 
+neon sites.
 
 This script is only for neon site and we will develop a more general
 code later.
 
-This script first creates and builds a generic base case. 
+This script first creates and builds a generic base case.
 Next, it will clone the base_case for different neon sites and run
-types to reduce the need to build ctsm everytime. 
+types to reduce the need to build ctsm everytime.
 
 This script will do the following:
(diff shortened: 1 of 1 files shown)


tools/site_and_regional/run_neon.py
```diff
@@ -534,6 +535,9 @@ def run_case(
                     elif not setup_only:
                         print("Resubmitting case {}".format(case_root))
                         case.submit(no_batch=no_batch)
+                        logger.info("-----------------------------------")
+                        logger.info("Successfully submitted case!")
+                        logger.info("Use 'qstat -u <user-name>' to check its run status")
```
TeaganKing, line 540: This is really helpful to have the 'qstat' suggestion printed here. I think this will be a good way to inform users that things are running and what the best methods are to check on case status going forward.

wwieder (approved):
Thanks @adrifoster . @ekluzek can we just merge this with #2015?

tools/site_and_regional/run_neon.py
```diff
@@ -747,7 +754,7 @@ def parse_neon_listing(listing_file, valid_neon_sites):
 
             tmp_df = tmp_df[tmp_df[7].str.contains(latest_version)]
             # -- remove .nc from the file names
-            tmp_df[9] = tmp_df[9].str.replace(".nc", "")
+            tmp_df[9] = tmp_df[9].str.replace(".nc", "", regex=False)
```
ekluzek, line 757: For education I'd love to hear about the reason for this. Where there some cases that were failing that requires this addition?

tools/site_and_regional/run_neon.py
```diff
@@ -611,6 +615,9 @@ def run_case(
             case.check_all_input_data()
             if not setup_only:
                 case.submit(no_batch=no_batch)
+                logger.info("-----------------------------------")
+                logger.info("Successfully submitted case!")
+                logger.info("Use 'qstat -u <user-name>' to check its run status")
```
ekluzek, line 620: The problem with this is that you are assuming that PBS is being used, which won't necessarily be the case on non NCAR machines. I think run-neon.py is agnostic in terms of machine, so we shouldn't make this assumption here. The build does actually specify what the status command is, and there is a script to check the status (although it was non functional the last time I looked). So I think with a little more work we could get the right name.

Or you could add something to the effect of

Use the batch status checker on your system to check it's run status

That's not nearly as good for a naive user though so...

tools/site_and_regional/run_neon.py
ekluzek: See my comment below...

ekluzek (changes requested):
My main issue is building in the assumption that you are running on cheyenne. I think we can work around this, but will take a little investigation.

ekluzek:
@wwieder I'll plan on merging a few PR's to actually make a tag. But, outside of that I think we should keep this separate.

@adrifoster I think there might be something that gives a success status in case.submit, but will need to investigate.

tools/site_and_regional/run_neon.py
adrifoster: Not failing, Python was giving a warning:

`[FutureWarning: The default value of regex will change from True to False in a future version](https://stackoverflow.com/questions/66603854/futurewarning-the-default-value-of-regex-will-change-from-true-to-false-in-a-fu)`

I don't think we actually need this to be a regex evaluation?


according to: https://pandas.pydata.org/docs/whatsnew/v1.2.0.html#deprecationse, I updated this function accordingly. I don't think we need

tools/site_and_regional/run_neon.py
adrifoster: Yes true this does rely on having PBS.

ekluzek: It looks like you just need to do this...

`./xmlquery batch_query`

to get the batch q-status. So that's easy enough.

There is also a ./case.qstatus script, but it doesn't seem to ever work correctly for me so I don't use it. We really should get that working in cime if it populates cases with it.

tools/site_and_regional/run_neon.py
adrifoster: In the interest of getting this in before the flux workshop maybe we should just remove this line for now? I am not immediately sure how to call the xmlquery from this script since right now everything is done with the actual python objects.

but I imagine if we had that we could make some kind of function...


````
def get_queue_command(batch_system):
  match batch_system.lower():
      'pbs':
           queue_command = 'qstat -u <user_name>'
      'slurm':
           queue_command = 'squeue -u <user_name>'
      'cobalt':
          queue_command = 'qstat -u <user_name>'
      'lsf':
            queue_command = 'bqueues -u <user_name>'
      case _:
          queue_command = ''

    return queue_command
```

I think these are all correct...

TeaganKing: I'm wondering if another temporary fix option would be just to print something like "if using PBS, check case status with qstat -u user_name; if using slurm, check case status with ..."? Maybe it would look clunky, but it would at least inform users that it is useful to check output status. Just an idea.

ekluzek (approved):
Awesome. Thanks for that update @adrifoster looks good. It was good for us to look at that together earlier.

tools/site_and_regional/run_neon.py
ekluzek: Awesome, that is a great future protection to have. Thanks for bringing this in. Resolving this thread...

tools/site_and_regional/run_neon.py
ekluzek: @adrifoster and I met and worked out what this needed to be and she implemented it. So this is resolved.

tools/site_and_regional/run_neon.py
ekluzek: This was worked out...

adrifoster:
> Thanks @ekluzek , for some reason it's not printing the "success" lines so I'm diagnosing this now...

Okay just needed to change the `logger.info`s to `print`s! working now!

On Cheyenne:

```
varname year_align is already in stream mod dictionary
varname dtlimit is already in stream mod dictionary
submit_jobs case.run
Submit job case.run
Submit job case.st_archive
-----------------------------------
Successfully submitted case!
Use qstat to check its run status
```

PR #2022, opened 2023-06-14
Delete unused hist_add_subscript method (no references in the CTSM repo)

johnpaulalex:
### Description of changes
Delete unused hist_add_subscript method

### Specific notes


ekluzek:
Interesting that you found this. It does look like you are right this is code that is totally not used. There's also some data that goes along with this that can be removed as well num_)subs, sub_* etc. at the top of the module. It defines these as dimensions, but since num_subs==0 no dimensions get set.

This does go back to code in clm4_0_* so it's quite old. I'm guessing it must be something needed in clm4_0, that we don't need anymore. Since, this seems to be about defining dimensions, maybe these "Subscripts" were dimensions that needed to be defined in the past and were taken out.

ekluzek (changes requested):
This does indeed look like old code that isn't used anymore. There's some data that goes along with it that we should remove as well, that I comment about earlier. That's the only thing to additionally do. I do think we should remove the data, but I can also do that myself.

How did you catch this? Just by looking at the code, or by use of any tools that pointed it out to you?

ekluzek (approved):

Issue #1696, opened 2022-04-06
Implementation of flood irrigation and paddy irrigation as well as the irrigation method dataset


YiYaoVUB:
- [1] Currently, drip and sprinkler irrigation methods have been implemented in CTSM. However, the most widely used method, flood irrigation, has not yet been developed. The difference between drip and sprinkler irrigation is where water is applied (under canopy for drip and over canopy for sprinkler). We think for flood irrigation, the water should also be applied under the canopy, but the amount of water should be different. Considering that the soil normally will be saturated in a short period, we propose that we change the amount of water to the deficit between saturated soil water and current soil water.

- [2] To simulate irrigation-induced impacts more realistically, a surface dataset with the spatial distribution of the irrigation method will be needed. In #565 @billsacks has posted the enhancement required. Here we propose to use the data from https://hess.copernicus.org/articles/19/3073/2015/, in which they generated a global map of irrigation method methods at 0.5x0.5 resolution. For each CFT in every pixel, they provided the fraction of three irrigation methods (flood, sprinkler, and drip). In mksurfdata tool, we propose to remap the fraction of irrigation methods, then choose the dominant one to limit the computation cost.

- [3] Rice is a CFT that requires special irrigation. During the growth season, the surface water needs to be held in the field for some periods. We propose to change the surface water module of rice paddies, so water could be held at a given point (at the saturated point or a given depth).

wwieder:
Thanks for creating this issue @YiYaoVUB.  Next steps are to open a pull request that addresses each of the topics above.  Let's start with the with flood irrigation work and move on from there.  

Thanks again for these contributions!

PR #2026, opened 2023-06-15
Irr 2022 james

YiYaoVUB:
### Specific notes

Testing performed, if any:

### DIFF
--- a/src/biogeophys/IrrigationMod.F90
+++ b/src/biogeophys/IrrigationMod.F90
@@ -481,7 +481,7 @@ contains
          irrig_method_default_int = irrig_method_drip
       case ('sprinkler')
          irrig_method_default_int = irrig_method_sprinkler
-	  case ('flood')
+	   case ('flood')
          irrig_method_default_int = irrig_method_flood													  
       case default
          write(iulog,*) 'ERROR: unknown irrig_method_default: ', trim(irrig_method_default)
@@ -1321,7 +1321,7 @@ contains
           waterflux_inst%qflx_irrig_drip_patch(p)      = qflx_irrig_tot
        else if(this%irrig_method_patch(p) == irrig_method_sprinkler) then
           waterflux_inst%qflx_irrig_sprinkler_patch(p) = qflx_irrig_tot
-	   else if(this%irrig_method_patch(p) == irrig_method_flood) then
+	    else if(this%irrig_method_patch(p) == irrig_method_flood) then
           waterflux_inst%qflx_irrig_drip_patch(p) = qflx_irrig_tot																 														  
        else
           call endrun(subgrid_index=p, subgrid_level=subgrid_level_patch, &
@@ -1428,7 +1428,7 @@ contains
 
     ! target soil moisture for this layer [kg/m2] (target for drip and spri; target_satu for flood and paddy)
     real(r8) :: h2osoi_liq_target
-	real(r8) :: h2osoi_liq_target_satu							   
+	 real(r8) :: h2osoi_liq_target_satu							   
 
     ! soil moisture at wilting point for this layer [kg/m2]
     real(r8) :: h2osoi_liq_wilting_point
@@ -1438,7 +1438,7 @@ contains
 
     ! Total of h2osoi_liq_target down to the depth of irrigation in each column [kg/m2]
     real(r8) :: h2osoi_liq_target_tot(bounds%begc:bounds%endc)
-	real(r8) :: h2osoi_liq_target_satu_tot(bounds%begc:bounds%endc)															
+	 real(r8) :: h2osoi_liq_target_satu_tot(bounds%begc:bounds%endc)															
 
     ! Total of h2osoi_liq at wilting point down to the depth of irrigation in each column
     ! [kg/m2]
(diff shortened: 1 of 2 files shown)


YiYaoVUB:
Hi Sam @samsrabin , please check this draft :)
Cheers,
Yi

src/biogeophys/SurfaceWaterMod.F90
```diff
@@ -472,15 +472,25 @@ subroutine QflxH2osfcSurf(bounds, num_hydrologyc, filter_hydrologyc, &
        c = filter_hydrologyc(fc)
 
        if (h2osfcflag==1) then
-          if (frac_h2osfc_nosnow(c) <= params_inst%pc) then
-             frac_infclust=0.0_r8
-          else
-             frac_infclust=(frac_h2osfc_nosnow(c)-params_inst%pc)**params_inst%mu
-          endif
+	   ! for rice we set the runoff at 0
+         if (col%itype(c)==(200+cft_lb+46) .or. col%itype(c) == (200+cft_lb+47)) then
```
samsrabin, line 476: This will need to use the standard CLM irrigated rice types, `nrice` and `nirrig_rice`. (Although why did you include rainfed rice?)

src/biogeophys/SurfaceWaterMod.F90
```diff
@@ -472,15 +472,25 @@ subroutine QflxH2osfcSurf(bounds, num_hydrologyc, filter_hydrologyc, &
...
+            frac_infclust=0.0_r8
+         else
+            frac_infclust=(frac_h2osfc_nosnow(c)-params_inst%pc)**params_inst%mu
+         endif
        endif
 
        ! limit runoff to value of storage above S(pc)
-       if(h2osfc(c) > h2osfc_thresh(c) .and. h2osfcflag/=0) then
+	   ! if surface water exceeds 10cm, then do drainage in the next step
+       if (col%itype(c)==(200+cft_lb+46) .or. col%itype(c)==(200+cft_lb+47)) then
```
samsrabin, line 487: This will need to use the standard CLM irrigated rice types, `nrice` and `nirrig_rice`. (Although why did you include rainfed rice?)

samsrabin:
Thanks for this, Yi! It looks pretty simple—just adding a new `flood` irrigation option that's used worldwide, in the style of the existing `drip` and `sprinkler`. It seems that there's some special behavior for rice, where runoff is set to 0 to reflect the use of ponding. So it looks like it addresses parts 1 and 3 of #1696.

This doesn't include the ability to read in maps of irrigation method, but it might be a good idea anyway to do that in a separate PR, just because small PRs are more tractable. I'm assuming you did this intentionally.

However, it looks like you're using your new PFT numbering to detect rice, which wouldn't work in our version. I've added comments in the code diff where this happens.

If I recall correctly, the main use of that new numbering scheme would be to allow multiple types of irrigation in one gridcell-year. That is probably the lowest priority for us in terms of integration. I'd think, in decreasing order of priority:
1. This PR,
2. A PR allowing the reading of irrigation method map files,
3. (Maybe someday) A PR allowing multiple irrigation methods in a single gridcell-year.

Before I send this around, please do the following:
1. Fill in the requested fields in the PR description, and
2. Change the title to something meaningful, like "Add flood irrigation option".

Thanks!

src/biogeophys/SurfaceWaterMod.F90
YiYaoVUB: Hey Sam, this one is a bit complicated. I guess I would need to use col%patch(bounds%begp) as there is only one patch for each irrigated crop column right? Will try and test it.

src/biogeophys/SurfaceWaterMod.F90
YiYaoVUB: Hey Sam, this one is a bit complicated. I guess I would need to use col%patch(bounds%begp) as there is only one patch for each irrigated crop column right? Will try and test it. Based on my experience, farmers normally change micro topography to hold the water in the field, so I also did it for rainfed rice. I would be very surprised if they don't do it for rainfed rice (check fig1: https://link.springer.com/article/10.1007/s13593-021-00710-2)

YiYaoVUB:
Thanks Sam! Actually this pull request does not include allowing different irrigation techniques for one CFT in one gridcell :) I will open a PR for that in future if they want it. All stuff in this PR is from my paper Yao_et_al_2022_JAMES, that is why I named it this, but i agree with you that i need to change the name to something more clear. By the way, the map reading part has already been finished by Sean Swenson if I am not wrong :) Anyway, I will work on this and try to redo the pr.
Cheers,
Yi

src/biogeophys/SurfaceWaterMod.F90
samsrabin: I think `bounds%begp` refers to the first patch on the processor, in which case that wouldn't work. To get the crop type associated with a column, here's an adaptation on something I've done before (although I'd hope there's a more elegant way)

```fortran
vegtype_thiscol = 0
if (col%itype(c) == istcrop) then
    n_noncrop = 0
    do p = col%patchi(c),col%patchf(c)
        if (patch%active(p)) then
            if (patch%itype(p) >= npcropmin) then
                if (vegtype_thiscol > 0) then
                    call endrun('ERROR multiple active crop patches found in this column')
                end if
                vegtype_thiscol = patch%itype(p)
            else
                n_noncrop = n_noncrop + 1
            end if
        end if
    end do
    
    ! Check
    if (n_noncrop > 0 .and. vegtype_thiscol > 0) then
        call endrun('ERROR Active crop and non-crop patches found in this active column')
    end if
end if
```

src/biogeophys/SurfaceWaterMod.F90
samsrabin: Ah okay, that makes sense about the microtopography. In that case, it'd be helpful to add a comment explaining your reasoning so that future developers don't get confused like I did.

tools/mksurfdata_map/src/mkirrigation_methodMod.F90
```diff
@@ -0,0 +1,274 @@
+module mkirrigation_methodMod
+!-----------------------------------------------------------------------
+!BOP
+!
+! !MODULE: mklai
+!
+! !DESCRIPTION:
+! Make LAI/SAI/height data
```
samsrabin, line 8: This and the rest of the description will need to be changed.

src/biogeophys/IrrigationMod.F90
```diff
@@ -478,6 +481,8 @@ subroutine translate_irrig_method_default
          irrig_method_default_int = irrig_method_drip
       case ('sprinkler')
          irrig_method_default_int = irrig_method_sprinkler
+      case ('flood')
+         irrig_method_default_int = irrig_method_flood
```
samsrabin, line 485: I know you didn't invent `irrig_method_default_int`, but it doesn't appear to be used for anything.

samsrabin:
Closing this, as it is superseded by #2102.

PR #2035, opened 2023-06-21
doc that masterlist actflag is just h0 in practice, doc that masterlist has all the fields. 

johnpaulalex:
### Description of changes

document that masterlist actflag is just h0 in practice, document that masterlist has all the fields.

### Specific notes


ekluzek (approved):

samsrabin:
Just a heads up: I expect this to have conflicts with #2087, which is due to be brought in soon. They shouldn't be too hard to resolve, though.

ekluzek:
I looked at the conflicts and there's only one with #2087 and it should be easy to resolve.

PR #2044, opened 2023-06-28
More confined regular expression for NEON

### LINKED ISSUE #1499: File written by hist_printflds is missing some fields
### Brief summary of bug

I noticed that the file written by the hist_printflds routine (e.g., master_list_nofates.rst) is missing some fields.

### General bug information

**CTSM version you are using:** ctsm5.1.dev057

**Does this bug cause significantly incorrect results in the model's science?** NO

**Configurations affected:** ??

### Details of bug

I noticed that all of the `PCT_*` fields (defined in subgridWeightsMod) are missing from https://github.com/ESCOMP/CTSM/blob/master/doc/source/users_guide/setting-up-and-running-a-case/master_list_nofates.rst and from the file in the test, `ERP_P36x2_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs`, even though they are present in the history file for that test. I haven't looked to see if any other fields may be affected.

The problem is that hist_printflds is called a little too early in initialization – before init_subgrid_weights_mod and possibly before other relevant subroutines. This could probably be safely moved to the end of initialize2.

We should move that routine, then regenerate the documentation from these history field lists.

### LINKED ISSUE #1506: Remove extra mapping files from namelist_defaults_ctsm.xml
With ctsm5.1.dev040 many of the mapping files in namelist_defaults_ctsm.xml are now uneeded and can be removed. As you can see from the discussion in #1501 this can cause confusion. So it would be best if the extra entries were removed.

### LINKED ISSUE #2028: Refactor to move setting of NEON surface datasets deeper into the build-namelist system
This is similar to #2027, but not as obvious as something needed to do. Right now NEON surface datasets are set using user-mods. We have a line that gives the creation dates for all NEON files and we update all files even when we only need one to be updated. We also have a separate copy of this setting for FATES.

If we refactor to move this to inside of namelist_defaults, we could easily have a line for each site. Which would make it easily possible to change date stamps for individual sites. And the logic for FATES could be embedded at that level as well. Which would remove some of the need to replicate the user-mods for FATES.

The downside of creating all files isn't too bad, since these datasets are small. But, for dataset providence it would be beneficial to only update the datasets that really need to, so then you can understand the history of needed changes more clearly.

### LINKED ISSUE #2039: Conditional for NEON usermods is too broad
This statement is still true in a BGC case because of the extra wildcard

https://github.com/ESCOMP/CTSM/blob/bb2a8d2c0c05ebb5417724f98fb0586dc7584ae6/cime_config/usermods_dirs/NEON/defaults/shell_commands#L35C1-L36C1

Something like the following would be more accurate

`if [[ $compset =~ .*CLM[0-9]+%SP* ]]; then`

### LINKED ISSUE #2103: Inconsistency in how resolution/grid is set in build-namelist vs. buildnml
This is an inconsistency in how the resolution/grid is set in the buildnml and in build-namelist. It's been in place for a very long time. And in a sense it doesn't matter in terms of how it works -- but it consistency would likely be better.

However, these is one case where it doesn't work correctly and that's in the build-namelist unit tester which makes the wrong assumption so it is an issue there.

The issue is that for buildnml: 
- LND_GRID is set fo CLM_USRDAT meaning that a user defined grid name is going to be used
- CLM_USRDAT_NAME is the name of the set of data (i.e. NEON) or the specific site

While for build-namelist:
- Both "--res" and "--clm_usr_name" are set to $CLM_USRDAT_NAME (from within buildnml)

To be consistent "--res" should be CLM_USRDAT and "--clm_usr_name" should be $CLM_USRDAT_NAME. This is helpful because that's closer to how the case is actually set to. And the case can NOT set LND_GRID to an arbitrary name -- it HAS to be CLM_USRDAT for the general structure to work with the cime, cdeps and cmeps externals (as well as the MCT code).

ekluzek:
### Description of changes

More confined regular expression for NEON. And a few simple fixes.

### Specific notes

  Fixes #2039
  Fixes #2103
  Fixes #2028
  Fixes #1506
  Fixes #1499

Are answers expected to change (and if so in what way)? Yes? For some NEON cases?

Testing performed, if any: None so far, but regular testing will be run


wwieder:
@ekluzek I wonder what a timeline for merging this change could be?  I ask because @danicalombardozzi is working on revisions to our NEON paper and assumed that we'll also need @briandobbins to make a new docker image with the code modifications that have evolved since our last NEON branch tag?

ekluzek:
Hi Will. Let's talk more tomorrow. But, we've been delaying on this to get other work in. And we want the BGC sequencing work to come in as soon as it's ready so difficult merges won't have to happen. This is simple enough and orthogonal to that, that it's going to be OK to come in before that if needed. I think we should be able to get it in sometime next week.

wwieder:
@slevis-lmwg can you add this to the merge tag you're working on, it addressed a bug we need fixed for the container used for NEON runs.

PR #2045, opened 2023-06-28
Surface roughness modifications (cont'd)

### LINKED ISSUE #1316: New parameterization of vegetation surface roughness for forests

slevis-lmwg:
### Description of changes

This PR continues from where #1596 left off.

### Specific notes

@ekluzek @RonnyMeier 

Fixes #1596 
Fixes #1316 

Yes

Yes, NLCOMP diffs in the test-suites

Testing performed, if any:
Test-suite results reported below


slevis-lmwg:
TODO carrying over from #1596:
Refactor some of the "magic numbers" with parameter names from the corresponding publication

src/main/lnd2glcMod.F90
```diff
@@ -204,7 +204,7 @@ subroutine update_lnd2glc(this, bounds, num_do_smb_c, filter_do_smb_c, &
       ! Make sure we haven't already assigned the coupling fields for this point
       ! (this could happen, for example, if there were multiple columns in the
       ! istsoil landunit, which we aren't prepared to handle)
-      if (fields_assigned(g,n)) then
+      if (1==2) then
```
slevis-lmwg, line 207: Looking through the surface roughness mods, I came across this modified if-statement.

I don't know the reason for the change, but @ekluzek I'm wondering whether you would approve this change.

ekluzek: This looks like a mistake to me. So I'm glad you noticed and pointed it out. I would put it back and make sure things work. If not, we'll need to investigate further. My hope is that this was a test that was accidentally committed, and shouldn't have been. If it is needed it would need to be changed.

slevis-lmwg:
I will soon commit my refactor of "magic numbers" to parameters.
I refactored in two steps:
1) Bit-for-bit identical (bfb) refactor, as checked by comparing 2002-12-31 h0 files from `ERS_Ly3.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-Meier2022_surf_rough`.
2) Answer-changing refactor to replace divisions with multiplications, as checked by comparing the same output as in (1). The answer-changing refactor remains bfb for tests without the `Meier2022_surf_rough` testmods, as checked by comparing h0 cprnc files from `ERP_P72x2_Lm25.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-monthly`.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -17,7 +17,7 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
-  <test name="SMS_D_Ln1" grid="f09_g17" compset="I1850Clm50BgcNoAnthro" testmods="clm/Meier2022_surf_rough">
+  <test name="SMS_D_Ln6" grid="f09_g17" compset="I1850Clm50BgcNoAnthro" testmods="clm/Meier2022_surf_rough">
```
slevis-lmwg, line 20: I should have kept this as a separate commit...

This test was failing on SUBMIT with this error:
`Runtime (1800.0 s) must be a multiple of the longest coupling interval ROF_NCPL (10800.0s).  Adjust runtime or ROF_NCPL`
...so I changed the length accordingly.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -865,13 +884,53 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+            z0mv(p)   = exp(egvf * log(z0mv(p)) + (1._r8 - egvf) * log(z0mg(c)))
+
+         case ('Meier2022')
+            lt = max(0.00001_r8,elai(p)+esai(p))
+            displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(cd1_param * lt)**0.5_r8)) / (cd1_param*lt)**0.5_r8)
+
+            lt = min(lt,z0v_LAImax(patch%itype(p)))
+            delt = 2._r8
+            U_ustar_ini = (z0v_Cs(patch%itype(p)) + z0v_Cr(patch%itype(p)) * lt / 2._r8)**(-0.5_r8) &
+                      *z0v_c(patch%itype(p)) * lt / 4._r8
```
slevis-lmwg, line 903: Found more divisions that could be * 0.5 and * 0.25 but will hold off for now...

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -516,26 +557,61 @@ subroutine SetRoughnessLengthsAndForcHeightsNonLake(this, bounds, &
...
+                if ( snomelt_accum(c) < 1.e-5_r8 )then
+                    z0mg(c) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8
+                else
+                    z0mg(c) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
+                end if
+             else
+                z0mg(c) = this%zsno
+             end if                    
+          else if (lun%itype(l) == istice) then
+             z0mg(c) = this%zglc
```
slevis-lmwg, line 605: The above seems unnecessarily repetitive just to apply this `else if`. Refactoring should be bfb.

src/biogeophys/LakeFluxesMod.F90
```diff
@@ -324,20 +336,62 @@ subroutine LakeFluxes(bounds, num_lakec, filter_lakec, num_lakep, filter_lakep,
...
+               if(use_z0m_snowmelt) then
+                  if ( snomelt_accum(c) < 1.e-5_r8 ) then
+                     z0mg(p) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8
+                  else
+                     z0mg(p) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
+                  end if
+               else
+                  z0mg(p) = params_inst%zsno
+               end if                      
+               z0hg(p) = z0mg(p) / exp(params_inst%a_coef * (ust_lake(c) * z0mg(p) / nu_param)**params_inst%a_exp) ! Consistent with BareGroundFluxes
```
slevis-lmwg, line 377: I think I can also reduce repetition in the above with bfb results.

src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -488,6 +491,7 @@ subroutine BulkDiag_NewSnowDiagnostics(bounds, num_c, filter_c, &
        ! all snow falls on ground, no snow on h2osfc (note that qflx_snow_h2osfc is
        ! currently set to 0 always in CanopyHydrologyMod)
        newsnow(c) = qflx_snow_grnd(c) * dtime
+       snomelt_accum(c) = max(0._r8, snomelt_accum(c) - newsnow(c)/1000._r8)
```
slevis-lmwg, line 494: ...another division that could be a multiplication.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1381,6 +1382,7 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
                      if (imelt(c,j) == 1 .AND. j < 1) then
                         qflx_snomelt_lyr(c,j) = max(0._r8,(wice0(c,j)-h2osoi_ice(c,j)))/dtime
                         qflx_snomelt(c)       = qflx_snomelt(c) + qflx_snomelt_lyr(c,j)
+                        snomelt_accum(c)      = snomelt_accum(c) + qflx_snomelt_lyr(c,j) * dtime/1000._r8
```
slevis-lmwg, line 1385: ...one more

src/biogeophys/LakeFluxesMod.F90
slevis-lmwg: Testing the refactors here and in FrictionVelocityMod with the test that I used in [this earlier post.](https://github.com/ESCOMP/CTSM/pull/2045#issuecomment-1636423638)

slevis-lmwg:
@ekluzek I consider this PR ready, unless:
1) we prefer to back out my answer-changing mods that replace divisions with multiplications
2) conversely we prefer to include a few additional answer-changing mods (replace more divisions with multiplications and remove repetitive code)

slevis-lmwg:
In this morning's software meeting @ekluzek explained that test-suite DIFFs from baseline should be limited to clm51, which agrees with what I found in this [earlier post](https://github.com/ESCOMP/CTSM/pull/2045#issuecomment-1612224410).

I should rerun these tests without the surface roughness mods to confirm BFB.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -865,13 +884,53 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+          case default
+            write(iulog,*) 'ERROR: unknown z0para_method: ', z0param_method
+            call endrun(msg = 'unknown z0param_method', additional_msg = errMsg(sourcefile, __LINE__))
+          end select
+
+          z0hv(p)   = z0mv(p)
+          z0qv(p)   = z0mv(p)
+
+          ! Update the forcing heights
+          ! TODO(KWO, 2022-03-15) Only for Meier2022 for now to maintain bfb with ZengWang2007
```
slevis-lmwg, line 927: Does this comment mean that we should remove the following if statement (here and elsewhere) going forward?

ekluzek: Hmmm, good question!

ekluzek: Who is KWO? That wouldn't be Ronny Meier... And it's not Keith Oleson either. We can't ask Ronny, but should look into this...

ekluzek: @olyson do you know what's going on here with this comments and the other's like it?

ekluzek: @swensosc do you know anything about this comment?

olyson: That's me, W is my middle initial.  In the process of working on this PR I found that forcing heights weren't being updated properly, independent of the Meier changes.  But I wanted to maintain bfb for the original code (which uses ZengWang2007).  At some point (now?), yes, we will want to remove the if statement so that the forcing heights are updated for both Meier and ZengWang2007.

ekluzek: Ahh, OK so Meier is doing the right thing, and we ought to change it so both are consistent. We probably should do that in a future update as an answer changing tag.

ekluzek: We should make an issue to remind us to do that. @olyson could you start that one?

olyson: To be clear, Meier is only doing the right thing because I added that code in.  I'll open an issue.

slevis-lmwg: I clarified with @ekluzek 
- We will merge with these lines as they are.
- @olyson will start an issue about it to update later.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1329,11 +1393,39 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
 
             if (zeta(p) >= 0._r8) then     !stable
-               zeta(p) = min(zetamax,max(zeta(p),0.01_r8))
+               ! remove stability cap when biomass heat storage is active 
+               if(use_biomass_heat_storage) then 
+                  ! TODO(KWO, 2022-03-15) Only for Meier2022 for now to maintain bfb with ZengWang2007
+                  if (z0param_method == 'Meier2022') then
+                     zeta(p) = min(zetamax,max(zeta(p),0.01_r8))
+                  else
+                     zeta(p) = min(100._r8,max(zeta(p),0.01_r8))
```
slevis-lmwg, line 1402: Revert new lines 1396-1406 to old line 1332. That should bring back BFB in clm51 cases.

src/biogeophys/CanopyFluxesMod.F90
slevis-lmwg: done

src/biogeophys/CanopyFluxesMod.F90
slevis-lmwg: This test 
ERI_C2_Ld9.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_gnu.clm-default
confirms that we get BFB against dev129 again. Next I have submitted the cheyenne test-suite.

slevis-lmwg: Still getting some (though fewer) diffs from the baseline. Results filtered by this command:
`./cs.status.fails | grep -v NLCOMP | grep -v 'otherwise bit-for-bit' | grep -v 'does not exist' | grep -v 'EXPECTED FAILURE' | grep FAIL`
```
SMS_D.1x1_brazil.I2000Clm51FatesSpCruRsGs.cheyenne_gnu.clm-FatesColdDryDepSatPhen BASELINE ctsm5.1.dev129: DIFF
SMS_D.1x1_brazil.I2000Clm51FatesSpCruRsGs.cheyenne_gnu.clm-FatesColdMeganSatPhen BASELINE ctsm5.1.dev129: DIFF
ERP_P36x2_Ld30.f45_f45_mg37.I2000Clm51FatesSpCruRsGs.cheyenne_intel.clm-FatesColdSatPhen BASELINE ctsm5.1.dev129: DIFF
ERS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm51FatesRs.cheyenne_intel.clm-FatesCold BASELINE ctsm5.1.dev129: DIFF
ERS_Ly5_P144x1.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-cropMonthOutput BASELINE ctsm5.1.dev129: DIFF
SMS.f45_f45_mg37.I2000Clm51FatesSpRsGs.cheyenne_nvhpc.clm-FatesColdSatPhen BASELINE ctsm5.1.dev129: DIFF
```
I will investigate.

slevis-lmwg: The fates tests' lnd_in files differ from the baseline (I don't know why, yet):
```
101,102d100
<  use_z0m_snowmelt = .false.
<  z0param_method = 'ZengWang2007'
168c166
<  zetamaxstable = 2.0d00
---
>  zetamaxstable = 0.5d00
```
The one non-fates test does not have the zetamaxstable difference, so I don't know why it differs from the baseline. I will investigate.

slevis-lmwg: The non-fates diffs count as roundoff I think, maybe because we're looking at output after 5 years?
The largest normalized diffs are:
```
RMS ERRH2OSNO                        2.8529E-18            NORMALIZED  5.6909E-03
RMS ERRSEB                           2.7629E-18            NORMALIZED  4.9140E-04
RMS ERRSOI                           1.1648E-14            NORMALIZED  3.7069E-05
RMS ERRSOL                           7.7001E-19            NORMALIZED  1.4330E-03
RMS ERRH2O                           8.1752E-26            NORMALIZED  1.6803E-10
```
and the rest are 1e-12 or smaller. I will bring this up at Monday's stand-up to decide whether to dig further.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -213,7 +213,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <zetamaxstable               use_biomass_heat_storage=".true." >2.0d00</zetamaxstable>
 <zetamaxstable phys="clm4_5"                                   >2.0d00</zetamaxstable>
 <zetamaxstable phys="clm5_0"                                   >0.5d00</zetamaxstable>
-<zetamaxstable phys="clm5_1"                                   >0.5d00</zetamaxstable>
+<zetamaxstable phys="clm5_1"                                   >2.0d00</zetamaxstable>
+<zetamaxstable phys="clm5_1" z0param_method="ZengWang2007" >0.5d00</zetamaxstable>
+<zetamaxstable phys="clm5_1" z0param_method="Meier2022" >2.0d00</zetamaxstable>
```
slevis-lmwg, line 218: @ekluzek based on our conversation last week that we want this PR to NOT be answer-changing for tests not using Meier2022, I am reverting line 216 to 0.5, because it fixes the lnd_in files of the Fates tests that differed from baseline (dev129).

cime_config/testdefs/ExpectedTestFails.xml
```diff
@@ -30,6 +30,13 @@
 
   <!-- aux_clm test suite failures -->
 
+  <test name="ERS_D.f10_f10_mg37.I1850Clm51Sp.cheyenne_intel.clm-Meier2022_surf_rough">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>Error: Forcing height is below canopy height for patch index</issue>
+    </phase>
+  </test>
+
```
slevis-lmwg, line 39: Removing these lines because I have not encountered a RUN failure with this test.

src/main/clm_varcon.F90
```diff
@@ -154,6 +154,15 @@ module clm_varcon
...
   ! real(r8) :: c14ratio = 1._r8  ! debug lets set to 1 to try to avoid numerical errors
 
+  !------------------------------------------------------------------
+  ! Surface roughness constants
+  !------------------------------------------------------------------
+  real(r8), public, parameter :: beta_param = 7.2_r8  ! Meier et al. (2022) https://doi.org/10.5194/gmd-15-2365-2022
+  real(r8), public, parameter :: nu_param = 1.5e-5_r8  ! Meier et al. (2022) kinematic viscosity of air
+  real(r8), public, parameter :: b1_param = 1.4_r8  ! Meier et al. (2022) empirical constant
+  real(r8), public, parameter :: b4_param = -0.31_r8  ! Meier et al. (2022) empirical constant
+  real(r8), public, parameter :: cd1_param = 7.5_r8  ! Meier et al. (2022) originally from Raupach (1994)
```
slevis-lmwg, line 164: By reverting the code to magic numbers rather than using these parameters in their place, I have matched the baseline (dev129) for the last failing test.

slevis-lmwg: I now hypothesize that the last failing test had roundoff diffs due to a change in order of operations when using `b4_param` from -0.31 to + (-0.31). Not tested.

slevis-lmwg:
Test-suites:
Izumi IN PROGRESS `./run_sys_tests -s aux_clm -c ctsm5.1.dev131 -g ctsm5.1.dev133`
Cheyenne WAIT for compute nodes to become available

src/main/clm_varcon.F90
slevis-lmwg: When cheyenne's compute nodes come up:
Try the same test, but reverting only b4_param to its hardwired number rather than all the above parameters.

slevis-lmwg: Repeated the test as recommended in the previous post:
```
+++ b/src/biogeophys/FrictionVelocityMod.F90
@@ -579,9 +579,9 @@ contains
           if (frac_sno(c) > 0._r8) then
              if(use_z0m_snowmelt) then
                 if ( snomelt_accum(c) < 1.e-5_r8 )then
-                    z0mg(c) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8
+                    z0mg(c) = exp(-b1_param * rpi * 0.5_r8 -0.31_r8  ) * 1.e-3_r8
                 else
-                    z0mg(c) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
+                    z0mg(c) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) -0.31_r8  ) * 1.e-3_r8
                 end if
              else
                 z0mg(c) = this%zsno
@@ -594,9 +594,9 @@ contains
           if (frac_sno(c) > 0._r8) then ! Do snow first because ice could be snow-covered
              if(use_z0m_snowmelt) then
                 if ( snomelt_accum(c) < 1.e-5_r8 )then
-                    z0mg(c) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8
+                    z0mg(c) = exp(-b1_param * rpi * 0.5_r8 -0.31_r8  ) * 1.e-3_r8
                 else
-                    z0mg(c) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
+                    z0mg(c) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) -0.31_r8  ) * 1.e-3_r8
```
```
+++ b/src/biogeophys/LakeFluxesMod.F90
@@ -351,9 +351,9 @@ contains
          else                          ! use roughness over snow as in Biogeophysics1
             if(use_z0m_snowmelt) then
                if ( snomelt_accum(c) < 1.e-5_r8 ) then
-                  z0mg(p) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8
+                  z0mg(p) = exp(-b1_param * rpi * 0.5_r8 -0.31_r8  ) * 1.e-3_r8
                else
-                  z0mg(p) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
+                  z0mg(p) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) -0.31_r8  ) * 1.e-3_r8
                end if
             else
                z0mg(p) = params_inst%zsno
@@ -604,9 +604,9 @@ contains
             else ! Snow layers
                if(use_z0m_snowmelt) then
                   if ( snomelt_accum(c) < 1.e-5_r8 )then
-                      z0mg(p) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8 
+                      z0mg(p) = exp(-b1_param * rpi * 0.5_r8 -0.31_r8  ) * 1.e-3_r8 
                   else
-                      z0mg(p) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
+                      z0mg(p) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) -0.31_r8  ) * 1.e-3_r8
```
and had a successful result (showing the FAILs only):
```
20230804_093711_tp00hq: 1 test
    FAIL ERS_Ly5_P144x1.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-cropMonthOutput NLCOMP
    FAIL ERS_Ly5_P144x1.f10_f10_mg37.IHistClm51BgcCrop.cheyenne_intel.clm-cropMonthOutput BASELINE /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev131: FIELDLIST field lists differ (otherwise bit-for-bit)
```

ekluzek: This has been verified so resolving.

cime_config/testdefs/ExpectedTestFails.xml
ekluzek: Awesome, that the test is working. So resolving.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: Yep, we will make Meier2022 the default at a later answer changing tag.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: That's certainly roundoff, especially with a 5-year case. So resolving.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: The issue is #2071 
So resolving.

src/main/lnd2glcMod.F90
ekluzek: This needs to be worked on, we need to make sure the if statement is correct. This might take some time to make sure it's the right statement.

Although I wonder if just reverting to

`if (fields_assigned(g,n) )then`

is what should be done.

slevis-lmwg: Yes, seemed most logical to revert this back to what we have in main. I have already done so in this commit:
https://github.com/ESCOMP/CTSM/pull/2045/commits/548e8be5d32dd6dc1d0a9db3ff9a8531c20b7935

slevis-lmwg:
I have updated to dev134 and submitted test-suites.

Izumi IN PROGRESS
Cheyenne ...waiting in the queue

src/main/lnd2glcMod.F90
ekluzek: Oh, awesome. When I brought up the file in github, it must have showed me an earlier version. But, looking at it on Cheyenne I see that this was reverted. Thanks for doing that.

src/biogeophys/HumanIndexMod.F90
```diff
@@ -1014,6 +1014,15 @@ subroutine Wet_BulbS (Tc_6,rh,wbt)
 ! !LOCAL VARIABLES:
 !EOP
 !
+#ifndef NDEBUG
```
ekluzek, line 1017: Do we really want this #ifdef here?

ekluzek (changes requested):
OK, awesome. I have some questions and things to do, more parameters mostly. Some of this could be made into an issue and done later.

We'll decide on this when we talk it over together.

src/biogeophys/HumanIndexMod.F90
ekluzek: This was on purpose. But, I'm not sure if really should be done.

Was it expensive?

ekluzek: This was on purpose. But, I'm not sure if really should be done (the #if...)

Was it expensive? No not more expensive than the expression below which is expensive, and can blow up for bad values.

And this WetBulb stuff is already known to be expensive, hence we don't turn it on that often.

So final bottom-line remove the #ifndef and the #endif

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -516,26 +557,61 @@ subroutine SetRoughnessLengthsAndForcHeightsNonLake(this, bounds, &
...
-       if (frac_sno(c) > 0._r8) then
-          z0mg(c) = this%zsno
-       else
-          z0mg(c) = this%zlnd
-       end if
+
+       select case (z0param_method)
+       case ('ZengWang2007')
+          if (frac_sno(c) > 0._r8) then
+             if(use_z0m_snowmelt) then
```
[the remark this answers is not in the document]
ekluzek: For code maintenance unless this option is needed for science we'd like to simplify it and not add this option for ZengWang.

@olyson or @swensosc do you have opinions on how useful this would be for running with ZengWang?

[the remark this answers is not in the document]
ekluzek: If we simplify though we should change the buildnml so that you can NOT use this option with ZengWang.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -214,6 +214,8 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <zetamaxstable phys="clm4_5"                                   >2.0d00</zetamaxstable>
 <zetamaxstable phys="clm5_0"                                   >0.5d00</zetamaxstable>
 <zetamaxstable phys="clm5_1"                                   >0.5d00</zetamaxstable>
```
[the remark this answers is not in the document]
ekluzek: We should make sure z0param_method HAS to be selected in the buildnml.

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -337,20 +349,42 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
...
             qstar = temp2(p)*dqh(p)
-            z0hg_patch(p) = z0mg_patch(p) / exp(params_inst%a_coef * (ustar(p) * z0mg_patch(p) / 1.5e-5_r8)**params_inst%a_exp)
+
+            select case (z0param_method)
+            case ('ZengWang2007')
+               z0hg_patch(p) = z0mg_patch(p) / exp(params_inst%a_coef * (ustar(p) * z0mg_patch(p) / nu_param)**params_inst%a_exp)
+            case ('Meier2022')
+
+               ! After Yang et al. (2008)
+               z0hg_patch(p) = 70._r8 * nu_param / ustar(p) * exp( -beta_param * ustar(p)**(0.5_r8) * (abs(tstar))**(0.25_r8))
```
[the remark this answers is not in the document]
ekluzek: We think we should leave the **0.5 and **0.25 should stay as they are. The **0.25 should probably stay that way as that's better than sqrt(sqrt(...)), so the other one should as well.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -865,13 +884,53 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+            lt = max(0.00001_r8,elai(p)+esai(p))
+            displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(cd1_param * lt)**0.5_r8)) / (cd1_param*lt)**0.5_r8)
+
+            lt = min(lt,z0v_LAImax(patch%itype(p)))
+            delt = 2._r8
+            U_ustar_ini = (z0v_Cs(patch%itype(p)) + z0v_Cr(patch%itype(p)) * lt * 0.5_r8)**(-0.5_r8) &
+                      *z0v_c(patch%itype(p)) * lt * 0.25_r8
+            U_ustar = U_ustar_ini
+
+            do while (delt > 0.0001_r8)
```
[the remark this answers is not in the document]
ekluzek: Since, they aren't repeated we think we'll leave them alone.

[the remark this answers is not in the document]
ekluzek: This expression is confusing though. For example the **(-0.5) which means 1/sqrt(...).

We think adding a comment about this would be helpful.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -865,13 +884,53 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+         select case (z0param_method)
+         case ('ZengWang2007')
+            lt = min(elai(p)+esai(p), tlsai_crit)
+            egvf =(1._r8 - alpha_aero * exp(-lt)) / (1._r8 - alpha_aero * exp(-tlsai_crit))
+            displa(p) = egvf * displa(p)
+            z0mv(p)   = exp(egvf * log(z0mv(p)) + (1._r8 - egvf) * log(z0mg(c)))
+
+         case ('Meier2022')
+            lt = max(0.00001_r8,elai(p)+esai(p))
+            displa(p) = htop(p) * (1._r8 - (1._r8 - exp(-(cd1_param * lt)**0.5_r8)) / (cd1_param*lt)**0.5_r8)
```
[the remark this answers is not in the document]
ekluzek: Let's not do this. But, possibly add some comments about it....

src/biogeophys/FrictionVelocityMod.F90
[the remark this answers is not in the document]
olyson: I'm not aware of any need for this. Might be better to simplify.

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -337,20 +349,42 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
...
+
```
ekluzek, line 360: Let's remove this commented out code. We can't maintain it when commented out. And without scientific effort we can't add it, as we don't know if it really offers scientific value.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -1007,7 +1066,12 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
             ! changed by K.Sakaguchi from here
             ! transfer coefficient over bare soil is changed to a local variable
             ! just for readability of the code (from line 680)
-            csoilb = vkc / (params_inst%a_coef * (z0mg(c) * uaf(p) / 1.5e-5_r8)**params_inst%a_exp)
+            ! RM: Does this need to be updated if Ya08 is used too? Proposed formulation (definitely double-check!)
```
ekluzek, line 1069: Remove this commented out alternative, that we can't maintain.

src/main/clm_varctl.F90
```diff
@@ -245,6 +245,14 @@ module clm_varctl
   ! this error-check.
   logical, public :: for_testing_allow_interp_non_ciso_to_ciso = .false.
 
+  !----------------------------------------------------------
+  !  Surface roughness parameterization
+  !----------------------------------------------------------
+
+  character(len=64), public :: z0param_method
```
ekluzek, line 252: This should have a comment...

ekluzek:
This has more work to be done on it, so won't be done immediately. But, we'll schedule it for next week....

ekluzek:
When I do a diff I see a lot of red for white-space at the ends of lines. It would be good to remove that for the new code coming in. The red makes me hate that white-space at the end.

slevis-lmwg:
Addressed most suggestions and now running two preliminary tests before submitting test-suites:
```
./create_test ERS_Ly3.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-Meier2022_surf_rough
./create_test ERP_P72x2_Lm25.f10_f10_mg37.I2000Clm51BgcCrop.cheyenne_intel.clm-monthly -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev134
```

slevis-lmwg:
Test-suites IN PROGRESS

src/biogeophys/HumanIndexMod.F90
slevis-lmwg: Gone!

src/biogeophys/BareGroundFluxesMod.F90
slevis-lmwg: Removed

src/biogeophys/CanopyFluxesMod.F90
slevis-lmwg: Removed

src/main/clm_varctl.F90
slevis-lmwg: Added

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -337,20 +349,42 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
...
             z0qg_patch(p) = z0hg_patch(p)
+            ! Update the forcing heights for new roughness lengths
+            ! TODO(KWO, 2022-03-15) Only for Meier2022 for now to maintain bfb with ZengWang2007
+            if (z0param_method == 'Meier2022') then
+               forc_hgt_u_patch(p) = forc_hgt_u_patch(g) + z0mg_patch(p) + displa(p)
+               forc_hgt_t_patch(p) = forc_hgt_t_patch(g) + z0hg_patch(p) + displa(p)
+               forc_hgt_q_patch(p) = forc_hgt_q_patch(g) + z0qg_patch(p) + displa(p)
+            end if
             thvstar = tstar*(1._r8+0.61_r8*forc_q(c)) + 0.61_r8*forc_th(c)*qstar
```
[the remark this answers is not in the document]
slevis-lmwg: #2118

src/biogeophys/CanopyFluxesMod.F90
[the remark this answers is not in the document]
slevis-lmwg: Addressed with comments

src/biogeophys/CanopyFluxesMod.F90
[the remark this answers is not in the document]
slevis-lmwg: Replaced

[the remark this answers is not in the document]
slevis-lmwg: Hmm, I didn't realize that they were different, and made them both 1.e-4_r8. I should probably back this change out...

src/biogeophys/FrictionVelocityMod.F90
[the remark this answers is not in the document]
slevis-lmwg: Removed

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -516,26 +557,61 @@ subroutine SetRoughnessLengthsAndForcHeightsNonLake(this, bounds, &
...
+                if ( snomelt_accum(c) < 1.e-5_r8 )then
+                    z0mg(c) = exp(-b1_param * rpi * 0.5_r8 + b4_param) * 1.e-3_r8
+                else
+                    z0mg(c) = exp(b1_param * (atan((log10(snomelt_accum(c))+0.23_r8)/0.08_r8)) + b4_param) * 1.e-3_r8
```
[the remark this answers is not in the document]
slevis-lmwg: Replaced

apcraig:
I'd like to ask a question about this PR.

I am adding CTSM to RASM (CESM1.2).  I'm currently doing so with release-clm5.0.35, which I think is the last release from several months ago.  Anyway, we need to couple the (logz0) to WRF.  Several years I did the same with an earlier version of CTSM (for the NNA project, still ongoing), but I'm trying to shift to a more recent and robust release.  In the older version, there was a subroutine called SetActualRoughnessLengths which computed z0's from a call in clm_drv.  In release-clm5.0.35, that subroutine is gone and it looks like z0's are only computed in Fates, if Fates is on (just from looking at the code).  Is that correct?

In our test run, use_fates=.false. and it looks like z0's are never computed.  Several questions.
- Does this PR and recent changes address this shortcoming?  Are z0's now computed when fates is off?
- Should I be running with fates on?
- More generally, would you recommend another version more recent than release-clm5.0.35 for us to use?  I prefer to use a more robust version than some recent hash in master, but there are trade-offs.

Thanks.

slevis-lmwg:
Hi @apcraig ,

My interpretation of this PR's mods:
This PR modified how z0 terms were calculated (eg see description of changes in [the ChangeLog ](doc/ChangeLog)). So z0 terms were already being calculated a different way before this PR (eg see changes in [BaregroundFluxes](src/biogeophys/BareGroundFluxesMod.F90), [BiogeophysPreFluxCalcs](src/biogeophys/BiogeophysPreFluxCalcsMod.F90) and elsewhere). Looking at this PR's mods may help you more easily find where z0 is calculated. This PR does not turn on the new z0 calculations by default, so it keeps answers unchanged from the last baseline.

You should have access to z0 calculations with or without Fates. I do not know why Fates does its own calculation of z0. Maybe it was easier (or better) than using the clm calculation.

As you suggest, it may not make sense to work with this most recent version UNLESS you wish to use the new z0 calculations (z0param_method = "Meier2022").

apcraig:
Thanks @slevis-lmwg.  OK, I'll have a look at this version.  I think an important point is that in release-clm5.0.35, z0 is only calculated by fates.  At some point since that version, it sounds like this was updated and z0 is calculated with or without fates.  It sounds like I don't need the latest version, but I need some version since release-clm5.0.35.  Can you recommend a particular ctsm tags or hash that is relatively robust and tested that might meet my needs, I assume something newer than release-clm5.0.35 and something older than the current master.  Thanks!

slevis-lmwg:
When I type `git grep z0` in a copy of clm5.0.00 that I have, I see calculations of z0, so I wonder why release-clm5.0.35 is different. I don't feel I have followed the model development closely enough to be able to recommend the right version for your work.

Issue #2041, opened 2023-06-22
LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart FAIL in ctsm5.2


slevis-lmwg:
### Brief summary of bug

This test fails in ctsm5.2
`LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart` 
with
`ERROR in CompareBulkToTracer: tracer does not agree with bulk water`

### General bug information

**CTSM version you are using:** [output of `git describe`]
alpha-ctsm5.2.mksrf.16_ctsm5.1.dev123

**Does this bug cause significantly incorrect results in the model's science?** [Yes / No]
Yes

**Configurations affected:**
See test name above.

### Details of bug

This is the first time that we have tried to run the test suites with ctsm5.2. The failing test uses a new fsurdat in ctsm5.2:
`/glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_esmf/ctsm5.2.0/surfdata_10x15_hist_78pfts_CMIP6_2000_c230517.nc`
...relative to ctsm5.1:
`/glade/p/cesmdata/cseg/inputdata/lnd/clm2/surfdata_map/release-clm5.0.18/surfdata_10x15_hist_78pfts_CMIP6_simyr2000_c190214.nc`

@billsacks 
@ekluzek suggested that you may have quicker insight into this failure than we would.

slevis-lmwg:
Met with @billsacks and @ekluzek and decided the following:
- understand why ctsm5.1 does not trigger same error
- see possible connection with this function/subroutine: UpdateState_RemoveCanfallFromCanopy and also Bill's notes here: UpdateState_RemoveSnowUnloading
- try running _D_ (debug) mode
- apply truncate small values function? Increase error tolerance?

slevis-lmwg:
@billsacks @ekluzek 
I have promising news and am interested in your thoughts:
1) diff between the lnd_in files of the ctsm5.1 and ctsm5.2 tests showed that the latter was setting `convert_ocean_to_land = .false.`
2) In both versions, the test uses clm50 (I2000Clm50BgcCrop).
3) In ctsm5.2 I saw that, if I changed to clm51, I would get `convert_ocean_to_land = .true.`
4) I tried that and the test passed!

billsacks:
Thanks for your work on this and for providing this update @slevis-lmwg !

[...]

PR #2048, opened 2023-07-07
Get rid of tiny snocan (tracer and bulk) that causes test failure

slevis-lmwg:
### Description of changes

Test failure described in issue #2041 goes away when we reset tiny snocan (tracer and bulk) to zero.

### Specific notes

@billsacks @ekluzek 

CTSM Issues Fixed (include github issue #): #2041 

Are answers expected to change (and if so in what way)? Yes, larger than roundoff

Testing performed, if any:
1) From my ctsm5.2 branch this test now passes and diffs from baseline are expected due to new ctsm5.2 fsurdat relative to baseline ctsm5.1 fsurdat...
`./create_test LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev123`
2) From ctsm5.1 main plus the same changes that I pushed to this PR, I get diffs from baseline...
`./create_test LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev129`
Largest NORMALIZED RMS diffs are of order 10-1 for variables ERRH2O*, ERRSEB, ERRSOL

### DIFF
--- a/src/biogeophys/CanopyHydrologyMod.F90
+++ b/src/biogeophys/CanopyHydrologyMod.F90
@@ -200,7 +200,6 @@ contains
      integer  :: i     ! index of water tracer or bulk
      real(r8) :: dtime ! land model time step (sec)
 
-     real(r8) :: snocan_baseline(bounds%begp:bounds%endp)                    ! baseline of snocan variable for use in truncate_small_values function
      real(r8) :: qflx_liq_above_canopy_patch(bounds%begp:bounds%endp)        ! liquid water input above canopy (rain plus irrigation) [mm/s]
      real(r8) :: tracer_qflx_liq_above_canopy_patch(bounds%begp:bounds%endp) ! For one tracer: liquid water input above canopy (rain plus irrigation) [mm/s]
      real(r8) :: forc_snow_patch(bounds%begp:bounds%endp)                    ! atm snow, patch-level [mm/s]
--- a/src/biogeophys/CanopyHydrologyMod.F90
+++ b/src/biogeophys/CanopyHydrologyMod.F90
@@ -33,7 +33,6 @@ module CanopyHydrologyMod
   use WaterTracerUtils        , only : CalcTracerFromBulk
   use ColumnType      , only : col, column_type
   use PatchType       , only : patch, patch_type
-  use NumericsMod     , only : truncate_small_values, rel_epsilon
   !
   ! !PUBLIC TYPES:
   implicit none
@@ -198,7 +197,7 @@ contains
      type(water_type)       , intent(inout) :: water_inst
      !
      ! !LOCAL VARIABLES:
-     integer  :: i, p  ! index of water tracer or bulk, index of patch-level
+     integer  :: i     ! index of water tracer or bulk
      real(r8) :: dtime ! land model time step (sec)
 
      real(r8) :: snocan_baseline(bounds%begp:bounds%endp)                    ! baseline of snocan variable for use in truncate_small_values function
@@ -303,23 +302,6 @@ contains
      ! Update snocan and liqcan based on interception, for bulk water and each tracer
      do i = water_inst%bulk_and_tracers_beg, water_inst%bulk_and_tracers_end
         associate(w => water_inst%bulk_and_tracers(i))
-        ! Get rid of tiny snocan (tracer and bulk)
-        ! Very small snocan (< rel_epsilon**2) will be set to zero
-        ! See NumericsMod for rel_epsilon value
-        snocan_baseline(bounds%begp:bounds%endp) = rel_epsilon  ! this gets multiplied by rel_epsilon in truncate_small_values
-        call truncate_small_values(num_soilp, filter_soilp, begp, endp, &
-           snocan_baseline(begp:endp), b_waterstate_inst%snocan_patch(begp:endp))
-        ! Explanation for why we use this do-loop instead of call
-        ! truncate_small_values for w%waterstate_inst%snocan_patch:
-        ! call truncate_small_values would set w%waterstate_inst%snocan_patch
-        ! to zero only when w%waterstate_inst%snocan_patch < rel_epsilon**2
-        ! while it needs to be zero every time that
-        ! b_waterstate_inst%snocan_patch is zero
-        do p = begp, endp
-           if (b_waterstate_inst%snocan_patch(p) == 0._r8) then
-              w%waterstate_inst%snocan_patch(p) = b_waterstate_inst%snocan_patch(p)
-           end if
-        end do
         call UpdateState_AddInterceptionToCanopy(bounds, num_soilp, filter_soilp, &
              ! Inputs
              dtime                 = dtime, &
--- a/src/biogeophys/CanopyHydrologyMod.F90
+++ b/src/biogeophys/CanopyHydrologyMod.F90
@@ -33,6 +33,7 @@ module CanopyHydrologyMod
   use WaterTracerUtils        , only : CalcTracerFromBulk
   use ColumnType      , only : col, column_type
   use PatchType       , only : patch, patch_type
+  use NumericsMod     , only : truncate_small_values, rel_epsilon
   !
   ! !PUBLIC TYPES:
   implicit none
@@ -197,9 +198,10 @@ contains
      type(water_type)       , intent(inout) :: water_inst
      !
      ! !LOCAL VARIABLES:
-     integer  :: i     ! index of water tracer or bulk
+     integer  :: i, p  ! index of water tracer or bulk, index of patch-level
      real(r8) :: dtime ! land model time step (sec)
 
+     real(r8) :: snocan_baseline(bounds%begp:bounds%endp)                    ! baseline of snocan variable for use in truncate_small_values function
      real(r8) :: qflx_liq_above_canopy_patch(bounds%begp:bounds%endp)        ! liquid water input above canopy (rain plus irrigation) [mm/s]
      real(r8) :: tracer_qflx_liq_above_canopy_patch(bounds%begp:bounds%endp) ! For one tracer: liquid water input above canopy (rain plus irrigation) [mm/s]
      real(r8) :: forc_snow_patch(bounds%begp:bounds%endp)                    ! atm snow, patch-level [mm/s]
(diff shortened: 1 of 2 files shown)


src/biogeophys/CanopyHydrologyMod.F90
```diff
@@ -301,6 +303,23 @@ subroutine CanopyInterceptionAndThroughfall(bounds, &
...
+           snocan_baseline(begp:endp), b_waterstate_inst%snocan_patch(begp:endp))
+        ! Explanation for why we use this do-loop instead of call
+        ! truncate_small_values for w%waterstate_inst%snocan_patch:
+        ! call truncate_small_values would set w%waterstate_inst%snocan_patch
+        ! to zero only when w%waterstate_inst%snocan_patch < rel_epsilon**2
+        ! while it needs to be zero every time that
+        ! b_waterstate_inst%snocan_patch is zero
+        do p = begp, endp
+           if (b_waterstate_inst%snocan_patch(p) == 0._r8) then
+              w%waterstate_inst%snocan_patch(p) = b_waterstate_inst%snocan_patch(p)
```
slevis-lmwg, line 320: Oh, probably better to replace the if-statement with
`w%waterstate_inst%snocan_patch(p) = max(0._r8, b_waterstate_inst%snocan_patch(p))`

Or not?

slevis-lmwg: ...actually, no, because the left-hand-side (tracer) may need to equal bulk * ratio.

billsacks:
Thanks a lot for your work on this, @slevis-lmwg !

I have been trying to do truncations like this based on relative differences from a meaningful starting point rather than absolute differences, because that feels less ad hoc and more tied to considerations of floating point numerical precision. What that means in practice is: rather than truncating to 0 based on some (arbitrary) absolute threshold, instead truncate to 0 if the new state (after a state update) is "very" close to 0 in a relative sense, relative to the state before the update. For example, if the new state is 13 orders of magnitude smaller than the old state, then it feels safe to assume that the new state was supposed to go all the way to 0 - regardless of whether the absolute value of the new state is 1e-100, 1e-10, 1, 1e10, etc.: in any of those cases, a new state being 13 orders of magnitude smaller than the old state indicates that the fluxes brought the new state to approximately 0, given the precision available with floating point arithmetic - and then you don't need to worry about figuring out what a typical magnitude is for each state variable.

To see what this looks like in practice, look for the other uses of truncate_small_values in the code. I have put these truncations right after state updates: first saving the state prior to the state update, then comparing the post-update state with the pre-update state. A benefit of doing it this way is that you can rely on the same truncation for tracers as for bulk, because with this relative comparison, they should either both be truncated or both not be truncated.

(I guess in principle we could hit a situation where one of tracer or bulk just barely met the threshold for truncation whereas the other just barely did not. I think my assumption has been that if the bulk meets the threshold then it does so by enough that the tracers will as well, and vice versa, but that might rarely get violated and maybe this implementation needs some rethinking.)

slevis-lmwg:
Pls ignore latest commit for now.

slevis-lmwg:
Latest update passes, while accomplishing @billsacks recommendation I think.
Largest normalized diffs from baseline are greater than before:
```
 RMS FH2OSFC                          1.7754E-02            NORMALIZED  3.1800E+01
 RMS H2OSFC                           5.1145E-02            NORMALIZED  3.1783E+01
 RMS H2OSFC_H2OTR                     5.1145E-02            NORMALIZED  3.1783E+01
 RMS H2OSFC_TESTBIG                   5.1145E-01            NORMALIZED  3.1783E+01
 RMS H2OSFC_TESTMED                   5.1145E-03            NORMALIZED  3.1783E+01
 RMS H2OSFC_TESTSMALL                 5.1145E-12            NORMALIZED  3.1783E+01
 RMS HEAT_FROM_AC                     9.2309E-07            NORMALIZED  1.2312E+01
 RMS QDRAI_PERCH                      2.8024E-13            NORMALIZED  1.7616E+01
 RMS QDRAI_PERCH_H2OTR                2.8024E-13            NORMALIZED  1.7616E+01
 RMS QDRAI_PERCH_TESTBIG              2.8024E-12            NORMALIZED  1.7616E+01
 RMS QDRAI_PERCH_TESTMED              2.8024E-14            NORMALIZED  1.7616E+01
 RMS QDRAI_PERCH_TESTSMALL            2.8024E-23            NORMALIZED  1.7616E+01
 RMS QH2OSFC                          6.3497E-08            NORMALIZED  3.1790E+01
 RMS URBAN_AC                         9.2349E-07            NORMALIZED  1.2268E+01
```

slevis-lmwg:
Next I want to find the smallest my_rel_epsilon for which the test passes, in order to reduce diffs from baseline.

billsacks:
Thanks for your continued and careful work on this, @slevis-lmwg . A few thoughts:
- This new implementation looks more like what I would expect. Typically I have put the truncation around just a single state update (or maybe a bit more than that), and that feels ideal to me in order to know exactly which calculation(s) we're doing the truncation on, but the way you've done it here seems okay.
- This does seem like larger-than-expected changes due to truncating small values... I'd be interested in a little more investigation to convince ourselves that this is working as intended. **Oh, I just saw your edit (after writing most of the rest of this post). I'll keep the rest of this post, but your edit to your last comment maybe explains the big diffs.**
- That said, I don't think that using a smaller value of my_rel_epsilon is necessarily the way to go here. The default relative value (1e-13) is about 3 orders of magnitude greater than machine epsilon (1e-16)... I could see going down to something like 1e-14, but anything smaller than that feels like we're going to be fine-tuning it for one specific case but that this could cause problems in other situations by being too small.
- Instead, I think it may be worth trying to understand why truncation of tiny snocan values can lead to significant differences in h2osfc (and other variables). This certainly seems unintuitive to me. Is this a bug, or will we discover that this is expected behavior and we can sign off on these answer changes?
- Finally, given the answer changes in this PR - at least, assuming they end up being greater than roundoff once you use the appropriate baselines - I wonder if it would be worth splitting this out into a separate PR into master to keep this separate from the other 5.2 work

slevis-lmwg:
Thank you for your continued feedback @billsacks, and I'm glad you noticed the update to my earlier post. 

I have now repeated the ctsm5.1 test with the code changes that I pushed to this PR (i.e. same as test2 described in the PR's initial post):
`./create_test LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart -c /glade/p/cgd/tss/ctsm_baselines/ctsm5.1.dev129`

I found max NORMALIZED diffs in clm.h0 (h1 and cpl diffs were smaller):
```
 RMS ERRH2O                           1.3353E-15            NORMALIZED  5.9706E-01
 RMS ERRH2OSNO                        2.8949E-17            NORMALIZED  3.2638E-01
 RMS ERRH2O_H2OTR                     1.6152E-15            NORMALIZED  7.4016E-01
 RMS ERRH2O_TESTBIG                   1.6152E-14            NORMALIZED  7.4016E-01
 RMS ERRH2O_TESTMED                   1.6152E-16            NORMALIZED  7.4016E-01
 RMS ERRH2O_TESTSMALL                 1.6152E-25            NORMALIZED  7.4016E-01
 RMS ERRSEB                           1.4818E-15            NORMALIZED  1.9952E-01
 RMS ERRSOL                           4.1268E-16            NORMALIZED  1.5098E-01
```
As expected, these are smaller than the max NORMALIZED ctsm5.2 diffs in clm.h0:
1) Do you find these acceptable or would you recommend deeper investigation?
2) I like your suggestion of bringing this in via ctsm5.1 rather than ctsm5.2.

billsacks:
I typically ignore these ERR terms when looking for the biggest RMS differences, because these ERR terms are just indications of balance errors, and hence are differences between two similar quantities, so it's common for the diff from baseline in these ERR terms to be large even for just roundoff-level changes.

I checked what I assume is the correct case (/glade/scratch/slevis/LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart.C.20230710_160106_v2tn7v) for other diffs, and for me at least, they fall into the uncomfortable territory of being fairly small but not obviously just roundoff-level. But given the changes here, it's hard to see how they could lead to changes that are anything other than fundamentally roundoff-level (i.e., truncating a roundoff-level-different-from-zero term to exactly zero), so I'm comfortable with this – though (as for other changes) it should be noted in the list of answer-changing tags that are eventually investigated to ensure there aren't any significant changes that crept in.

One thing you could change that would increase my confidence a bit would be to put the setting of snocan_baseline(p) around line 1528 (the comment, "Update dew accumulation (kg/m2)"): this would put the setting of the baseline immediately before what looks to be the relevant state update, so there's no question that it's accidentally surrounding some other changes in snocan that are happening elsewhere.

slevis-lmwg:
> One thing you could change that would increase my confidence a bit would be to put the setting of snocan_baseline(p) around line 1528 (the comment, "Update dew accumulation (kg/m2)"): this would put the setting of the baseline immediately before what looks to be the relevant state update, so there's no question that it's accidentally surrounding some other changes in snocan that are happening elsewhere.

I made the proposed change (placed it in line 1526, just before the comment "Update dew acc...") instead of in line 1377. The result is bfb identical.

billsacks:
> I made the proposed change (placed it in line 1526, just before the comment "Update dew acc...") instead of in line 1377. The result is bfb identical.

Awesome, thanks for confirming that @slevis-lmwg . My vote would be to change the final code to that version.

slevis-lmwg:
I wanted to confirm next steps:
- Open new PR to merge this change into ctsm5.1 main.
- Run clm diagnostic pkg, as we do for "significant" changes to model answers.
- Is there a reason to not merge this PR into the ctsm5.2 branch, as well?

Am I forgetting anything?

billsacks (approved):

billsacks:
Awesome, looks good now!

> I wanted to confirm next steps:
> 
> * Open new PR to merge this change into ctsm5.1 main.

Yes. (As far as I'm concerned you can do it without a PR since we've already gone over the changes here, but you could open a PR and link back to this one.)

> * Run clm diagnostic pkg, as we do for "significant" changes to model answers.

Yes, although we often have been batching together multiple changes for efficiency, to avoid needing to run and analyze the results on every answer-changing tag. I think that's up to you.

> * Is there a reason to not merge this PR into the ctsm5.2 branch, as well?

I think the sanest workflow is to merge this into ctsm5.1 (i.e., master), then merge the latest ctsm5.1 tag into the ctsm5.2 branch: doing a separate merge of this PR into ctsm5.2 will (in my mind) make it harder to track what is and isn't in ctsm5.2 compared with master. But either way could work.

PR #2053, opened 2023-07-12
Change small snocan to zero

### LINKED ISSUE #2041: LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart FAIL in ctsm5.2

slevis-lmwg:
### Description of changes

#2041 and #2048 discuss and resolve a test failure in the ctsm5.2 branch. The failure goes away when we reset small snocan to zero.

@billsacks recommended merging this change in ctsm5.1 and then updating the ctsm5.2 branch to the latest ctsm5.1.

### Specific notes

@billsacks @ekluzek 

Fixes #2041 #2048 

Yes, more than round off.
Should not change climate.
I will update the Answer-changing tags wikipage as [recommended below](https://github.com/ESCOMP/CTSM/pull/2053#issuecomment-1634112613) and will follow up with a test simulation.

No.

Testing performed, if any:
The test that was failing in ctsm5.2
`LWISO_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.cheyenne_gnu.clm-coldStart`
now passes.


billsacks (approved):

billsacks:
Thanks @slevis-lmwg ! I would probably not call this "significant" myself. In fact, I'd probably say that the answer changes from this change are *expected* to be roundoff-level, but then (like many initially-roundoff-level answer changes) grow over time. It's true that the end result is showing answer changes that are perhaps somewhat bigger than I'd expect due to initially roundoff-level answer changes, but fundamentally I would still consider this to be a roundoff-level change, since it just truncates roundoff-level-greater-than-zero states (which probably were intended to be zero) to exactly zero. Reasons this may be greater-than-roundoff-level are (1) if there is some issue with this implementation that you and I aren't seeing; or (2) if there is some significant behavior change that depends on whether snocan is exactly zero vs. non-zero. But I might still argue that (2) doesn't change this from being fundamentally a roundoff-level change.

These thoughts don't really impact the path forward, except for how you document this change. Of course, feel free to document it how you want to, but I raise these points if you wants to stay consistent with how I've classified changes in the past.

slevis-lmwg:
@billsacks thank you, this helps. I do prefer to stay consistent with how you have classified changes before.

slevis-lmwg:
Test-suites:
Izumi `./run_sys_tests -s aux_clm -c ctsm5.1.dev131 -g ctsm5.1.dev136.slevis` IN PROGRESS

slevis-lmwg:
I'm running the test-suites in prep. for the merge.

slevis-lmwg:
Test-suites
Cheyenne: roundoff diffs expected
Izumi: roundoff diffs expected

slevis-lmwg:
Follow-up PR to this one:
#2457

PR #2072, opened 2023-07-25
Add ability to remove crop residues

samsrabin:
### Description of changes
This PR introduces the ability to remove crop stem and leaf biomass left over after harvest to a 1-year product pool rather than leaving them as litter. This is done by setting `crop_residue_removal_frac`, range [0, 1], to something greater than the default 0.

Based on work performed by Michael Graham, although reworked from the ground up to behave similarly to harvested biofuels.

### Specific notes

Contributors other than yourself, if any: @mwgraham, @danicalombardozzi 

CTSM Issues Fixed: **None**

Are answers expected to change (and if so in what way)? **Not with residue removal disabled.**

Any User Interface Changes (namelist or namelist defaults changes)? **Adds namelist parameter `crop_residue_removal_frac`.**

Testing performed, if any: **None so far.**


samsrabin:
Closing this, to be replaced with separate PRs for answer-changing and non-answer-changing parts.

PR #2074, opened 2023-07-26
New outputs: N to food and seed

samsrabin:
### Description of changes
In response to a user request for `GRAINN_TO_FOOD` outputs, this adds `*_N_TO_FOOD` and `*_N_TO_SEED` outputs for reproductive N pools. These are off by default, unlike their C counterparts.

### Specific notes

Testing performed, if any: A simple 5-day test gave the following for the gridcell average of the last day:
screenshot_6117
Spatial patterns match, but it's weird that the C:N ratio is so high—on the order of 100:1. This could just be a result of looking in what might be the middle of the growing season, before N retranslocation is complete.


samsrabin:
- [ ] Should probably do `aux_clm` testing to see whether any diffs are introduced. There shouldn't be any, but one never knows.
- [ ] Update master_list_nofates.rst (wait on #2083)

samsrabin:
Will be unblocked by #2087.

samsrabin:
`aux_clm` OK; baselines saved as `grainN_to_food.a26a5d1b2`.

samsrabin:
Tasks from co-review with @slevis-lmwg:
- [x] ~~Determine why I added `get_repr_hist_fname` to `CNVegNitrogenFluxType%InitHistory()`; remove if unneeded.~~ Needed to get `_N_TO_FOOD` variable names depending on which reproductive pools exist.
- [ ] Add `GRAINN_TO_FOOD` to first history file in `crop` testmod.
- [ ] Do a longer test, examining at-harvest C:N ratios of both `TO_FOOD` and `TO_SEED`.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -3106,8 +3111,15 @@ subroutine CNOffsetLitterfall (num_soilp, filter_soilp, &
                          repr_grainc_to_food_thisyr(p,k) = repr_grainc_to_food_thisyr(p,k) &
                              + repr_grainc_to_food_perharv(p,h,k)
                      end if
+                     repr_grainn_to_food_thispool = npool_to_reproductiven(p,k) - repr_grainn_to_seed(p,k)
                      repr_grainn_to_food(p,k) = t1 * reproductiven(p,k) &
                           + npool_to_reproductiven(p,k) - repr_grainn_to_seed(p,k)
```
slevis-lmwg, line 3116: I asked @samsrabin to remind me the diff between "grain to seed" and "grain to food" and I see here that the two terms summed are the reproductive pool. I'm not sure "seed" is the right term for what's left when you subtract food from the reproductive pool, but that's out of scope for this PR.

src/biogeochem/CNVegNitrogenFluxType.F90
```diff
@@ -14,7 +15,7 @@ module CNVegNitrogenFluxType
   use SoilBiogeochemDecompCascadeConType , only : decomp_cascade_con
   use dynSubgridControlMod               , only : get_do_grossunrep
   use CropReprPoolsMod                   , only : nrepr, repr_grain_min, repr_grain_max, repr_structure_min, repr_structure_max
-  use CropReprPoolsMod                   , only : get_repr_rest_fname, get_repr_longname
+  use CropReprPoolsMod                   , only : get_repr_hist_fname, get_repr_rest_fname, get_repr_longname
```
slevis-lmwg, line 18: @samsrabin you do seem to use this below, so you may need to keep here.

src/biogeochem/CNVegNitrogenFluxType.F90
```diff
@@ -1034,6 +1039,50 @@ subroutine InitHistory(this, bounds)
...
+               long_name=get_repr_longname(k)//' N to seed', &
+               ptr_patch=data1dptr, &
+               default='inactive')
+          data2dptr => this%repr_grainn_to_food_perharv_patch(:,:,k)
+          call hist_addfld2d ( &
+               ! e.g., GRAINN_TO_FOOD_PERHARV
+               fname=get_repr_hist_fname(k)//'N_TO_FOOD_PERHARV', &
+               units='gN/m^2', &
+               type2d='mxharvests', &
+               avgflag='I', &
```
slevis-lmwg, line 1072: I mentioned PR #2019 which, as written, eliminates 'I' hardwiring in the code; however, we are waiting for a different coding solution from CAM, so no need to worry about that right now.

slevis-lmwg (approved):

samsrabin:
Hmm. I'm concerned that the way CLM does C and N allocation to seed vs. grain makes it not fit for purpose with regard to harvested (food) N content. I don't exactly understand all the machinations, but the result is plain enough: At harvest, to-food N is sometimes zero while C is positive. This is not the case for to-seed N and C content (or, of course, seed + food).

Moreover, even if we combine both seed and food to get all grain biomass, C:N looks off. Soybeans typically have C:N ratio of < 10, but from a 5-year run, I'm not seeing mean at-harvest grain C:N below 45:
image

I still think it's worth adding these outputs, as without them I don't think there's a way to calculate aboveground crop N biomass. But perhaps some metadata of GRAINN_TO_FOOD should disclaim that it's not necessarily realistic.

Tagging @billsacks in case you have any thoughts, as I see you worked on the seed allocation and deficit code, albeit back in 2017.

samsrabin:
aux_clm OK as of af4a688.

samsrabin:
@billsacks Thanks for that context—it's so great that you have notes going back that far!

As far as I can tell, everything that happens to `repr_grainn_to_food_patch` happens to the C equivalent and vice versa: e.g., "Send the remaining grain to the food product pool", the update of `reproductive[cn]_patch` in `[CN]StateUpdate1()`. I think the difference must come in how the seed deficit is calculated. But again, things seem to be the same mechanically—although I haven't taken as close a look.

Does it all derive from `leafcn`, then? That seems to be the driver of seed deficit via `seed[cn]_to_leaf`.

billsacks:
Interesting, thanks for your analysis. I don't remember enough about this to weigh in further, but let me know if you want to talk about it.

samsrabin:
@slevis-lmwg This is ready for re-review (commits beginning b95b986).

slevis-lmwg (approved):
I looked over the latest changes and have no concerns.

Issue #942, opened 2020-03-12
"use_cn" is used inconsistently in soilbiogechem for FATES than in the rest of the model


ekluzek:
For most of the model the logical namelist variable "use_cn" is used as a flag to say that the above ground Carbon-Nitrogen model is turned on. However, when FATES is on use_cn=.false. is used in the soil biogeochemistry to signal that Nitrogen isn't in use.

Here's a grep of the code for soil-BGC:

``` fortran
SoilBiogeochemLittVertTranspMod.F90:  use clm_varctl                         , only : iulog, use_c13, use_c14, spinup_state, use_vertsoilc, use_fates, use_cn
SoilBiogeochemLittVertTranspMod.F90:            if (use_cn ) then
SoilBiogeochemPrecisionControlMod.F90:    use clm_varctl , only : iulog, use_c13, use_c14, use_nitrif_denitrif, use_cn
SoilBiogeochemPrecisionControlMod.F90:                  if (use_cn) then
SoilBiogeochemPrecisionControlMod.F90:            if (use_cn) then
SoilBiogeochemStateType.F90:  use clm_varctl     , only : use_vertsoilc, use_cn 
SoilBiogeochemStateType.F90:    if (use_cn) then
```

The thin that's done in LittVertTransfer is...

``` fortran
      !------ loop over litter/som types
      do i_type = 1, ntype

         select case (i_type)
         case (1)  ! C
            conc_ptr          => soilbiogeochem_carbonstate_inst%decomp_cpools_vr_col
            source            => soilbiogeochem_carbonflux_inst%decomp_cpools_sourcesink_col
            trcr_tendency_ptr => soilbiogeochem_carbonflux_inst%decomp_cpools_transport_tendency_col
         case (2)  ! N
            if (use_cn ) then
               conc_ptr          => soilbiogeochem_nitrogenstate_inst%decomp_npools_vr_col
               source            => soilbiogeochem_nitrogenflux_inst%decomp_npools_sourcesink_col
               trcr_tendency_ptr => soilbiogeochem_nitrogenflux_inst%decomp_npools_transport_tendency_col
            endif
```

So it's basically used to tell soil-BGC that FATES isn't using Nitrogen.

This is first of all confusing to use it for two purposes like this. Second, FATES is going to be adding nutrients in via PARTEH and one of the first nutrients will be Nitrogen. So there should be a different flag that can be used to tell soil-BGC that Nitrogen is on or off.

[...]

PR #2076, opened 2023-07-27
Add FATES land use change module to pass LUH2 data to FATES

### LINKED ISSUE #1077: Read in full LUH2 dataset for use by FATES
FATES is going to use the area option for harvest variables from a streams file rather than the mass option (on the landuse.timeseries files). Hence, we'll need to support both variables on the new landuse.timeseries files.

It looks like what we really need is to read in the raw LUH2 harvest data for FATES as a stream file. In the landuse.timeseries files harvest is per PFT, but what FATES needs is the total, rather than resolved by PFT.

Definition of Done: CTSM/FATES reads in full LUH2 data and uses it in FATES

glemieux:
### Description of changes
This pull request adds a new module, the structure of which is adopted from the dyn_subgrid code, to enable hlms to ingest raw luh2 data to be passed to FATES. A new namelist variable is introduced use_fates_luh, to engage the module functionality, which is off by default. The luh2 dataset to be ingested is a minimally processed concatenation of the luh2 state, transition and management datasets.

The luh2 dataset is generated by the fates data tool found in https://github.com/NGEET/fates/pull/1032.  This pull request is associated with https://github.com/NGEET/fates/pull/1040.

### Specific notes

  Fixes #1077

Are answers expected to change (and if so in what way)? No change expected for non-luh2 run modes

Any User Interface Changes (namelist or namelist defaults changes)?  Yes, this adds a new set of namelist variables for the luh2 data file and 

Testing performed, if any: Will do standard


bld/CLMBuildNamelist.pm
```diff
@@ -4302,21 +4302,33 @@ sub setup_logic_fates {
...
               } elsif ( ! -f "$fname" ) {
                  $log->fatal_error("$fname does NOT point to a valid filename" );
               }
            }
         }
+        my $var = "use_fates_luh";
+        if ( defined($nl->get_value($var))  ) {
+           if ( &value_is_true($nl->get_value($var)) ) {
+              $var = "fluh_timeseries";
+              my $fname = substr $nl->get_value($var), 1, -1;  # ignore first and last positions of string because those are quote characters
```
ekluzek, line 4324: Use the remove_leading_and_trailing_quotes function to do this.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -718,6 +718,23 @@ Full pathname to the inventory initialization control file.
 (Only relevant if FATES is on).
 </entry>
 
+<!-- This may eventually migrate to the dynamic_subgrid group-->
+<entry id="use_fates_luh" type="logical" category="physics"
+       group="elm_inparm" valid_values="" >
```
ekluzek, line 723: Define LUH in here as well as in the filename.

bld/CLMBuildNamelist.pm
```diff
@@ -4302,21 +4302,33 @@ sub setup_logic_fates {
...
+              if ( ! defined($nl->get_value($var))  ) {
+                 $log->fatal_error("$var is required when use_fates_luh is set" );
```
ekluzek, line 4326: We should add a test to the  build-namelist_test.pl for these two error conditions. Probably should add a test that turns it on as well.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -718,6 +718,23 @@ Full pathname to the inventory initialization control file.
...
+If TRUE, enable use of land use state and transition data from luh_timeseries file.
+(Only valid for fates land use change runs, where there is a luh_timeseries file.)
```
ekluzek, line 725: I think I'd remove this middle sentence here. I see the dependence the other way -- if this is set to true, then you have to have a fluh_timeseries file. So there isn't a need for this one.

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
```diff
@@ -0,0 +1,194 @@
...
+  use dynFileMod            , only : dyn_file_type
+  use dynVarTimeUninterpMod , only : dyn_var_time_uninterp_type
+  use clm_varcon            , only : grlnd
+  use clm_varctl            , only : iulog
+
+  implicit none
+
+  private
+
+  real(r8), allocatable, public :: landuse_transitions(:,:)
```
ekluzek, line 23: comments about what the local variables are about...

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
```diff
@@ -0,0 +1,194 @@
...
+    ! Since fates needs state data during initialization, make sure to call
+    ! the interpolation routine at the start
+    call dynFatesLandUseInterp(bounds,init_state=.true.)
+
+  end subroutine dynFatesLandUseInit
+
+
+  !-----------------------------------------------------------------------
+  subroutine dynFatesLandUseInterp(bounds, init_state)
+
```
ekluzek, line 146: Give a description.

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
```diff
@@ -0,0 +1,194 @@
...
+    use dynTimeInfoMod , only : time_info_type
+    use clm_varctl     , only : use_cn
+
+    ! !ARGUMENTS:
+    type(bounds_type), intent(in) :: bounds       ! proc-level bounds
+    logical, optional, intent(in) :: init_state   ! fates needs state for initialization
+
+    ! !LOCAL VARIABLES:
+    integer                     :: varnum
```
ekluzek, line 155: Description of variables.

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
```diff
@@ -0,0 +1,194 @@
...
+    integer                     :: i
+    logical                     :: init_flag
+    real(r8), allocatable       :: this_data(:)
+    character(len=*), parameter :: subname = 'dynFatesLandUseInterp'
+    !-----------------------------------------------------------------------
+    SHR_ASSERT_ALL(bounds%level == BOUNDS_LEVEL_PROC, subname // ': argument must be PROC-level bounds')
+
+    ! This shouldn't be called by cn currently, but return if it is
+    if (use_cn) return ! Use this as a protection in lieu of build namelist check?
```
ekluzek, line 164: I'd either make this an abort here, or remove it. It's a mistake for this to be called for non-FATES, so either abort here if so, or just remove the check. I'd side with the abort as it is a mistake if called...

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
```diff
@@ -0,0 +1,194 @@
...
+
+    ! Get the current year
+    call dynFatesLandUse_file%time_info%set_current_year()
+
+    if (dynFatesLandUse_file%time_info%is_before_time_series() .and. .not.(init_flag)) then
+       ! Reset the land use transitions to zero for safety
+       landuse_transitions(1:num_landuse_transition_vars,bounds%begg:bounds%endg) = 0._r8
+       landuse_states(1:num_landuse_state_vars,bounds%begg:bounds%endg) = 0._r8
+    else
+       ! Right now we don't account for the topounits
```
ekluzek, line 179: I'm not sure what you mean by topounits?

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
```diff
@@ -0,0 +1,194 @@
...
+  type(dyn_file_type), target   :: dynFatesLandUse_file
+
+  ! Land use name arrays
+  character(len=5), public, parameter  :: landuse_state_varnames(num_landuse_state_vars) = &
+                    [character(len=5)  :: 'primf','primn','secdf','secdn','pastr','range', &
+                                          'urban','c3ann','c4ann','c3per','c4per','c3nfx']
+
+  character(len=14), public, parameter :: landuse_transition_varnames(num_landuse_transition_vars) = &
+                    [character(len=14) :: 'primf_to_secdn','primf_to_pastr','primf_to_range','primf_to_urban', &
+                                          'primf_to_c3ann','primf_to_c4ann','primf_to_c3per','primf_to_c4per','primf_to_c3nfx', &
```
ekluzek, line 38: This list is so long that I wonder if there might be another way to manage it? Maybe in a different file? Or maybe you process all the variables and use the variable list from the file?

But, that might only matter if we think this might change in the future? How likely is anything here to change?

src/main/controlMod.F90
```diff
@@ -1101,6 +1105,8 @@ subroutine control_print ()
        write(iulog, *) '    use_fates_fixed_biogeog = ', use_fates_fixed_biogeog
        write(iulog, *) '    use_fates_nocomp = ', use_fates_nocomp
        write(iulog, *) '    use_fates_sp = ', use_fates_sp
+       write(iulog, *) '    use_fates_luh= ', use_fates_sp
+       write(iulog, *) '    fluh_timeseries = ',fluh_timeseries
```
ekluzek, line 1109: Add a trim around character strings...

src/main/controlMod.F90
```diff
@@ -1101,6 +1105,8 @@ subroutine control_print ()
...
        write(iulog, *) '    fates_inventory_ctrl_filename = ',fates_inventory_ctrl_filename
```
ekluzek, line 1110: This isn't yours but could use a trim function around it.

ekluzek (changes requested):
@glemieux this is an exciting upgrade for CTSM/FATES so it's exciting it's starting to come in.

I have a few requests that I think should be straight forward and easy. Let me know if you want to meet to discuss.

I have one other question. How will the workflow to create these files work? I think we will add that to the CTSM5.2 Makefile right? And it will exercise FATES python scripts to make them right? That's something that we should discuss a bit...

ekluzek:
@ckoven sorry I didn't get this until now. We had this slated as coming in after cross-seed dispersal as it was looking more complete. We might need to think about how to relay general FATES priorities to CTSM to make sure we are in sync. Something we are trying to do in CTSM now is bring things in that are ready to go. A goal with that is to bring things in ASAP when possible, and hopefully will get us bringing in PR's quicker than we have been. But, at the same time some work has a higher priority than something that is already ready to go. So we are going to need to balance between those at times.

ckoven:
Thanks @ekluzek and no worries!

wwieder:
Excited for this to come in. Thanks for working on it @glemieux  and @ckoven .  

After this is merged, it seems like a logical next step to try and transient, historical no-comp baseline?  I wondered if it makes sense for @samsrabin to do this run?  We'll also likely need to start building up capacity to post-process results.  These seem like high FATES priorities for the LMWG and I'm wondering how best to make them a reality?

ckoven:
Thanks @wwieder — in principle I totally agree; in practice I think it might make sense to wait until the second set of land-use-change mods are ready, which I am still working on. Basically the set of changes in this PR provides the infrastructure for land-use change, but won't yet really be scientifically meaningful for a few reasons. The main one of which is that there is in this PR no enforcement of land cover properties on patches with a given land use. I.e. there is nothing that makes pasture lands or croplands behave differently from secondary lands here. In the short term, the way to accomplish this is (1) the linkage of the land-use classes with the nocomp logic, to enforce a different PFT composition on patches with different land use classes, and (2) starting to apply land management to the different land use classes, e.g. grazing on pasture and rangelands.  In the long term, number 2 above can be expanded while number 1 can be relaxed, so that land cover follows more fully as a result of management rather than direct imposition of land cover (except in cases like crops imposition of land cover is a direct land management practice). So once this first land-use PR is in, I have subsequent branches that will start to address both of those points, as well as other things needed to do transient runs such as the capability to spin up the model under potential vegetation, and then transition from that to historical land use. A less-fundamental but still relevant point is that we are working towards initial calibrated version of FATES in a nocomp configuration, but not yet a full-FATES version; so we don't have a parameter set to run through a global full-FATES with land-use transient run, but we are much closer to having an initial nocomp global calibration.  So as part of this second set of land-use modifications, I do intend to do some global transient historical runs in nocomp configuration.

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: This is unlikely to change unless the [LUH](https://luh.umd.edu/) update for LUH3 changes the naming or number (longer term consideration).  Thinking about how to manage these sort of lists generally is not a bad idea though.  Maybe I can make an issue/discussion for it?

bld/CLMBuildNamelist.pm
glemieux: addressed via 3bbbc53590a480a73c541d4454818102a591d255

bld/CLMBuildNamelist.pm
glemieux: adapted the inventory tests via 96b827c90568af10f420530f31b1d765d16c061c

bld/namelist_files/namelist_definition_ctsm.xml
glemieux: acronym definition added via f795147d6d1c61df54e88bd603cf5f818200013a

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: addressed via 87b149f86fe87a0e07ef627d64febb304306fd7a

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: addressed via 87b149f86fe87a0e07ef627d64febb304306fd7a

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: addressed via 87b149f86fe87a0e07ef627d64febb304306fd7a

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: addressed via 87b149f86fe87a0e07ef627d64febb304306fd7a

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: This is a copy/paste holdover from e3sm.  Addressed via 87b149f86fe87a0e07ef627d64febb304306fd7a

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: @ekluzek I have an alternate method of approaching this on a different branch (that I haven't tested yet), but I'm wondering what you think: 8434cbd2f48e6f03dc6c24ebb45f7dd5292a50a8

src/main/controlMod.F90
glemieux: addressed in b06181632f3e7609d6a71fd3b55fa0b4fe5fc800

src/main/controlMod.F90
glemieux: addressed via b06181632f3e7609d6a71fd3b55fa0b4fe5fc800

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
glemieux: Per our discussion today, @ekluzek and I agree to move forward with the long list of parameters as is currently.  I will have a forthcoming refactor of the LUH2 data tool (to be included with https://github.com/NGEET/fates/pull/1107) in which I will explore incorporating the metadata approached discussed here: https://github.com/glemieux/fates/issues/30#issuecomment-1703210904

glemieux:
I should have noted https://github.com/NGEET/fates/pull/1040#issuecomment-1853081300 here instead; after discussing the carbon balance checking with @rgknox, we've realized that the wood product should not be included in the check for fates runs.  As a result this has been removed via 1a1ec94.

glemieux:
@ekluzek Generating fates baselines prior to testing this PR branch with `ctsm5.1.dev159` and comparing against the previous fates baseline based on `ctsm5.1.dev158` is B4B with one exception: the `FatesCold` 13 month `1x1_brazil` smoke test is showing diffs.  What's interesting is the `gnu` version of this test is B4B against the previous tag baseline.  

Also note that for `aux_clm`, while we have testmods that run over a year, we only have reduced complexity tests, which might be the reason the previous tag was coming back B4B.  Maybe we should migrate this particular single site smoke test to `aux_clm` as well?

glemieux:
> @ekluzek Generating fates baselines prior to testing this PR branch with `ctsm5.1.dev159` and comparing against the previous fates baseline based on `ctsm5.1.dev158` is B4B with one exception: the `FatesCold` 13 month `1x1_brazil` smoke test is showing diffs. What's interesting is the `gnu` version of this test is B4B against the previous tag baseline.

@rgknox noted that the issue is that that one test for the previous fates baseline was rebuilt with the wrong fates version so should be ignored.

ekluzek:
Ahh, good I'm glad @rgknox figured it out. I'll redo that baseline test. Sorry about that, that's my bad.

glemieux:
I'm seeing failing RUNs with fates NEON testmods on izumi.  The issue appears to be that the `weight` has the wrong dimensions (one beyond what is expected):
```
Runtime Error: /home/glemieux/ctsm/src/main/surfrdUtilsMod.F90, line 238: Invalid reference to procedure SURFRDUTILSMOD:COLLAPSE_TO_DOMINANT - Dummy array WEIGHT (number 1) has 18 elements but actual argument only has 17 elements
Program terminated by fatal error
/home/glemieux/ctsm/src/main/surfrdUtilsMod.F90, line 238: Error occurred in SURFRDUTILSMOD:COLLAPSE_TO_DOMINANT
/home/glemieux/ctsm/src/main/surfrdMod.F90, line 854: Called by SURFRDMOD:SURFRD_VEG_ALL
/home/glemieux/ctsm/src/main/surfrdMod.F90, line 192: Called by SURFRDMOD:SURFRD_GET_DATA
/home/glemieux/ctsm/src/main/clm_initializeMod.F90, line 247: Called by CLM_INITIALIZEMOD:INITIALIZE2
/home/glemieux/ctsm/src/cpl/nuopc/lnd_comp_nuopc.F90, line 659: Called by LND_COMP_NUOPC:INITIALIZEREALIZE
/home/glemieux/ctsm/components/cmeps/cime_config/../cesm/driver/esmApp.F90, line 128: Called by ESMAPP
```

I'm testing non-fates NEON sites right now via the izumi `aux_clm` run, but looking at recent commits, I'm kind of at a loss for what could have caused these failures.  Note that all other fates testmods are passing.

glemieux:
> I'm testing non-fates NEON sites right now via the izumi `aux_clm` run, but looking at recent commits, I'm kind of at a loss for what could have caused these failures. Note that all other fates izumi testmods are passing and the `gnu` compiler versions of these NEON tests passed on derecho.

I think I know what this is: in NGEET/fates#1040, the number of maximum fates patches per site went up by one (to 17) which sets the `natpft_ub` to 17 for fates (note lower bound is 0).  The NEON surface dataset is read for setting the `surfpft_ub`, which is then used to allocate the pft bounds for `wt_nat_patch`; the data set has 16 pfts.  The issue crops up in that the `collapse_to_dominant` call passes in `wt_nat_patch` and `natpft_ub` to set the input argument bounds and then finds the discrepency.

The simplest fix to align is to update the new fates parameter file coming in with this PR to reduce the number of maximum patches back to 16.  @ckoven does dropping the max number of patches for the primary landunit  by one (to 9) sound appropriate?  

As an aside, @ekluzek should the `collapse_to_dominant` call actually be passing the `surfpft` bounds as input since that's how it's being allocated?

ekluzek:
@glemieux offhand that does sound right to me. Do you want to look at the code together? Or do you want to put together the code changes you think should happen and have us look it over?

glemieux:
@ekluzek the code change would be pretty straightforward, but it'll be easier to walk through it together.  The thing that I want to test out first is what the failure mode would look like when this fix is in.  Let me give that a run and I'll tag up with you via chat later.

ckoven:
> @ckoven does dropping the max number of patches for the primary landunit by one (to 9) sound appropriate

@glemieux yes that makes sense to me. thanks!

glemieux:
Status update: `aux_clm` and `fates` restesting on izumi has been kicked off and is in the queue.  I'll test `aux_clm` (and retest `fates`) on derecho once izumi testing comes back clean.

glemieux:
Regression testing the `aux_clm` suite on izumi is complete.  All expected tests passed B4B with the expected exception of the fates testmods.  This is expected as  the reorganization of the disturbance code will result in small differences showing up in the longer test (like the one Ld30 fates testmod).  All other fates diffs are `FIELDLIST` differences due to known history output name changes.

Izumi folder location: `/home/glemieux/scratch/ctsm-tests/tests_pr2076_aux-clm`

glemieux:
`aux_clm` testing on derecho is complete.  All expected tests pass B4B with the known exception of the fates test mods, which is due to known science updates since the API 31.0 update (including the landuse PR).

derecho location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr2079_aux_clm`

A quick note that I needed to rebuild the `DAE` test to get it to pass `RUN` even though the logs suggest everything ran fine and the `COMPARE_base_rest` passed.  The Teststatus.log had the following which was tripping up the test reporter immediately after the `COMPARE_base_rest` output:

```
 620  ---------------------------------------------------
 621 2023-12-18 17:25:19: ERROR: ERROR: Unrecognized line ('/bin/bash: module: line 1: syntax error: unexpected end of file
 622 ') found in /glade/u/home/glemieux/scratch/ctsm-tests/tests_pr2079_aux_clm/DAE_C2_D_Lh12.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-DA_multidrv.GC.pr2079_aux_clm_int/run/case2run     /da.log.2668444.desched1.231218-172242.gz
 623  ---------------------------------------------------
```

Hat tip to @slevis-lmwg for pointing me to the fact that he was seeing the same thing as note with https://github.com/ESCOMP/CTSM/pull/2253#issuecomment-1852255124 and that a resubmitted fixed his case.

src/dyn_subgrid/dynFATESLandUseChangeMod.F90
ekluzek: Added this to https://github.com/ESCOMP/CTSM/issues/942#issuecomment-1872704298

bld/CLMBuildNamelist.pm
ekluzek: Added this in and committed it to the branch.

ekluzek (approved):
I worked on a couple issues here, filing a couple issues and resolving the conversations that were addressed.

src/main/controlMod.F90
```diff
@@ -780,9 +782,11 @@ subroutine control_spmd()
     call mpi_bcast (use_fates_fixed_biogeog, 1, MPI_LOGICAL, 0, mpicom, ier)
     call mpi_bcast (use_fates_nocomp, 1, MPI_LOGICAL, 0, mpicom, ier)
     call mpi_bcast (use_fates_sp, 1, MPI_LOGICAL, 0, mpicom, ier)
+    call mpi_bcast (use_fates_luh, 1, MPI_LOGICAL, 0, mpicom, ier)
     call mpi_bcast (use_fates_bgc, 1, MPI_LOGICAL, 0, mpicom, ier)
     call mpi_bcast (fates_inventory_ctrl_filename, len(fates_inventory_ctrl_filename), MPI_CHARACTER, 0, mpicom, ier)
     call mpi_bcast (fates_paramfile, len(fates_paramfile) , MPI_CHARACTER, 0, mpicom, ier)
+    call mpi_bcast (fluh_timeseries, len(fates_paramfile) , MPI_CHARACTER, 0, mpicom, ier)
```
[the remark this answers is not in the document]
ekluzek: Issue #2305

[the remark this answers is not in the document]
glemieux: Good catch!  Thanks for making the issue.  Maybe we can get the fix and test in with #2265.

PR #2077, opened 2023-07-27
Add FATES cross-grid seed dispersal mechanism

glemieux:
### Description of changes
This PR enables FATES to disperse seeds across neighboring grid cells using MPI. The API update includes calls to new fates dispersal procedures and as such should be coordinated with https://github.com/NGEET/fates/pull/1005.

The design documentation can be found in the [FATES Developer's Guide](https://fates-users-guide.readthedocs.io/en/latest/developer/elm-fates_seed-dispersal_design_doc.html).

### Specific notes

Contributors other than yourself, if any: @YanlanLiu

Are answers expected to change (and if so in what way)? Answer should not change for non seeddispersal mode

  New fates_seeddisp_cadence item and requires changes to the FATES parameter file

Testing performed, if any: FATES testing, will do aux_clm


glemieux:
@ekluzek just a heads up that from the FATES software meeting today, I'm going to move forward with this as the next FATES api update.

glemieux:
Cross posting that running fates regression tests on cheyenne, I'm seeing all `ERP` tests failing regardless of grid size, debug mode or threading: https://github.com/NGEET/fates/pull/1005#issuecomment-1700310191

As discussed in the ctsm software meeting this morning, this seems likely to be due the MPI_AllReduce not being able to maintain b4b when layouts change.

glemieux:
> Cross posting that running fates regression tests on cheyenne, I'm seeing all `ERP` tests failing when seed dispersal is on regardless of grid size, debug mode or threading: [NGEET/fates#1005 (comment)](https://github.com/NGEET/fates/pull/1005#issuecomment-1700310191)
> 
> As discussed in the ctsm software meeting this morning, this seems likely to be due the MPI_AllReduce not being able to maintain b4b when layouts change.

@billsacks was spot on with this being due to the non-associativity of floating point sums with MPI.  I refactored the code to utilize a non-reducing MPI collective and it appears that the ERP tests are passing now.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -228,8 +237,10 @@ module CLMFatesInterfaceMod
       procedure, public  :: wrap_hydraulics_drive
       procedure, public  :: WrapUpdateFatesRmean
       procedure, public  :: wrap_WoodProducts
+      procedure, public  :: WrapSeedGlobal
+      procedure, public  :: wrap_seed_dispersal
       procedure, public  :: UpdateCLitterFluxes
-      procedure, public  :: UPdateNLitterFluxes
+      procedure, public  :: UpdateNLitterFluxes
```
glemieux, line 243: Note that this is simply a case change for style alignment

glemieux:
After much struggling with the ERP tests, it was determined that there is still an inconsistency between runs that have different pelayouts.  Per discussion at the fates software meeting, it was agreed that I should create an issue noting this and move forward with review, testing, and integration.  The issue is https://github.com/NGEET/fates/issues/1089.  This PR also adds a seed dispersal testmod that is used with a PEM testtype.  This has been identified as an expected failure to aid future tracking.

Per recommendation from @adrifoster, I've refactored the code on both the hlm and fates side to utilize a namelist switch to handle the seed dispersal cadence, which will also allow the user to turn off seed dispersal without modifying the parameter file.

Testing via the `fates` test suite on izumi and cheyenne is complete and b4b in expected cases.   `aux_clm` testing still TBD.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
...
+   endif
+   call t_stopf('fates-seed-mpi_reduce')
+
+ end subroutine WrapSeedGlobal
+
+ ! ======================================================================================
+
+ subroutine wrap_seed_dispersal(this,bounds_clump)
+
+    ! This subroutine pass seed_id_global to bc_in and reset seed_out
```
ekluzek, line 2719: A little hard to understand the comment.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
...
+
+   call t_startf('fates-seed-mpi_reduce')
+
+   if (IsItDispersalTime(setdispersedflag=set_flag)) then
+
+      ! Re-initialize incoming seed buffer for this time step
+      this%fates_seed%incoming_global(:,:) = 0._r8
+      this%fates_seed%outgoing_global(:,:) = 0._r8
+
+      ! Distribute outgoing seed data from all nodes to all nodes
```
ekluzek, line 2685: I think by node you really mean task right?

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -638,6 +638,13 @@ Switch deciding which nutrient model to use in FATES.
 (Only relevant if FATES is on)
 </entry>
 
+<entry id="fates_seeddisp_cadence" type="integer" category="physics"
+        group="clm_inparm" valid_values="0,1,2,3">
+Switch defining the cadence at which seeds are dispersed across
+gridcells.  Setting the switch value to zero turns off dispersal.
+(Only relevant if FATES is on)
```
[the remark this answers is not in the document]
ekluzek: Also add a note that "1" is only for quick testing, but is slow. Is there a scientific need for monthly?

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
...
+      call MPI_Allgatherv(this%fates_seed%outgoing_local, procinfo%ncells*numpft_fates, MPI_REAL8, &
+                          this%fates_seed%outgoing_global, this%fates_seed%ncells_array*numpft_fates, this%fates_seed%begg_array*numpft_fates, &
+                          MPI_REAL8, mpicom, ier)
+      write(iulog,*) 'WSG: MPI_Allgatherv ier: ', ier
```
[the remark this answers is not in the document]
ekluzek: I would add a note that this is inefficient and a performance hog. Each task really only needs some range of its neighbors information, but giving everything to each task is easier to code up. This is OK though for something done very seldom though. 

If I understand this method you don't just need the immediate neighbors, but a range of neighbor grid-cells, which also makes it hard to just send the immediate neighbors.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
...
+    call t_startf('fates-seed-disperse')
+
+    nc = bounds_clump%clump_index
+
+    do s = 1, this%fates(nc)%nsites
+       c = this%f2hmap(nc)%fcolumn(s)
+       g = col%gridcell(c)
+
+       ! Check that it is the beginning of the current dispersal time step
+       if (IsItDispersalTime()) then
```
ekluzek, line 2739: This should be done outside the loop for performance.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
...
+      this%fates_seed%incoming_global(:,:) = 0._r8
+      this%fates_seed%outgoing_global(:,:) = 0._r8
+
+      ! Distribute outgoing seed data from all nodes to all nodes
+      call MPI_Allgatherv(this%fates_seed%outgoing_local, procinfo%ncells*numpft_fates, MPI_REAL8, &
+                          this%fates_seed%outgoing_global, this%fates_seed%ncells_array*numpft_fates, this%fates_seed%begg_array*numpft_fates, &
+                          MPI_REAL8, mpicom, ier)
+      write(iulog,*) 'WSG: MPI_Allgatherv ier: ', ier
+
+      ! zero outgoing local for all gridcells outside threaded region now that we've passed them out
```
ekluzek, line 2691: If calling outside of a threaded region is required should we test if that's the case here?

cime_config/testdefs/testlist_clm.xml
```diff
@@ -1835,6 +1835,22 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
+  <test name="ERS_D_Ld15" grid="5x5_amazon" compset="I2000Clm50FatesRs" testmods="clm/FatesColdSeedDisp">
+    <machines>
+      <machine name="cheyenne" compiler="gnu" category="fates"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:40:00</option>
```
ekluzek, line 1843: Add a comment about this.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -1835,6 +1835,22 @@
...
+    </options>
+  </test>
+  <test name="PEM_D_Ld15" grid="5x5_amazon" compset="I2000Clm50FatesRs" testmods="clm/FatesColdSeedDisp">
+    <machines>
+      <machine name="cheyenne" compiler="gnu" category="fates"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:40:00</option>
```
ekluzek, line 1851: Comment here as well.

ekluzek: Should we also add a failing ERP test to aux_clm for this?

cime_config/testdefs/testmods_dirs/clm/FatesColdSeedDisp/README
```diff
@@ -0,0 +1,23 @@
...
+values to the fates parameter file for the following variables:
+
+fates_seed_dispersal_fraction
+fates_seed_dispersal_max_dist
+fates_seed_dispersal_pdf_scale
+fates_seed_dispersal_pdf_shape
+
+Given that the default fates parameter file has the above variables as unset,
+a custom fates parameter file must be supplied to appropriately test this mode.
+This testmod itself addresses CTSM issue #2151.
```
ekluzek, line 12: Here I think you should maybe just say the content of #2151 rather than reference a number that doesn't link to github. In the issue we should start marking what parts of it are addressed with this one.

ekluzek (changes requested):
@glemieux nice to see this work come in. We talked about this a few years ago and I think it's a cool project to bring in. I also like the clear names you've used here in the FORTRAN code, it makes it so much more readable.

Some things I think should be done to bring in:

- [x] Update FATES tag
- [ ] Add more documentation on what the integer values mean in XML, write statements, comments etc.
- [x] Add an abort on the MPI_Allgatherv call if ier indicates an error.

I think it would be good to know something about timing for this when this feature is on. My guess is that daily is going to be dog slow, but monthly and yearly might be acceptable. But, it would be good to know that and document that daily is just for software testing of this feature.

There's some other suggestions that I ask about, but you can decide how important they are. It might be good for us to go over this together as well. So please schedule a chat if you'd like that.

Another long term concern I have here is having MPI code embedded fairly deep down into the FATES interface. Possibly the MPI call is something that should be moved into shr_mpi_mod.F90. That's fairly simple to do, but requires messing with more externals. So maybe only something to think about for later. But, we are trying to do MPI things at a higher level, so this is a departure from our current desired architecture. So a shorter term solution might be to add some comments about the hidden MPI call at higher levels...

bld/namelist_files/namelist_definition_ctsm.xml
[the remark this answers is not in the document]
ekluzek: Maybe also add a note that turning this feature on, makes the memory for the feature get larger with the number of processors. Which is in the opposite direction of the rest of the code, where adding more processors makes memory go down. Since, it's only one process and amount of memory so won't necessarily swamp out the overall model memory profile.

cime_config/testdefs/testlist_clm.xml
ekluzek: Let's add the ERP test when the PEM test starts passing. You could add a comment about that.

cime_config/testdefs/testmods_dirs/clm/FatesColdSeedDisp/README
glemieux: Per our discussion, I'll add the url linking to the issue.

src/utils/clmfates_interfaceMod.F90
ekluzek: Yes, use the specific term for MPI task.

src/main/clm_driver.F90
```diff
@@ -1267,6 +1269,12 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
     end do
     !$OMP END PARALLEL DO
 
+
+    ! Pass fates seed dispersal information to all nodes
+    if (fates_seeddisp_cadence .ne. fates_dispersal_cadence_none) then
+       if (use_fates .and. is_beg_curr_day()) call clm_fates%WrapSeedGlobal()
```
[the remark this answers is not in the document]
glemieux: I'll move the `is_beg_curr_day()` inside the call here.

src/utils/clmfates_interfaceMod.F90
glemieux: I'll update this to read "across all MPI_tasks"

src/utils/clmfates_interfaceMod.F90
ekluzek: Let's look into this.

src/utils/clmfates_interfaceMod.F90
glemieux: I'll update this to remove the out-of-date variable names

cime_config/testdefs/ExpectedTestFails.xml
```diff
@@ -109,7 +109,12 @@
     </phase>
   </test>
 
-
+  <test name="PEM_D_Ld15.5x5_amazon.I2000Clm50FatesRs.cheyenne_gnu.clm-FatesColdSeedDisp">
+    <phase name="COMPARE_base_modpes">
```
glemieux, line 113: For some reason this isn't showing up as an expected failure during fates suite regression testing.  @billsacks originally found a discrepancy between the testlist definition and this, but I've since fixed that without seeing the "expected failure" appended to the `./cs.status` output.

src/main/clm_driver.F90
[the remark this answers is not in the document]
glemieux: Also, per @ekluzek, I'll add some more comments providing information to future developers that there is an MPI call wrapped in this `WrapSeedGlobal` call that they may not be aware of.  This should hopefully mitigate any issues with future refactors that might try and move this call somewhere inappropriate, such as inside a threaded region.

src/main/clm_varctl.F90
```diff
@@ -260,6 +260,8 @@ module clm_varctl
 
   ! These are INTERNAL to the FATES module
 
+  integer, public            :: fates_seeddisp_cadence = -9             ! 0 => no seed dispersal
```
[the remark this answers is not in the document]
adrifoster: It looks like we have -9 used as an unset for other things, so perhaps we can just use/create an `integer_unset` parameter?

src/main/controlMod.F90
```diff
@@ -1109,6 +1111,7 @@ subroutine control_print ()
        write(iulog, *) '    use_fates_fixed_biogeog = ', use_fates_fixed_biogeog
        write(iulog, *) '    use_fates_nocomp = ', use_fates_nocomp
        write(iulog, *) '    use_fates_sp = ', use_fates_sp
+       write(iulog, *) '    fates_seeddisp_cadence = ', fates_seeddisp_cadence
```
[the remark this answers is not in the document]
adrifoster: My kingdom for an enumerator type...

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
...
+   logical :: set_flag ! local logical variable to pass to IsItDispersalTime
+
+   type (neighbor_type),  pointer :: neighbor
+
+   ! If WrapSeedGlobal is being called at the end a fates restart call,
+   ! pass .false. to the set_dispersed_flag to avoid updating the 
+   ! global dispersal date
+   set_flag = .true.
+   if (present(is_restart_flag)) then
+      if (is_restart_flag) set_flag = .false.
```
adrifoster, line 2674: can you rename `set_flag`? at first glance it is unclear what is going on

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2593,6 +2643,116 @@ end subroutine wrap_canopy_radiation
 
  ! ======================================================================================
 
+ subroutine WrapSeedGlobal(this,is_restart_flag)
```
adrifoster, line 2646: Can we rename `WrapSeedGlobal` and `wrap_seed_dispersal` to be more informative?

As far as I understand `wrap_seed_dispersal` gives seed to each gridcell/site (and resets outgoing seeds) from the gathered data and `WrapSeedGlobal` actually disperses the seeds? 

I would suggest something like `update_site_seed_inout` for `wrap_seed_dispersal` and then `WrapSeedDispersalGlobal` for `WrapSeedGlobal`

adrifoster (changes requested):

src/utils/clmfates_interfaceMod.F90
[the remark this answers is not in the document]
glemieux: I converted the debugging into an actual check of `ier` so that it will fail out if it's non-zero

glemieux:
Per discussion today around #2151, I'm going to update the testmod to adopt a shell script from @adrifoster to call `ncgen` to avoid having to manually build and store the necessary parameter file for this test everytime the API updates.

- [ ] Update testmod with shell script calling `ncgen`

src/utils/clmfates_interfaceMod.F90
glemieux: We might be able to use `[omp_in_parallel](https://www.openmp.org/spec-html/5.0/openmpsu115.html)` here.  I'll give it a shot.

glemieux:
@ekluzek and @adrifoster I think I've resolved all the requested changes.  Would you take a look when you get a chance to see if they look good?

cime_config/testdefs/ExpectedTestFails.xml
glemieux: This worked with the latest retest of fates.

glemieux:
Regression testing with `aux_clm` is coming back with all expected tests as b4b on cheyenne.  The only differences are `NLCOMP` diffs due to the new `fates_seeddisp_cadence` option.

Cheyenne location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr2077-aux_clm`

glemieux:
Regression testing on izumi completed over the weekend.  All expected tests are b4b/

Location: `/home/glemieux/scratch/ctsm-tests/tests_pr2077-aux_clm`

adrifoster (approved):

Issue #1127, opened 2020-08-27
Remove use of "master" in CTSM


ekluzek:
We currently have about a thousand instances of the word "master" in CTSM (and about 2k in cime). Most of this is the use of the variable "masterproc". It would be good to get away from this terminology and use a different term in its place. Doing this should be bit-for-bit and relatively straight forward to do. Although because it's used a lot it won't be a trivial thing to do.

billsacks:
Let's keep #1048 for the documentation-related pieces of this, which relate to "master" as in "master branch". This issue can be used for pieces that involve Fortran code changes, such as the use of "master" in the MPI sense.

ekluzek:
Note, that CAM is working on this same thing. They are suggesting to move to root_proc for masterproc. Other terms I'm seeing suggested in general to replace master are: primary, director, initiator, controller, leader, conductor, source, or main.

https://github.com/ESCOMP/CAM/issues/327

DavidBergeron-eng:
The vault is classified documents not allowed entry by anyone but me or designer...ME!!!

DavidBergeron-eng:
What are you wanTing to pull from vault

ekluzek:
CAM closed their issue as a "not planned". #1048 will deal with the part of this for github branches.

But, the big outstanding piece would be to change masterproc to something else in our MPI code. Which is a big change.

There is also a use of master_doc in one place lilac/docs/conf.py.

Uses of "master" I see in github are: master_doc, masterproc and mastertask in the code. And the following phrases in documentation or comments:
- master proc
- master task
- master restart
- master clock
- documentation master file
- master input namelist file
-  Repository URL paths that include "master" as the branch for it

PR #2087, opened 2023-08-03
Rework master_list* files etc.

### LINKED ISSUE #2083: Rework master_list_(no)?fates.rst?
My first big-kid PR merge got me thinking about the files `master_list_nofates.rst` and `master_list_fates.rst`. Primarily, I'm wondering if we really need the line numbers included, as doing so means editing the list causes huge diffs in the git history. For example, adding 11 variables in #1863 caused [over 750 lines of diffs](https://github.com/ESCOMP/CTSM/pull/1863/files#diff-8609dcf1357cc3e508de7c54449b7169dcbd926eeb575a221da0286acd08584f).

Some other thoughts:
* We should probably remove "master" from the filenames.
* The filenames should more clearly describe what is in the files.

samsrabin:
### Description of changes
* Renames `master_list*.rst` files to `history_fields*.rst`.
* Updates those files with latest output lists and descriptions.
* Removes line numbers from those files.
* Renames various variables, changes comments, etc. to avoid use of "master" (in references to these files only).

Will unblock #2074.

### Specific notes

CTSM Issues Fixed: 
  Resolves #2083.
  Some work on #1127

Any User Interface Changes (namelist or namelist defaults changes)? Yes: `hist_master_list_file` parameter changed to `hist_fields_list_file`.

Testing performed, if any: Running with `hist_fields_list_file = .true.` produces `history_fields_nofates.rst`, which looks fine.


samsrabin:
@ekluzek What do you think of adding this to [your upcoming simple bit-for-bit tag](https://github.com/ESCOMP/CTSM/projects/6#card-89718468)? It'd be nice to get it in as it's blocking #2074.

ekluzek (approved):

ekluzek:
@samsrabin yes, I'll add it to the list.

samsrabin:
- [ ] Resolve issue where history field list file is saved without `rst` extension.

samsrabin:
@ekluzek I'm requesting a re-review because I discovered some more places to change variable names and references. I also added a testmod `SaveHistFieldList`.

Finally, I updated the .rst files that we ship, using the outputs of the following tests:
```
SMS_Ld1.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_intel.clm-SaveHistFieldList
SMS_Ld1.f10_f10_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-SaveHistFieldList
```
These tests use the default output variables (i.e., without specifying any additional outputs in `user_nl_clm`). The .rst files thus indicate "Active?" with T/F based on the default list. I felt this would be the most useful for people referring to those files. 

However, it differs from the description @slevis-lmwg wrote in [README_master_list_files](https://github.com/ESCOMP/CTSM/blob/ctsm5.1.dev133/doc/source/users_guide/setting-up-and-running-a-case/README_master_list_files) (original text below; diff with my updated version [here](https://github.com/ESCOMP/CTSM/pull/2087/files#diff-a941d96b36c2914e61f8c3c1c87c64613f44e6dd422625cee53a7bd9df571c2b)):
> 2021/9/8 slevis
> 
>The files master_list_nofates.rst and master_list_fates.rst each contain a table of the history fields, active and inactive, available in the CTSM cases that get generated by these tests:
ERP_P36x2_D_Ld3.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_gnu.clm-extra_outputs
ERS_Ld9.f10_f10_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-FatesColdDefCH4
> 
>To reproduce these .rst files, run the above tests and the files will appear in the corresponding run directories.

What do y'all think of my change to indicate "Active?" based on the default settings?

slevis-lmwg:
> What do y'all think of my change to indicate "Active?" based on the default settings?

If I'm understanding correctly that this will indicate active/inactive for each field, yet all the fields will still get listed, then I like that.

samsrabin:
Yep, that's right!

samsrabin:
- [ ] Merge in changes from #2115 once that's complete.

billsacks:
@samsrabin - thanks for your recent commit adding the level dimension to this table!

ekluzek (approved):
This is great. The change to remove the field number from the files will really help prevent git conflicts in those files. I also like your removal of master for the field lists.  You probably didn't even know, but that is work towards #1127 which is really awesome. There are some technical terms in use that have negative historical connotations and we should remove them from our use. So it's good to be making some progress in that regard.

ekluzek:
Oh, something we should check if this interfere's with some of @johnpaulalex PR's that are still outstanding. So I'll try a merge and see what I see...

ekluzek:
It turns out @samsrabin had already checked @johnpaulalex PR's, but I ended up rechecking and came to the same conclusion. There's one PR that has a conflict, but it won't be hard to resolve. But, it might have been good for me to check to know what to do to resolve it.

PR #2108, opened 2023-08-15
Improve robustness of run_python_script()

### LINKED ISSUE #2109: conda run -n can fail if a conda environment is already active
### Brief summary of bug

For tests that invoke `cmds_to_setup_conda()`, manually calling the script invoking that function (e.g., `case.build` for `FSURDATMODIFYCTSM`) could fail if doing so with a conda environment already activated. The problem is that

```bash
conda run -n ctsm_pylib ...
```

seems to not actually use ctsm_pylib if, for instance, the conda base environment is active.

Note that this is a _different_ issue (and gives a different sort of error) from #2111.

### General bug information

**CTSM version you are using:** `ctsm5.1.dev133`

**Configurations affected:** FSURDATMODIFYCTSM and RXCROPMATURITY tests.

Calling `case.build` for an `FSURDATMODIFYCTSM` test fails for me if I've previously done `conda activate base`.

In TestStatus.log, something like:
```
subprocess.CalledProcessError: Command '. /glade/scratch/afoster/tests_0814-163644ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.C.0814-163644ch_int/.env_mach_specific.sh;  conda run -n ctsm_pylib python3 /glade/work/afoster/CTSM/tools/modify_input_files/fsurdat_modifier /glade/scratch/afoster/tests_0814-163644ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.C.0814-163644ch_int/modify_fsurdat.cfg' returned non-zero exit status 1.
```
And in fsurdat_modifier.log, something like:
```
  File "/glade/u/home/samrabin/ctsm_standardize-time-metadata/tools/modify_input_files/fsurdat_modifier", line 15, in <module>
    from ctsm.modify_input_files.fsurdat_modifier import main
  File "/glade/u/home/samrabin/ctsm_standardize-time-metadata/tools/modify_input_files/../../python/ctsm/modify_input_files/fsurdat_modifier.py", line 20, in <module>
    from ctsm.modify_input_files.modify_fsurdat import ModifyFsurdat
  File "/glade/u/home/samrabin/ctsm_standardize-time-metadata/tools/modify_input_files/../../python/ctsm/modify_input_files/modify_fsurdat.py", line 12, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'

ERROR conda.cli.main_run:execute(47): `conda run python3 /glade/u/home/samrabin/ctsm_standardize-time-metadata/tools/modify_input_files/fsurdat_modifier /glade/scratch/samrabin/tests_0815-121457ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.0815-121457ch/modify_fsurdat.cfg` failed. (See above for error)
```
(The missing module could be something other than `numpy`.)

samsrabin:
Issue #2109: For tests that invoke `cmds_to_setup_conda()`, manually calling the script invoking that function (e.g., `case.build` for `FSURDATMODIFYCTSM`) could fail if doing so with a conda environment already activated. The problem is that
```bash
conda run -n ctsm_pylib ...
```
seems to not actually use ctsm_pylib if, for instance the conda base environment is active. 

Issue #2111: For some users, FSURDATMODIFYCTSM is failing because conda doesn't load.

### Description of changes

Putting `CONDA_PREFIX=;` before the `conda run` command fixes #2109.

### Specific notes

Remaining tasks:
- [ ] Figure out #2111. This may be an environment issue; @ekluzek will work with CISL to investigate.
- [ ] Split up the long list of commands into chunks for better error tracing.
- [ ] Use `subprocess.run(..., shell=False)` instead of `True`

Contributors other than yourself, if any: @adrifoster, @ekluzek 

CTSM Issues Fixed:
- Fixes #2109

Testing performed, if any: None so far; will do tests from `aux_clm` that exercise this code (`FSURDATMODIFYCTSM` and `RXCROPMATURITY`). I will ask @adrifoster, @ekluzek, and @rgknox to try `FSURDATMODIFYCTSM` as well, once I have fixes in place.

### DIFF
--- a/cime_config/SystemTests/systemtest_utils.py
+++ b/cime_config/SystemTests/systemtest_utils.py
@@ -5,7 +5,7 @@ Reduce code duplication by putting reused functions here.
 import os, subprocess
 
 
-def cmds_to_setup_conda(caseroot):
+def cmds_to_setup_conda(caseroot, test_conda_retry=True):
     # Add specific commands needed on different machines to get conda available
     # Use semicolon here since it's OK to fail
     #
@@ -19,7 +19,12 @@ def cmds_to_setup_conda(caseroot):
         subprocess.run("which conda", shell=True, check=True)
     except subprocess.CalledProcessError:
         # Remove python and add conda to environment for cheyennne
-        conda_setup_commands += " module unload python; module load conda;"
+        unload_python_load_conda = "module unload python; module load conda;"
+        # Make sure that adding this actually loads conda
+        if test_conda_retry:
+            subprocess.run(unload_python_load_conda + "which conda", shell=True, check=True)
+        # Save
+        conda_setup_commands += " " + unload_python_load_conda
 
     return conda_setup_commands
 
--- a/cime_config/SystemTests/systemtest_utils.py
+++ b/cime_config/SystemTests/systemtest_utils.py
@@ -29,15 +29,21 @@ def cmds_to_setup_conda(caseroot, test_conda_retry=True):
     return conda_setup_commands
 
 
-def run_python_script(caseroot, this_conda_env, command, tool_path):
-
+def cmds_to_run_via_conda(caseroot, conda_run_call, command, test_conda_retry=True):
     # Run in the specified conda environment
-    conda_setup_commands = cmds_to_setup_conda(caseroot)
-    conda_setup_commands += f" conda run -n {this_conda_env}"
+    conda_setup_commands = cmds_to_setup_conda(caseroot, test_conda_retry)
+    conda_setup_commands += " " + conda_run_call
 
     # Finish with Python script call
     command = conda_setup_commands + " " + command
     print(f"command: {command}")
+
+    return command
+
+
+def run_python_script(caseroot, this_conda_env, command_in, tool_path):
+
+    command = cmds_to_run_via_conda(caseroot, f"conda run -n {this_conda_env}", command_in)
 
     # Run with logfile
     tool_name = os.path.split(tool_path)[-1]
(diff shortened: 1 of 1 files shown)


ekluzek:
@samsrabin and I went over this and we showed how each of our environments are working and causing problems here.

I have some issues with my environment that I likely need to fix with the help of CISL. The module table gets corrupted when I use a different shell besides my default of tcsh. And my conda version for tcsh doesn't seem to be correct, but is for other shells. So I'm going to work on that.

I did some investigation into the systemtest_utils.py script and I saw a few things to improve. The long command it creates makes is hard to identify exactly what is going wrong. So he's going to have a trial test version of some of the commands beforehand to make sure they are working and having the desired effect. Then it can die with a nice explanation of what went wrong. For example, when conda isn't found, to run the module load and see if conda is available at that point, if not it will abort.

The other thing I was looking into was the shell=True part of our usage of subprocess.run from python. That is useful if you want it to be interpreted by the shell (and are using shell specific interpretations, env variables, and things like ~). It also changes how arguments are given. If shell=False, you need to give it as a sequence, which can be put together with the "shlex" python command.

samsrabin:
@ekluzek I've added a bit of error traceability, as well as the fallback to the old method. Would you mind testing with the latest version in this PR?

PR #2125, opened 2023-08-29
Avoid using subprocess.run() in FSURDATMODIFYCTSM

### LINKED ISSUE #2109: conda run -n can fail if a conda environment is already active

### LINKED ISSUE #2111: conda fails to load for SystemTests
### Brief summary of bug

For some users, FSURDATMODIFYCTSM is failing because `conda` (called in a subprocess from Python) doesn't load. Note that this is a _different_ issue (and gives a different sort of error) from #2109.

### General bug information

**CTSM version you are using:** `ctsm5.1.dev133-47-g6925f8cc9`

**Configurations affected:** FSURDATMODIFYCTSM and RXCROPMATURITY tests.

### Details of bug

@rgknox and @ekluzek (#1959), as well as @slevis-lmwg (#2106), are all affected. This doesn't happen for me or, I think, @adrifoster. I think it's thus something to do with our shell environments.

We suspect this stems from changes introduced in `cime_config/SystemTests/` in `ctsm5.1.dev131` (as was the case for #2109). The `module unload python; module load conda;` step fails to load `conda`, so then the call of `conda run -n` fails and ends the test. However, that module un/loading step _was_ present before, so there's something subtle going on.

Unknown.

From TestStatus.log:
```
which: no conda in (/glade/u/apps/ch/opt/netcdf/4.9.0/intel/19.1.1/bin:/glade/u/apps/ch/opt/ncarcompilers/0.5.0/intel/19.1.1:/glade/u/apps/ch/opt/ncarcompilers/0.5.0/intel/19.1.1/mpi:/glade/p/cesmdata/cseg/PROGS/esmf/8.4.1b02/mpiuni/2.3.0/intel/19.1.1/bin/bing/Linux.intel.64.mpiuni.default:/glade/u/apps/opt/intel/2020u1/vtune_profiler/bin64:/glade/u/apps/opt/intel/2020u1/inspector/bin64:/glade/u/apps/opt/intel/2020u1/advisor/bin64:/glade/u/apps/opt/intel/2020u1/compilers_and_libraries/linux/bin/intel64:/glade/u/apps/ch/opt/cmake/3.22.0/bin:/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/bin:/glade/u/apps/opt/vncmgr:/glade/u/apps/opt/globus-utils:/glade/u/apps/ch/opt/usr/bin:/glade/u/apps/ch/opt/lmod/8.7.13/bin:/glade/u/apps/opt/intel/2020u1/vtune_amplifier/bin64:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/bin/X11:/usr/games:/opt/ibutils/bin:/usr/lib/mit/bin:/opt/pbs/bin:/glade/u/apps/opt/hpss/hpss-client.5.0.2.p5/bin:/glade/u/apps/ch/os/usr/bin:/glade/u/apps/ch/opt/hpe-cpe/22.02/bin)
  File "./case.build", line 256, in <module>
    _main_func(__doc__)
  File "./case.build", line 216, in _main_func
  File "/glade/u/home/rgknox/ctsm/cime_config/SystemTests/fsurdatmodifyctsm.py", line 46, in __init__
    self._run_modify_fsurdat()
  File "/glade/u/home/rgknox/ctsm/cime_config/SystemTests/fsurdatmodifyctsm.py", line 77, in _run_modify_fsurdat
    tool_path,
  File "/glade/u/home/rgknox/ctsm/cime_config/SystemTests/systemtest_utils.py", line 39, in run_python_script
    command, shell=True, check=True, text=True, stdout=f, stderr=subprocess.STDOUT
  File "/glade/u/apps/ch/opt/python/3.7.9/gnu/9.1.0/lib/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '. /glade/scratch/rgknox/tests_0814-134713ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.GC.0814-134713ch_int/.env_mach_specific.sh;  module unload python; module load conda; conda run -n ctsm_pylib python3 /glade/u/home/rgknox/ctsm/tools/modify_input_files/fsurdat_modifier /glade/scratch/rgknox/tests_0814-134713ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.GC.0814-134713ch_int/modify_fsurdat.cfg' returned non-zero exit status 127.
```
And fsurdat_modifier.log:
```
Lmod has detected the following error: Unable to load module because of error
when evaluating modulefile:
     /glade/u/apps/ch/modulefiles/default/idep/conda/latest.lua: [string
"require("posix")..."]:54: attempt to call global 'source_sh' (a nil value)
     Please check the modulefile and especially if there is a the line number
specified in the above message
While processing the following module(s):
    Module fullname  Module Filename
    ---------------  ---------------
    conda/latest     /glade/u/apps/ch/modulefiles/default/idep/conda/latest.lua

/bin/sh: conda: command not found
```

samsrabin:
Issues #2109 and #2111 seem to stem from idiosyncrasies of different users' Cheyenne environments that become important when `subprocess.run()` is called. This PR removes the use of `subprocess.run()` from the FSURDATMODIFYCTSM test, improving robustness and resolving those issues.

### Description of changes

Instead of starting a new subprocess shell in which `fsurdat_modifier` is called, this PR makes it so the command is called directly by Python itself. This does require that all the Python dependencies are loaded, which can be accomplished by activating the `ctsm_pylib` environment before calling `run_sys_tests` or `cime/scripts/create_test`.

Unfortunately, this method can't be used to fix RXCROPMATURITY failing for some users, even though that's also due to environment. Hopefully that will resolve itself once we move to Derecho.

### Specific notes

**Remaining tasks** (not including testing):
- [x] Make FSURDATMODIFYCTSM call `fsurdat_modifier` directly.

**CTSM Issues Fixed:**
- Fixes #2109 
- Fixes #2111 
- Supersedes PR #2108.

### Testing
`FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel`:
- [x] Works for me.
- [x] Works for a user who experienced #2109.
- [x] Works for a user who experienced #2111.
- [x] `clm_pymods` test suite passes.


samsrabin:
Before I continue down this path, I want to make sure FSURDATMODIFYCTSM is working for everyone. It does work for me, but I didn't experience either of the previous issues.

@adrifoster @glemieux Would one of y'all be willing to test this to see if you hit issue #2109?
@billsacks @ekluzek @rgknox @slevis-lmwg  Would one of y'all be willing to test this to see if you hit issue #2111?

Specifically, what we need to see is that `fsurdat.nc` and `done_FSURDATMODIFYCTSM_setup.txt` are generated in the case directory of test `FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel`. This happens early in the `SHAREDLIB_BUILD` phase, so it should be pretty quick to tell one way or the other.

Note that you must have all the requisite Python dependencies loaded. To ensure this, please test with the `ctsm_pylib` conda environment active when calling `run_sys_tests` or `cime/scripts/create_test`. If you could test with both of those methods, that'd be great. Thanks!

ekluzek:
I tried it again and the build phase works!

However, it failed at the run phase, because numpy wasn't loaded. This is because the run phase is sent to the share queue as a separate thing, so it probably needs to load the conda env again. I also tried running case.submit with --no-batch, but that didn't work either. I think it might still spawn off a separate process even in that case, although it doesn't go into the queue. I'm actually also suspicious that there is a bug in cime for no-batch.

I also tried Derecho, but that had multiple problems at this point.

My case for Cheyenne is here:

/glade/scratch/erik/tests_0901-095751ch/FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel.0901-095751ch

samsrabin:
Unfortunately I'm seeing this as well. I guess the run phase re-imports `fsurdatmodifyctsm.py`, even though it doesn't extend `run_phase()`. (Of course, there's no way for CIME to know that, so this behavior makes sense.)

I've unchecked "works for me" in the PR description.

ekluzek:
By the way @samsrabin thanks for working on this. It's tricky for all of us, but important to get figured out.  If it's helpful to get some of us together to brainstorm let us know.

ekluzek:
Awesome. Latest update now works for me!

PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel CREATE_NEWCASE
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel XML
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel SETUP
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel SHAREDLIB_BUILD time=252
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel MODEL_BUILD time=48
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel SUBMIT
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel RUN time=135
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel MEMLEAK insuffiencient data for memleak test
PASS FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.cheyenne_intel SHORT_TERM_ARCHIVER

python/ctsm/crop_calendars/cropcal_module.py
```diff
@@ -1,13 +1,17 @@
...
 import sys
 import os
 import glob
 
+# Import the CTSM Python utilities
+_CTSM_PYTHON = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), os.pardir, os.pardir, os.pardir, "python"
+)
+sys.path.insert(1, _CTSM_PYTHON)
+import ctsm.crop_calendars.cropcal_utils as utils
```
billsacks, line 13: Can you speak to why this is needed? In other places we can import stuff throughout the ctsm python library without doing path manipulation - e.g., see job_launcher_qsub.py (for an example in a subdirectory of python/ctsm, like this one is). Is it possible that you need an `__init__.py` in the crop_calendars directory to enable this??

python/ctsm/run_sys_tests.py
```diff
@@ -692,12 +696,31 @@ def _run_test_suite(
         )
 
 
+def _try_systemtests(testname_list):
```
billsacks, line 699: Cool, thanks for adding this! I see that it ended up being a little more involved than I imagined because you're checking for whether either of the relevant tests is being run... and I like that addition. Thanks!

python/ctsm/crop_calendars/cropcal_module.py
samsrabin: Good point; not necessary! I had that bit in my head from adding it to `fsurdatmodifyctsm.py`, but that was only necessary because it's not in `python/`. My latest commit removes these unneeded lines.

billsacks: Cool, thanks.

samsrabin:
It's looking like this solution can't be applied to RXCROPMATURITY because it calls external Python scripts during the run phase; see discussion [here](https://github.com/ESMCI/cime/discussions/4482). I'm leaning towards bringing in the FSURDATMODIFYCTSM solution now, hoping that Derecho magically fixes things for RXCROPMATURITY—I'm not sure banging my head against this anymore is worth it for a test that hardly anyone will run.

ekluzek:
That makes sense to me @samsrabin. The RxCropMaturity test will only be run in the ctsm_sci test list right? Or maybe another special test list right?

We also might want to try this out on Derecho and see if it's a problem there. If it isn't there isn't a need to worry about this for Cheyenne.

Some of these things can be difficult to figure out, so there's a point when we should just punt and move on...

samsrabin:
That's right @ekluzek, it's only run in the `ctsm_sci` suite.

I agree that it's worth testing on Derecho, but I guess we have some work to do before that's possible.

cime_config/SystemTests/systemtest_utils.py
```diff
@@ -39,19 +53,32 @@ def run_python_script(caseroot, this_conda_env, command, tool_path):
...
-        print("get past this error by running ./py_env_create ")
-        print("in your ctsm directory and trying this test again. ")
-        print("(2) If conda is not available, install and load conda, ")
-        print("run ./py_env_create, and then try this test again. ")
-        print("(3) If (1) and (2) are not the issue, then you may be ")
-        print(f"getting an error within {tool_name} itself. ")
-        print("Default error message: ")
-        print(error.output)
-        raise
+        # First, retry with the original method
```
ekluzek, line 56: In. this context I don't know what "original method" means. But, I think you are saying run with "conda activate" rather than "conda run". So it tries with "conda run" and if that fails, it tries again with "conda activate".  Being more explicit in the comment about that's what it's doing would be helpful here.

python/ctsm/crop_calendars/cropcal_module.py
ekluzek: Wait I still see the path logic, am I missing something?

cime_config/SystemTests/systemtest_utils.py
```diff
@@ -39,19 +53,32 @@ def run_python_script(caseroot, this_conda_env, command, tool_path):
...
+        command = cmds_to_run_via_conda(
+            caseroot, f"conda activate {this_conda_env} && ", command_in, test_conda_retry=False
```
ekluzek, line 58: Oh this is where test_conda_retry is set to False. But, I'm not sure I see why it's false here? What is the reason it's False here?

ekluzek (changes requested):
OK, I have at least one suggestion about a comment. And a question about the test_conda_retry logical. I also reopened the conversation Bill had about the CTSM path stuff, which looks like it's still there. And I think it can be removed in that place as well as others. In any case I'm looking forward to our discussion to go over this.

cime_config/SystemTests/systemtest_utils.py
samsrabin: In the original (`conda activate`) method, the line that's nested beneath `if test_conda_retry:` wasn't run. So when falling back on the original method, I thought it'd be best to skip that.

cime_config/SystemTests/systemtest_utils.py
samsrabin: Good point; I improved comments in d1ed672.

cime_config/SystemTests/systemtest_utils.py
```diff
@@ -5,32 +5,46 @@
 import os, subprocess
 
 
-def cmds_to_setup_conda(caseroot):
+def cmds_to_setup_conda(caseroot, test_conda_retry=True):
```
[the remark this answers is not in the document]
samsrabin: (See [here](https://github.com/ESCOMP/CTSM/pull/2125#discussion_r1329225749).)

samsrabin:
`clm_pymods` tests still pass after latest commits.

ekluzek (approved):
@samsrabin and I went over this earlier. And he made changes according to my suggestions.

The one thing that I thought we should discuss as a group is how to handle setting the path for python for these type of system tests that now need to manipulate the path. We didn't have this before because the top level tool skeleton handled it. So there it's in one place. Here we need a better way to put it in one place. One way to do that would be to set the path for python using an env variable. I can think of other ways to do it as well. We should decide as a group and then make an issue to change it to that method.

python/ctsm/crop_calendars/cropcal_module.py
ekluzek: This is the thing we should discuss as a group.

PR #2154, opened 2023-09-12
Rearrange calculation of leaf/livestem C and N to biofuels/litter

samsrabin:
### Description of changes

This PR rearranges the calculation of how much leaf and livestem C and N goes to biofuels vs. litter, in anticipation of me adding crop residue removal. This PR also makes the affected subroutine easier to read.

### Specific notes

Specifically, this PR introduces four new variables in `CNOffsetLitterfall()`:
- `leafc_remaining`
- `leafn_remaining`
- `livestemc_remaining`
- `livestemn_remaining`

Those variables are initialized with the pool remaining after biofuel harvest. When I introduce crop residue removal, it will be able to act on (and update) those variables directly.

**Contributors other than yourself, if any:** @billsacks, @ekluzek 

**CTSM Issues Fixed:** None

**Are answers expected to change (and if so in what way)?** Yes, but only roundoff-level due to order-of-operations changes.

**Testing performed:** `aux_clm` showed diffs from `ctsm5.1.dev132`.

### DIFF
--- a/src/biogeochem/CNPhenologyMod.F90
+++ b/src/biogeochem/CNPhenologyMod.F90
@@ -3136,6 +3136,10 @@ contains
                   leafc_to_litter(p)  = t1 * leafc_remaining  + cpool_to_leafc(p)
                   livestemc_to_litter(p)   = t1 * livestemc_remaining  + cpool_to_livestemc(p)
                   livestemn_to_litter(p)   = t1 * livestemn_remaining
+                  ! Sam Rabin 2023-09-11:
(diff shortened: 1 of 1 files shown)


billsacks:
I haven't looked closely at this, but I love to see cleanup like this that makes the code easier to understand - thanks!

slevis-lmwg (approved):
Looks good to me.

@samsrabin did this small refactor to set up a template for biofuel removal that can be repeated in another PR for residue removal.

PR #2156, opened 2023-09-13
Moving critical toolchain scripts to /python and creating wrapper scripts for them

### LINKED ISSUE #1441: Move critical toolchain scripts out of tools/contrib and their guts into the python subdirectory and unit tests added
We've introduced some critical single-point toolchain scripts at this point, and the top level scripts should be moved as a skeleton wrapper out of the tools/contrib directory to the main tools directory in a single_point subdirectory. The NEON scripts should be in the same directory with a _neon ending in the name. The guts should go into python modules that are under the python subdirectory. Since the guts are outside in the python subdirectory, there seems to be no reason to have subdirectories for each of these tools, and it's OK to have some duplication between the small top level skeleton scripts≥

There should also be unit tests added to important logic for the python modules in the python subdirectory. The unit tests should be connected to the python subdirectory unit-testing structure and be triggered with the other unit-tests that are there.

The top level tools/README file should also be updated to talk about how these new scripts fit into the workflow.

The system tools/tests should also continue to work for these tools in their new directory.

The first scripts that need to be moved are:
 - modify_singlept_site_neon.py (to tools/single_point)
- subset_data.py (to tools/single_point)
- modify_singlept_site to tools/single_point

We talked about this at out July 22nd/2021 software meeting.

Definition of Done:
- [x] Move subset_data out
- [x] Add testing for subset_data
- [x] Move run_neon.py out
- [x] Add testing for run_neon.py
- [x] Move neon_surf_wrapper.py out
- [x] Add testing for neon_surf_wrapper.py
- [x] Move modify_singlept_site_neon.py out
- [x] Add testing for modify_singlept_site_neon.py
- [x] Move new script when it comes in: tools/site_and_regional/plumber2_surf_wrapper.py
- [x] Add testing for plumber2_surf_wrapper.py

slevis-lmwg:
### Description of changes

- Move scripts to /python/ctsm/site_and_regional
subset_data.py (DONE before this PR)
run_neon.py
neon_surf_wrapper.py
modify_singlept_site_neon.py
- Make wrapper scripts for them in /tools/site_and_regional
subset_data (DONE before this PR)
run_neon
neon_surf_wrapper
modify_singlept_site_neon
- Add testing

### Specific notes

@TeaganKing and hackathon participants (2023/9/13)

Fixes #1441 

No

The wrapper scripts do not have .py suffixes.

Testing performed, if any:
./run_neon --help
./run_neon --neon-sites ABBY
...the latter still running as we speak.


TeaganKing:
A few comments on unit test additions that I made after the hackathon today:
1. For `neon_surf_wrapper`, `execute(subset_command)` and `execute(modify_command)` should probably be in the sys tests; the `parser` test probably isn't really worth unit testing, but I added it anyways so that the unit testing structure for `neon_surf_wrapper` is ready for any future additions.
2. Perhaps `valid_neon_sites` in `modify_singlept_site_neon`should come from a similar list as that in `run_neon` rather than a hardcoded list within `modify_singlept_site_neon`; this may be outside the scope of this PR but also probably not too hard to change.
3. Some of the `modify_singlept_site_neon` tests would require pre-existing files. Where can we find these? I marked a few notes where this could be useful with "TODO".
4. These still need pylint/black checking. I'll plan to do that in a separate commit (along with run_neon formatting) in order to avoid the 'blame' issue for those files.

ekluzek:
@TeaganKing this all sounds really great. I just wanted to comment on your item "3" above. 

What I've done with files that we need for the python system tests is I'll usually copy them from existing files that we have from inputdata. Since, we can't guarantee inputdata find files will be accessible, it's best to explicitly add the files that are needed for testing with the names we need for the tests. We've set this up using "git lfs" which does require a small amount of setup. See the README files in that directory for this. We still try to only put files in there that are as small as possible.

TeaganKing:
Got it! Thanks for these details. I did see some files that were stored with Git LFS. I think we'll just need to determine which files we want to use (or if we can use any files that are already there), and then I'll be sure to add them with large file storage.

TeaganKing:
@slevis-lmwg and I just met to discuss these tests. A few notes:
- System tests are probably not needed for `neon_surf_wrapper`; the main functionality of this scripts comes from submitting commands to other scripts which are thoroughly system tested. I will add unit tests to confirm that we are generating a working command.
- We will use the Mexico city 1x1 surface file in these unit tests since it is already stored in the test inputs directory, `python/ctsm/test/testinputs/surfdata_1x1_mexicocityMEX_hist_16pfts_Irrig_CMIP6_simyr2000_c221206.nc`, and meets the needs of our testing.
- We will not unit test `sort_print_soil_layers` within `modify_singlepoint_site_neon` because it prints out information but does not actually change the result of running a NEON case.
- We would prefer to not include an example of the raw dataset that is used in `modify_singlept_site_neon.find_soil_structure`, but will test the functionality of finding this attribute on the Mexico 1x1 file in order to test that the attribute name remains consistent.

I will update these unit tests per our discussion, as well as work on fleshing out `test_sys_run_neon` (the latter of which might require new files in the test inputdata directory).

python/ctsm/site_and_regional/modify_singlept_site_neon.py
```diff
@@ -550,7 +502,7 @@ def main():
 
     # --  directory structure
     current_dir = os.getcwd()
-    parent_dir = os.path.dirname(current_dir)
+    # parent_dir = os.path.dirname(current_dir)
```
slevis-lmwg, line 505: ...if not used, I would delete.
Same comment for obs_top a few lines down.

slevis-lmwg: Same possibly for a few print statements further down.

python/ctsm/test/test_unit_neon_surf_wrapper.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Basic class for testing neon_surf_wrapper.py.
+    """
+
+    def test_parser(self):
+        """
+        Test that parser has same defaults as expected
+        """
+
+        self.assertEqual(get_parser().argument_default, None, "Parser not working as expected")
```
slevis-lmwg, line 37: This one is new, right? 
If I'm understanding correctly, if blank, then fail. Seems good.

slevis-lmwg (commented):
@TeaganKing this looks good!

I included a suggestion regarding the failing xr.open_dataset, which may or may not be helpful...

python/ctsm/test/test_unit_modify_singlept_site_neon.py
```diff
@@ -0,0 +1,138 @@
...
+            find_surffile(surf_dir, site_name, pft_16)
+
+    # TODO: need to fix unkown file type error?
+    # def test_find_soil_structure(self):
+    #    """
+    #    Test to ensure that correct attributes are found for find_soil_structure.
+    #    soil_texture_raw_data_file_name should be found, and test should go through sysexit.
+    #    """
+    #    surf_file = "testinputs/surfdata_1x1_mexicocityMEX_hist_16pfts_Irrig_CMIP6_simyr2000_c221206.nc"
+    #    f1 = xr.open_dataset(surf_file)
```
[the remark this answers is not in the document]
slevis-lmwg: Also may be worth confirming that `test_1x1_mexicocity` still passes.

python/ctsm/test/test_unit_neon_surf_wrapper.py
TeaganKing: This is actually saying that there should be no default argument, as the argument_default should be equal to None. If some default argument were added, we would want to be aware of how that would affect `run_neon` when the execute command is called. I could also change it to assert some other part of the parser is working as expected.

python/ctsm/site_and_regional/modify_singlept_site_neon.py
TeaganKing: Good points. There are a lot of print statements in this file. @wwieder , it looks like you had some TODO's related to the print sections that maybe got resolved? Do you want these print statements still? I'm personally not sure what's relevant/useful for folks looking at soil levels.

python/ctsm/test/test_unit_modify_singlept_site_neon.py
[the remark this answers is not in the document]
TeaganKing: Good idea. Thanks for pointing out that this test also uses the 1x1 Mexico City file.

It looks like the `test_1x1_mexicocity` within `test_sys_fsurdat_modifier.py` is not working. 

I'm getting the same error as what I was seeing above:
`OSError: [Errno -51] NetCDF: Unknown file format: b'/glade/work/tking/ctsm/python/ctsm/test/testinputs/surfdata_1x1_mexicocityMEX_hist_16pfts_Irrig_CMIP6_simyr2000_c221206.nc'`

Note that I did add the following lines in order to test with unittest:
`# -- add python/ctsm  to path (needed if we want to run the test stand-alone)
_CTSM_PYTHON = os.path.join(os.path.dirname(os.path.realpath(__file__)), os.pardir, os.pardir)
sys.path.insert(1, _CTSM_PYTHON)`

I ran the test from within the `ctsm_pylib` environment with `python -m unittest test_sys_fsurdat_modifier.py`.

A number of other tests within `test_sys_fsurdat_modifier` also failed-- we should probably revisit that testing script, too.

@ekluzek had you run these tests in a different way than how I describe above in order to correctly find/open the netcdf files?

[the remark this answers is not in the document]
ekluzek: Hi @TeaganKing this is a "git lfs" issue. I could tell because when you can that NetCDF file it is actually a text file that gives the LFS info for that file.

So go through the directions in the 

/glade/work/tking/ctsm/python/ctsm/test/testinputs/README.md

and hopefully that will get you going. If not let's meet to get you going. This should all be resolved after you get that figured out. Unfortunately this is one of those things where the errors you see don't make the actual problem obvious. It would be good to improve that...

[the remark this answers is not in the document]
TeaganKing: Oh, got it! That makes a lot of sense that I don't actually have the files locally if they are from git lfs. I'll check out those directions. Thanks @ekluzek !

python/ctsm/test/test_unit_neon_surf_wrapper.py
slevis-lmwg: Oh, I misunderstood, thank you for explaining. I think it's fine to leave this way for now.

python/ctsm/test/test_unit_modify_singlept_site_neon.py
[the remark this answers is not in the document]
TeaganKing: With LFS, the unit tests for `modify_singlept_site_neon` are now working! I decided to just test that the file was modified today instead of changed, because if the file was changed more than once, it would fail a test that checked that the file hadn't been updated today.

@ekluzek , regarding `test_sys_fsurdat_modifier`, there does still seem to be a test failure on "fsurdat_out is specified in both the command line and the config file, pick one". If these system tests are being tested in another manner, other than with `python unitttest`, do they still run into this issue?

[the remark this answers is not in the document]
slevis-lmwg: @TeaganKing this may answer your last question...
Run the full suite of tests from the /python directory:
`./run_ctsm_py_tests`
Try it with --help first to see available options.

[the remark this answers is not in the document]
ekluzek: LFS issue that was resolved.

ekluzek (approved):
@TeaganKing and I went over this briefly just now. I had looked at it earlier, and she also had shown us what her and @slevis-lmwg had been working on earlier.

This  is really great and does the critical move we need to enable using our CTSM python development tools and testing. I don't see anything to change here. I did have the small comment that making unit tests obvious is a good thing. So for example for the test of a bad site name, make the site name an obvious failure. I was impressed they were able to do more tests than I though they might have been able to do which is great.

@TeaganKing also has some system tests on Cheyenne, that she'll work on when Cheyenne is up, and that would complete this work. I've also added this to upcoming tags since we want to bring this in right away in order to facilitate other work to happen that will build upon this point. We will discuss how to bring this to main-dev CTSM right away at our CTSM software meeting.

python/ctsm/site_and_regional/modify_singlept_site_neon.py
wwieder: Looks like these were from Negin [NS]. I'm not sure if we should keep them or not?

python/ctsm/site_and_regional/modify_singlept_site_neon.py
```diff
@@ -709,9 +659,9 @@ def main():
         print("Updated  : ", f2.PCT_CROP.values)
 
         print("Updating PCT_NAT_PFT")
-        #print (f2.PCT_NAT_PFT)
+        # print (f2.PCT_NAT_PFT)
         print(f2.PCT_NAT_PFT.values[0])
-        #f2.PCT_NAT_PFT.values[0] = [[100.0]]
+        # f2.PCT_NAT_PFT.values[0] = [[100.0]]
         print(f2.PCT_NAT_PFT[0].values)
```
wwieder, line 662: These unused lines should be removed?

wwieder (approved):

python/ctsm/site_and_regional/modify_singlept_site_neon.py
TeaganKing: Ah, yes, thanks @wwieder , I do see some of these comments are from Negin, too. I think there's the bit marked "TODO Will: if I sum them up , are they 3.5? (m) YES" and some discussion on updates that should instead be made to subset_data-- were these resolved?

python/ctsm/site_and_regional/modify_singlept_site_neon.py
```diff
@@ -586,7 +538,7 @@ def main():
     soil_mid = 0.5 * (soil_bot - soil_top) + soil_top
     # print ("Cumulative sum of soil bottom depths :", sum(soil_bot))
 
-    obs_top = df["biogeoTopDepth"] / 100
+    # obs_top = df["biogeoTopDepth"] / 100
```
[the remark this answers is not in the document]
TeaganKing: I don't think this is used-- at least it doesn't appear again in this file. However, `obs_bot` is used.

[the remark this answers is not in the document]
TeaganKing: If it is used elsewhere, that would be good to know, though.

python/ctsm/site_and_regional/modify_singlept_site_neon.py
```diff
@@ -635,12 +587,10 @@ def main():
         # -- Check to make sure the rounded oc is not higher than carbon_tot.
         # -- Use carbon_tot if estimated_oc is bigger than carbon_tot.
 
-        if estimated_oc > carbon_tot:
-            estimated_oc = carbon_tot
+        estimated_oc = min(estimated_oc, carbon_tot)
```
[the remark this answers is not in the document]
TeaganKing: Would the carbon total ever be less than the estimated OC? In that case, it looks like it would give a different result?

python/ctsm/site_and_regional/modify_singlept_site_neon.py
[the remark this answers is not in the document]
wwieder: I'd just delete it then

python/ctsm/site_and_regional/modify_singlept_site_neon.py
```diff
@@ -575,18 +524,12 @@ def main():
     # better suggestion by WW to write dzsoi to neon surface dataset
     # This todo needs to go to the subset_data
 
-    # TODO Will: if I sum them up , are they 3.5? (m) YES
-    print("soil_top:", soil_top)
-    print("soil_bot:", soil_bot)
-    print("Sum of soil top depths    :", sum(soil_top))
-    print("Sum of soil bottom depths :", sum(soil_bot))
-
     soil_top = np.cumsum(soil_top)
```
wwieder, line 527: Ok, I see this comment now,  Yes fine to remove

wwieder (approved):
Look good, I think you resolved other comments @TeaganKing .  Thank you!

TeaganKing:
Following up on a conversation yesterday with @ekluzek --

I think "--setup-only" doesn't end up taking too long (just a minute or two), so I'm going to suggest keeping this as a regular 'sys' test instead of including a 'long_sys' test. I have noted a few other items in a `TODO` in `tests_sys_run_neon` that would be great to test, but would involve running multiple cases and thus a significant amount of additional run time. So, perhaps these are not ideal for system testing.

The `test_sys_modify_singlept_site_neon` doesn't seem to support relative paths (from either the python or test directories); this test currently just asserts that the modification process starts but the paths cause issues in finding surface datasets. I included a `TODO` with more details on what I think we should ideally actually test; however, my impression is that this may be dependent on some changes to the `run_neon` code.

python/ctsm/test/test_sys_modify_singlept_site_neon.py
```diff
@@ -0,0 +1,75 @@
...
+        TODO: The primary items to test here are the following:
+        1) Fields are overwritten with site-specific data for neon sites
+        2) Downloaded data is used in surface dataset
+        3) Check specific fields listed in update_metadata for correct output
+        4) Check that a netcdf with correct formatting is created
+        """
+        sys.argv = [
+            "modify_singlept_site_neon",
+            "--neon_site",
+            "/glade/work/tking/ctsm/cime_config/usermods_dirs/NEON/ABBY",
```
slevis-lmwg, line 60: Talking with @ekluzek:
@TeaganKing pls replace personal path with path_to_ctsm_root.

slevis-lmwg: Beyond that, could we have a short meeting one last time and then I will run testing and merge.

TeaganKing: Thanks for going over this @slevis-lmwg ! I'll update that path, check calendars, and set up a quick meeting with you shortly.

TeaganKing:
@slevis-lmwg and I just met and realized that while system tests for run_neon are running smoothly for `./run_ctsm_py_tests -p test_sys_run_neon*`, they don't pass with `./run_ctsm_py_tests -s` due to an error in the parser utils function that calls on the current working directory. We're thinking this has to do with this being run from a tempdir that gets updated part way through the case setup. I'm planning to dig into why this is causing issues with running all tests at once but not the singular test by itself.

slevis-lmwg:
While meeting with @TeaganKing we both tried these tests:
```
./run_ctsm_py_tests --unit
./run_ctsm_py_tests -p test_sys_run_neon.py
./run_ctsm_py_tests --sys
```
Lines 1 and 2 worked. Line 3 failed with
```
  File "/glade/work/slevis/git/single_point/python/ctsm/test/test_sys_run_neon.py", line 58, in test_one_site
    parser = get_parser(sys.argv, "description_for_parser", valid_neon_sites)
  File "/glade/work/slevis/git/single_point/python/ctsm/site_and_regional/run_neon.py", line 90, in get_parser
    CIME.utils.setup_standard_logging_options(parser)
  File "/glade/work/slevis/git/single_point/cime/CIME/utils.py", line 1605, in setup_standard_logging_options
    print('cwd =', os.getcwd())
FileNotFoundError: [Errno 2] No such file or directory
```
@TeaganKing will troubleshoot and will follow up with me.

slevis-lmwg:
Meanwhile I'm running
`./run_sys_tests -s clm_pymods -c ctsm5.1.dev140 --skip-generate`
Later, I will need to update to the latest tag and rerun all the python tests.

samsrabin:
My notes from the CTSM SE standup 2023-07-17 re: [this discussion item](https://github.com/ESCOMP/CTSM/discussions/2059):

> Bill: probably not chdir'ing our of a temporary directory that is destroyed
> ```
> test_sys_mesh_modifier.py:        os.chdir(self._tempdir)  # cd to tempdir
> test_unit_run_sys_tests.py:        os.chdir(self._curdir)
> test_unit_run_sys_tests.py:        os.chdir(self._original_wd)
> ```
> 
> `test_sys_mesh_modifier.py` probably needs a `chdir` back to original

I also have a note that says it works with `make test` runs system vs unit tests in a different order from how `run_ctsm_py_tests` does it

ekluzek:
@TeaganKing @samsrabin has the solution above! 3 point swish for him!

I tried his suggestion out and it's now working for me. So let's still meet this afternoon and go over what to do for the final version.

TeaganKing:
Sounds great! And that's fantastic, thanks so much for sharing your previous experience and solutions to this issue, @samsrabin !

samsrabin:
All credit to @billsacks for the trace there!

python/ctsm/test/test_sys_run_neon.py
```diff
@@ -58,7 +58,7 @@ def test_one_site(self):
 
         # assert that BART directories were created during setup
         self.assertTrue("BART" in glob.glob(self._tempdir + "/BART*")[0])
-        #self.assertTrue("BART" in glob.glob("BART*")[0])
+        # self.assertTrue("BART" in glob.glob("BART*")[0])
```
slevis-lmwg, line 61: Do we need to keep this alternate line around as a comment, or could we remove? I'm fine either way if there's justification.

TeaganKing: Ah, yes, I can remove that! Thanks for catching this.

TeaganKing:
After discussing some additional changes with @ekluzek and @slevis-lmwg , I have made those changes and updated the ChangeLog. So, this is ready for a new tag!

.git-blame-ignore-revs
```diff
@@ -10,6 +10,9 @@ b88e1cd1b28e3609684c79a2ec0e88f26cfc362b
 b771971e3299c4fa56534b93421f7a2b9c7282fd
 9de88bb57ea9855da408cbec1dc8acb9079eda47
 8bc4688e52ea23ef688e283698f70a44388373eb
+4ee49e3e516ca7dee5df378f65664f93a7db4415
+0207bc98dd5c75cd69a0e788bc53e41093712f5c
```
slevis-lmwg, line 14: Reminder to add e4d38681df23ccca0ae29581a45f8362574e0630
@TeaganKing I'm happy to do this unless you get to it first.

slevis-lmwg:
Sorry, I forgot to ask something else:
Does this PR introduce a new /testinputs file? If so, is there anything that I need to do about that during the merge?

.git-blame-ignore-revs
TeaganKing: Done

TeaganKing:
No updates to `testinputs`; we just used pre-existing test inputs to perform all of these tests.

TeaganKing:
We will need to update the tag to `dev153` in `ChangeLog`.

slevis-lmwg:
make all passes (in a meeting with @TeaganKing and @ekluzek we decided to ignore pylint suggestions)

clm_pymods testing IN PROGRESS

PR #2091, opened 2023-08-09
Fixes to methane Tech Note

samsrabin:
### Description of changes

Various minor fixes to the methane section of the Tech Note:
- Added missing citation links
- Added missing equation references
- Corrected superscript syntax
- Corrected units in a table
- Corrected use of $\phi$ to $\varphi$

Equation references still missing:
- [x] "The gaseous concentration is calculated with Henry’s law as described in equation ."
- [x] "… with the same representation as Equation but with the gas diffusivity of oxygen."
- [x] "The solution to equation is solved in several sequential steps: …"

### Specific notes

Contributors other than yourself, if any: @jinmuluo

CTSM Issues Fixed: Item 3 in #1990.

Are answers expected to change (and if so in what way)? No

Any User Interface Changes (namelist or namelist defaults changes)? No

Testing performed, if any: Rendered HTML looks good.

PR #2157, opened 2023-09-14
Soil gas diffusivity bugfix

### LINKED ISSUE #1990: Problems about the soil gas diffusivity in methane code and nitrification-denitrification mod
Hi CTSM community,

I and my advisor have found some potential scientific problems in CLM and we hope to discuss these issues with you here.

**(1)** The formulation for diffusion appears incorrect in the methane code. This diffusion parameter taken from the methane code: (src/biogeochem/ch4Mod.F90,  line 3872) and explained in section 2.25.7.3. of the CLM tech note. 

https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Methane/CLM50_Tech_Note_Methane.html#aqueous-and-gaseous-diffusion

image

This appears to be coded incorrectly. The diffusion coefficient here is given as a linear combination of two terms weighted by (om_frac) and (1-om_frac) in the equation above. The equations for these terms are correctly given in the tech note.
From the code:

**eps=theta_a** 
**fa=theta_a/theta_s**

where **theta_a** is the air-filled pore space and **theta_s** is the total porosity. The code should be eps**(10./3.) instead of f_a**(10./3.)  in the second line. See 2.25.13 in the tech note.

**(2)** Correction to the nitrification, denitrification code. nitri-denitri code (src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90, line 283).
There are 2 corrections to how the diffusion is specified here to obtain the correct N2O emissions:
  (i) In the Parton formulation it is the normalized diffusion parameterization that is used (see Parsons et al., 1996): D/D_o.
The specification for the diffusion used presently, src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90, line 283 is taken from the CH4 code. However, the diffusion coefficient is not normalized here as it should be to give D/D_o
Thus, **(d_const_g(s,1) + d_const_g(s,2)^t_soisno_c)^1.e-4** should be omitted in the specification for diffusion only in the Nitrif-Denitrif module.
 (ii) The remainder of the equation should be corrected as detailed in (1) above in the Nitrif-DenitrifMod.

**(3)** CLM5 tech note correction. Table 25.2 gives the gas diffusivity in air in units of "m2 s-1". This should be “cm2 s-1”. For example, if the temperature is 273K, the O2 diffusivity in air is around 0.4762 m2 s-1 from the table. Standard values are about 10^-4 smaller. Values of these constants in the code seem to be correct so this only applies to the tech note.

Best,

Jinmu Luo

samsrabin:
### Description of changes
Resolves two bugs in the calculation of diffusion in `SoilBiogeochemNitrifDenitrif()`. Also does some rearranging and renaming to improve clarity.

### Specific notes
From #1990:
- 3dfa6aa resolves issue 1
- 8e65ee1 resolves issue 2
- (Issue 3 is resolved in #2091.)

Contributors other than yourself, if any: @jinmuluo, @wwieder 

CTSM Issues Fixed:
- Fixes #1990 

Are answers expected to change (and if so in what way)? **Yes.** Of outputs in aux_clm tests, only `F_N2O_DENIT` is affected.

#### Testing performed:
aux_clm:
1. 7d2979c (several commits comprising bug fixes and comments) shows diffs from parent `ctsm5.1.dev133` only in history field `F_N2O_DENIT`. (At least on Cheyenne; my Izumi test has been wiped. Can rerun if needed.)
2. 728b7e6 adds some parentheses to simulate upcoming order-of-operations changes. Diffs from 7d2979c, but again (for Cheyenne), only in `F_N2O_DENIT`.
3. Commits through 3695c60 do some rearranging and renaming. Bit-for-bit identical to 728b7e6.


samsrabin:
@jinmuluo Would you mind having a look at the changes here to make sure they all look good?

src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90
```diff
@@ -168,10 +168,11 @@ subroutine SoilBiogeochemNitrifDenitrif(bounds, num_soilc, filter_soilc, &
     real(r8) :: mu, sigma
     real(r8) :: t
     real(r8) :: pH(bounds%begc:bounds%endc)
+    real(r8) :: D0  ! temperature dependence of gaseous diffusion coefficients
     !debug-- put these type structure for outing to hist files
     real(r8) :: co2diff_con(2)                      ! diffusion constants for CO2
-    real(r8) :: eps
-    real(r8) :: f_a
+    real(r8) :: fc_air_frac ! Air-filled fraction of soil volume at field capacity
```
wwieder, line 174: are we trying to line up comments for variable descriptions any more?

src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90
```diff
@@ -278,9 +282,20 @@ subroutine SoilBiogeochemNitrifDenitrif(bounds, num_soilc, filter_soilc, &
...
+
+               ! Diffusivity after Millington & Quirk (1961)
+               ! Eq. 9 in Riley et al. (2011, Biogeosciences)
+               diffus_millingtonquirk = fc_air_frac**(10._r8/3._r8) / watsat(c,j)**2
+
+               ! First, get diffusivity as a unitless constant, which is what's needed to
+               ! calculate ratio_k1 below.
+               diffus (c,j) = &
+                    (om_frac        * diffus_millingtonquirk + &
+                    (1._r8-om_frac) * diffus_moldrup )
```
wwieder, line 298: I like how this breaks out the two diffusing calculations following equations in the paper, but what happened to the first part of the diffus equation `(d_con_g(2,1) + d_con_g(2,2)*t_soisno(c,j)) * 1.e-4_r8 *`?

wwieder: Oh, I see it's added with calculation of `D0`, below.

wwieder: Now I'm confused because  `D0` is being calculated on the `else` side of the if `(use_lch4)` logical?  Is the intended?

wwieder (commented):
Have you run tests to evaluate the impact of this bug fix on fluxes simulated in the model?

src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90
samsrabin: I thought we decided that was more trouble than it's worth; see "Lower priority" section of [notes from 2023-09-07](https://docs.google.com/document/d/1Qn9c7NiGxYMN3p47Fpa0QINLwwxzRpGXG61GsuGjVM0/edit#heading=h.3vr45vczwma).

src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90
samsrabin: `D0` is actually being calculated at line 389, which is _after_ that if-else block (ends at line 313). The code at line 390 needs to happen after the calculation of `ratio_k1` at line 381.

In the "else" case (i.e., `use_lch4` false), `diffus` is zero by line 390 so it doesn't matter what `D0` is.

samsrabin:
@wwieder If you're asking about scientific tests of the impact, no, I haven't. I do have all the (Cheyenne) aux_clm outputs, though, so we could look at those.

wwieder:
I think it's worth running a longer case to understand the scientific impacts on the bug fix, or is the answer changing part of this just due to roundoff changes in the order of operations?

src/soilbiogeochem/SoilBiogeochemNitrifDenitrifMod.F90
wwieder: Yeah, when I see what that actually looks like in practice I wonder if it's the right decision? 
- I thought when @ekluzek raised the question it was WRT going through and correcting existing code that's not being modified (except for the vertical alignment).  Avoiding this seems OK to prevent scope creep on our work.  
- Introducing new text that's not aligned (especially in code blocks that are otherwise tidy and easy to read) seems misguided.
Just my 2-cents, but I suggest aligning this now, merging the PR and discussing the topic at an SE meeting.  This is bleeding into a discussion about 'reflections on our process', with is outside the scope of a bug fix.

samsrabin: Done! (2bb68ad)

samsrabin:
A longer test makes sense, @wwieder, since the diffs aren't just roundoff-level. How long do you think, and at what resolution?

wwieder:
My default is for a full spinup from cold start and then historical simulation at 1 degree, but lately we've been testing things out more with the sparse grid from the PPE.  What do you think @slevis-lmwg, @dlawrenncar and @linniahawkins?

slevis-lmwg:
@wwieder I don't have first-hand experience with the PPE sparse grid; however, it seems like a reasonable time-saving approach to use it.

linniahawkins:
Probably depends on the application - for global scales and primary variables the sparse grid is quite accurate but issues may arise at regional scales or for specific PFTs. A practical issue might be that it doesn't play nice with NCL, Daniel has some nice python code to project the sparse grid back to full grid, but it adds an extra step. Here's a visual of where the 400 cells are located (blue points): 
Sparsegrid_cell_locations

samsrabin:
Thanks, @linniahawkins! @wwieder, looks good to me; what do you think?

wwieder:
Sorry, what are you asking here, @samsrabin?  If it's OK to evaluate effects of the bug fix on ecosystem states and fluxes using the sparse grid instead of a full global simulation?   I think this seems like a reasonable way to quantify the magnitude of the bug fix.  Or at least to try... My question more on the process side of things, (e.g. how to we understand / quantify answer changing tags & can we leverage the sparse grid)?  I don't think we can use the PPE branch tag here, as it's not up to date with main, but you still should be able to use the sparse grid surface dataset.

At some point I'd be curious to have a closer look at differences between Daniel's extrapolation from the 400 points back to the globe and a gridded simulation, but that's not really the question here.

samsrabin:
Marking as blocked because I want to run with the PPE sparse grid, but sparse grids are incompatible with NUOPC.

dlawrenncar:
Just so I understand fully, are sparse grids completely incompatible with
NUOPC or just incompatible at the moment until someone resolves the issue?

samsrabin:
@dlawrenncar Just for now. @adrifoster, @slevis-lmwg, and @ekluzek are working on adding the capability.

ekluzek:
@dlawrenncar see #1731 to see where we are at in the process. I did get a test case working with an earlier version of this. But, we didn't bring it in as something that we test. And since that time mesh_modifier was completed and will be the way we make the process replicatable.

Once, @adrifoster has this working, we will save it as a user-mod directory and start testing with it. And we will update the notes on the process to make sure we have a process that can be replicated. @adrifoster also found a file in the PPE work that we can base these files off of.

slevis-lmwg:
> @dlawrenncar Just for now. @adrifoster, @slevis-lmwg, and @ekluzek are working on adding the capability.

Generally speaking, it works, e.g. Xiulin's work with Charlie and Lara. 
The workflow involves a somewhat long list of steps. 
We (@adrifoster @ekluzek @slevis-lmwg) are working on implementing on the PPE sparse grid now.

samsrabin:
I'll go ahead and test with a 4x5 grid instead of waiting for the NUOPC sparse grid.

samsrabin:
It looks like this results in a reduction of mean global diffusivity by about 12.7%:
screenshot_6517

This ranges from about 0–35% across gridcells in two years I looked at:
screenshot_6518

@wwieder Because of how quickly diffusivity levels out, it doesn't seem necessary to do another run but from a spinup with the fix applied. What do you think?

wwieder:
I agree, seems like a pretty minor difference.  What do trace gas fluxes
look like (CO2, CH4, Denitrif, or N2O)?  I'm assuming they also aren't
large and this can be merged.

samsrabin:
The only variable that had any diffs in (Izumi tests of) aux_clm was `F_N2O_DENIT`. This has larger relative differences, with the new code resulting in a greater flux, especially in the boreal zone:
screenshot_6519

screenshot_6520

wwieder:
Great.  I think this is good to merge.  Thanks, Sam.

jinmuluo:
Hi Sam,

I've seen the same changes in my outputs.

Best,


will wieder ***@***.***> 于2023年10月26日周四 19:57写道：

PR #2158, opened 2023-09-14
Add is_doy_in_interval() function

samsrabin:
### Description of changes

This PR adds a function, `is_doy_in_interval()`, that checks whether a given date (in day-of-year, DOY, format) is inclusive-within an interval given by start and end dates (also DOY).

### Specific notes

This is in preparation for upcoming work on crop calendars.

I considered doing something fancier, using the ESMF date/time types, but eventually decided to just replicate the existing behavior of `CropPhenology()`. Specifically, there's no special consideration of leap days. This has two effects:
1. Intervals that include a leap day end, in leap years, on the calendar day before they end in non-leap years.
2. Intervals that occur entirely after a leap day are, in leap years, shifted one calendar day earlier than they occur in non-leap years.

I don't think this behavior is necessarily a bug, anyway.

CTSM Issues Fixed: **None.**

Are answers expected to change (and if so in what way)? **No.**

Testing performed: **`aux_clm` tests all pass bit-for-bit (except for expected failures).**

### DIFF
--- a/src/utils/clm_time_manager.F90
+++ b/src/utils/clm_time_manager.F90
@@ -59,6 +59,7 @@ module clm_time_manager
         is_beg_curr_year,         &! return true on first timestep in current year
         is_end_curr_year,         &! return true on last timestep in current year
         is_perpetual,             &! return true if perpetual calendar is in use
+        is_doy_in_interval,       &! return true if day of year is in the provided interval
         is_near_local_noon,       &! return true if near local noon
         is_restart,               &! return true if this is a restart run
         update_rad_dtime,         &! track radiation interval via nstep
@@ -1756,6 +1757,44 @@ contains
     is_perpetual = tm_perp_calendar
 
   end function is_perpetual
+
+  !=========================================================================================
+
+  logical function is_doy_in_interval(start, end, doy_in)
+
+    ! Return true if day of year is in the provided interval.
+    ! Does not treat leap years differently from normal years.
+    ! Arguments
+    integer, intent(in) :: start ! start of interval (day of year)
+    integer, intent(in) :: end ! end of interval (day of year)
+    integer, optional, intent(in) :: doy_in ! day of year to query
+    
+    ! Local variables
+    integer :: doy
+    logical :: window_crosses_newyear
+
+    character(len=*), parameter :: sub = 'clm::is_doy_in_interval'
+
+    ! Get doy of beginning of current timestep if doy_in is not provided
+    if (present(doy_in)) then
+       doy = doy_in
+    else
+       doy = get_prev_calday()
+    end if
+
+    window_crosses_newyear = end < start
+
+    if (window_crosses_newyear .and. &
+        (doy >= start .or. doy <= end)) then
+       is_doy_in_interval = .true.
+    else if (.not. window_crosses_newyear .and. &
+        (doy >= start .and. doy <= end)) then
+       is_doy_in_interval = .true.
+    else
+       is_doy_in_interval = .false.
+    end if
+    
+  end function is_doy_in_interval
 
   !=========================================================================================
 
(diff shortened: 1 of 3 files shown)


src/utils/clm_time_manager.F90
```diff
@@ -1759,6 +1760,44 @@ end function is_perpetual
...
   !=========================================================================================
 
+  logical function is_doy_in_interval(start, end, doy_in)
+
+    ! Return true if day of year is in the provided interval.
+    ! Does not treat leap years differently from normal years.
+    ! Arguments
+    integer, intent(in) :: start ! start of interval (day of year)
+    integer, intent(in) :: end ! end of interval (day of year)
+    integer, optional, intent(in) :: doy_in ! day of year to query
```
billsacks, line 1770: - [ ] Do you anticipate using the 'optional' aspect here? If just "maybe", I'd suggest making this required because having it optional makes it possible to accidentally misuse the function (forgetting to specify `doy_in`, but it compiles and runs... but does the wrong thing). If yes, then please add some documentation up here in the argument list saying what happens if doy_in isn't provided. But actually, maybe a less error-prone way to handle this would be: make doy_in required, then introduce a new function, `is_today_in_interval(start, end)` which simply calls `is_doy_in_interval(start, end, get_prev_calday())`

src/utils/test/clm_time_manager_test/test_clm_time_manager.pf
```diff
@@ -577,4 +577,35 @@ contains
...
+    @assertFalse(is_doy_in_interval(start, end, 200))
+    @assertTrue(is_doy_in_interval(start, end, 35))
+    @assertTrue(is_doy_in_interval(start, end, 350))
+
+    start = 300
+    end = 300
+    @assertTrue(is_doy_in_interval(start, end, start))
+    @assertTrue(is_doy_in_interval(start, end, end))
+    @assertFalse(is_doy_in_interval(start, end, 200))
+    @assertFalse(is_doy_in_interval(start, end, 350))
```
billsacks, line 607: This is a great set of unit tests of this function – I'm glad you added all of these to cover a variety of conditions.

I don't know whether to make this next comment, but I'll go ahead and make it and then walk it back a little afterwards: I tend to subscribe to the (common, I think) unit testing guidance of having each test just test a single scenario. The advantages of this are: (1) if you get a failure, the test is (hopefully) clearly named so that you can see based on the failure exactly what condition failed; and (2) all scenarios are run, in contrast to a situation like this one where the first failed assertion will cause the test to bail, so you have a less clear picture of what works and what doesn't. That said, I have come to feel less strongly about this over time, and in this particular case the (small but still non-zero) overhead of splitting each test into its own subroutine may not be worth it. So my overall feeling is: we should strive to follow the one-scenario-per-test guidance in general, but there are always times when it's worth breaking a rule, and this might be one of those times.

billsacks (changes requested):
This looks great. I agree that using the ESMF methods would have been overkill for this usage. I have one small request along with another comment that is more like rumination than request.

src/utils/clm_time_manager.F90
samsrabin: Good thinking! I added (7b57111) a new function, `is_today_in_doy_interval(start, end)` and am now requiring the third argument to `is_doy_in_interval()`.

src/utils/test/clm_time_manager_test/test_clm_time_manager.pf
samsrabin: I definitely see what you're saying, but since you've provided me an out from having to split this into 15 separate tests (plus 6 more for `check_is_today_in_doy_interval()`), I'm taking it! (I'm sure I could reduce the number of tests without compromising coverage, but I'd rather err on the side of more tests than necessary.)

However, I did split `check_is_doy_in_interval()` into three separate tests based on the relative values of `start` and `end` (eded771). Hopefully that's a good compromise.

ekluzek (approved):
@billsacks already approved so this is good to go. But, it's short enough for me to also give a thumbs up as well.

Issue #1878, opened 2022-10-25
Convert wetlands to bare ground


billsacks:
To remove one source of negative runoff, we are planning to use bare ground in place of wetlands. (Note that wetlands are currently just used to represent ocean points: we currently don't run with inland wetlands.)

The original proposal was to change the surface datasets to specify bare ground rather than wetlands. However, I realized that that plan would make it impossible to do the rigorous treatment of coastal areas described in #1716 . So my plan is to add a flag that allows making this change at runtime. The other advantage of this approach is that this conversion can be turned on or off with the same surface datasets.

One other small change I'll make is to remove the code in mksurfdata that forces soil properties to hard-coded default values in wetland grid cells. This forcing seems unnecessary since soil properties are already given the same default values if no data are found for a given output cell; by removing this forcing for wetland grid cells, we will be able to use any existing soil data for a grid cell where soil properties are available despite the grid cell being considered ocean by the PFT data.

@ekluzek @wwieder @slevisconsulting

billsacks:
@ekluzek @wwieder @slevisconsulting - I have a few questions:

1. Are you okay with my proposal of doing this at runtime rather than on the surface datasets?

2. Do you agree with the idea that the default for this behavior should depend on the model version, so that CTSM52 uses the new convert-wetlands-to-bare-ground behavior but earlier versions remain unchanged? Or should we change this for all physics versions? If this will depend on model version, it looks like I'll need to add a new 5_2 phys option and compset (at a glance these don't seem to have been added yet).

3. I thought I'd do this on the ctsm5.2.mksurfdata branch, since it will involve some minor changes to mksurfdata. Do you agree? If so, is there any standard testing being done for changes on that branch?

ekluzek:
1. I agree with the plan to remove this out of the datasets, to the model so that the solution is flexible. It's so much effort to create a new set of datasets for all resolutions, run time flexibility makes this much better.

2. We had proposed that the datasets would be updated for all physics versions, so we could do this either way. We might need a clm5_2 physics option anyway, so nothing wrong with adding that. Although if ctsm5.2 is just the new datasets and we change the datasets for all versions -- maybe ctsm5.2 doesn't really mean much. So we could just do a couple ctsm5.2 tags and then make a new tag for ctsm5.3 where we start changing answers beyond surface datasets.
3. Yes, that makes the most sense to put it on the branch
[...]

PR #2164, opened 2023-09-15
Distinguish between ocean and wetland in fsurdat files

### LINKED ISSUE #1886: Distinguish between ocean and true wetlands on surface dataset
Currently we use the wetland landunit to represent ocean points as well as true (inland) wetlands. This could cause problems for the changes in #1878 as well as for those in #1716. For the sake of those changes and possibly other reasons, we should distinguish between ocean points and true wetlands on the surface dataset by adding a new PCT_OCEAN field and using that rather than the PCT_WETLAND field for points where we don't have any land cover data. At runtime these could still be combined into the wetland landunit (though with the changes for #1878 ocean would instead be changed to bare ground), but separating them would allow us to distinguish between these two very different land covers for #1878 and #1716 – and we could possibly also have some metadata describing the separation of the wetland landunit into ocean vs. true wetland for each grid cell.

Currently we (by default) do not include inland wetlands on the surface dataset, so this isn't a problem, but it would become problematic if we started including inland wetlands again.

slevis-lmwg:
### Description of changes

Add pctocn to mksurfdata_esmf and write PCT_OCEAN to fsurdat files.

### Specific notes

@billsacks @ekluzek @mvertens 

Fixes #1886 

No

fsurdat contains PCT_OCEAN now

Testing performed, if any:
Code worked to generate the following fsurdat file `/glade/work/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf/surfdata_1.9x2.5_hist_78pfts_CMIP6_1850_c230914.nc`
I will also need to change clm code to successfully read the new fsurdat file.


slevis-lmwg:
Note to self that I need to run all python tests for approval of this PR (in addition to aux_clm).

slevis-lmwg:
Currently seeing diffs in a single grid cell (ncview indices (i,j) = (20,68) and (lat, lon) = (38.8,50)
I haven't identified a reason, yet, so I will next resort to write statements in the code...

src/main/surfrdMod.F90
```diff
@@ -475,9 +481,9 @@ subroutine surfrd_special(begg, endg, ncid, ns)
 
     topo_glc_mec(:,:) = max(topo_glc_mec(:,:), 0._r8)
 
-    pctspec = pctwet + pctlak + pcturb_tot + pctgla
+    pctspec = pctwet + pctlak + pcturb_tot + pctgla + pctocn
```
slevis-lmwg, line 484: In reviewing this with @swensosc:

Sean asked whether this code should be backwards compatible with previous fsurdat files that do not include pct_ocean.

As written, the code is not backwards compatible, and maybe we need to consider this.

src/main/surfrdMod.F90
```diff
@@ -497,22 +503,26 @@ subroutine surfrd_special(begg, endg, ncid, ns)
...
 
-       wt_lunit(nl,istdlak)     = pctlak(nl)/100._r8
+       wt_lunit(nl,istdlak) = pctlak(nl) / 100._r8
 
-       wt_lunit(nl,istwet)      = pctwet(nl)/100._r8
+       ! Until ctsm5.1 we would label ocean points as wetland in fsurdat
+       ! files. Starting with ctsm5.2 we label ocean points as ocean
+       ! (always 100%) and wetland points as wetland. Here we merge them
+       ! again to keep model behavior unchanged for now.
+       wt_lunit(nl,istwet) = (pctocn(nl) + pctwet(nl)) / 100._r8
```
slevis-lmwg, line 512: @swensosc:
As written now this exactly replicates what the code was doing before introducing pct_ocean.
What we want is
`wt_lunit(nl, istocn) = pctocn(nl) / 100._r8`
...and throughout the code we want to change istwet to istocn, not everywhere, but where Bill made changes for #1878 .

tools/mksurfdata_esmf/src/mksurfdata.F90
```diff
@@ -1222,12 +1235,12 @@ subroutine normalize_and_check_landuse(ns_o)
          ! Confirm that we have done the rescaling correctly: now the sum of all landunits
          ! should be 100% within tol_loose
          suma = pctlak(n) + pctwet(n) + pctgla(n) + pcturb(n) + pctcft(n)%get_pct_l2g()
-         suma = suma + pctnatpft(n)%get_pct_l2g()
+         suma = suma + pctnatpft(n)%get_pct_l2g() + pctocn(n)
```
slevis-lmwg, line 1238: Here and in line 1267, merge the two separate suma lines.

src/main/surfrdMod.F90
slevis-lmwg: @swensosc suggested starting at subr. apply_convert_ocean_to_land in surfrdUtilsMod.F90 [in Bill's PR merge](https://github.com/ESCOMP/CTSM/commit/075820c1902e19a586a1bdb8e31655123048360f#diff-2cf8bfd0e51dc881e54df728b909f59a91eafdefb381760c38f970f9ccbedd43)

src/main/surfrdMod.F90
slevis-lmwg: > In reviewing this with @swensosc:
> 
> Sean asked whether this code should be backwards compatible with previous fsurdat files that do not include pct_ocean.
> 
> As written, the code is not backwards compatible, and maybe we need to consider this.

@ekluzek @billsacks 
Sean and I didn't have strong opinions on this. I, in particular, was not worried about backwards compatibility between ctsm5.2 fsurdat files and ctsm5.1 code, because we already need a short list of hacks for such backwards compatibility, so this could be added to the list if necessary.

I am interested in your feedback, and I can raise the question on Thursday morning if you prefer.

ekluzek: I agree -- we really can't use CTSM5.2 fsurdat files with CTSM5.1 or vice versa. So I don't want to worry about backwards compatibility. It always adds complexity. and it's complexity that becomes less useful over time.

@swensosc is right to bring it up though, because we should think about it. And in terms of CTSM5.1 datasets we did try to keep them backwards compatible. And I think we'll try to do the same in the future, until we have another major change in datasets.

I also think that having PCT_OCEAN on the datasets is a real improvement in making the datasets understandable to users. PCT_WETLAND made users assume it really meant inland wetlands, and certainly not actually ocean. So these datasets with PCT_OCEAN is much easier for users to understand.

billsacks: I agree with Erik regarding backwards compatibility - specifically, skipping it in this instance.

slevis-lmwg:
Notes to self:
While working on izumi, I will try this test when looking for bfb answers:
`SMS.f10_f10_mg37.I2000Clm51BgcCrop.izumi_gnu.clm-crop`
First generate new fsurdats, one without ocean and one with ocean.

slevis-lmwg:
My latest modifications have not changed answers.
@swensosc would you like to look at my last two commits for confirmation that I did what you suggested?
If you have no further comments, this PR can be merged with the ctsm5.2.mksurfdata branch.

src/main/landunit_varcon.F90
```diff
@@ -18,9 +18,10 @@ module landunit_varcon
 
   integer, parameter, public :: istsoil    = 1  !soil         landunit type (natural vegetation)
   integer, parameter, public :: istcrop    = 2  !crop         landunit type
-  ! Landunit 3 currently unused (used to be non-multiple elevation class glacier type:
-  ! istice, and landunit 4 was istice_mec; now they are combined into a single landunit
-  ! type, 4)
+  integer, parameter, public :: istocn     = 3  !ocean        landunit type
+  ! Landunit 3 used to be non-multiple elevation class glacier type, istice
+  ! Landunit 4 was istice_mec; now landunit 4 and old landunit 3 are combined
+  ! into landunit 4
```
slevis-lmwg, line 24: Reviewing with @ekluzek:

Remove these comments that refer to past code history.

slevis-lmwg:
> Run all python tests
> 
> * aux_clm will fail so no need to test it
> * Unit tests failing
>   -- replace files in testinputs, see /tools/modify_input_files/bkups/bkup_of_new_items for cfg files I may have used
>   Don't think I lost mods I made in python/ctsm/test/test_unit_fsurdat_modifier.py?
>   -- ask Erik about creating ...1x1_mex...
> * Sys tests failing
> * pymod not tested, yet
> 
> Generate new (DONE) and old (NEXT) 1-deg fsurdat and landuse files and test.

@swensosc thanks for your email (@swensosc looked at my code and was ok with my mods).
@swensosc suggested an additional test, where we work with a raw dataset that includes inland wetlands and confirm that everything works as intended. I like the suggestion. First though I plan to resolve all the python test failures.

slevis-lmwg:
Updating to latest ctsm5.2.mksurfdata...
Conflict in python/ctsm/test/testinputs/modify_fsurdat_short.cfg suggests that I may need to remake this file 
`ctsm/test/testinputs/surfdata_5x5_amazon_hist_16pfts_CMIP6_2000_c231031.nc` to a 78-pft file, but we'll see if tests pass.

slevis-lmwg:
PASS: make all
FAIL: clm_pymods same expected fails as before (see [earlier post](https://github.com/ESCOMP/CTSM/pull/2164#issuecomment-1721971956))

PR #2176, opened 2023-09-27
Generate anomaly forcing files for historical period

samsrabin:
### Description of changes
Extends `ssp_anomaly_forcing_smooth` (i.e., the script that generates anomaly files allowing GSWP3 forcings to be adjusted to look more like CAM) to produce files for 1850–2014 or 1850–2100. This helps avoid inconsistencies between a run whose historical period was simulated with the unadjusted GSWP3 data but whose future period forcings are adjusted.

### Specific notes
When calling `ssp_anomaly_forcing_smooth`, instead of providing an integer [1, 4] as the first argument, provide a string. Valid values: `hist`, `(hist-)ssp126`, `(hist-)ssp245`, `(hist-)ssp370`, `(hist-)ssp585` (as well as any capitalization and some punctuation). Values without `hist` reproduce original SSP-only outputs. 

Other notes:
- 1850–2100 files (generated with, e.g., period `hist-ssp245`) are _not_ identical in the 1850–2014 period to the files generated with period `hist`. Similarly, the `hist-ssp245` and `hist-ssp585` (e.g.) files differ in the historical period. Differences arise as soon as the first timestep. I'm not sure why this is, but I think it has to do with the boxcar averaging and/or smoothing steps.
- Specifying only `hist` requires additional specification of `--climo-year`, because default 2015 is outside historical period.
- Also changes the format to which the output netCDF is initially written, avoiding "stalling out" for me as file grows in size. Requires that the shell opened by `subprocess.Popen()` support `module load nco`.

CTSM Issues Fixed (include github issue #): **None**

Are answers expected to change (and if so in what way)? **No.**

**Testing performed**
File produced for period `ssp370` is, according to `nccmp -dfs`, identical to `/glade/p/cesmdata/cseg/inputdata/atm/datm7/anomaly_forcing/CMIP6-SSP3-7.0/af.allvars.CESM.SSP3-7.0.2015-2100_c20220628.nc`.


samsrabin:
- [ ] Add ability to specify a single historical ensemble member to use, either for the period itself or as starting point of SSP ensemble members.

samsrabin:
Not needed! `ssp_anomaly_forcing_smooth` generates CAM anomalies for the SSP period _relative to historical-period CAM.

Will submit a smaller PR with the netCDF write format change.

PR #2181, opened 2023-10-02
ssp_anomaly_forcing write fix

samsrabin:
### Description of changes
For me, on Casper, `ssp_anomaly_forcing` would stall out after a few variables were written. This PR solves the issue by writing first to netCDF-4 and then doing `nccopy` to convert to CDF5.

### Specific notes

Minor other improvements:
- Removes unused variables.
- `num_ens` now automatically set.
- Can now specify `--output-dir`; default `./anomaly_forcing/` reproduces old behavior.
- Fixes bug where `--write_climo` file paths were missing a `/`.

Contributors other than yourself, if any: **None**

CTSM Issues Fixed: **None**

Are answers expected to change (and if so in what way)? **No**

### Testing performed
According to `nccmp -gf`,
- SSP1-2.6 output file is identical to before, aside from (as expected) `Created_on`, `gitdescribe`, and `Created_by` attributes.
- SSP1-2.6 `--write_climo` outputs are identical to before.


samsrabin:
@swensosc, would you mind looking this over and giving it a try?

ekluzek:
Unfortunately this is something I encouraged @samsrabin to go down, because of my misunderstanding of how the ssp_anomaly_forcing setup works. I misremembered how it works. I think I knew this at one point, but led @samsrabin down the wrong path. It doesn't make GSWP3 forcing look like CAM -- it applies the CAM anomaly from historical to SSP to make GWSP look like a future scenario. So the natural way to run is to run GSWP3 forcing and then go into the SSP period. So there is no reason to do the anomaly for historical. That could be something useful as a different thing, but isn't how this is currently setup. But, this is my fault rather than @samsrabin, I wanted to make sure the blame was given to me.

@samsrabin did get an improvement here in terms of performance, so he will likely bring that in, but not the major thrust of this PR.

samsrabin:
@ekluzek This PR is actually just the improvements; I've closed the previous PR (#2176).

samsrabin:
From CTSM SE meeting: @wwieder will test.

tools/contrib/ssp_anomaly_forcing_smooth
```diff
@@ -196,25 +199,15 @@ if num_ens != len(histdir):
     print("number of ensemble members not the same")
     sys.exit("number of members different")
 
-# test w/ 1 ensemble member
-num_ens = 3
-
 # Setup output directory
-sspoutdir = "anomaly_forcing/CMIP6-" + ssptag
+sspoutdir = "CMIP6-" + ssptag
```
wwieder, line 203: fwiw I like this anomaly_forcing subdirectory that was created previously

tools/contrib/ssp_anomaly_forcing_smooth
```diff
@@ -41,14 +41,21 @@ parser.add_argument(
     action="store_true",
     default=False,
 )
+parser.add_argument(
+    "--output_dir", "--output-dir",
+    help="Top-level output directory (default: ./anomaly_forcing/). Sub-directory will be created for the selected scenario.",
+    type=str,
+    default=os.path.join(".", "anomaly_forcing"),
+)
+
```
wwieder, line 50: this anomaly_forcing directory wasn't appended to my output-dir.  should it hav been?

tools/contrib/ssp_anomaly_forcing_smooth
```diff
@@ -244,6 +237,17 @@ field_out_wind = ["uas", "vas"]
...
+# Instead, write to netCDF-4 and then convert to CDF5.
+output_format = "NETCDF4_CLASSIC"
+def convert_to_cdf5(outfilename):
+    outfilename_cdf5 = outfilename.replace(".nc4", ".nc")
+    subprocess.Popen(
+        f"module load nco && nccopy -k 'cdf5' {outfilename} {outfilename_cdf5} && rm {outfilename}",
+        stdout=subprocess.PIPE,
+        shell="True",
+    )
+
```
wwieder, line 250: This also doesn't seem to have happened, as intended.  See /glade/scratch/wwieder/CMIP6-SSP1-2.6

wwieder (commented):
Minor comments, otherwise I was able to get this to run.

tools/contrib/ssp_anomaly_forcing_smooth
samsrabin: `./anomaly_forcing` is just the default. I didn't want to append that automatically to whatever is specified as `output_dir` because I figured some people might want to name their directory something else.

tools/contrib/ssp_anomaly_forcing_smooth
samsrabin: (See [above](https://github.com/ESCOMP/CTSM/pull/2181/files#r1380747997).)

samsrabin:
@wwieder, could you give this another try?

wwieder:
seems like the code gets stuck

```
Beginning ensemble number  3
b.e21.BHIST.f09_g17.CMIP6-historical.004.clm2.h0.WIND.185001-201412.nc
b.e21.BSSP585cmip6.f09_g17.CMIP6-SSP5-8.5.103.clm2.h0.WIND.201501-206412.nc
b.e21.BSSP585cmip6.f09_g17.CMIP6-SSP5-8.5.103.clm2.h0.WIND.206501-210012.nc
uas
vas
```

The SSP3 case seemed to get hung up at pbot too.  I'm not sure if they finished writing as expected, but the files are large.
/glade/scratch/wwieder/CMIP6-SSP3-7.0

My command was ./ssp_anomaly_forcing_smooth 3 --output_dir /glade/scratch/wwieder in bash with ctsm_pylib activated.

samsrabin:
Seems like trying to switch to `xarray` just gave you the issue I experienced with `netCDF4`. I think I will just live with this and ask for help if I ever need to run this script. I'm closing this PR and replacing it with one for refactoring only: #2235.

PR #2192, opened 2023-10-10
Add sowing window input files

samsrabin:
### Description of changes
Previously, one could run crops with either (a) sowing windows defined by the hemisphere-specific start and end dates on the paramfile or (b) prescribed sowing dates specified by input file `stream_fldFileName_sdate`. This PR replaces the latter with two new input files, `stream_fldFileName_swindow_start` and `stream_fldFileName_swindow_end`.

### Specific notes
- If a gridcell-crop has the same date in both sowing window input files, that date is treated as a prescribed sowing date. This means that the usual "Can we plant in this sowing window?" rules are discarded. It may be desirable to add an option to treat those as one-day-long sowing windows instead.
- Adds `sowingDates` testmod, which uses sowing windows derived from the static GGCMI phase 3 crop calendar dataset.

**Any User Interface Changes (namelist or namelist defaults changes)?** Removes namelist parameter `stream_fldFileName_sdate`. Adds namelist parameters `stream_fldFileName_swindow_start` and `stream_fldFileName_swindow_end`.

**Testing performed, if any:** A version of 862f02c [with some additional verbosity](https://github.com/samsrabin/CTSM/compare/862f02c..7a5afde) passed `aux_clm` against `ctsm5.1.dev142`. `RXCROPMATURITY` test also works.


samsrabin:
Something went wrong in the merge of 142. Closing to replace with a better version.

PR #2193, opened 2023-10-10
Add sowing window input files

samsrabin:
### Description of changes
Previously, one could run crops with either (a) sowing windows defined by the hemisphere-specific start and end dates on the paramfile or (b) prescribed sowing dates specified by input file `stream_fldFileName_sdate`. This PR replaces the latter with two new input files, `stream_fldFileName_swindow_start` and `stream_fldFileName_swindow_end`.

### Specific notes
- By default, CLM still uses the same old paramfile values. Eventually I hope to use more realistic sowing window input files by default, but there's a lot of work still to be done in that vein.
- If a gridcell-crop has the same date in both sowing window input files, that date is treated as a prescribed sowing date. This means that the usual "Can we plant in this sowing window?" rules are discarded. It may be desirable to add an option to treat those as one-day-long sowing windows instead, or maybe to only treat as prescribed sowing date if _all_ crops in all gridcells have the same start and end date.
- Adds `sowingDates` testmod, which uses sowing windows derived from the static GGCMI phase 3 crop calendar dataset.
- Includes unit testing.

**Any User Interface Changes (namelist or namelist defaults changes)?** Removes namelist parameter `stream_fldFileName_sdate`. Adds namelist parameters `stream_fldFileName_swindow_start` and `stream_fldFileName_swindow_end`.

**Testing performed, if any:** A version of 1cade99 [with some additional verbosity](https://github.com/samsrabin/CTSM/compare/1cade99..7a5afde) passed `aux_clm` against `ctsm5.1.dev142`. `RXCROPMATURITY` test also works.


slevis-lmwg (changes requested):
During our paired review of this PR, @samsrabin kept notes that he will address.

samsrabin:
Tasks from code review with @slevis-lmwg:

`cropcalStreamMod`:
- [ ] Remove/rework `! If read-in value is invalid, allow_unprescribed_planting in CropPhenology()` comment
- [ ] Throw error if invalid start or end sowing window date, instead of setting to -1. Or at least warn
- [ ] Don't hard-wire 365.
- [ ] Don't allocate `rx_swindow_starts_thisyr_patch` and `rx_swindow_ends_thisyr_patch` twice.

`CNPhenologyMod`:
- [ ] `! Ensure that a window was found`: Add comment that this is not currently reachable, but having it here will catch more things in the case of future code changes.
- [ ] In `test_was_sown_in_this_window_startend` and partner: Add comments explaining logic for why we have assertTrue and assertFalse for some tests.
- [ ] In `do_plant_prescribed`, replace `sowing_count (p) < mxsowings` with check of whether was sown in current window
- [ ] `! Save these diagnostic variables only on the first day of the window`: Actually **last** day of window

Misc.:
- [ ] Add input validity check to `get_doy_tomorrow()` (and add unit test for error)
- [ ] Add note in `namelist_definition_ctsm.xml`, descriptions of sowing window start and end date parameters: If a cell has the same start and end dates, the normal climate checks are bypassed.

samsrabin:
8d1537c broke RXCROPMATURITY; investigating.

samsrabin:
@slevis-lmwg Ready for re-review! Everything should be self-explanatory, but if you'd like to go through the diffs together, let me know.

src/utils/clm_time_manager.F90
```diff
@@ -1274,6 +1275,29 @@ end function get_calendar
...
+  function get_doy_tomorrow(doy_today) result(doy_tomorrow)
+
+    !---------------------------------------------------------------------------------
+    ! Given a day of the year (doy_today), return the next day of the year
+
+    integer, intent(in) :: doy_today
+    integer             :: doy_tomorrow
+    character(len=*), parameter :: sub = 'clm::get_doy_tomorrow'
+
+    if ( doy_today < 1 .or. doy_today > 367 )then
```
slevis-lmwg, line 1287: Why 367? Is it 367 in leap years only? If so, I suggest replacing 367 with `curr_days_per_year + 1` and explaining why `+ 1` in a comment.

slevis-lmwg (changes requested):
@samsrabin you will find one more request before I approve. Also I think it would be best if we went over things together in a 30' meeting today or Monday.

src/utils/clm_time_manager.F90
samsrabin: Good point. I don't know why it was EVER 367; fixed in 54e5f2470.

slevis-lmwg (approved):

PR #2199, opened 2023-10-16
FATES restart fix for long runs

rgknox:
### Description of changes

FATES had been calling a routine (canopy_structure()) during the restart read process.  The purpose was to facilitate rebuilding the canopy. However, this had an inadvertent effect of changing results from base, because this procedure would not just rebuild, but modify the cohort compositions due to termination and promotion/demotion in the canopy. We realized that we could simply bypass this call, and add two variables to the restart file to achieve b4b restarts in tests.

This needs to be synchronized with: https://github.com/NGEET/fates/pull/1098

### Specific notes

Contributors other than yourself, if any: @mvdebolskiy identified the locus of the problem which enabled its fix, with team contributions and troubleshooting from @mvertens , @glemieux , @rosiealice and @ckoven 

https://github.com/NGEET/fates/issues/1051

No, this is just a bug fix.

Testing performed, if any:

We constructed a new test that was simply long enough to enable more canopy development (at least that is the hypothesis).  The key is to have a FATES ERS test that is an LM5 (5 month). However, the need for a long test would likely be circumvented if we had an initial condition file that had been generated after long simulation (ie at least a few months of cold start spinup, but longer would probably be better (ie years)).


ekluzek:
@rgknox @glemieux @adrifoster @ckoven do we want to expedite bringing this in as it's a critical fix?

Externals_CLM.cfg
```diff
@@ -1,8 +1,8 @@
 [fates]
 local_path = src/fates
 protocol = git
-repo_url = https://github.com/NGEET/fates
-tag = sci.1.67.2_api.27.0.0
+repo_url = https://github.com/rgknox/fates
```
ekluzek, line 4: This will need to be updated to a tag before coming into main-dev (as per usual...)

src/utils/clmfates_interfaceMod.F90
```diff
@@ -1054,7 +1054,7 @@ subroutine dynamics_driv(this, nc, bounds_clump,      &
 
             call ed_update_site(this%fates(nc)%sites(s), &
                   this%fates(nc)%bc_in(s), &
-                  this%fates(nc)%bc_out(s))
+                  this%fates(nc)%bc_out(s),.false.)
```
ekluzek, line 1057: I'd prefer seeing the .false. with a keyword argument in front of it, so I have some idea of what this is in regard to.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -1716,7 +1716,7 @@ subroutine restart( this, bounds_proc, ncid, flag, waterdiagnosticbulk_inst, &
 
                   call ed_update_site( this%fates(nc)%sites(s), &
                         this%fates(nc)%bc_in(s), &
-                        this%fates(nc)%bc_out(s) )
+                        this%fates(nc)%bc_out(s),.true. )
```
ekluzek, line 1719: Same as above.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -1940,7 +1940,7 @@ subroutine init_coldstart(this, waterstatebulk_inst, waterdiagnosticbulk_inst, &
 
               call ed_update_site(this%fates(nc)%sites(s), &
                     this%fates(nc)%bc_in(s), &
-                    this%fates(nc)%bc_out(s))
+                    this%fates(nc)%bc_out(s),.false.)
```
ekluzek, line 1943: above.

src/utils/clmfates_interfaceMod.F90
ekluzek: Since this the one that uses .true. it would be good to have a comment that explains why this is true and the others are not.

ekluzek (changes requested):
I request a small change for readability that won't be hard to bring in.

I'm really glad you were able to figure this out! Thanks to everyone who worked on this! @mvdebolskiy @glemieux @mvertens I know for sure...

With testing there are tests that will start passing that can be removed from the expected fails. It also sounds like you setup a special test to find this -- should that come in as well?

So the things I see:
- [ ] Update FATES tag
- [ ] Add keyword syntax to the .false./.true. calls.
- [ ] Add a comment to the one that's different the .true. one.
- [ ] Remove expected fails that now work.
- [ ] Add a new test for this?

rgknox:
Those all sound like great suggestions @ekluzek . I'm working through designing a new test right now. I had originally wanted to go all-out and create a finidat file to be used in the new test's initialization (which would allow us to encounter the salient model mechanics in a shorter run), but I'm going to circumvent that for now because that is a larger topic/issue. I'll check back in when I have a new test created. In short I'm hoping I can trigger the original error with an f10 grid and a few months. Ideally, I'd like to see a test similar to an ERS f10 Lm13 in there, if it does not take much longer than the other tests.

src/utils/clmfates_interfaceMod.F90
mvdebolskiy: I think all of this can be actually done with:
`hlm_is_restart` without passing any extra arguments.

rgknox: I was thinking about using that too @mvdebolskiy , it might work, but I wasn't sure if that flag stays on just during the restart sequence (its been a while since I've worked with it), or during the time-stepping following restarts.  If it also stays true during the latter, it won't work because we will be turning of all canopy_structure calls during the run.  But I agree, it would be nicer to use the flag if possible.

rgknox: @mvdebolskiy , I checked the code and hlm_is_restart will not work for our needs in this scenario.  That flag tells us if the run (as a whole) was restarted.  But, what we need to know is if the current call to ed_update_site was called during the restart setup sequence.  The time-stepping that follows will also be true for hlm_is_restart, on a restarted run.

Externals_CLM.cfg
rgknox: yup, will update at time of integration since this is a synchronized PR

src/utils/clmfates_interfaceMod.F90
rgknox: @ekluzek I added in the keywords

mvdebolskiy (approved):

mvdebolskiy (commented):
Just a reminder to update externals, and I think it is good to go. Will just need to make an issue for adding a test that starts with really spun-up restart file.

rgknox:
I tested a bunch of configurations to get a 25 month test that completes in a reasonable amount of time, I found that using the FatesColdNoComp testdef is fairly effective at doing this.  I tested different pe layouts and found that with:
72 cores = 38 minutes
144 cores = 29 minutes
252 cores = 27 minutes
My take is that 144 core setup is the sweet spot and will move forward with this

rgknox:
all aux_clm pass, izumi and cheyenne, preparing changelogs

src/utils/clmfates_interfaceMod.F90
ekluzek: Perfect

ekluzek:
@rgknox I realized this isn't pointing to a FATES tag. And it looks like the FATES tag hasn't been made yet. Could you do that and update the branch? Thanks...

ekluzek (approved):

ekluzek:
@rgknox I'm making the tag now. But, we will need you to rename the test directory to under the baselines for both Cheyenne and izumi to ctsm5.1.dev151. There were several directories and I wasn't sure which was the right one, and there seemed to be permission issues when I tried to copy them over. So if you could do that Monday that would be great.

ekluzek:
@rgknox also was the FATES test list run for this? I expect that some of the FATES expected fails, might now work? Or is that not the case? Monday would be good to look into this and it probably would be good to run the FATES test suite then as well. If more of the FATES test pass, we can mark them as working in the next tag. For now I'll check that thing off the list with a note about it.

PR #2240, opened 2023-11-08
Use baset_latvary parameters

### LINKED ISSUE #2127: baset_latvary_slope and baset_latvary_intercept not used
### Brief summary of bug

Namelist parameters `baset_latvary_slope` and `baset_latvary_intercept` don't seem to actually be used. Instead, the default values (4 and 12, respectively) are hard-coded in `CropType%InitCold`.

### General bug information

**CTSM version you are using:** `ctsm5.1.dev139-1-g6282180d9`

**Does this bug cause significantly incorrect results in the model's science?** Not unless the user wanted to set those, which I can't imagine happens often.

**Configurations affected:** Crop compsets.

### Details of bug
They're hard-coded here:
https://github.com/ESCOMP/CTSM/blob/1e2e2c35d9568c94eac6cb606f37c18294158682/src/biogeochem/CropType.F90#L373-L379

In addition, I think that whole `if` block could be refactored into a one-liner that would additionally avoid hard-coding the ±30° boundaries:
```fortran
this%latbaset_patch(p) = pftcon%baset(ivt) + baset_latvary_intercept &
    - max(baset_latvary_intercept, baset_latvary_slope * abs(grc%latdeg(g)))
```

In addition, for ease of user understanding, I think it'd be preferable to replace `baset_latvary_slope` as a namelist parameter with `baset_latvary_maxlat`, which would allow them to specify the absolute latitude where `baset` is no longer affected. `baset_latvary_slope` would still be used in the above equation and calculated like so:
```fortran
baset_latvary_slope = baset_latvary_intercept / baset_latvary_maxlat
```

samsrabin:
### Description of changes
Namelist parameters `baset_latvary_slope` and `baset_latvary_intercept` were never actually used, with values of 0.4 and 12 being hard-coded in the relevant subroutine instead. This PR fixes that, and also adds unit testing of a refactored function that uses them.

### Specific notes

- Fixes #2127

**Are answers expected to change (and if so in what way)?** No; those values are the default and aren't changed in any test. Things will only change if users had changed either of those parameters in the namelist—i.e., their settings will now be obeyed.

**Testing performed:**

- [X] Unit testing of the new `latbaset()` function passes. This initially failed when `baset_latvary_slope` and `baset_latvary_intercept` weren't set to their default values.

- [X] The following tests (comparing to ctsm5.1.dev150) pass on Cheyenne:
```
SMS_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.cheyenne_gnu.clm-crop
```
- [X] The following tests (comparing to ctsm5.1.dev150) pass on Izumi:
```
SMS_D.f10_f10_mg37.I2000Clm51BgcCrop.izumi_gnu.clm-crop
SMS_D.f10_f10_mg37.I2000Clm51BgcCrop.izumi_intel.clm-crop
SMS_D_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.izumi_nag.clm-crop
```


src/biogeochem/CropType.F90
```diff
@@ -383,13 +385,7 @@ subroutine InitCold(this, bounds)
...
        this%nyrs_crop_active_patch(p) = 0
 
-       if ( grc%latdeg(g) >= 0.0_r8 .and. grc%latdeg(g) <= 30.0_r8) then
-          this%latbaset_patch(p)=pftcon%baset(ivt)+12._r8-0.4_r8*grc%latdeg(g)
-       else if (grc%latdeg(g) < 0.0_r8 .and. grc%latdeg(g) >= -30.0_r8) then
-          this%latbaset_patch(p)=pftcon%baset(ivt)+12._r8+0.4_r8*grc%latdeg(g)
-       else
-          this%latbaset_patch(p)=pftcon%baset(ivt)
-       end if
+       this%latbaset_patch(p) = latbaset(pftcon%baset(ivt), grc%latdeg(g), this%baset_latvary_intercept, this%baset_latvary_slope)
```
slevis-lmwg, line 388: @samsrabin thank you for your work on this.

To make sure I'm understanding correctly, baset_latvary_* get their values from the namelist?

src/biogeochem/CropType.F90
```diff
@@ -972,4 +968,31 @@ subroutine checkDates( )
...
+    !     else
+    !         latbaset = baset + baset_latvary_intercept - baset_latvary_slope*abs(latdeg)
+    !     end if
+    ! But the one-liner below should improve efficiency, at least marginally.
+
+    latbaset = baset + baset_latvary_intercept - min(baset_latvary_intercept, baset_latvary_slope * abs(latdeg))
+
+  end function latbaset
+
+  real(r8) function latbaset_max_lat(intercept, slope)
```
slevis-lmwg, line 991: Is this function only used by the unit tests? If so, I recommend adding a comment about that, so nobody deletes this function thinking that it serves no purpose.

slevis-lmwg (changes requested):
@samsrabin looks good. My comments should be simple to address. Let me know if you have questions.

src/biogeochem/CropType.F90
samsrabin: Good point! In 239aa87, I moved it to the test module that uses it.

src/biogeochem/CropType.F90
samsrabin: Yep, that's right.

samsrabin:
Thanks for the review, @slevis-lmwg! I did some rearranging of `CropType%InitCold()` in 35478a3, then moved `latbaset_max_lat()` in 239aa87.

slevis-lmwg (commented):
One more question...

src/biogeochem/CropType.F90
```diff
@@ -371,43 +372,32 @@ subroutine InitCold(this, bounds)
...
-       else
-          this%latbaset_patch(p)=pftcon%baset(ivt)
-       end if
-       if ( trim(this%baset_mapping) == baset_map_constant ) then
-          this%latbaset_patch(p) = nan
-       end if
-    end do
-!DLL -- end of mods
-
-    if (use_crop) then
```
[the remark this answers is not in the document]
samsrabin: Sorry, should have explained: the call itself of this subroutine is wrapped in an `if (use_crop)`, so the one inside the subroutine was redundant.

Note that `use_crop` refers to whether crops are turned on in the run, not to any particular LU or PFT.

samsrabin:
- [ ] Run `aux_clm` against `ctsm5.1.dev154`

PR #2270, opened 2023-11-29
Update Externals.cfg to work on Derecho

### LINKED ISSUE #2217: Tags for building CTSM library on Derecho [WRF-CTSM]
Hello, 

I was wondering if the in what CTSM tags, `build_ctsm` script works on Derecho. 

@irismuz is interested in running WRF-CTSM on Derecho and she was wondering which tag she should use to build the ctsm library on Derecho. 

This is relevant to #2147.

samsrabin:
### Description of changes

Update Externals.cfg to version from CESM3_dev aba49b6.

### Specific notes

- Resolves #2217

Testing performed, if any: With this change, my tillage & residue branch works on Derecho.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -8,7 +8,7 @@ required = True
 local_path = components/cism
 protocol = git
 repo_url = https://github.com/ESCOMP/CISM-wrapper
-tag = cismwrap_2_1_95
+tag = cismwrap_2_1_96
 externals = Externals_CISM.cfg
 required = True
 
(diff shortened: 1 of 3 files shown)


samsrabin:
Unfortunately it looks like even the first commit here (10b3985) causes all Izumi nag tests to fail, with this in `atm.log`:
```
 ERROR: reading input namelist, datm_in iostat= 186
 ERROR: (atm_comp_nuopc):(InitializeAdvertise) : namelist read error datm_in
```

ekluzek:
Actually @samsrabin with our current plan I suggest that you don't worry about izumi, and that I fix it in ctsm5.1.dev158 with the tag with more comprehensive testing. Not that many people use izumi, and you can just put a disclaimer that it won't work there. I wouldn't normally recommend that, but with our current plan I think that makes sense.

I've been testing on Derecho and little on Cheyenne, but I don't think I've tried izumi yet. I just assumed it would work...

PR #2080, opened 2023-07-28
Improvements to processing of crop calendar files

samsrabin:
### Description of changes

In `python/ctsm/crop_calendars/`, `process_ggcmi_shdates.py` and `regrid_ggcmi_shdates.py` are used to convert the raw GGCMI crop calendar files into CTSM-compatible versions. This update fixes some bugs, removes dependencies on the `nco` utilities, and enables the use of surface datasets as template files (in addition to the existing ability to use CTSM output files).

### Specific notes

Contributors other than yourself, if any: None

CTSM Issues Fixed (include github issue #): None

Are answers expected to change (and if so in what way)? No

Any User Interface Changes (namelist or namelist defaults changes)? No

Testing performed, if any: I have used the scripts successfully.

PR #2292, opened 2023-12-14
Improvements to processing of crop calendar files (v2)

samsrabin:
### Description of changes

In `python/ctsm/crop_calendars/`, `process_ggcmi_shdates.py` and `regrid_ggcmi_shdates.py` are used to convert the raw GGCMI crop calendar files into CTSM-compatible versions. This update fixes some bugs, removes dependencies on the `nco` utilities, and enables the use of surface datasets as template files (in addition to the existing ability to use CTSM output files).

### Specific notes

Testing performed, if any: I have used the scripts successfully.


samsrabin:
- [ ] Incorporate comments from @ekluzek [here](https://github.com/ESCOMP/CTSM/pull/2080#pullrequestreview-1773073914).

ekluzek (approved):
Just went over this with @samsrabin. These scripts are going to mainly be used by CTSM SE's when updates happen (and for now just by Sam), so they don't need to be totally robust for general use. He is going to add some comments about where the data is located, and we plan to just leave it on /glade/campaign for our own internal CESM use.

This does all the things I wanted to see and makes the scripts follow the pattern of other scripts, which is great. Thanks for the update @samsrabin

PR #2302, opened 2023-12-28
Add new scheme Li2024 of ozone plant damage

lifang0209:
### Description of changes
Changes include: 
OzoneMod.F90: add an option for scheme of Li2024:
Include add: (i) CalcOzoneUptakeLi2024OnePoint to calculate ozone uptake for a single point, and change the name of old module for Lombardozzi2015 and Falk schemes as CalcOzoneUptakeLFOnePoint. (ii) add CalcOzoneStressLi2024 and CalcOzoneStressLi2024OnePoint to calculate ozone stress. 

Change CanopyFluxesMod.F90, OzoneBaseMod.F90, OzoneOffMod.F90 to read the variable sabv that CalcOzoneUptake use.
subroutine CalcOzoneUptake(this, bounds, num_exposedvegp, filter_exposedvegp, &
       forc_pbot, forc_th, rssun, rssha, rb, ram, tlai, forc_o3, sabv)

In bld/namelist_files/namelist_definition_ctsm.xml: add stress_li2024 as the new option

### Specific notes

Contributors other than yourself, if any: @slevis-lmwg @dlawrenncar @danicalombardozzi @ekluzek 

Yes, when using the new scheme

new namelist option: stress_li2024 

Testing performed, if any:
@lifang0209 ran I2000Sp and I2000BgcCrop

### DIFF
--- a/src/biogeophys/OzoneMod.F90
+++ b/src/biogeophys/OzoneMod.F90
@@ -68,7 +68,12 @@ module OzoneMod
      procedure, private :: InitCold
 
      ! Calculate ozone uptake for a single point, for just sunlit or shaded leaves
-     procedure, private, nopass :: CalcOzoneUptakeOnePoint
+     procedure, private, nopass :: CalcOzoneUptakeLi2024OnePoint
+     procedure, private, nopass :: CalcOzoneUptakeLFOnePoint
+
+     ! Ozone stress functions Fang Li 2024
+     procedure, private         :: CalcOzoneStressLi2024           ! Stress parameterization
+     procedure, private, nopass :: CalcOzoneStressLi2024OnePoint   ! Ozone stress calculation 
 
      ! Original ozone stress functions from Danica Lombardozzi 2015
      procedure, private         :: CalcOzoneStressLombardozzi2015           ! Stress parameterization
@@ -81,23 +86,15 @@ module OzoneMod
   end type ozone_type
 
   ! !PRIVATE TYPES:
-  integer, parameter :: stress_method_lombardozzi2015 = 1
-  integer, parameter :: stress_method_falk = 2
+  integer, parameter :: stress_method_li2024 = 1
+  integer, parameter :: stress_method_lombardozzi2015 = 2
+  integer, parameter :: stress_method_falk = 3
 
   ! TODO(wjs, 2014-09-29) The following parameters should eventually be moved to the
   ! params file. Parameters differentiated on veg type should be put on the params file
   ! with a pft dimension.
 
-  ! o3:h2o resistance ratio defined by Sitch et al. 2007
-  real(r8), parameter :: ko3 = 1.67_r8
-
-  ! LAI threshold for LAIs that asymptote and don't reach 0
-  real(r8), parameter :: lai_thresh = 0.5_r8
-
-  ! threshold below which o3flux is set to 0 (nmol m^-2 s^-1)
-  real(r8), parameter :: o3_flux_threshold = 0.8_r8
-
-  ! o3 intercepts and slopes for photosynthesis
+  ! o3 intercepts and slopes for photosynthesis in Lombardozzi2015
   real(r8), parameter :: needleleafPhotoInt   = 0.8390_r8  ! units = unitless
   real(r8), parameter :: needleleafPhotoSlope = 0._r8      ! units = per mmol m^-2
   real(r8), parameter :: broadleafPhotoInt    = 0.8752_r8  ! units = unitless
@@ -105,7 +102,7 @@ module OzoneMod
   real(r8), parameter :: nonwoodyPhotoInt     = 0.8021_r8  ! units = unitless
   real(r8), parameter :: nonwoodyPhotoSlope   = -0.0009_r8 ! units = per mmol m^-2
 
-  ! o3 intercepts and slopes for conductance
+  ! o3 intercepts and slopes for conductance in Lombardozzi2015
   real(r8), parameter :: needleleafCondInt    = 0.7823_r8  ! units = unitless
   real(r8), parameter :: needleleafCondSlope  = 0.0048_r8  ! units = per mmol m^-2
   real(r8), parameter :: broadleafCondInt     = 0.9125_r8  ! units = unitless
@@ -146,10 +143,11 @@ contains
     type(bounds_type), intent(in)    :: bounds
     character(len=*),  intent(in)    :: o3_veg_stress_method
     !-----------------------------------------------------------------------
-
-    if (o3_veg_stress_method=='stress_lombardozzi2015') then
+   if (o3_veg_stress_method=='stress_li2024')then
+     this%stress_method = stress_method_li2024
+   else if (o3_veg_stress_method=='stress_lombardozzi2015') then
        this%stress_method = stress_method_lombardozzi2015
-    else if (o3_veg_stress_method=='stress_falk') then
+   else if (o3_veg_stress_method=='stress_falk') then
        this%stress_method = stress_method_falk
        if (.not. use_luna ) call endrun(' use_luna=.true. is required when o3_veg_stress_method = stress_falk.')
     else
@@ -223,7 +221,8 @@ contains
          avgflag='A', long_name='total ozone flux into shaded leaves', &
          ptr_patch=this%o3uptakesha_patch)
 
-    if (this%stress_method==stress_method_lombardozzi2015) then
+    if (this%stress_method==stress_method_li2024 .or. &
+        this%stress_method==stress_method_lombardozzi2015) then
        ! For this and the following variables: how should we include leaf area in the
        ! patch averaging?
        !
@@ -354,7 +353,7 @@ contains
 
   !-----------------------------------------------------------------------
   subroutine CalcOzoneUptake(this, bounds, num_exposedvegp, filter_exposedvegp, &
-       forc_pbot, forc_th, rssun, rssha, rb, ram, tlai, forc_o3)
+       forc_pbot, forc_th, rssun, rssha, rb, ram, tlai, forc_o3, sabv)
     !
     ! !DESCRIPTION:
     ! Calculate ozone uptake.
@@ -372,6 +371,7 @@ contains
     real(r8) , intent(in) :: ram( bounds%begp: )       ! aerodynamical resistance (s/m)
     real(r8) , intent(in) :: tlai( bounds%begp: )      ! one-sided leaf area index, no burying by snow
     real(r8) , intent(in) :: forc_o3( bounds%begg: )   ! ozone partial pressure (mol/mol)
+    real(r8) , intent(in) :: sabv( bounds%begp: )      ! solar radiation absorbed by vegetation (W/m**2)
     !
     ! !LOCAL VARIABLES:
     integer  :: fp             ! filter index
@@ -391,6 +391,7 @@ contains
     SHR_ASSERT_ALL_FL((ubound(rb) == (/bounds%endp/)), sourcefile, __LINE__)
     SHR_ASSERT_ALL_FL((ubound(ram) == (/bounds%endp/)), sourcefile, __LINE__)
     SHR_ASSERT_ALL_FL((ubound(tlai) == (/bounds%endp/)), sourcefile, __LINE__)
+    SHR_ASSERT_ALL_FL((ubound(sabv) == (/bounds%endp/)), sourcefile, __LINE__)
 
     associate( &
          o3uptakesha => this%o3uptakesha_patch                , & ! Output: [real(r8) (:)] ozone dose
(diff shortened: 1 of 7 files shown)


lifang0209:
update Y in PODY and response functions in Li2024 in OzoneMod.F90

src/biogeophys/OzoneMod.F90
```diff
@@ -547,6 +692,118 @@ subroutine CalcOzoneStress(this, bounds, &
...
+
+    if (o3uptake == 0._r8) then
+       ! No o3 damage if no o3 uptake
+       o3coefv = 1._r8
+       o3coefg = 1._r8
+    else
+       ! Determine parameter values for this pft
+      if(pft_type >= 1 .and. pft_type <= 3)then  !Needleleaf tree
+        o3coefv = max(0._r8, min(1._r8, 1.005_r8 - 0.0064_r8 * o3uptake))
+        o3coefg = max(0._r8, min(1._r8, 0.965_r8 * o3uptake ** (-0.041)))
```
slevis-lmwg, line 786: These right-hand-sides do not match what I see in Li et al. (in prep), which says
```
1.014-0.010 POD0.5
0.975-0.037 ln(POD0.5)
```

src/biogeophys/OzoneMod.F90
```diff
@@ -547,6 +692,118 @@ subroutine CalcOzoneStress(this, bounds, &
...
+      end if
+      if(pft_type >= 4 .and. pft_type <= 8)then  !Broadleaf tree
+        o3coefv = max(0._r8, min(1._r8, 0.943_r8 * exp(-0.0085*o3uptake)))
+        o3coefg = max(0._r8, min(1._r8, 0.943_r8 * exp(-0.0058*o3uptake)))
+      end if
+      if(pft_type >= 9 .and. pft_type <= 11)then !Shrub
+        o3coefv = max(0._r8, min(1._r8, 1.000_r8-0.074_r8 * log(o3uptake)))
+        o3coefg = max(0._r8, min(1._r8, 0.991_r8-0.060_r8 * log(o3uptake)))
```
slevis-lmwg, line 794: These right-hand-sides also do not match what I see in Li et al. (in prep), which says
```
1.142 POD5**(-0.140)
1.075 POD5**(-0.104)
```

src/biogeophys/OzoneMod.F90
```diff
@@ -547,6 +692,118 @@ subroutine CalcOzoneStress(this, bounds, &
...
+      end if
+      if(pft_type >= 12 .and. pft_type <= 14)then !Grass
+        o3coefv = max(0._r8, min(1._r8, 0.997_r8 - 0.016_r8 * o3uptake))
+        o3coefg = max(0._r8, min(1._r8, 0.989_r8 - 0.045_r8 * log(o3uptake)))
+      end if
+      if (pft_type >= 15)then !Crop
+        o3coefv = max(0._r8, min(1._r8, 0.909_r8 - 0.028_r8 * log(o3uptake)))
+        o3coefg = max(0._r8, min(1._r8, 1.005_r8 - 0.169_r8 * tanh(o3uptake)))
```
slevis-lmwg, line 802: Grasses and crops seem to match what I see in Li et al. (in prep).

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -1934,9 +1934,9 @@ subroutine Photosynthesis ( bounds, fn, filterp, &
 
                ! Convert gs_mol (umol H2O/m**2/s) to gs (m/s) and then to rs (s/m)
 
-               gs = gs_mol(p,iv) / cf
+               gs = gs_mol(p,iv) / cf * o3coefg(p)
                rs_z(p,iv) = min(1._r8/gs, rsmax0)
-               rs_z(p,iv) = rs_z(p,iv) / o3coefg(p)
```
slevis-lmwg, line 1939: @lifang0209 explained to me that moving the effect from the resistance to the conductance, also moves it to the current time step rather than the following time step. I have not checked the validity of that statement.

However @lifang0209 you said that you had tested and found the effect of this change on the simulation to be very small. Could you post some sort of documentation here (possibly a figure) that would make this point clear?

src/biogeophys/OzoneMod.F90
```diff
@@ -424,9 +445,120 @@ subroutine CalcOzoneUptake(this, bounds, num_exposedvegp, filter_exposedvegp, &
...
+    real(r8) :: o3_flux_threshold   ! threshold below which o3flux is set to 0 (nmol m^-2 s^-1)
+    real(r8) :: lai_thresh     !LAI threshold for growing season
+    ! o3:h2o resistance ratio from Li et al. 2024
+    real(r8), parameter :: ko3 = 1.51_r8
+
+    character(len=*), parameter :: subname = 'CalcOzoneUptakeOnePoint'
+    !-----------------------------------------------------------------------
+
+    ! convert o3 from mol/mol to nmol m^-3
+    o3concnmolm3 = forc_ozone * 1.e9_r8 * (forc_pbot/(forc_th*SHR_CONST_RGAS*0.001_r8))
```
slevis-lmwg, line 495: When you finalize this conversion (see my other comment), then replace any remaining hardwiring with parameter names. If the conversion parameters already exist elsewhere, please reuse them, and otherwise add them locally.

src/biogeophys/OzoneMod.F90
```diff
@@ -424,9 +445,120 @@ subroutine CalcOzoneUptake(this, bounds, num_exposedvegp, filter_exposedvegp, &
...
+
+    ! calculate instantaneous flux
+    o3flux = o3concnmolm3/ (ko3*rs+ rb + ram)
+    ! set lai_thresh    
+        if (pftcon%evergreen(pft_type) == 1) then
+         lai_thresh=0._r8 ! evergreens grow year-round
+       else  ! for deciduous vegetation
+        if(pft_type == 10)then !temperate shrub
```
slevis-lmwg, line 503: As you did with "evergreen" above, please replace 10 with the shrub name from pftconMod.F90. 

Similarly replace numbers below with pft names and use `is_shrub, is_grass` for these types in particular.

For 15 I think you can use nc3crop. All of these are in pftconMod.F90.

slevis-lmwg (changes requested):
@lifang0209 thank you for this nice update to the ozone code. 
- After we talked, I reviewed the code more carefully, posted a few comments, and requested a few improvements. 
- Scientifically the most pressing is to resolve the inconsistency between the formulas appearing in the code versus those documented in Li et al. (in prep).

src/biogeophys/OzoneMod.F90
```diff
@@ -424,9 +445,120 @@ subroutine CalcOzoneUptake(this, bounds, num_exposedvegp, filter_exposedvegp, &
...
+     if(sabv > 0._r8)then  !daytime
+       o3fluxperdt = o3fluxcrit * dtime * 0.000001_r8
+     else
+       o3fluxperdt = 0._r8
+     end if
+
+     if (tlai > lai_thresh) then
+      ! o3 uptake decay
+      if (pftcon%evergreen(pft_type) == 1) then
+       leafturn = 1._r8/(pftcon%leaf_long(pft_type)*365._r8*24._r8)
```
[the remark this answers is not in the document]
lifang0209: done

src/biogeophys/OzoneMod.F90
```diff
@@ -547,6 +692,118 @@ subroutine CalcOzoneStress(this, bounds, &
...
+       ! Determine parameter values for this pft
+      if(pft_type >= 1 .and. pft_type <= 3)then  !Needleleaf tree
+        o3coefv = max(0._r8, min(1._r8, 1.005_r8 - 0.0064_r8 * o3uptake))
+        o3coefg = max(0._r8, min(1._r8, 0.965_r8 * o3uptake ** (-0.041)))
+      end if
+      if(pft_type >= 4 .and. pft_type <= 8)then  !Broadleaf tree
+        o3coefv = max(0._r8, min(1._r8, 0.943_r8 * exp(-0.0085*o3uptake)))
+        o3coefg = max(0._r8, min(1._r8, 0.943_r8 * exp(-0.0058*o3uptake)))
+      end if
+      if(pft_type >= 9 .and. pft_type <= 11)then !Shrub
```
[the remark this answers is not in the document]
lifang0209: Done

src/biogeophys/OzoneMod.F90
lifang0209: formulas in the code are consistent with the Eqs.2-3 in our submitted MS that I sent you on Jan 17. You used an old MS version. The difference is in the data pre-processing which changed the sample of observations we used to generate the response functions shown in Eqs. 2-3. According to suggestion of ecophysiologist Felicity Hayes (our coauthor and the chair of LRTAP ozone impacts on vegetation in observations), in pre-processing, we conducted linear regression using the relative values and corresponding PODY for individual plant species in a study, and data with intercept falling outside the range of 0.9 to 1.1 are removed based on LRTAP convention to ensure a usable response relationship.

src/biogeophys/OzoneMod.F90
lifang0209: see above reply

src/biogeophys/OzoneMod.F90
lifang0209: see above reply

src/biogeophys/PhotosynthesisMod.F90
lifang0209: I have added comment in the PhotosynthesisMod.F90:
! Fang Li revised o3 impact on gs
! now: multiply gs response factor (o3coefg) with gs
! clm5.0: skipped the effect on gs and had rs divided by o3coefg,
! which led to a one timestep delay in the o3 impact on gs,
! although the impact on an annual scale is quite small

src/biogeophys/OzoneMod.F90
lifang0209: add as local variables

src/biogeophys/OzoneMod.F90
lifang0209: Done

ekluzek:
@lifang0209 thanks so much for your work here and this contribution! There is some more work to be done on this, so I'm converting it to a draft. That doesn't change the importance of this, it just marks it more clearly as there's more work to do here....

src/biogeophys/OzoneMod.F90
```diff
@@ -453,16 +590,25 @@ subroutine CalcOzoneUptakeOnePoint( &
...
     real(r8) :: o3fluxcrit     ! instantaneous o3 flux beyond threshold (nmol m^-2 s^-1)
     real(r8) :: o3fluxperdt    ! o3 flux per timestep (mmol m^-2)
     real(r8) :: heal           ! o3uptake healing rate based on % of new leaves growing (mmol m^-2)
-    real(r8) :: leafturn       ! leaf turnover time / mortality rate (per hour)
+    real(r8) :: leafturn       ! leaf turnover time / mortality rate (per sec)
     real(r8) :: decay          ! o3uptake decay rate based on leaf lifetime (mmol m^-2)
 
-    character(len=*), parameter :: subname = 'CalcOzoneUptakeOnePoint'
+    ! o3:h2o resistance ratio defined by Sitch et al. 2007
+    real(r8), parameter :: ko3 = 1.67_r8
```
ekluzek, line 603: @lifang0209 should any of these be added to the parameter file, so they can be modified rather than hardcoded here?

src/biogeophys/OzoneMod.F90
```diff
@@ -494,21 +640,21 @@ subroutine CalcOzoneUptakeOnePoint( &
        endif
 
        if (pftcon%evergreen(pft_type) == 1) then
-          leafturn = 1._r8/(pftcon%leaf_long(pft_type)*365._r8*24._r8)
+          leafturn = 1._r8/(pftcon%leaf_long(pft_type)*avg_dayspyr*secspday)
```
ekluzek, line 643: I very much like using functions and constants to get things like the number of days in a year, and the seconds per day. That's an important improvement.

src/biogeophys/OzoneMod.F90
```diff
@@ -494,21 +640,21 @@ subroutine CalcOzoneUptakeOnePoint( &
...
        else
           leafturn = 0._r8
        endif
 
        ! o3 uptake decay based on leaf lifetime for evergreen plants
-       decay = o3uptake * leafturn * dtimeh
+       decay = o3uptake * leafturn * dtime
```
ekluzek, line 649: Is this OK to do? Because dtimeh is in hours and dtime is in seconds, won't this change answers for the older versions?

src/biogeophys/OzoneMod.F90
```diff
@@ -424,9 +444,125 @@ subroutine CalcOzoneUptake(this, bounds, num_exposedvegp, filter_exposedvegp, &
     end associate
 
   end subroutine CalcOzoneUptake
+    !-----------------------------------------------------------------------
+  subroutine CalcOzoneUptakeLi2024OnePoint( &
```
ekluzek, line 448: I think this should actually be spun off into it's own file. That would be better in keeping with the OO nature of how this is built. But, @slevis-lmwg and I will discuss, and I need to look at it more closely.

ekluzek (changes requested):
@lifang0209 thanks so much for the contribution. I have a few suggestions. Mostly that @slevis-lmwg and I will discuss and work on. I had one question for you about dtimeh though...

lifang0209:
> @lifang0209 thanks so much for the contribution. I have a few suggestions. Mostly that @slevis-lmwg and I will discuss and work on. I had one question for you about dtimeh though...

src/biogeophys/OzoneMod.F90
lifang0209: Hi Erik, thanks for your time and suggestions.
The correct (observational) value for ko3 should be  
 real(r8), parameter :: ko3 = 1.51_r8
which used in CalcOzoneUptakeLi2024OnePoint in the same file (OzoneMod.F90).
I'm not sure if it's good to set different values of a parameter for different schemes in the parameter file.

src/biogeophys/OzoneMod.F90
lifang0209: thanks

src/biogeophys/OzoneMod.F90
lifang0209: According to Sam's suggestion, leafturn is changed from per hour to per second. Correspondingly, we should use dtime (in seconds) instead of dtimeh (in hours).

src/biogeophys/OzoneMod.F90
lifang0209: Thanks. That's not my area of expertise; I'll leave it to you and Sam to decide. 
If you do this, don't forget move CalcOzoneUptakeLFOnePoint.
CalcOzoneUptakeLi2024OnePoint and CalcOzoneUptakeLFOnePoint are used to calculate the ozone uptake for a single point. The former is used for the new scheme (li2024), and the latter for the lombardozzi2015 and falk schemes.

PR #2309, opened 2024-01-08
Fix izumi run_sys_tests

samsrabin:
### Description of changes
ctsm5.1.dev162 broke `run_sys_tests` on Izumi. This PR fixes that by moving the `import` of `numpy` to a file that's not called from `run_sys_tests`.

### Specific notes

ctsm5.1.dev162 added functions to `python/ctsm/utils.py` that depend on `numpy`. Because that's not available on Izumi, when `run_sys_tests` (indirectly) `import`ed from `utils.py`, it would fail.

**CTSM Issues Fixed:** None

**Any User Interface Changes (namelist or namelist defaults changes)?** None

Testing performed, if any:
- `LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac` passes
- `make test` passes
- `run_ctsm_py_tests` passes
- `run_sys_tests` no longer fails on Izumi


ekluzek (approved):
Ahh so this separates out the part of utils that require CTSM pylint conda environment to be loaded and those that are bare python.

This is a good change and I see why it's needed.

The name of the file is good as it's descriptive. It is tied to the name we are using for the conda environment right now. That might change, and the real requirement is for xarray. So I wonder about changing the name. But that's the only comment.

PR #2314, opened 2024-01-11
NEONv3 updates

wwieder:
### Description of changes
Updates so that run_neon works with the latest tag on derecho.  Additional modifications to use NEON_v3 data, but these may not be critical right now.

### Specific notes
Out of the box run_neon is broken.

### DIFF
--- a/python/ctsm/site_and_regional/run_neon.py
+++ b/python/ctsm/site_and_regional/run_neon.py
@@ -62,14 +62,11 @@ import sys
 import time
 import pandas as pd
 
-from standard_script_setup import *
-
 # Get the ctsm util tools and then the cime tools.
 _CTSM_PYTHON = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "python"))
 sys.path.insert(1, _CTSM_PYTHON)
 
 from ctsm import add_cime_to_path
-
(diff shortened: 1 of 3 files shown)


python/ctsm/site_and_regional/run_neon.py
```diff
@@ -62,14 +62,11 @@
 import time
 import pandas as pd
 
-from standard_script_setup import *
```
wwieder, line 65: I think this is really the only line that has to change for the script to work, but it also likely needs to be fixed in black, to prevent it from being moved again.  Not sure if there's another tag in the queue where this fix could come in...?

wwieder:
@slevis-lmwg and @ekluzek, we decided not to include this PR right now.  Can we convert to draft for now (or should we close the PR with a won't fix?)

slevis-lmwg:
I don't know the answer, but the button to convert to draft is near the top under Reviewers. I tried it and do not seem to have permission to change to draft.

wwieder:
PS #2315 should fix the issue with run_neon so that it runs out of the box.

wwieder:
As such, I'm going to close this with a won't fix, as more robust modifications are needed to support NEON v3 data for the CLM6 release.

PR #2342, opened 2024-01-31
Fix "Assign to project" action (?)

samsrabin:
### Description of changes
Removes backticks from the YAML file that were causing syntax errors

### Specific notes

**Contributors other than yourself, if any:** @mvdebolskiy, @ekluzek 

**Testing performed, if any:** None

### DIFF
--- a/.github/workflows/assign-to-project.yml
+++ b/.github/workflows/assign-to-project.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Assign to High Priority project
     steps:
-    - name: Assign issues and pull requests with priority: high label to project 25
+    - name: Assign issues and pull requests with priority-high label to project 25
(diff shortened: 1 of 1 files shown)


ekluzek (changes requested):
You also need to remove the : after priority: on line 16.

We should just approve this and send it to master without worrying about a tag.

wwieder:
seems good, Erik. I don't think this needs a tag.

mvdebolskiy:
Can you rerun `[Merge tag 'ctsm5.1.dev166' into CESM3_dev](https://github.com/ESCOMP/CTSM/actions/runs/7750813536)` action to see if it doesn't fail?

samsrabin:
Hmm. Not sure how I would do that.

mvdebolskiy:
Oh, you can't apparently. https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs

Issue #2338, opened 2024-01-25
Soil temperature initialization not implemented correctly in ctsm5.1.dev058 and thus subsequent tags


olyson:
### Brief summary of bug

Soil temperature was intended to be initialized at 272K, replacing the previous setting of 274K, as described in issue #1460, but was not implemented correctly

### General bug information

**CTSM version you are using:** ctsm5.1.dev058 (bug is in latest ctsm tag as well)

**Does this bug cause significantly incorrect results in the model's science?** Yes

**Configurations affected:** BGC

### Details of bug

This issue here:
https://github.com/ESCOMP/CTSM/issues/1460

requested a change of initial soil temperature from 274 to 272 in TemperatureType.F90.  It looks like that was implemented as part of a bunch of small changes in Oct, 2021 through this pull request:

https://github.com/ESCOMP/CTSM/pull/1497/files

However, it looks like the wrong line was changed.  The original issue #1460 describes the required change like this:

The code modification required is in TemperatureType.F90, changing this line:

this%t_soisno_col(c,1:nlevgrnd) = 274._r8

to this:

this%t_soisno_col(c,1:nlevgrnd) = 272._r8

See /glade/work/oleson/PPE.n08_ctsm5.1.dev023/cime/scripts/ctsm51c6_PPEn08ctsm51d023_2deg_GSWP3V1_Sparse400_cs_ts_tsoisno272_2000AD/SourceMods/src.clm/TemperatureType.F90

The code referred to here was correct, however, what ended up being changed as part of the combined PR was a different line of code that sets t_soisno for pervious and impervious road only, i.e.,

                  if (col%itype(c) == icol_road_perv .or. col%itype(c) == icol_road_imperv) then
                     this%t_soisno_col(c,1:nlevgrnd) = 272._r8

### Important output or errors that show the problem

We recently noted large deposits of deep soil carbon at high latitudes in recent simulations, which was a feature that the soil temperature change was intended to correct.

wwieder:
This won't be b4b, but should it come in on a b4b-dev tag, or do we need to combine it with other answer changing tags?  Also tagging @samsrabin and @adrifoster

samsrabin:
b4b-dev merges should only include PRs from the b4b-dev branch, and PRs on that branch should always be bit-for-bit. As far as whether it needs its own tag on master or if it can be brought in with others, I think that's something we can assess once the PR fixing this issue is ready.

ekluzek:
[...]

PR #2345, opened 2024-02-01
Add run_neon SystemTest

samsrabin:
### Description of changes
Adds `RUNNEON` SystemTest, which is designed to replicate a simple `./run_neon` call in a way that can be added to aux_clm.

Not yet working!

### Specific notes

Testing with
```
./run_sys_tests -t RUNNEON_Ld1.CLM_USRDAT.HIST_DATM%1PT_CLM51%BGC_SICE_SOCN_SROF_SGLC_SWAV_SESP.derecho_intel.clm-default --skip-compare --skip-generate
```
fails in the `SHAREDLIB_BUILD` phase. See `/glade/derecho/scratch/samrabin/tests_0201-142855de/RUNNEON_Ld1.CLM_USRDAT.HIST_DATM%1PT_CLM51%BGC_SICE_SOCN_SROF_SGLC_SWAV_SESP.derecho_intel.clm-default.0201-142855de/ABBY/bld/cesm.bldlog.240201-143035` for example log file.

### Remaining tasks:
- [ ] Get `build_phase()` to work
- [ ] Ensure `run_phase()` works
- [ ] Rework `run_phase()` to not submit a separate job, but rather to run in the existing job

### DIFF
--- a/cime_config/SystemTests/runneon.py
+++ b/cime_config/SystemTests/runneon.py
@@ -1,0 +1,101 @@
+"""
+CTSM-specific test that uses the run_neon.py methods for setting up, building, and running a NEON
+site
+"""
+
+import os
+import sys
+from CIME.SystemTests.system_tests_common import SystemTestsCommon
+
+_CTSM_PYTHON = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), os.pardir, os.pardir, "python"
+)
+sys.path.insert(1, _CTSM_PYTHON)
+import ctsm.site_and_regional.run_neon as rn  # python: disable=wrong-import-position
+
+
+class RUNNEON(SystemTestsCommon):
+    # pylint: disable=too-many-instance-attributes
+    """
+    CTSM-specific test that uses the run_neon.py methods for setting up, building, and running a
+    NEON site
+    """
+
+    def __init__(self, case):
+        # initialize an object interface to the SMS system test
+        SystemTestsCommon.__init__(self, case)
+
+        caseroot = self._case.get_value("CASEROOT")
+        argv = [
+            "run_neon",
+            "--neon-sites",
+            "ABBY",
+            "--output-root",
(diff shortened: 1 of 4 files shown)


samsrabin:
@wwieder: Here's my initial stab at a SystemTest that exercises run_neon.

@ekluzek: I can't figure out why the build is failing…

cime_config/SystemTests/runneon.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+
+        if run_from_postad:
+            self._neon_site.finidat = None
+
+    def build_phase(self, sharedlib_only=False, model_only=False):
+        """
+        Builds the case
+        """
+        if not model_only:
```
samsrabin, line 76: I've tried with `if sharedlib_only`, `if model_only`, etc., as well as with no conditional here. I get the same build error no matter what.

samsrabin:
Oddly, the build seems to work on Izumi but not Derecho. (It then fails in SUBMIT with `Build complete is not True`, but that's not entirely unexpected.)

ekluzek:
With #2338 and the other issues this comes back to being something that would be good to bring in if we can figure it out...

samsrabin:
@ekluzek Did you mean #2438?

ekluzek:
Yep! Thanks for the correction @samsrabin

ekluzek:
Weirdly fails in the build phase, let's try to figure this out.

samsrabin:
Re-pinging @ekluzek for review, since it would be really good to have this test done before #2686.

samsrabin:
This will be an important part of making our Python system tests portable; see Issue #3236.

samsrabin:
This could probably be simplified by just calling `run_tower` in the run phase, and not splitting up into its actual build vs. run phases. This would have the side benefit of testing `run_tower` as a whole.

PR #2350, opened 2024-02-07
Delete _FillValue and history from parameter files

### LINKED ISSUE #2347: Parameter file formatting has changed
### Brief summary of bug

Changes to the formatting of the default parameter file make the file difficult to work with outside of python.

### General bug information
for example, an ncdump on the new parameter file produces this:
```
ncdump -v bgc_rf_l1s1 /glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/paramdata/ctsm51_params.c240105.nc 
bgc_rf_l1s1 = _ ;
```

whereas doing this for an older file shows:
```
ncdump -v bgc_rf_l1s1 /glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/paramdata/ctsm51_params.c231117.nc 
bgc_rf_l1s1 = 0.55
```

Besides making it difficult to quickly query default parameter values, this bug makes modifying a parameter file as we [describe on the wiki](https://github.com/ESCOMP/CTSM/wiki/Moving-a-parameter-to-the-params-file) impossible.  It also makes it difficult to use nco commands, like ncap2, because it's difficult to then check that modified parameter values have been changed as intended.

latest master tag, ctsm5.1.dev166

**Does this bug cause significantly incorrect results in the model's science?** 
No, but it does make modifying the parameter file with ncdump or nco commands difficult

All

samsrabin:
### Description of changes

Updates parameter files to `c240207`. These are the same as `c240105` except:
- Attribute `_FillValue` has been removed from all variables
- Global attributes `history`, `history_of_appended_files`, and `latest_git_log` have been removed

### Specific notes

CTSM Issues Fixed:
- Fixes #2347

- aux_clm in progress
- `ncdump -v bgc_rf_l1s1` prints the value instead of `_`

### DIFF
--- a/bld/namelist_files/namelist_defaults_ctsm.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm.xml
@@ -484,9 +484,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- The default filenames are given relative to the root directory
      for the CLM2 data in the CESM distribution -->
 <!-- Plant function types (relative to {csmdata}) -->
-<paramfile phys="clm5_1" >lnd/clm2/paramdata/ctsm51_params.c240105.nc</paramfile>
-<paramfile phys="clm5_0" >lnd/clm2/paramdata/clm50_params.c240105.nc</paramfile>
-<paramfile phys="clm4_5" >lnd/clm2/paramdata/clm45_params.c240105.nc</paramfile>
+<paramfile phys="clm5_1" >lnd/clm2/paramdata/ctsm51_params.c240207.nc</paramfile>
+<paramfile phys="clm5_0" >lnd/clm2/paramdata/clm50_params.c240207.nc</paramfile>
+<paramfile phys="clm4_5" >lnd/clm2/paramdata/clm45_params.c240207.nc</paramfile>
 
 <!-- ================================================================== -->
 <!-- FATES default parameter file                                       -->
--- a/cime_config/testdefs/testmods_dirs/clm/ciso_cwd_hr/user_nl_clm
+++ b/cime_config/testdefs/testmods_dirs/clm/ciso_cwd_hr/user_nl_clm
@@ -1,2 +1,2 @@
-paramfile = '$DIN_LOC_ROOT/lnd/clm2/paramdata/ctsm51_ciso_cwd_hr_params.c240105.nc'
+paramfile = '$DIN_LOC_ROOT/lnd/clm2/paramdata/ctsm51_ciso_cwd_hr_params.c240207.nc'
 hist_fincl1 = 'CWDC_HR','C13_CWDC_HR','C14_CWDC_HR','CWD_HR_L2','CWD_HR_L2_vr','CWD_HR_L3','CWD_HR_L3_vr'
--- a/py_env_create
+++ b/py_env_create
@@ -23,7 +23,7 @@ if [ $error != 0 ]; then
    exit -1
 fi
 rm condahelp.txt
-ctsm_python=ctsm_pylib
+ctsm_python=ctsm_pylib3
 
 
 condadir="$dir/python"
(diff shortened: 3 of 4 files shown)


samsrabin:
@wwieder and @olyson: New parameter files are available for you to try:
- `clm45_params.c240207.nc`
- `clm50_params.c240207.nc`
- `ctsm51_params.c240207.nc`
- `ctsm51_ciso_cwd_hr_params.c240207.nc`

olyson:
Thanks, that seems to be working for me.

ekluzek (approved):
Thanks for working on this @samsrabin. This will be nice to get in ASAP, since it's causing trouble for people. I'm also glad that you removed the history attribute as we talked about since it wasn't working in cprnc. history is a standard CF standard variable that some tools also update automatically when a change is made (for example NCO), so it's useful in general, but it had become too long. But, we can add it back in when it makes sense, which as we said will likely be when we start tuning things.

This is of course really simple, since it just changes the parameter files. I'm hitting this as approve so as to not slow it down. But, I do have an important question.

This removes _FillValue that were nan, and _FillValue for scalars both of which make sense. What I'm not so sure about is that it also removes _FillValue for arrays and it's also removing _FillValue's that were consciously set to something. Here's a list of non NaN _FillValues...

``` text
- 		allconsl:_FillValue = 0. ;
- 		allconss:_FillValue = 0. ;
- 		arootf:_FillValue = 0. ;
- 		arooti:_FillValue = 0. ;
- 		astemf:_FillValue = 0. ;
- 		bfact:_FillValue = 0. ;
- 		cc_dstem:_FillValue = -999.99 ;
- 		cc_leaf:_FillValue = -999.99 ;
- 		cc_lstem:_FillValue = -999.99 ;
- 		cc_other:_FillValue = -999.99 ;
- 		declfact:_FillValue = 0. ;
- 		fd_pft:_FillValue = -999.99 ;
- 		fleafi:_FillValue = 0. ;
- 		fm_droot:_FillValue = -999.99 ;
- 		fm_dstem:_FillValue = -999.99 ;
- 		fm_leaf:_FillValue = -999.99 ;
- 		fm_lroot:_FillValue = -999.99 ;
- 		fm_lstem:_FillValue = -999.99 ;
- 		fm_other:_FillValue = -999.99 ;
- 		fm_root:_FillValue = -999.99 ;
- 		fsr_pft:_FillValue = -999.99 ;
- 		gddmin:_FillValue = 0. ;
- 		graincn:_FillValue = 0. ;
- 		grnfill:_FillValue = 0. ;
- 		hybgdd:_FillValue = 0. ;
- 		laimx:_FillValue = 0. ;
- 		lfemerg:_FillValue = 0. ;
- 		max_NH_planting_date:_FillValue = 0 ;
- 		max_SH_planting_date:_FillValue = 0 ;
- 		mergetoclmpft:_FillValue = 0 ;
- 		min_NH_planting_date:_FillValue = 0 ;
- 		min_SH_planting_date:_FillValue = 0 ;
- 		min_planting_temp:_FillValue = 1000. ;
- 		mxmat:_FillValue = 0 ;
- 		mxtmp:_FillValue = 0. ;
- 		pftpar20:_FillValue = 9999.9 ;
- 		pftpar28:_FillValue = 9999.9 ;
- 		pftpar29:_FillValue = 1000. ;
- 		pftpar31:_FillValue = 1000. ;
- 		planting_temp:_FillValue = 1000. ;
- 		pprodharv10:_FillValue = 0. ;
- 		variantnames:_FillValue = "" ;
- 		ztopmx:_FillValue = 0. ;
```

I think those ones likely have some meaning and should probably remain as they were. Other's of the arrays probably should have _FillValue as well, but it would take some time to analyze which ones and what the value should be. So that would be a job for later...

samsrabin:
@ekluzek Good point. I've now made new files, `*c240207b.nc`, with `_FillValue` only removed from variables that had it erroneously added in `c240105`. Does that work?

PR #2353, opened 2024-02-08
Do not crash "make all" even if pylint isn't clean

### LINKED ISSUE #2316: "make all" doesn't run black if lint fails
As of ctsm5.1.dev163, `make lint` errors:
```
...
Your code has been rated at 8.80/10 (previous run: 8.80/10, +0.00)

make: *** [Makefile:50: lint] Error 30
```

This means that, when calling `make all`, the `black` test isn't reached.

Doing either of the following in `python/Makefile` fixes the issue:
1. Add `--fail-under=0` to `PYLINT_ARGS`
2. Change the order of the tests in `all` from `test lint black` to `test black lint`.

samsrabin:
### Description of changes
Ensures that all steps of `make all` are run, even if pylint isn't clean.

### Specific notes

- Fixes #2316 

**Testing performed, if any:** `make all` now reaches "Successfully ran all standard tests" even though pylint isn't clean.

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -27,7 +27,7 @@ PYLINT_SRC = \
 #       ../cime_config/buildlib \
 #       ../cime_config/buildnml
 
-all: test lint black
+all: test black lint
 	@echo
 	@echo
 	@echo "Successfully ran all standard tests"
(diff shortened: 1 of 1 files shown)


ekluzek (approved):
This is great!

PR #2355, opened 2024-02-08
Set initial t_soisno=272 for soils and 274K for urban road

olyson:
### Description of changes

Set initial t_soisno=272K for soils and 274K for urban road

### Specific notes

Soil temperature was intended to be initialized at 272K, replacing the previous setting of 274K, as described in issue https://github.com/ESCOMP/CTSM/issues/1460, but was not implemented correctly.  Instead urban road was set to 272K and soils were unchanged.

Contributors other than yourself, if any: @wwieder @dlawrenncar 

CTSM Issues Fixed (include github issue #): https://github.com/ESCOMP/CTSM/issues/2338

Are answers expected to change (and if so in what way)? Yes. Less soil carbon at high latitudes.  See here for a set of simulations evaluating this: https://github.com/NCAR/LMWG_dev/issues/45

Testing performed, if any:
No system testing performed yet.


slevis-lmwg:
[Link to discussion ](https://github.com/ESCOMP/CTSM/pull/2371#issuecomment-1947508620)of the problem.

olyson:
I have added a use_fates that encompasses the code in question as suggested by @wwieder in issue #2373 .  The test now passes. However, a couple of questions.
1. The setting of t_soisno to 5deg below the freezing temperature when excess_ice is on is outside of the use_fates check and so if the test is changed so that excess_ice is on ( the current default is off), this test will fail (confirmed).  I'm not sure what the future plans for excess_ice are and if it is intended for use with FATES.
2.  There is an indication from the excess_ice PR (https://github.com/ESCOMP/CTSM/pull/1787) that it might be ok to set the initial soil temperature to 272K instead of tfrz-5.  This would simplify the code, particularly if we want to run FATES with excess_ice on.  "src/biogeophys/TemperatureType.F90 -- initial soil temperature set to 268.15 K at the cold start (might be redundant if https://github.com/ESCOMP/CTSM/issues/1460 is added)"

olyson:
Per the discussion at CTSM software meeting, my understanding is that we will not put in a use_fates statement and will simply mark the test as an expected fail.  This will be revisited once #2384 is resolved.  Also, @olyson will test a potential fix from @rgknox from the FATES side.

olyson:
I've backed out the use_fates change and added the FATES test as an expected failure.  I ran the full test suite on both derecho and izumi to make sure there weren't any other problems.  All tests PASSed or FAILed as expected.

ekluzek (approved):
Looks great.

@olyson is this ready to come to master then?

olyson:
Yes, I believe it is ready, thanks.

doc/ChangeLog
```diff
@@ -1,4 +1,71 @@
...
+
+Answer changes
+--------------
+
+Changes answers relative to baseline: YES
+
+  Summarize any changes to answers, i.e.,
+    - what code configurations: all
+    - what platforms/compilers: all
+    - nature of change: roundoff
```
olyson, line 60: I would say larger than roundoff but not climate-changing.  The original diagnostics for this change are here:
https://webext.cgd.ucar.edu/I2000/ctsm51c6_PPEn08ctsm51d023_2deg_GSWP3V1_Sparse400_cs_ts_tsoisno272_2000AD/lnd/ctsm51c6_PPEn08ctsm51d023_2deg_GSWP3V1_Sparse400_cs_ts_tsoisno272_2000AD.381_400-ctsm51c6_PPEn08ctsm51d023_2deg_GSWP3V1_Sparse400_cs_ts_2000AD.381_400/setsIndex.html

olyson (commented):
Looks good other than see my comment about summarizing changes to answers, thanks.

doc/ChangeLog
slevis-lmwg: Thank you, Keith.

PR #2344, opened 2024-02-01
[WIP] Refactoring of `neon_site` into `tower_site`, and `neon_site`

TeaganKing:
### Description of changes
This PR will create a `BaseSite` class which defines common structures that are in both NeonSite and Plumber2Site classes. The `TowerSite` class (child of `BaseSite`) will define common
functionalities that are in both `NeonSite` and `Plumber2Site` classes. With this structure, we will be able to implement a `Plumber2Site` class parallel to the `NeonSite` class that utilizes structure and functionalities in the parent classes.

Contributors other than yourself, if any:
@ekluzek
@adrifoster 

CTSM Issues Fixed (include github issue #):
This will address part of #1487

Are answers expected to change (and if so in what way)?
No, this PR should be BFB.

Any User Interface Changes (namelist or namelist defaults changes)?
No

Issue #1487, opened 2021-09-16
Run Supported Towers capabilities


wwieder:
Expand capabilities from `run_neon` to simulate additional supported tower sites (e.g. PLUMBER2 & AmeriFlux for RUBISCO).

Since we already have surface datasets and datm input for additional tower sites we should be able to build on the `run_neon` script to automate the generation of spinup, postAD and transient simulations for different tower sites.

From our conversation at the SE meeting it sounds like this could be done by creating the appropriate user_mods directories for each network / tower site and then adding some additional functionality to a more generic `run_tower` script that facilitates running one or multiple supported tower sites.

@olyson noted that PUMBER has some additional complications because all tower sites:
- Run with local time, meaning they have to initialize the model at different starting times
- Span different time periods, length of records
- Forcing data are at 30 or 60 minute time intervals
- Use LAI stream files for SP simulations (these have also been created)
@ekluzek suggested this can all be handled by the user_mod directories.

This issue is closely related to #1354, but has more/  updated specifics

ekluzek:
Adding next to this so that we decide who will be working on this and about when we want this to happen.

wwieder:
I don't see us doing this in the short term, so I'll close this issue for now.

wwieder:
I'm reopening this issue because it seems that we'd like to broaden capabilities of single point runs to include PLUMBER2 sites.  This can now build on the subset_data and run_neon capabilities we've already created and can hopefully be extended to this broader set of towers that are being processed through ONEFLUX and evaluated with modelevaluation.org.

wwieder:
Assigning Keith and Sam to this for now, as they were in our meeting yesterday, but we should discuss prioritization for this and who's best suited to do the work.

wwieder:
@ekluzek noted that we want to bring in testing for this python script.  The high level design would to use the same infrastructure to call either NEON or PLUMBER sites.  

We can also use this as an opportunity to clean up some of the logic / duplication of features handled by this python script vs. those in usermod_dirs. I'll set up a meeting to work on this code design.

A timeline for this could be to complete this by December 2023.

ekluzek:
This is part of the work the team will be doing in our hackathon to get this going for tower sites.

Myself and the following...

@adrifoster 
@olyson 
@slevis-lmwg
@TeaganKing 
@wwieder

slevis-lmwg:
[...]

PR #2315, opened 2024-01-11
Refactoring run_neon for PLUMBER2 part1

TeaganKing:
### Description of changes
- [x] 1. Remove unused functions and arguments 
- [x] 2. Move arg_parse to its own file and NeonSite to own file
- [x] 3. Unit tests for arg_parse and NeonSite

Additional (not-yet-ordered) expected changes (for future PR):
-  Build abstract TowerSite class
-  Generic class 
-  Add NEON class with NEON-specific behavior
-  Create new methods and unit tests
-  Add PLUMBER class and behaviors
-  Integration testing

Contributors other than yourself, if any:
@ekluzek @adrifoster 

### Some additional resources
https://docs.google.com/document/d/19QAvUSJY0QJdFF5LrOFhU0JVhfT3QwvFQaf8IiPtXxU/edit#heading=h.cyie9ucyj2pr
https://docs.google.com/document/d/1i57JZgu6vWtdvr8rNMEnrtNIMIl8hB4jUeE4UvxNh38/edit

CTSM Issues Fixed (include github issue #):

Are answers expected to change (and if so in what way)? No. This will work towards supporting additional tower sites.

Note, this is a WIP PR. I changed branches from the original PR-- #2259

PR #2363, opened 2024-02-13
Refactoring of neon_site into `tower_site` and `neon_site`

TeaganKing:
### Description of changes
This PR will create a TowerSite class which defines common functionalities that are in both NeonSite and Plumber2Site classes. With this structure, we will be able to implement a Plumber2Site class parallel to the NeonSite class that utilizes structure and functionalities in the parent classes.

#### Contributors other than yourself, if any:
@ekluzek
@adrifoster

#### CTSM Issues Fixed (include github issue #):
This will address part of https://github.com/ESCOMP/CTSM/issues/1487
 Some work on #2405

#### Are answers expected to change (and if so in what way)?
No, this PR should be BFB.

#### Any User Interface Changes (namelist or namelist defaults changes)?
No

Note: I updated the branch I was using, so this is a continuation of the work done in #2344


samsrabin:
Just a heads up that I've moved this to point at the new b4b-dev branch instead of master. This should allow it to avoid backlog and get merged sooner once it's ready.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -484,9 +484,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- The default filenames are given relative to the root directory
      for the CLM2 data in the CESM distribution -->
 <!-- Plant function types (relative to {csmdata}) -->
-<paramfile phys="clm5_1" >lnd/clm2/paramdata/ctsm51_params.c240105.nc</paramfile>
-<paramfile phys="clm5_0" >lnd/clm2/paramdata/clm50_params.c240105.nc</paramfile>
-<paramfile phys="clm4_5" >lnd/clm2/paramdata/clm45_params.c240105.nc</paramfile>
+<paramfile phys="clm5_1" >lnd/clm2/paramdata/ctsm51_params.c240207b.nc</paramfile>
```
ekluzek, line 487: I think this is showing some updates that aren't in your set of changes. This should be resolved with an update. We'll talk about that in a sec...

cime_config/buildnml
```diff
@@ -144,8 +144,8 @@ def buildnml(case, caseroot, compname):
                 or clm_usrdat_name is "NEON.PRISM"
             ):
                 logger.warning(
-                    "WARNING: Do you have approriprate initial conditions for this simulation?"
-                    + " Check that the finidat file used in the lnd_in namelist is apprporiately spunup for your case"
+                    "WARNING: Do you have appropriate initial conditions for this simulation?"
+                    + " Check that the finidat file used in the lnd_in namelist is appropriately spunup for your case"
```
ekluzek, line 148: Nice spelling correction. Thank you!

python/ctsm/site_and_regional/neon_site.py
```diff
@@ -27,126 +27,33 @@
...
-                        """Existing base case should be a historical type and is not
-                        --rerun with the --overwrite option""",
-                    )
-                # reset the case
-                case.case_setup(reset=True)
-            case_path = case.get_value("CASEROOT")
-
-            if setup_only:
-                return case_path
+        if user_mods_dirs is None:
```
ekluzek, line 45: Ahh this part is specific to NEON because of the path, there will be similar code to this for PLUMBER2, just with a different path.

A possible refactoring would be to have this as a generic function in the TowerSite class and you then give the specific name for. There isn't a lot of code here for that though...

ekluzek: But, we should also add an error check that the directory is valid and that would be a good reason to have a generic function in TowerSite for this.

python/ctsm/site_and_regional/neon_site.py
```diff
@@ -27,126 +27,33 @@
...
-
-        Args:
-        case:
-            case object
-        """
-
-        if case.get_value("BATCH_SYSTEM") == "none":
-            return "none"
-        return case.get_value("batch_query")
+    # def get_batch_query(self, case):
```
ekluzek, line 55: Remove the comments, since we just use the base method.

python/ctsm/site_and_regional/neon_site.py
```diff
@@ -191,204 +100,20 @@ def run_case(
...
-        for reffile in glob.iglob(refrundir + "/{}{}.clm2.r.*.nc".format(self.name, root)):
-            m_searched = re.search(r"(\d\d\d\d-\d\d-\d\d)-\d\d\d\d\d.nc", reffile)
-            if m_searched:
-                refdate = m_searched.group(1)
-            symlink_force(reffile, os.path.join(rundir, os.path.basename(reffile)))
-        logger.info("Found refdate of %s", refdate)
-        if not refdate:
-            logger.warning("Could not find refcase for %s", case_root)
-            return False
+        super().set_ref_case(case)
```
ekluzek, line 109: Since, this just runs super -- I'm not seeing the need for this?

python/ctsm/site_and_regional/run_neon.py
```diff
@@ -175,7 +175,7 @@ def main(description):
     cesmroot = path_to_ctsm_root()
     # Get the list of supported neon sites from usermods
     valid_neon_sites = glob.glob(
-        os.path.join(cesmroot, "cime_config", "usermods_dirs", "NEON", "[!d]*")
+        os.path.join(cesmroot, "cime_config", "usermods_dirs", "NEON", "[!Fd]*")
```
ekluzek, line 178: What does the F do?

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
+"""
+This module includes the definition for the TowerSite class,
+which has NeonSite and Plumber2Site child classes. This class defines common
+functionalities that are in both NeonSite and Plumber2Site classes.
+"""
```
ekluzek, line 5: Maybe mention the ability to add other child classes that could be done here? GenericTowerSite, or another set of random sites...

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+        """
+        self.name = name
+        self.start_year = int(start_year)
+        self.end_year = int(end_year)
+        self.start_month = int(start_month)
+        self.end_month = int(end_month)
+        self.cesmroot = path_to_ctsm_root()
+        self.finidat = finidat
+
+    def __str__(self):
```
ekluzek, line 57: This is something that could be unit tested.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+            initial_time = time.time()
+            build.case_build(case_path, case=case)
+            end_time = time.time()
+            total = end_time - initial_time
+            print("Time required to building the base case: {} s.".format(total))
+            # update case_path to be the full path to the base case
+        return case_path
+
+    # pylint: disable=no-self-use
+    def get_batch_query(self, case):
```
ekluzek, line 161: This could be unit tested.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+        Args:
+        case:
+            case object
+        """
+
+        if case.get_value("BATCH_SYSTEM") == "none":
+            return "none"
+        return case.get_value("batch_query")
+
+    def modify_user_nl(self, case_root, run_type, rundir, site_lines=None):
```
ekluzek, line 174: This could be unit tested.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+                "hist_nhtfrq = -8760",
+                "hist_empty_htapes = .true.",
+            ] + site_lines
+
+        if user_nl_lines:
+            with open(user_nl_fname, "a") as nl_file:
+                for line in user_nl_lines:
+                    nl_file.write("{}\n".format(line))
+
+    def set_ref_case(self, case):
```
ekluzek, line 199: This could be unit tested.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+                print("---- cloning the base case in {}".format(case_root))
+                #
+                # EBK: 11/05/2022 -- Note keeping the user_mods_dirs argument is important. Although
+                # it causes some of the user_nl_* files to have duplicated inputs. It also ensures
+                # that the shell_commands file is copied, as well as taking care of the DATM inputs.
+                # See https://github.com/ESCOMP/CTSM/pull/1872#pullrequestreview-1169407493
+                #
+                basecase.create_clone(case_root, keepexe=True, user_mods_dirs=user_mods_dirs)
+
+        with Case(case_root, read_only=False) as case:
```
ekluzek, line 360: This section is long and should be broken out into a separate method (could be private). That might make it testable as well.

This would be something done later.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+
+        print("using this version:", version)
+
+        if experiment is not False:
+            self.name = self.name + "." + experiment
+        case_root = os.path.abspath(os.path.join(base_case_root, "..", self.name + "." + run_type))
+
+        rundir = None
+        if os.path.isdir(case_root):
+            if overwrite:
```
ekluzek, line 305: This section seems long and should be broken out, which might make it testable.

This would be something for later.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+
+        logger.info("base_case_name : %s", self.name)
+        logger.info("user_mods_dir  : %s", user_mods_dirs[0])
+
+        if overwrite and os.path.isdir(case_path):
+            print("Removing the existing case at: {}".format(case_path))
+            shutil.rmtree(case_path)
+
+        with Case(case_path, read_only=False) as case:
+            if not os.path.isdir(case_path):
```
ekluzek, line 106: Maybe this should be brought out into a separate method?

This would be a future update...

ekluzek (changes requested):
Cool this is great. Nice start here!

I have a few questions we'll go over in a bit. I'd also like to know how the pylint does with this? And is:

cd python
make all

Clean?

Some of the methods are long and could maybe be broken up to be shorter and also to make them unit-testable. That could also be a future refactoring. I added some notes about that.

I also added some notes about things that look like they could be unit-testable. They might already be, but wanted to ask.

It'll be good to go over this in a bit...

python/ctsm/site_and_regional/tower_site.py
ekluzek: Let's leave this off for now.

python/ctsm/site_and_regional/tower_site.py
ekluzek: @TeaganKing believes this already is tested for the run_neon unit test, which is great. So just confirm and if so resolve this conversation.

python/ctsm/site_and_regional/tower_site.py
ekluzek: I'll point you to a way to do this with a dummy case as a way to do this in a unit test.

And we will do this at a later PR.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,413 @@
...
+        ):
+            symlink_force(os.path.join(refrundir, "inputdata"), os.path.join(rundir, "inputdata"))
+
+        case.set_value("RUN_REFDATE", refdate)
+        if case_root.endswith(".postad"):
+            case.set_value("RUN_STARTDATE", refdate)
+        # NOTE: if start options are set, RUN_STARTDATE should be modified here
+        return True
+
+    # pylint: disable=too-many-statements
```
ekluzek, line 245: We should try to avoid disabling too-many-statements as that tells us our code is too long and too complex and should be broken up.

ekluzek: Add a TODO.

python/ctsm/site_and_regional/run_neon.py
ekluzek: This says don't get cases that start with either F or d. It was confusing in FATES cases.

This does mean if we add cases that do start with it, it will fail in possibly a weird way.

So add a comment about that...

TeaganKing:
There are some tests for modifying user namelists in `CTSM/python/ctsm/test/test_unit_neon_site.py`. 
Some additional discussion on unit tests' current implementation is in #2315.

I have addressed other comments that relate to what we wanted to do now.

Regarding 'make all', I'm hoping to test this but am currently out of scratch space and am thus running into errors... Fred is making room elsewhere for a simulation that's taking up the majority of that space, so hopefully I can test this soon.

TeaganKing:
Quick update, I'm currently running into some errors with `test_sys_run_neon.TestSysRunNeon` in `make all` but not with `python3 ./run_ctsm_py_tests  --sys`. In Sam Rabin's comment mentioned [here](https://github.com/ESCOMP/CTSM/pull/2156), changing directories out of the temp dir resolved this issue previously; however, that doesn't seem to be resolving the issue in this case. I am thinking this may be a result of building in tower_site instead of neon_site, but I still need to look into that in more detail to confirm.

The following pylint errors are also present:
`pylint -j 4 --rcfile=ctsm/.pylintrc --fail-under=0 ctsm
************* Module ctsm.crop_calendars.process_ggcmi_shdates
ctsm/crop_calendars/process_ggcmi_shdates.py:450:4: C0206: Consider iterating with .items() (consider-using-dict-items)
ctsm/crop_calendars/process_ggcmi_shdates.py:501:8: C0206: Consider iterating with .items() (consider-using-dict-items)
************* Module ctsm.crop_calendars.grid_one_variable
ctsm/crop_calendars/grid_one_variable.py:25:16: R1735: Consider using {} instead of dict() (use-dict-literal)
************* Module ctsm.test.test_unit_singlept_data_surfdata
ctsm/test/test_unit_singlept_data_surfdata.py:489:8: E1137: "self.ds_test['PCT_URBAN']" does not support item assignment (unsupported-assignment-operation)
ctsm/test/test_unit_singlept_data_surfdata.py:1051:8: E1137: "self.ds_test['PCT_URBAN']" does not support item assignment (unsupported-assignment-operation)
`

ekluzek:
@TeaganKing do you need some help with this or can you continue to figure this out? Maybe we could meet say Friday sometime?

The changing out of temp-dir thing you mention is sometimes tricky because you have to do it on the test that's before the one where it breaks. So it might appear in the files you worked with, but it really originates in whatever test was done before yours. Hopefully, that makes sense enough to fix it, but otherwise let's meet later...

TeaganKing:
A few additional notes on the `make all` test error:
- Running `python run_ctsm_pytests -s` worked, but `make all` failed.
- Commenting out `test_sys_run_neon.test_one_site` results in all tests passing; so the error must be encountered during that particular test rather than as a result of other tests.
- This error is not encountered when running neon at a particular site, eg with `./run_neon --neon-sites DEJU`, the model builds and runs successfully.
- Setting up real (permanent throughout the test) directories instead of temporary directories (that are removed) in `test_sys_run_neon` also did not resolve the issue.

**Error details:**
Occurs after this expected message:
```
---- building a base case -------
---- creating a base case -------
---- base case created ------
---- base case setup ------
---- base case build ------
--- This may take a while and you may see WARNING messages ---
```

Error:
```
======================================================================
ERROR: test_one_site (test.test_sys_run_neon.TestSysRunNeon)
This test specifies a site to run
----------------------------------------------------------------------
  File "/glade/work/tking/temp/CTSM/cime/CIME/utils.py", line 704, in run_sub_or_cmd
    getattr(mod, subname)(*subargs)
  File "/glade/work/tking/temp/CTSM/cime/CIME/build_scripts/buildlib.mpi-serial", line 74, in buildlib
    run_bld_cmd_ensure_logging(cmd, logger)
  File "/glade/work/tking/temp/CTSM/cime/CIME/utils.py", line 2611, in run_bld_cmd_ensure_logging
    expect(stat == 0, filter_unicode(errput))
  File "/glade/work/tking/temp/CTSM/cime/CIME/utils.py", line 175, in expect
    raise exc_type(msg)
CIME.utils.CIMEError: ERROR: cat: Filepath: No such file or directory
/glade/derecho/scratch/tking/tmp/tmp7i5f_fyy/BART/Tools/Makefile:195: *** NETCDF not found: Define NETCDF_PATH or NETCDF_C_PATH and NETCDF_FORTRAN_PATH in config_machines.xml or cmake_macros.  Stop.

During handling of the above exception, another exception occurred:

  File "/glade/work/tking/temp/CTSM/python/ctsm/test/test_sys_run_neon.py", line 61, in test_one_site
    main("")
  File "/glade/work/tking/temp/CTSM/python/ctsm/site_and_regional/run_neon.py", line 227, in main
    cesmroot, output_root, res, compset, user_mods_dirs, overwrite, setup_only
  File "/glade/work/tking/temp/CTSM/python/ctsm/site_and_regional/neon_site.py", line 51, in build_base_case
    case_path = super().build_base_case(cesmroot, output_root, res, compset, user_mods_dirs)
  File "/glade/work/tking/temp/CTSM/python/ctsm/site_and_regional/tower_site.py", line 154, in build_base_case
    build.case_build(case_path, case=case)
  File "/glade/work/tking/temp/CTSM/cime/CIME/build.py", line 1329, in case_build
    return run_and_log_case_status(functor, cb, caseroot=caseroot)
  File "/glade/work/tking/temp/CTSM/cime/CIME/utils.py", line 2477, in run_and_log_case_status
    rv = func()
  File "/glade/work/tking/temp/CTSM/cime/CIME/build.py", line 1320, in <lambda>
    dry_run,
  File "/glade/work/tking/temp/CTSM/cime/CIME/build.py", line 1212, in _case_build_impl
    complist,
  File "/glade/work/tking/temp/CTSM/cime/CIME/build.py", line 840, in _build_libraries
    logfile=file_build,
  File "/glade/work/tking/temp/CTSM/cime/CIME/utils.py", line 716, in run_sub_or_cmd
    expect(False, "{} FAILED, cat {}".format(cmd, logfile))
  File "/glade/work/tking/temp/CTSM/cime/CIME/utils.py", line 175, in expect
    raise exc_type(msg)
CIME.utils.CIMEError: ERROR: /glade/work/tking/temp/CTSM/cime/CIME/build_scripts/buildlib.mpi-serial FAILED, cat /glade/derecho/scratch/tking/tmp/tmp7i5f_fyy/BART/bld/mpi-serial.bldlog.240227-101229
```

@ekluzek I will reach out to you for some help sorting out these details!

TeaganKing:
I wanted to note a few more thoughts on these test failures after meeting with @ekluzek earlier today.

1) We noticed that the tests were succesful prior to some derecho updates / machine maintenance, so this could possibly be tied to the updates to a newer version of CIME.
2) Case.build within the test case directory works fine; this shows that the case setup is done as expected, but for some reason the method of calling case build from within run_neon with the test setup seems to be failing.
3) We considered that there could be an absolute path error; however, using the same absolute paths as in the test and running by hand is successful (eg, `/glade/u/home/tking/work/temp/CTSM/tools/site_and_regional/run_neon --neon-sites BART --setup-only --output-root /glade/derecho/scratch/tking/tmp/temp
`).
4) Looking into Makefile,in the `python` directory it looks pretty straightforward and does not change directories or do anything that would affect a full path.
4) Note that `make all` runs `make utest` and `make stest` (the latter being the relevant part to this issue). `make stest` fails.
5) `./run_ctsm_py_tests -s` succeeds and uses `python3` but through the shebang line
6) Running `module load nco` and then `python3 ./run_ctsm_py_tests --sys` is also successful (which is essentially what make stest does).

This issue is probably tied to details regarding how our testing is performed, and may require some additional brainstorming with folks who are more familiar with our testing suite.

ekluzek:
I replicated the above behavior in my own sandbox. So it's reproducible. I don't know what might be different  inside the Makefile environment than outside, which makes this strange.

A previous problem we've run into is that we are removing directory trees, but not making sure that isn't the current directory. So I tried making that update to other places, and it acts the same way.  I do see something that does need to be fixed in the main code that I'll add a comment for. I'm doubtful that's the issue, but it is something good to fix.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,414 @@
...
+        if not output_root:
+            output_root = os.getcwd()
+        case_path = os.path.join(output_root, self.name)
+
+        logger.info("base_case_name : %s", self.name)
+        logger.info("user_mods_dir  : %s", user_mods_dirs[0])
+
+        if overwrite and os.path.isdir(case_path):
+            print("Removing the existing case at: {}".format(case_path))
+            shutil.rmtree(case_path)
```
ekluzek, line 103: It's critical to test and make sure the directory being removed is NOT the current one. I'm doubtful this is the current problem, but still something that should be checked.

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -0,0 +1,414 @@
...
+
+        if experiment is not False:
+            self.name = self.name + "." + experiment
+        case_root = os.path.abspath(os.path.join(base_case_root, "..", self.name + "." + run_type))
+
+        rundir = None
+        if os.path.isdir(case_root):
+            if overwrite:
+                print("---- removing the existing case -------")
+                shutil.rmtree(case_root)
```
ekluzek, line 308: Another rmtree to check that it isn't the current directory.

TeaganKing:
I did also notice that with a fresh CTSM checkout, if I use the tower_site.py and other files that I've changed, this fails, but if I use the old files, 'make all' does not fail... So this probably is also a result of splitting up run_neon into multiple classes/files. @ekluzek when you say you tested this in your own sandbox, I assume that is with these code changes specifically rather than the current CTSM code?

ekluzek:
@TeaganKing yes exactly. It fails in my sandbox where I checked out your branch, but it passes in master or b4b-dev.

I don't think it's really about anything that you did, but something more subtle in the testing. What I'd like to do is to see if we can step through what's going on with a python debugger, or add more print statements to each thing going on to see how the Makefile environment is different from running directly. One guess I have is that Maybe there are different environment variable settings when run from inside the Makefile, than when run outside of make. Just checked that and the Makefile adds the following env settings:

``` text
> MAKEFLAGS=
> MAKELEVEL=1
> MAKE_TERMERR=/dev/pts/1
> MAKE_TERMOUT=/dev/pts/1
144a149
> MFLAGS=
```

Which doesn't look like it should matter...

Sometimes when new code is added the code can be fine, but it exposes an existing problem. This is rare, but when it happens it becomes really hard to track down.

python/ctsm/site_and_regional/tower_site.py
TeaganKing: This seems like perhaps a separate issue? Shall I make an additional issue ticket, or did you want to make those changes in this PR? I'm thinking it should be in a different PR due to it addressing a separate issue?

TeaganKing:
In a discussion at the CTSM software meeting, we decided to remove the `make stest` and `make utest` components of the makefile due to redundancy with the python tests, so we will not resolve the `make all` issue mentioned above. The makefile changes will be done in a separate PR. The python tests were successful, so this should be good to merge in. @ekluzek I made some comments to your suggestions above.

TeaganKing:
Thanks Erik. Just to put everything in one place, I will describe all of the unresolved conversations here.

**Testing-related issues which we decided to leave as is for now:**
In `neon_site.py`, we could add an error check that the `user_mods_dirs` directory is valid.
TowerSite.__str__ could be unit tested, although it's probably not essential at the moment.
`TowerSite.get_batch_query` and `set_ref_case` could be unit tested.

**Testing outside the scope of this work:**
We should generally add a utility for removing a directory tree since this problem seems to be repeatedly causing errors. This should be a new issue ticket that will probably also be relevant in other parts of CTSM, in addition to `TowerSite.build_base_case` and `TowerSite.run_case`.

**Further Refactoring:**
The `TowerSite.build_base_case` section that starts with `with Case(case_path, read_only=False) as case:` could be brought out into a separate method. `run_case` should also be broken down further in the future.

ekluzek:
Cool, thanks @TeaganKing. I want to get this into b4b-dev so it'll be on the next tag there that we move to main. Which means I need to get it on by tomorrow. So I'll just open those issues talked about above, and close the conversations and all that.

I also did some work on the branch to deal with the temporary directory issue, so I'll check that in. It didn't solve it, but should improve the future. So I might as well check that in here.

Thanks for the work on this. Looking forward to the next piece here.

TeaganKing:
Awesome, thank you Erik!

python/ctsm/site_and_regional/tower_site.py
ekluzek: Yep, let's do this later. I've filed the issue in #2412

python/ctsm/site_and_regional/tower_site.py
ekluzek: #2413

python/ctsm/site_and_regional/tower_site.py
ekluzek: #2413 talks about refactoring to make things unit-testable.

python/ctsm/site_and_regional/tower_site.py
ekluzek: #2413

python/ctsm/site_and_regional/tower_site.py
ekluzek: #2413

python/ctsm/site_and_regional/tower_site.py
ekluzek: #2412

python/ctsm/site_and_regional/neon_site.py
ekluzek: #2412 and #2413

python/ctsm/site_and_regional/tower_site.py
ekluzek: #2413

ekluzek (approved):
Resolved all the conversations, so this can be merged in.

ekluzek:
python testing is good and clm_pymods tests on Derecho pass. So merging in.

PR #2389, opened 2024-02-26
Minor docs updates

samsrabin:
### Description of changes
Fixes a table in the Methane tech note, and updates instructions for changing tillage settings.

### Specific notes

Testing performed, if any: Build works with both Docker method and VSCode.


ekluzek (approved):
I can't really comment on this. But, it's simple and I have to assume you are correcting things that need to be updated. So I wanted to make sure you have at least one approval here.

PR #2390, opened 2024-02-28
Hillslope merge

samsrabin:
See #1715. This makes one minor change and updates ChangeLog/Sum.


ekluzek (approved):
The approvals for this are really happening in #1715. And this has been extensively been looked at by others. This is just bringing in the main parts of Hilslope there, while approval and other smaller changes happen in #1715.

So I am approving this as is.

Issue #2400, opened 2024-03-05
Add hillslope scripts


samsrabin:
@swensosc has various scripts related to making files/variables for use in hillslope runs. Some of these are not CTSM-specific, so they're in a separate [Representative_Hillslopes repo](https://github.com/swensosc/Representative_Hillslopes). Some are CTSM-specific, though, so I need to add those to `python/ctsm/`. This will also require adding the Representative_Hillslopes repo as an optional external to CTSM.

PR #2401, opened 2024-03-05
Add hillslope scripts (to python/ctsm/) and external

samsrabin:
### Description of changes
Sean Swenson has various scripts related to making files/variables for use in hillslope runs. Some of these are not CTSM-specific, so they're in a separate [Representative_Hillslopes repo](https://github.com/swensosc/Representative_Hillslopes). Some are CTSM-specific, though, so this PR will add those to `python/ctsm/`. It also adds the Representative_Hillslopes repo as an optional external to CTSM.

### Specific notes

**Contributors other than yourself, if any:** Sean Swenson (@swensosc)

**CTSM Issues Fixed (include github issue #):** #2400

**Testing performed, if any:** None so far.


samsrabin:
Should be rebased onto current `master` after the next `master` merge into `b4b-dev` (March 7).

src/main/initVerticalMod.F90
```diff
@@ -80,9 +81,75 @@ subroutine readParams( ncid )
...
+    ! (e.g.) 1 on the source file correspond to layers with a value of 1 on the
+    ! destination file. So if you change the values of these constants, you either need to
+    ! adequately inform users of this change, or build in some translation mechanism in
+    ! init_interp (such as via adding more metadata to the restart file on the meaning of
+    ! these different values).
+    !
+    ! The distinction between "shallow" and "deep" bedrock is not made explicitly
+    ! elsewhere. But, since these classes have somewhat different behavior, they are
+    ! distinguished explicitly here.
+    integer, parameter :: LEVGRND_CLASS_STANDARD        = 1
```
ekluzek, line 110: Technically our standard says we ONLY use uppercase for CPP tokens. But, CPP tokens are now very rare in CTSM, and that was long enough ago that we had extensive use of CPP tokens, that I think it's OK to change our standard to allow it for parameters. This is a common code convention now and it's useful.

But, we should agree to this and change the standard on the wiki.

src/main/clm_instMod.F90
```diff
@@ -268,6 +271,14 @@ subroutine clm_instInit(bounds)
          urbanparams_inst%thick_wall(begl:endl), &
          urbanparams_inst%thick_roof(begl:endl))
 
+    ! Set hillslope column bedrock values
+    if (use_hillslope) then
+       call SetHillslopeSoilThickness(bounds,fsurdat, &
+            soil_depth_lowland_in=8.5_r8,&
```
ekluzek, line 277: What's the meaning of the hardcoded constants here (8.5 and 2.0)? Does it improve readability to make them a parameter?

src/main/controlMod.F90
```diff
@@ -257,6 +257,11 @@ subroutine control_init(dtime)
 
     namelist /clm_inparm/ use_biomass_heat_storage
 
+    namelist /clm_inparm/ use_hillslope
+
+    namelist /clm_inparm/ downscale_hillslope_meteorology
```
ekluzek, line 262: It seems like downscale_hillslope_meteorology and use_hillslope_routing could be in the Hillslope specific namelist rather than here.

use_hillslope is an overall control variable, so needs to be here, but it looks like to me these two are only used if use_hillslope is on.

ekluzek: By having namelist controls in the most specific location as possible, it makes it easier to change parameterization packages (add new versions, or remove old ones). The control variables at this level are kind of out of control (too many of them), so I like to make them specific if it can be done.

ekluzek (changes requested):
Thanks for the work on this @samsrabin and @swensosc! Nice to see this coming in.

I have some questions, requests for changes, and things I point out to do later. I think this likely would be good to go over this.

To be honest -- I thought this might be smaller than it really was -- so I took on reviewing it. But, might not have had I realized the size. LOL. That's not bad, just a personal confession.

I do wonder if someone else should review it as well? And it might be good to just go through together. So maybe the second person doesn't do much looking beforehand, just goes through the walkthrough with us?

samsrabin:
@ekluzek Thanks for the comments! This is still very much a draft, though; I really only just copied in Sean's scripts.

The commit will be much smaller once it's properly rebased as I mentioned [here](https://github.com/ESCOMP/CTSM/pull/2401#issuecomment-1979413932).

samsrabin:
Replacing this with #2425. @ekluzek, if you'd like to preserve your review comments, please file a new issue.

ekluzek:
Thanks @samsrabin I'll leave them here for now. We may discuss if they come up after review of the new PR. Although I'll probably make an issue of all uppercase in the FORTRAN code for constants.

PR #2427, opened 2024-03-15
Merge ctsm5.2 branch up to ctsm5.1.dev174

ekluzek:
### Description of changes

Merge ctsm5.2 branch up to the latest ctsm5.1 tag.

Testing performed, if any: regular testing will be done


ekluzek:
Testing showed that the Hillslope tests failed, MKSURFDATAESMF_P64x1.f10_f10_mg37.I1850Clm50BgcCrop.derecho_intel, and the USUMB test. The Hillslope tests fail, because they need surface datasets that work with ctsm5.2.0. MKSUIRFDATAESMF fails because when run on a compute node the hostname match fails. And the USUMB test is one we removed, but was still in the test list.

I've got a hacked ctsm5.2.0 hillslope surface dataset that seems to be working. So this is almost there.

ekluzek:
@samsrabin and I met. We agreed that the best way forward is for him to create a new surface dataset based on the new ctsm5.2.0 one. We realized that we want to base it on the ctsm5.2.0 data in general, and then add the hillslope information to it. The hillslope information is tied to the surface dataset landmask and natural-vegetation mask (where PCT_NATVEG > 0) so closely tied to the surface dataset. Since those masks might have changed in ctsm5.2.0, we might as well use the current processing that makes those datasets.

ekluzek:
The other problem I'm having is with the MKSURFDATAESMF. I fixed the first problem, but now there is a problem with running mksurfdata_esmf in the test.

As a first check I'm running a f10 case by hand in the tools directory. The build and create namelist process is fine. But, there is trouble with running. So I need to figure that out.

This is likely due to the externals update. Most likely in ccs_config and/or in cime.

ekluzek:
OK, the problem is in using mpibind rather than mpiexec in the run. This was updated in cime for the externals.

I am able to get it to run with mpiexec with the following binding...

`mpiexec -n 64 -ppn 64 --cpu-bind depth -d 2 --  $EXEC < $NAMELIST`

passing the needed arguments to mpibind, so far isn't working for me.

ekluzek:
I talked with CISL (Rory and Daniel) about mpibind and mpiexec, and the thing that needs to be done is to ask for the full node of processors to PBS, but let mpibind figure out how to bind processors within the node. So remove the cpu-bind and depth parts of this.

mpibind will make sure that processors on each socket are used which is important to ensure full memory bandwidth is being used.

Here's what Rory told me for reference...

> Hi Erik,
> 
> On Derecho (in the main queue), jobs are scheduled exclusively, so you are charged in terms of whole nodes, whether or not you decide to use all of the processors / memory / GPUs that are on the nodes you have requested. For performance reasons, your requests should always have ncpus=128, because PBS will creates a cgroup with the cpus you request, and preferentially uses CPU cores on socket-0, and then socket-1. So, if you do:
> 
> `#PBS -l select=ncpus=64:mpiprocs=64 `
> you are potentially limiting yourself to 1/2 the memory bandwidth available on the node. Even if you only want to use 64 MPI ranks, you should make the request as
> 
> `#PBS -l select=ncpus=128:mpiprocs=64 `
> in this case mpibind will put 32 ranks on each socket, and still allow you to use the full memory bandwidth of the node.
> 
> Rory

ekluzek:
OK, now testing is looking reasonable with only the following tests failing:

SMS.f45_f45_mg37.I2000Clm51FatesSpRsGs.derecho_nvhpc.clm-FatesColdSatPhen
SMS_D_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.izumi_intel.clm-tracer_consistency		
SMS_D_Ld3.f10_f10_mg37.I2000Clm51Bgc.izumi_nag.clm-HillslopeD		
SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm51Fates.izumi_nag.clm-FatesFireLightningPopDens--clm-NEON-FATES-NIWO

python testing is passing, and the previous fails are now working. I also ensured that the mksurfdata_esmf generation is working -- by hand testing a few cases and running the pot-veg case from the Makefile (in addition to the python testing).

ekluzek:
I was able to get most of the above passing. Hillslope needed rimport run on a file. Some tests just needed to be resubmitted.

This test: SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm51Fates.izumi_nag.clm-FatesFireLightningPopDens--clm-NEON-FATES-NIWO
fails (for me) because of the filename issue in CDEPS..
https://github.com/ESCOMP/CDEPS/issues/263

This test also fails:

SMS_D_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.izumi_intel.clm-tracer_consistency

which looks like a legit problem.

ekluzek:
I'm going to go over some of the python code changes with @samsrabin a bit later.

I might flag the one failing test as a problem so I can move onto the next steps for ctsm5.2. And then fix it in one of the two remaining tags on the ctsm5.2 branch.

python/ctsm/test/test_unit_gen_mksurfdata_jobscript_single.py
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/env python3
```
ekluzek, line 1: @samsrabin this is the new unit tester I added that I'd like to go over.

python/ctsm/toolchain/gen_mksurfdata_jobscript_multi.py
```diff
@@ -5,11 +5,18 @@
 """
 import os
 import sys
-import argparse
+import logging
 
-from ctsm.path_utils import path_to_ctsm_root
```
ekluzek, line 10: @samsrabin this is the other script to go over.

python/ctsm/toolchain/gen_mksurfdata_jobscript_single.py
```diff
@@ -7,6 +7,7 @@
 import argparse
 import logging
 
+
```
ekluzek, line 10: And this is the last script to go over.

ekluzek:
Looks like the alpha-ctsm5.2.mksrf.22_ctsm5.1.dev168 test fails in alpha-ctsm5.2.mksrf.23_ctsm5.1.dev171 but passed in alpha-ctsm5.2.mksrf.22_ctsm5.1.dev168. So I'll file it as a problem.

python/ctsm/toolchain/gen_mksurfdata_jobscript_single.py
ekluzek: @samsrabin and I went over this together. One idea to do later would be to separate single and multi into a class that has a base class used by both, and then extends it for each specific script. I'll add that to the issue that talks about this.

ekluzek: We also thought of an improvement in the name of the "part1" to include batch in the name. This is easy to do.

python/ctsm/toolchain/gen_mksurfdata_jobscript_multi.py
ekluzek: This is straight forward because it's mostly removing things in order to use the subroutines from single.

samsrabin (approved):

python/ctsm/toolchain/gen_mksurfdata_jobscript_multi.py
ekluzek: As part of this I did remove the capability of setting the queue (which isn't needed for Derecho). single does set the queue to a proper default for each machine listed. One issue that was there was that the queue is different depending on the machine, and the list of queues also varies. So this would need more infrastructure to support.

So removing it for now seemed like the best approach.

python/ctsm/test/test_unit_gen_mksurfdata_jobscript_single.py
ekluzek: Some suggestions here were to move the expected value setting from Setup to where it's actually used.

Some names of tests could be better to show they are failure tests.
There is a utility function that could be moved out.

python/ctsm/toolchain/gen_mksurfdata_jobscript_single.py
ekluzek: In issues #2446 and #2439

python/ctsm/toolchain/gen_mksurfdata_jobscript_multi.py
ekluzek: In issues #2446 and #2439

python/ctsm/test/test_unit_gen_mksurfdata_jobscript_single.py
ekluzek: In issues #2446 and #2439

PR #2425, opened 2024-03-15
[WIP] Add hillslope scripts (to python/ctsm/) and external

samsrabin:
# Work in progress! Do not review! At risk of force pushes!
### Description of changes
Sean Swenson has various scripts related to making files/variables for use in hillslope runs. Some of these are not CTSM-specific, so they're in a separate [Representative_Hillslopes repo](https://github.com/swensosc/Representative_Hillslopes). Some are CTSM-specific, though, so this PR will add those to `python/ctsm/`. It also adds the Representative_Hillslopes repo as an optional external to CTSM.

Replaces PR #2401, which I couldn't get to rebase properly.

### Specific notes

**Contributors other than yourself, if any:** Sean Swenson (@swensosc)

**CTSM Issues Fixed (include github issue #):**
- Resolves #2400

**Are answers expected to change (and if so in what way)?** No

**Any User Interface Changes (namelist or namelist defaults changes)?** No

**Testing performed, if any:** None so far.

PR #2434, opened 2024-03-22
Move hillslope variables off surface dataset

samsrabin:
### Description of changes

This PR moves the variables required for hillslope runs off the surface dataset and onto the new `hillslope_file`. This makes it so we don't have to maintain a hillslope-specific surface dataset for testing purposes. If/when running with hillslopes becomes the default, we can move these variables back onto the surface dataset.

### Specific notes

**Documentation?** User's Guide documentation will come in on my PR that adds the hillslope scripts to `python/ctsm/` (#2425).

- [x] Initial `hillslope` suite testing.

Waiting until this is next in queue
- [x] Merge in latest master tag
- [x] Derecho and Izumi aux_clm tests are b4b-identical with latest master tag, aside from hillslope tests, which should diff because of new mesh files.


samsrabin:
Blocked: Waiting on 5.2.

samsrabin:
5.2 is in, so this is no longer blocked.

samsrabin:
I have to change the `5x5_amazon` hillslope tests from `I1850` to `I2000` because there's no default 2000 surface dataset for that resolution. I've thus retargeted this branch at master instead of b4b-dev.

samsrabin:
@swensosc Would you mind having a look over this? Happy to walk you through it if you prefer.

swensosc:
I think most of it looks straightforward (changing input file).  You may wish to add a phrase to this warning "This means all hillslope columns in a gridcell will read identical values" to indicate that this will occur even if the finidat file has hillslope data on it.  Also, is there a check to ensure that hillslope_file is of the same spatial resolution as fsurdat?  If not, is that possible?

samsrabin:
Thanks, @swensosc!
- [ ] Add check that `fsurdat` and `hillslope_file` are the same resolution.
- [ ] Add text to warning message.

bld/CLMBuildNamelist.pm
```diff
@@ -2743,6 +2743,13 @@ SIMYR:    foreach my $sim_yr ( @sim_years ) {
         $log->fatal_error("$useinitvar is being set for you but a $var was not found, so $useinitvar, init_interp_attributes, and finidat must not be set correctly for this configuration in the namelist_default file" );
      }
   }
+
+  # this check has to be here and not earlier since use_init_interp is set here and hillslope is already set above in setup_logic_hillslope
+  if ( &value_is_true($nl->get_value($useinitvar)) && value_is_true($nl->get_value("use_hillslope")) ) {
+     $log->warning("WARNING: You have set use_hillslope while $useinitvar is TRUE.\n This means all hillslope columns in a gridcell will read identical values" .
+                   " from initial conditions. If you are sure you want this behaviour:")
```
slevis-lmwg, line 2750: The last sentence seems incomplete. Does the code give instructions for bypassing warnings after the ":"?

slevis-lmwg (approved):
Looks good to me, @samsrabin

bld/CLMBuildNamelist.pm
samsrabin: Good catch. Resolved with 4b032be5b.

bld/CLMBuildNamelist.pm
```diff
@@ -2759,8 +2759,7 @@ SIMYR:    foreach my $sim_yr ( @sim_years ) {
 
   # this check has to be here and not earlier since use_init_interp is set here and hillslope is already set above in setup_logic_hillslope
   if ( &value_is_true($nl->get_value($useinitvar)) && value_is_true($nl->get_value("use_hillslope")) ) {
-     $log->warning("WARNING: You have set use_hillslope while $useinitvar is TRUE.\n This means all hillslope columns in a gridcell will read identical values" .
-                   " from initial conditions. If you are sure you want this behaviour:")
+     $log->warning("WARNING: You have set use_hillslope while $useinitvar is TRUE.\n This means all hillslope columns in a gridcell will read identical values from initial conditions, even if the initial conditions (finidat) file has hillslope information. If you are sure you want this behaviour, add -ignore_warnings to CLM_BLDNML_OPTS.")
```
slevis-lmwg, line 2762: nit: I'm pretty sure that `$log->warning` will generate a default "-ignore_warnings" comment, though I forget whether it tells users where to add the -ignore_warnings...

src/main/surfrdMod.F90
```diff
@@ -46,6 +46,83 @@ module surfrdMod
 
 contains
 
+  subroutine check_domain_attributes(ncid, begg, endg, ldomain, info)
```
slevis-lmwg, line 49: Cool!

slevis-lmwg (approved):
Thanks, @samsrabin!

Issue #2222, opened 2023-10-26
Fixing Negative Ice Fluxes from Ocean to Glacier


Katetc:
The default method of running CISM in CESM is "No-Evolve" or non-active ice sheet. Because the ice sheet can't grow or change in this configuration, all water fluxes through the ice sheet must be balanced. As long as CISM is running in any form, CTSM will calculate and downscale surface mass balance over the ice sheet(s), including melt fluxes in the summer and high snow amounts in the winter. In order to balance summer time melt fluxes from the ice sheet, CTSM will pull water from the ocean (a negative ice flux) to counter ice sheet melting in the summer. 

A more realistic method for handling this need for balance would be to store the excess snow that is sent as run-off in the winter (due to the inactive ice sheet not being able to grow) in a water reservoir that can be accessed in the summer months to balance glacier melt. Our plan is to have excess snow capping over Greenland added to a "snow reservoir" that can be dealt out to balance melt in the summer months and decrease the need for water from the ocean.

I have started implementation of this and we would love to see it go into the CTSM version for CESM3.

ekluzek:
@Katetc thanks for filing this. Is this related to #423?

ekluzek:
@Katetc is this being worked on? Is there still desire for this for the CESM3 release?

ekluzek:
In discussion with @Katetc the plan now is to move this functionality out of CTSM and into a Data Glacier model dglc into CDEPS.

ekluzek:
We are going to do #1136 rather than this, so closing this one.

Issue #1136, opened 2020-09-01
Use data glc model where we used to use CISM2%NOEVOLVE


billsacks:
In compsets that currently use CISM2%NOEVOLVE – and will soon change to use SGLC – we'd like to instead use a data glc model in a time-constant mode. This should provide all of the benefits that we currently get from CISM2%NOEVOLVE: dictating glacier area and topographic heights over the ice sheet domain and providing a high-resolution grid onto which SMB and other fields can be downscaled.

We plan to develop this data glc model in CDEPS, which relies on NUOPC. So we will wait to make this switch until we are using NUOPC as a standard mode of operation (footnote this has happened now).

Definition of done:
- [x] Update CISM and CTSM to use GLC_USE_ANTARCTICA #2222
- [x] Update CTSM with a version of CDEPS that supports DGLC (https://github.com/ESCOMP/CDEPS/issues/25) and (https://github.com/ESCOMP/CDEPS/pull/268)
- [x] Update to a DGLC Verision that handles negative runoff for glaciers
- [ ] Add I compsets to CTSM that exercise DGLC
- [ ] Add testing to CTSM that test needed DGLC cases
- [ ] Change most I compsets/tests to use DGLC rather than SGLC

billsacks:
Blocked: depends on https://github.com/ESCOMP/CDEPS/issues/25

ekluzek:
LIWG is going to support a DGLC model rather than CISM%NOEVOLVE mode for CESM3. This means rather than fixing #2222 for CISM/CTSM the ability to run cases with DGLC will be setup. This is important for fully coupled compsets so that negative runoff is NOT sent to the ocean model. It looks right now the only CISM%NOEVOLVE compsets and tests are B1850 cases. So minimally we would only need to have a few tests around DGLC for 1850 with Bgc-Crop.

I added some notes about what needs to happen to the introduction for a "Definition of Done" above.

billsacks:
@ekluzek - responding to and clarifying some points:
- The current implementation of dglc does not yet deal with the negative runoff issue. That is going to involve some substantial development, but yes, I *think* the current tentative plan is now to do that development in dglc rather than CTSM.
[...]

PR #2449, opened 2024-03-30
change needed for the addition of a dglc component in cdeps

mvertens:
### Description of changes
change needed for the addition of a dglc component in cdeps

### Specific notes
This minor change to buildnml is needed in order to support the addition of dglc which will replace CISM running in noevolve mode.
This will enable having many of the current CTSM tests that use SGLC move to using DGLC%NOEVOLVE.

CTSM Issues Fixed: None

Testing performed, if any:
Ran a test case with 
compset 1850_DATM%GSWP3v1_CLM50%SP_SICE_SOCN_SROF_DGLC%NOEVOLVE_SWAV  and resolution  f10_f10_ais8gris4_mg37 
and compared to 
compset 1850_DATM%GSWP3v1_CLM50%SP_SICE_SOCN_SROF_CISM%NOEVOLVE_SWAV  and resolution  f10_f10_ais8gris4_mg37


ekluzek:
@mvertens thanks so much for working on this and bringing this in. I really like the idea of having a data GLC model to take the place of CISM NO-EVOLVE mode. This also resolves #2222 for us which is really important.

But, I have some questions for you on this:

- It looks like this is a work in progress correct? How close is this to being considered done by you?
- Will it require a CDEPS update to bring in?
- Right now this won't change answers for anything existing correct?

Since this is adding new capability and is fairly straight forward (and provided it doesn't change answers otherwise)-- I think I should rebase this to the b4b-dev branch. So I'll do that once you verify Doing that means it's easier to get in sooner. I'll work with you to do that once I hear more about this from the questions above.

I'm also going to make an issue about adding DGLC to CTSM which will allow us to close #2222 and track everything that will need to be done for this change. Again I'll wait until hearing and then I'll have you add anything there that I'm missing after the issue is started...

ekluzek:
Thanks for that update @billsacks. So there is only one more thing needed here and that is an update for CISM (once that is done). Other things will be done in future PR's and we can document those in the issue I'll open.

So things needed for this:

- [ ] Create an issue that documents all the things needed for DGLC (Erik will do this)
- [ ] Update CISM so it uses GLC_USE_ANTARCTICA (@Katetc and @billsacks will do this)
- [ ] Rebase this to b4b-dev (Erik can do this)

ekluzek (approved):
This is really great glad to see this coming in.

The code changes right now are very simple, so not much to review. But, having both DGLC and CISM use the same XML variable for similar behavior is a nice improvement.

ekluzek:
@billsacks can you comment on how much of our testing should change from using the stub-glacier model to DGLC? Are there reasons outside of fully coupled B1850 compsets that we should prefer running with DGLC over SGLC? Scientifically it's obviously an update to use DGLC rather than a stub, but in what ways will this matter for CTSM simulations?

wwieder:
Glad to see this can come in on b4b_dev.  Is that we want to get in this week, or can it wait another 2 weeks? I guess the logistics are a question for @billsacks and @ekluzek .  The urgency is more a question for @mvertens and @Katetc?

billsacks:
> can you comment on how much of our testing should change from using the stub-glacier model to DGLC? Are there reasons outside of fully coupled B1850 compsets that we should prefer running with DGLC over SGLC? Scientifically it's obviously an update to use DGLC rather than a stub, but in what ways will this matter for CTSM simulations?

I'll comment on that over in #1136 

> Glad to see this can come in on b4b_dev. Is that we want to get in this week, or can it wait another 2 weeks? I guess the logistics are a question for @billsacks and @ekluzek . The urgency is more a question for @mvertens and @Katetc?

It would be great to get this in this week, since this is required for the CDEPS change that we want to merge.

ekluzek:
OK in that case I've rebased it to b4b-dev and as long as we get the CISM tag in by Wednesday morning we can bring it into b4b-dev, which will then go to master on Thursday.

ekluzek:
Testing for aux_clm looks as expected. So merging.

PR #2457, opened 2024-04-05
Relax tolerance for truncating small snocan values in CanopyFluxes

### LINKED ISSUE #2444: Failing water isotope test on the ctsm5.2 branch
### Brief summary of bug

As seen here... https://github.com/ESCOMP/CTSM/pull/2427#issuecomment-2026011205

The test SMS_D_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.izumi_intel.clm-tracer_consistency fails in what will be alpha-ctsm5.2.mksrf.25_ctsm5.1.dev175 as well as alpha-ctsm5.2.mksrf.23_ctsm5.1.dev171 but passed in alpha-ctsm5.2.mksrf.22_ctsm5.1.dev168.

### General bug information

**CTSM version you are using:** alpha-ctsm5.2.mksrf.23_ctsm5.1.dev171

**Does this bug cause significantly incorrect results in the model's science?** [No

**Configurations affected:** with enable_water_tracer_consistency_checks = .true.

cesm.log:

 ```
This routine is depricated - use shr_log_setLogUnit instead        -134
 shr_file_mod.F90         912 
 This routine is depricated - use shr_log_setLogUnit instead        -132
 shr_file_mod.F90         912 
 This routine is depricated - use shr_log_setLogUnit instead        -134
ERROR in CompareBulkToTracer: tracer does not agree with bulk water
Called from: after first stage of hydrology
Variable: snocan_patch
First difference at index: 24
Bulk  :   0.40122337938811679-295
Tracer:   0.00000000000000000E+00
ratio:   0.10000000000000000E-09
Bulk*ratio:   0.40122337938811680-305
iam = 20: local  patch    index = 24
iam = 20: global patch    index = 2947
iam = 20: global column   index = 2211
iam = 20: global landunit index = 470
iam = 20: global gridcell index = 165
iam = 20: gridcell longitude    =   90.0000000
iam = 20: gridcell latitude     =   40.0000000
iam = 20: pft      type         = 8
iam = 20: column   type         = 1
iam = 20: landunit type         = 1
 ENDRUN:
 ERROR: CompareBulkToTracer: tracer does not agree with bulk water
cesm.exe           0000000004AE6556  Unknown               Unknown  Unknown
cesm.exe           000000000421343A  shr_abort_mod_mp_         114  shr_abort_mod.F90
cesm.exe           00000000042132A0  shr_abort_mod_mp_          61  shr_abort_mod.F90
cesm.exe           0000000000A3C63F  abortutils_mp_end          98  abortutils.F90
cesm.exe           00000000035962A7  watertracerutils_         401  WaterTracerUtils.F90
cesm.exe           00000000035B7728  watertype_mp_trac         953  WaterType.F90
cesm.exe           0000000000A5AAE0  clm_driver_mp_clm         588  clm_driver.F90
cesm.exe           00000000009915B2  lnd_comp_nuopc_mp         904  lnd_comp_nuopc.F90
libesmf.so         00007F242D0B3A7E  _ZNK5ESMCI13Metho         377  ESMCI_MethodTable.C
libesmf.so         00007F242D0B4EE2  _ZN5ESMCI11Method         563  ESMCI_MethodTable.C
libesmf.so         00007F242D0B3443  c_esmc_methodtabl         317  ESMCI_MethodTable.C


```

### PR DESCRIPTION
**Pull request by billsacks**
### Description of changes

The default of 1e-13 was letting through some state updates that look like they should have been truncated to 0: each time step, snocan was decreasing by about 12 orders of magnitude.

It looks like what's happening is that snocan is periodically getting set to a very small but non-zero value – maybe coming from a small value of liqcan (which probably needs its own truncate_small_values, which might solve this issue, but that's a matter for another time). Then the model seems to be "trying" to evaporate all of this small snocan in CanopyFluxes in the next time step: the fluxes have values that reduce snocan by something like 12 orders of magnitude. But that reduction isn't quite enough to trigger the truncate_small_values call that @slevis-lmwg introduced, since that uses the default relative epsilon of 1e-13. By loosening the tolerance on the epsilon value, we allow the state update to achieve what seems to be its "desired" reduction of the state to exactly 0 in a single time step.

I think the reason this was causing a failure in the tracer test is that eventually the snocan state got close to the smallest representable double precision value: the test failed when snocan reached 4e-296; the tracer in question is set up to be 10 orders of magnitude smaller, so it should have had a value of 4e-306 at this point. But that's getting very close to limit of representation of double precision, which is about 1e-308, so I think the tracer value lost precision and was rounded to 0, while the bulk value was still very slightly non-zero.

My hypothesis for why this just started appearing with the change that @ekluzek found in https://github.com/ESCOMP/CTSM/issues/2444#issuecomment-2035802437 is that the new freezing soil temperatures probably cause veg temperatures to be below freezing, which causes snocan to come into play where it maybe didn't before.

### Specific notes

- Resolves #2444 

Are answers expected to change (and if so in what way)? YES - roundoff level changes expected for some (maybe many) tests (even those without water tracers)

Testing performed, if any:
Ran SMS_D_Ld10.f10_f10_mg37.I2000Clm50BgcCrop.izumi_intel.clm-tracer_consistency, verified that it PASSes now

### CODE REVIEW

File: src/biogeophys/CanopyFluxesMod.F90 [hunk 1]
```diff
@@ -1624,7 +1624,8 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
       ! snocan < rel_epsilon * snocan_baseline will be set to zero
       ! See NumericsMod for rel_epsilon value
       call truncate_small_values(fn, filterp, begp, endp, &
-         snocan_baseline(begp:endp), snocan(begp:endp))
+         snocan_baseline(begp:endp), snocan(begp:endp), &
+         custom_rel_epsilon=1.e-10_r8)
```
billsacks, line 1628: The default value is 1e-13. A value of 1e-12 or 1e-11 would probably be sufficient here... maybe it wouldn't truncate the state to 0 in the first possible time step, but it probably would be triggered before the tiny value became an issue... but I felt like it would be safer to add a few orders of magnitude of slop. The risk of making this too big is that we could truncate values to zero that really are supposed to be non-zero, but even with 1e-10 we're still only triggering the truncation if the post-state-update value is 10 orders of magnitude smaller than the initial value, so this still seems *pretty* safe to me.

ekluzek (approved):
Awesome. Thanks for figuring this out @billsacks.

I'm thinking for visibility this should go to master rather than ctsm5.2 since it's an answer change, and I'd like it to be more obvious than bound up in all the ctsm5.2 changes. But, I'll ask other opinions...

billsacks:
> I'm thinking for visibility this should go to master rather than ctsm5.2 since it's an answer change, and I'd like it to be more obvious than bound up in all the ctsm5.2 changes. But, I'll ask other opinions...

That makes sense to me. I branched off of the 5.2 branch so that I could reproduce the issue and verify that this resolves it, but I agree with your approach here.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: We'll discuss this as part of #2458. Bringing this in like this makes sense to me for now.

I'm resolving this to free this PR up...

ekluzek:
I rebased to master, which means it now contains all of ctsm5.2, but that should be cleared up when ctsm5.2.0 is made on master.

slevis-lmwg:
@billsacks, could you give me collaborator permissions, so that I may update this PR to the latest ctsm tag in preparation for merging to master?

slevis-lmwg:
Test results here
/glade/derecho/scratch/slevis/tests_0706-120617de
Diffs exceed roundoff, but likely because they grow over time.
@billsacks could you give your assessment (and blessing if appropriate)?

billsacks:
@slevis-lmwg I have given you collaborator permissions on my fork. Thank you!

This PR now looks messed up. If just merging in the latest master doesn't clear things up, we could open a new PR with commit fef58b6d3 cherry-picked.

Given the minimal nature of these changes, I don't feel a need to look at the test results themselves: As you were seeing, I expect changes that grow from initially-roundoff-level to something bigger.

PR #2461, opened 2024-04-07
Fix crop phenology bugs with Gregorian calendar

### LINKED ISSUE #1595: Minor(?) bugs in crop phenology when running with a Gregorian calendar
### Brief summary of bug

From some code inspection, I see what look like minor bugs in crop phenology when running with a Gregorian calendar.

### General bug information

**CTSM version you are using:** latest master

**Does this bug cause significantly incorrect results in the model's science?** Maybe: I haven't thought carefully or investigated the magnitude of the impact. I suspect the impact is small, but I'm not positive about that.

**Configurations affected:** Runs with prognostic crops and a Gregorian calendar

### Details of bug

I see two issues that look like bugs (though there may be others as well):

1. Some crops have a `max_SH_planting_date` of 1231. Normally `get_calday` will turn this into the value 365. However, if the run uses a Gregorian calendar, then `get_calday` uses a year of 0 for this date (because the general format is `yyyymmdd`, so 1231 implicitly has a year of 0); year 0 is considered a leap year, so `get_calday` actually returns a value of 366 for these crops. This is true with or without the Gregorian calendar "hack" that I just removed in c432b6d0721dfb5b05b5bd417d64c0aaad8aff84 (since the value is exactly 366.0, that hack wasn't getting triggered for this value). I'm not sure what the implications of this are, but it seems possible that planting gets skipped entirely in some circumstances on non-leap years.

2. The calculation of days past planting:
    https://github.com/ESCOMP/CTSM/blob/57175712e10e443f37818c81b8e14c956f28de34/src/biogeochem/CNPhenologyMod.F90#L2033

    uses the number of days per year in the *current* year. However, I think what's relevant here is the number of days per year in the *previous* year. (Imagine that it's now Jan 2, and planting happened on Dec. 30 of the previous year. Whether this year is a leap year is irrelevant; what matters for doing this day subtraction is whether *last* year was a leap year, and so whether Dec 30 was day of year 364 or 365 last year.) I haven't followed this through the code to see the impact of this issue, though my gut feeling is that it's probably relatively small.

samsrabin:
### Description of changes

As @billsacks documented in #1595, there were two crop phenology bugs that could occur during leap years. This PR:
- Changes the calculation of days past planting to use the number of days in the _previous_ year (fixing [2] in that issue).
- Makes it so that the Julian dates (1-366) of the sowing windows are re-calculated at the beginning of each year (fixing [1] in that issue).

My latter solution is a bit more complicated than Bill's suggested solution of just adding 10000 to the read-in MMDD number. However, it has the (slight) advantage of always conforming to the specified date, rather than being shifted a day early in leap years.

### Specific notes

**Contributors other than yourself, if any:** @billsacks 

**CTSM Issues Fixed:**
- Resolves #1595 

**Are answers expected to change (and if so in what way)?** Yes, in leap years.

Unit tests of the behaviors in question failed initially and now pass.

`aux_clm` unexpectedly showed no diffs. This turned out to be because the only aux_clm test with a Gregorian calendar is `SMS_Ly5_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianRs.izumi_gnu.clm-gregorian_cropMonthOutput`, which only compares against the output from the `...h2.1854-12-31-00000.nc` file, which isn't a leap year. We thus should probably add a test with a Gregorian calendar that tests outputs from a leap year.

## Remaining tasks
- [ ] ? Add a test that compares outputs from a leap year.

### DIFF
--- a/src/biogeochem/CNPhenologyMod.F90
+++ b/src/biogeochem/CNPhenologyMod.F90
@@ -2502,6 +2502,29 @@ contains
   end function PlantDate_to_PlantJday
 
 
+  subroutine UpdateSowingWindows()
+    !
+    ! !DESCRIPTION:
+    ! Updates sowing windows with dates for this year.
+    !
+    ! !USES:
+    use pftconMod, only: npcropmin, npcropmax
+    !
+    ! !LOCAL VARIABLES
+    integer :: n
+
+    do n = npcropmin, npcropmax
+      if (pftcon%is_pft_known_to_model(n)) then
+         minplantjday(n, inNH) = PlantDate_to_PlantJday(pftcon%mnNHplantdate(n))
+         maxplantjday(n, inNH) = PlantDate_to_PlantJday(pftcon%mxNHplantdate(n))
+
+         minplantjday(n, inSH) = PlantDate_to_PlantJday(pftcon%mnSHplantdate(n))
+         maxplantjday(n, inSH) = PlantDate_to_PlantJday(pftcon%mxSHplantdate(n))
+      end if
+   end do
+  end subroutine UpdateSowingWindows
+
+
   !-----------------------------------------------------------------------
   subroutine CropPhenologyInit(bounds)
     !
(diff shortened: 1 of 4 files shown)


src/biogeochem/CNPhenologyMod.F90
```diff
@@ -2742,7 +2743,7 @@ function DaysPastPlanting(idop)
     else
        ! get_curr_days_per_year() or get_prev_days_per_year() would only differ in the last timestep
        ! of the year, but in that case this line is not reached.
-       DaysPastPlanting = jday - idop + get_curr_days_per_year()
+       DaysPastPlanting = jday - idop + get_curr_days_per_year(offset = -365*int(secspday))
```
samsrabin, line 2790: If called on Dec. 31 of a leap year, does this fail to look at last year?

ekluzek:
@samsrabin is this something that needs to come in with the cesm3_0_beta08 answer changes?

ekluzek:
Since Gregorian calendar isn't the default used, this doesn't absolutely need to come in for the release.

samsrabin:
Agreed; changing to "Eventually."

Issue #2446, opened 2024-03-29
A few improvements to gen_mksurfdata_jobscript_single/multi


ekluzek:
There are a few improvements to the gen_mksurfdata_jobscript_single/gen_mksurfdata_jobscript_mult scripts that would be helpful either for clarity, readability, or code cleanup.

This is in addition to #2439 and mostly are simple.

This mostly came up from review with @samrabin

https://github.com/ESCOMP/CTSM/pull/2427/files#r1543744893
https://github.com/ESCOMP/CTSM/pull/2427/files#r1543712951

- [ ] To make it more clear that the output shell scripts are generated have them end with .sh (then it's more clear what's what)
- [ ] Going with that change the name of the gen_mksurfdata_build.sh to gen_mksurfdata_build
- [ ] Improve names of some of the subroutines
- [ ] Improve names of the failure tests

PR #2464, opened 2024-04-10
Update ctsm5.2 branch to ctsm5.1.dev176 and some small updates

### LINKED ISSUE #2149: Change handling of LND_TUNING_MODE so user is warned of which option is used, add more supported options
LND_TUNING_MODE was added as an XML variable that helps us handle parameters that need to be tuned for different major physics versions of the model. When we added it in, the hope was that we would only add options that we have tuned the important parameters for that configuration. So options that weren't tuned wouldn't be available. This would then require the user to explicitly choose a different option that was tuned. And the idea is that we'd only add new options when we have it fully tuned.

We tuned clm4_5, clm5_0 with GSWP3, CRUv7, and CAM6.0. So there are tuning options for those setups. And we did add a clm5_1_GSWP3 option which of course is not tuned, so many of the parameters just used the clm5_0_GSWP3 version. That already goes against the idea of only adding fully tuned options. In #2117 I need to add LND_TUNING_MODE options for CAM4.0, CAM5.0 and CAM6.0 that's needed for dust, so the other settings have to come from the clm5_0_GSWP3 settings. So some of the options are now a mix of things that are truly tuned and that aren't. That's probably OK, but we do want it to be clearly documented. And it would be good if the user was aware of this. The question is how to do this?

The LND_TUNING_MODE is set by default according to the compset name. And it matches the CLM physics version with the ATM forcing. The clm physics version HAS to match, but the ATM forcing isn't required to, so that you can try different tuning options for the same clm physics.

One question I have is how untuned compset options should be handled? For example, clm5_1_CRUv7?

The way I propose is that when options are chosen that aren't fully tuned, there would be some documentation on it. And we could setup the buildnml to print a warning, for those options. I think that ALL of the LND_TUNING_MODE combinations of physics with forcing be provided, but the ones that aren't tuned give a warning, and say what they are based on.

What are other's reaction to this idea?

Definition of done:
- [x] Add all possible combinations for physics and Met forcing to config_components.xml (clm4_5, clm5_0, clm5_1 physics, and GSWP3, CRUv7, QIA, ERA5, CAM4, CAM5, CAM6, 1PT, and NLDAS2 forcing)
- [x] Have a reduced set in namelist_definition, and buildnml that points to the right one with a warning
- [x] Warn the user in buildnml if a tuning set is not fully tuned and based on another one.
- [x] Add the CAM4, CAM5 forcing options
- [x] Add warnings to buildnml that CAM4, CAM5 are based on the CAM6 tunings

ekluzek:
### Description of changes

Update the ctsm5.2 branch to ctsm5.1.dev176 and bring in some small updates. Copy initdata_map datasets to initdata_esmf/ctsm5.2. Copy ctsm51 parameter file to ctsm60. Change output jobscripts to include a .sh extension.
Use default PE layouts rather than assigned ones for ctsm_sci tests for odd grids.

### Specific notes

Contributors other than yourself, if any: @slevis-lmwg

 Fixes #2149
 Deals with a few things in  #2446

  New ctsm5.2 finidat files moved to the lnd/clm2/initdata_esmf/ctsm5.2 diretory
  ctsm6.0 parameter file copied from the ctsm5.1 one
  ctsm5.1.dev175 to ctsm5.1.dev176 update brings in dust-emiss datasets and LND_TUNING_MODE updates

Testing performed, if any:
  python Derecho PASS
  build_namelist tester PASS
  aux_clm -- Derecho  OK
  aux_clm -- Izumi OK
  ctsm_sci -- Derecho OK
  mosart -- Derecho OK
  rtm -- Derecho PASS


doc/ChangeLog
```diff
@@ -26,9 +26,9 @@ on coastal areas was also done.
 The following fields were added and removed to the list of fields on the datasets.
 
 Fields added:
-   ORGC, BULK, CFRAG, PHAQ (soil data)
+   ORGC, BULK, CFRAG, PHAQ (soil data) (currently NOT used by CTSM)
    mapunits (map units for the soil dataset)
```
slevis-lmwg, line 30:
```suggestion
   mapunits (map units from the soil dataset)
```

slevis-lmwg (commented):
I suggested a minor clarification.

ekluzek:
I messed up something in the merge of the namelist defaults file. So I'm going through and figuring out what I did wrong.

ekluzek:
Testing for the following test lists are all as expected:

- python Derecho
- build_namelist tester
- aux_clm -- Derecho
- aux_clm -- Izumi
- ctsm_sci -- Derecho
- mosart -- Derecho
- rtm -- Derecho

On izumi I had a few tests that fail because of an issue I'm having with conda:

ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.izumi_nag.clm-FatesColdLandUse
SMS_D_Ld5.f10_f10_mg37.I1850Clm50Bgc.izumi_nag.mosart-default

Note, also some NEON tests failed for me because of long filenames, because my testname is long.

Other's aren't having these problems though, so I'm moving this forward with those two small glitches.

PR #2465, opened 2024-04-11
Fix/excess ice cold start

### LINKED ISSUE #2373: SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_intel.clm-FatesColdHydro fails when t_soisno is initialized to 272K instead of 274K
### Brief summary of bug

When t_soisno is initialized to 272K instead of 274K per PR #2355 , the test SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_intel.clm-FatesColdHydro fails, as described in PR #2371.

### General bug information

The test fails at line 771 in FatesHydroWTFMod.F90:
`psi = this%psi_sat*(th/this%th_sat)**(-this%beta)`

where apparently th is zero.  

th is passed in as h2osoi_liqvol_shell at line 578 in FatesPlantHydraulicsMod.F90:

`cohort_hydr%psi_aroot(j) = csite_hydr%wrf_soil(j)%p%psi_from_th(csite_hydr%h2osoi_liqvol_shell(j,1))`

where h2osoi_liqvol_shell is zero presumably because the soil is frozen.

**CTSM version you are using:** ctsm5.1.dev167-1-ge4dde90ef

**Does this bug cause significantly incorrect results in the model's science?** Model doesn't run

**Configurations affected:** Presumably any FATES hydro cold start would fail

### Details of bug

This can be fixed by using a use_fates in TemperatureType.F90 such that the soil temperature is initialized to 272K if use_fates=.false., and 274K if use_fates=.true., but ideally it seems like fates should be able to cold start from an initial temperature of 272K.  Particularly since if excess ice is on, the soil temperature is currently initialized to 5K below freezing, so presumably that configuration would fail as well.

Traceback is:

```
forrtl: error (73): floating divide by zero
cesm.exe           00000000047FC6DB  Unknown               Unknown  Unknown
libpthread-2.17.s  00002B74A7C0D630  Unknown               Unknown  Unknown
cesm.exe           00000000047CE001  Unknown               Unknown  Unknown
cesm.exe           0000000003D270A5  fateshydrowtfmod_         771  FatesHydroWTFMod.F90
cesm.exe           0000000001DE07E5  fatesplanthydraul         578  FatesPlantHydraulicsMod.F90
cesm.exe           000000000178CF34  edcohortdynamicsm         264  EDCohortDynamicsMod.F90
```

### LINKED ISSUE #2384: Cold start soil temperature initialization when excess ice is on.
I have added a use_fates that encompasses the code in question as suggested by @wwieder in issue #2373 .  The test now passes. However, a couple of questions.
1. The setting of t_soisno to 5deg below the freezing temperature when excess_ice is on is outside of the use_fates check and so if the test is changed so that excess_ice is on ( the current default is off), this test will fail (confirmed).  I'm not sure what the future plans for excess_ice are and if it is intended for use with FATES.
2.  There is an indication from the excess_ice PR (https://github.com/ESCOMP/CTSM/pull/1787) that it might be ok to set the initial soil temperature to 272K instead of tfrz-5.  This would simplify the code, particularly if we want to run FATES with excess_ice on.  "src/biogeophys/TemperatureType.F90 -- initial soil temperature set to 268.15 K at the cold start (might be redundant if https://github.com/ESCOMP/CTSM/issues/1460 is added)"

_Originally posted by @olyson in https://github.com/ESCOMP/CTSM/issues/2355#issuecomment-1954837943_

mvdebolskiy:
### Description of changes

Changed the way soil temperature is initialized when excess ice is on and the model starts from cold.

### Specific notes

Added 2 parameters: `excess_ice_coldstart_depth` and `excess_ice_coldstart_temp` which control top depth and soil temperature for soil layers in columns where excess ice is present. Other columns get their default soil temperature.

`excessicestream_type` has been taken out of `waterstate_type` and its routines are called directly in `clm_inst%Init`. Checks for `UseExcessIceStreams()` are still in place in `WaterStateType.F90` for double-checking.

 Fixes #2384 
 Fixes #2373 

Yes. Anything that depends on deep soil temperature will be different.

Two namelist variables are added to `clm_inparm` group: `excess_ice_coldstart_depth` and `excess_ice_coldstart_temp` . Their default values are 0.5m and -2.15 deg C. 

Testing performed, if any: Ran aux_clm on Derecho some tests on izumi


bld/CLMBuildNamelist.pm
```diff
@@ -4564,14 +4564,18 @@ sub setup_logic_exice {
   my $use_exice = $nl->get_value( 'use_excess_ice' );
   my $use_exice_streams = $nl->get_value( 'use_excess_ice_streams' );
   # IF excess ice streams is on
+  if ( (not defined($use_exice_streams) && value_is_true($use_exice)) && ($nl_flags->{'clm_start_type'} == /cold/  || $nl_flags->{'clm_start_type'} == /arb_ic/ ) ) {
```
mvdebolskiy, line 4567: Not quite sure these are sufficient. Might need to use `finidat == ''` in the condition.

mvdebolskiy:
Also #2444 might be related

mvdebolskiy:
Here are results from I2000Clm60BgcCrop simulations from coldstart:
Given that the forcing gets stuck at 2015 for the rest of the run, it seems reasonable, though the initial temperatures can be lowered a little.

exice2000
tsoi2000
exice2100
tsoi2100

mvdebolskiy:
- [ ] add readNamelist subroutine into TemperatureMod and get rid of global clm_varctl declarations.
 - [ ] Make a test where finidat has no excess ice but use_excess_ice is true and use_excess_ice_streams = false.
 - [ ] Add endrun to WaterState, so the model stops if there is excess ice on the restart file but the use_excess_ice is set to false.

src/biogeophys/TemperatureType.F90
```diff
@@ -141,7 +141,7 @@ module TemperatureType
   !------------------------------------------------------------------------
   subroutine Init(this, bounds, &
```
ekluzek, line 142: As we want to read the namelist here, you'll need to add NLFilename to the arguments, here and in the call.

src/biogeophys/TemperatureType.F90
```diff
@@ -156,6 +156,7 @@ subroutine Init(this, bounds, &
     real(r8)          , intent(in) :: em_perroad_lun(bounds%begl:)
     logical           , intent(in) :: is_simple_buildtemp  ! Simple building temp is being used
     logical           , intent(in) :: is_prog_buildtemp    ! Prognostic building temp is being used
+    real(r8)          , intent(in) :: exice_init_stream_col(bounds%begc:)  ! initial excess ice concentration from the stream file
 
     call this%InitAllocate ( bounds )
```
ekluzek, line 161: Add a call to a new ReadNML subroutine

``` fortran
    call this%ReadNML( NLFilename )
```

src/biogeophys/WaterType.F90
```diff
@@ -240,14 +241,14 @@ subroutine Init(this, bounds, NLFilename, &
...
-         NLFilename = NLFilename)
+         NLFilename = NLFilename, exice_init_stream_col = exice_init_stream_col)
 
   end subroutine Init
 
   !-----------------------------------------------------------------------
   subroutine InitForTesting(this, bounds, params, &
        h2osno_col, snow_depth_col, watsat_col, &
-       t_soisno_col, use_aquifer_layer, NLFilename)
+       t_soisno_col, use_aquifer_layer, NLFilename, exice_init_stream_col)
```
ekluzek, line 251: This change, likely means the PF unit tests will need to change. But, I wonder if it's really needed there...

src/main/controlMod.F90
```diff
@@ -256,7 +256,8 @@ subroutine control_init(dtime)
 
     namelist /clm_inparm/ use_soil_moisture_streams
 
-    namelist /clm_inparm/ use_excess_ice
+    ! excess ice flag and parameters
+    namelist /clm_inparm/ use_excess_ice , excess_ice_coldstart_depth, excess_ice_coldstart_temp
```
ekluzek, line 260: As above more to TemperatureType.F90, as well as the changes below.

src/biogeophys/TemperatureType.F90
```diff
@@ -156,6 +156,7 @@ subroutine Init(this, bounds, &
     real(r8)          , intent(in) :: em_perroad_lun(bounds%begl:)
     logical           , intent(in) :: is_simple_buildtemp  ! Simple building temp is being used
     logical           , intent(in) :: is_prog_buildtemp    ! Prognostic building temp is being used
+    real(r8)          , intent(in) :: exice_init_stream_col(bounds%begc:)  ! initial excess ice concentration from the stream file
```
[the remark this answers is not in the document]
ekluzek: The one thing I wonder about though is the name -- I don't get a nice sense of what quantity it is by the name. Maybe include concentration (or conc) in the name? 

I'd also suggest removing "stream" from the name as it will still be set to something without streams.

ekluzek (changes requested):
Thanks for going over the code with me @mvdebolskiy. Good to see this come in so that we can turn excess-ice on by default for clm6_0 physics. And to have default finidat datasets that have spunup excess-ice in them.

I ask for some changes, some of which we already talked about. And some that probably just appear here:
- [ ] The changes you already talked about...
- [ ] Figure out a name for the new namelist that will be in TemperatureType
- [ ] Add a ReadNML subroutine to TemperatureType, with the new namelist variables in it.
- [ ] Make sure PF unit tests work (do what the README file in the "src" directory says on Derecho
- [x] Check the values of the new namelist items to make sure they are valid
- [ ] Think about changing the name of exice_init_stream_col
- [ ] There's a couple formatting and documentation suggestions I have

These should all be reasonable. Let me know if you'd like to meet to go over how to add the new namelist in. But, there are lots of examples so that should be straightforward to do.

ekluzek:
@mvdebolskiy and I talked about the procedure to bring this in. He's not up to making tags himself at this point. So I'll work with him on it. Right now he's going to run some excess ice tests on Derecho to make sure it's working correctly. Then I'll run aux_clm tests for Derecho and Izumi and point him to issues to fix.

This does change answers. But, we actually could make it bit-for-bit by setting the new namelist values appropriately. So that should be something we think about.

I do want to teach him to run aux_clm and create baselines on Derecho. But, we'll take that one step at a time...

mvdebolskiy:
> * [ ]  Figure out a name for the new namelist that will be in TemperatureType

    Added  `clm_temperature_inparm` group (for future handling of other coldstart temperature vars)
> * [ ]  Add a ReadNML subroutine to TemperatureType, with the new namelist variables in it.

 Added. Made NLFilename to be optional for unittests to be happy.
> * [ ]  Make sure PF unit tests work (do what the README file in the "src" directory says on Derecho
 Ran `../cime/scripts/fortran_unit_testing/run_tests.py --build-dir unit_tests.temp` : `100% tests passed, 0 tests failed out of 52`

> * [ ]  Check the values of the new namelist items to make sure they are valid

Checked both in `CLMBuildNamelist.pm` and in `TemperatureType%InitCold`
> * [ ]  Think about changing the name of exice_init_stream_col

Changed to `excie_init_conc_col`


Added a warning (instead of endrun) when starting with `use_excess_ice=.false.` from initial conditions that have it.
Note, simple `flag=read` would not work, since there is a call with `write` and `define` before (for init_interp) purposes, though the `RestartExcessIceIssue` helped.

New baseline on izumi: `/home/mvdebolskiy/bl/fix-excess-ice/ERS_D.f10_f10_mg37.I1850Clm60Sp.izumi_nag.clm-ExcessIceStreams`

src/biogeophys/WaterType.F90
ekluzek: Just tested and they are working, so this is good.

src/biogeophys/TemperatureType.F90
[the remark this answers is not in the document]
ekluzek: Changed to exice_init_conc_col

bld/CLMBuildNamelist.pm
ekluzek: Logic was updated to use finidat, which is likely more robust.

ekluzek: @mvdebolskiy The one thing here is that it looks like potentially the user might set use_excess_ice_streams one way and the logic here might end up overriding what they did. It's better to only set it if the user didn't touch it, and abort with an error if the user tried to do something nonsensical.

bld/CLMBuildNamelist.pm
```diff
@@ -4581,15 +4581,25 @@ sub setup_logic_exice {
...
-  # IF excess ice streams is on
+  my $finidat = $nl->get_value('finidat');
+  # If coldstart and use_excess_ice is on:
+  if ( ( (not defined($use_exice_streams)) && value_is_true($use_exice) ) && string_is_undef_or_empty($finidat) ) {
+   $nl->set_variable_value('exice_streams', 'use_excess_ice_streams' , '.true.');
+   $use_exice_streams = '.true.';
+   # if excess ice is turned off
+   } elsif ( (not defined($use_exice_streams)) && (not value_is_true($use_exice)) ) {
+   $nl->set_variable_value('exice_streams', 'use_excess_ice_streams' , '.false.');
+   $use_exice_streams = '.false.';
```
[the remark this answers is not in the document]
ekluzek: Brought in.

ekluzek (approved):

ekluzek:
@mvdebolskiy I'm planning on doing the following things to finish out the tag. Just FYI, unless you want to comment or contribute:

- [ ] Add some build-namelist unit tests for the new excess ice error conditions
- [ ] Make sure the build-namelist changes are solid, if not and easy go ahead and fix, otherwise file an issue
- [ ] Make sure we have the needed tests on Derecho for excess ice, if not add one
- [ ] Update to ctsm5.2.017
- [ ] Close all the conversations
- [ ] Take a stab at creating the ChangeLog for ctsm5.2.018 (have @mvdebolskiy review and vet)
- [ ] Run baselines for Derecho and Izumi for ctsm5.2.018
- [ ] Finalize the tag as normal

If any problems crop up, I'll consult with you and get a fix in and/or file an issue to fix later.

bld/CLMBuildNamelist.pm
ekluzek: OK, I adjusted a few things and think it's good now.

ekluzek: @mvdebolskiy I adjusted the namelist a bit, feel free to look it over and give feedback. And if you'd like to meet we can do that.

ekluzek: Resolving this for now so we'll be able to merge when needed.

doc/ChangeLog
```diff
@@ -1,4 +1,108 @@
 ===============================================================
+Tag name: ctsm5.2.018
+Originator(s): mvdebolskiy
+Date: Wed 31 Jul 2024 04:38:36 PM MDT
```
ekluzek, line 4: @mvdebolskiy this is a start at the ChangeLog please review and lets make any important changes needed.

ekluzek:
Baselines look as expected on Derecho and Izumi. Izumi shows differences for ERS_D.f10_f10_mg37.I1850Clm60Sp.izumi_nag.clm-ExcessIceStreams that are significant, and this is without a cold-start. Which we think shouldn't show differences. So we are investigating. 

I'll also run SMS_Lm12.f09_f09_mg17.I1850Clm60Sp.derecho_intel.clm-ExcessIceStartup_output_sp_exice on Derecho, creating a baseline for ctsm5.2.017 to check what happens there.

PR #2468, opened 2024-04-12
Dust namelist moved to cmeps

### LINKED ISSUE #2150: Add capability to flip between different dust emission parameterizations?
I introduce this idea here:

https://github.com/ESCOMP/CTSM/issues/1604#issuecomment-1712923802

I think it would be worthwhile to add the capability to flip between different dust emission parameterizations, for the reasons stated above. This would be done after #1836 is complete. And before #1897 comes in.

This would be done similarly as other OO options we have for code options is done. So a Factory that reads in a namelist to determine which one, and instantiate it. And then an abstract base class, and then specific versions based off the base class.
This doesn't actually add that much code. The abstract class and factory for Fire for example is only about 400 lines. With most of that being boilerplate. So I think adding this would not be a big development.

### LINKED ISSUE #2170: Update the drv_flds namelist definition file in CTSM to work with latest CAM/cmeps
### Brief summary of bug

I see this as a problem in trying SMS.f09_g17.2000_CAM%DEV_CLM51%SP_SICE_SOCN_MOSART_SGLC_SWAV.cheyenne_intel.clm-default
with ctsm5.1.dev125 (technically a ctsm5.2 branch of it, but shouldn't be relevant here).

### General bug information

**CTSM version you are using:** ctsm5.1.dev125-384-gaa3c5cc8a

**Configurations affected:** Some configurations with CAM

### Details of bug

I'm not really sure why this isn't a problem all of the time. As cases with CAM obviously work. So why is this an issue in this case?

This is the error I see when running preview_namelists for the above test. It does fail at this point and doesn't populate CaseDocs with namelists...

```
Create namelist for component drv
   Calling /glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/components/cmeps/cime_config/buildnml
Writing nuopc_runconfig for components ['CPL', 'ATM', 'LND', 'ROF']
ERROR: Variable 'atm_provides_lightning' from file '/glade/work/erik/ctsm_worktrees/ctsm5.2.mksurfdata/cime/scripts/SMS.f09_g17.2000_CAM%DEV_CLM51%SP_SICE_SOCN_MOSART_SGLC_SWAV.cheyenne_intel.clm-default.C.20230925_160609_ag813w/Buildconf/camconf/drv_flds_in' is not in the namelist definition.
```

### LINKED ISSUE #2376: Update CTSM with CMEPS update for controlling Dust emissions
This updates CTSM to use the namelist control in CMEPS (in https://github.com/ESCOMP/CMEPS/pull/429). So the CMEPS external needs to be updated, and the namelist control in CTSM changed to use CMEPS rather than the internal CTSM control settings and the CTSM ones removed.

### LINKED ISSUE #2524: Add option to NOT set the dust-emission drv_flds_in settings when coupled to CAM
As part of the work on dust emissions, we need a way to choose if either CAM or CTSM is going to set the CMEPS dust_emission namelist. The current logic for this is that if CAM and CTSM disagree on the fields in drv_fld_in -- CESM aborts with an error. This happens inside of CMEPS buildnml.

In the case of fire-emissions, MEGAN, or drydep these are all optional so we just don't set them when coupled to CAM. So there is no disagreement with CAM. In this case though dust_emissions -- must be set. So when coupled to CAM, CTSM needs to know not to set it, which means default values should NOT be set for this case.

ekluzek:
### Description of changes

Remove the dust emissions namelist items from CTSM and use the namelist in the drv_flds_in for CMEPS. Also
update the CMEPS version to handle this.

This updates CTSM to use the namelist control in CMEPS (in https://github.com/ESCOMP/CMEPS/pull/429). So the CMEPS external needs to be updated, and the namelist control in CTSM changed to use CMEPS rather than the internal CTSM control settings and the CTSM ones removed.

Add a unit test for the CMEPS code to make sure it's working correctly.

- [x] unit test
- [x] Add drv_flds_in namelist definition in CMEPS
- [ ] Finalize CMEPS tag point to it here
- [x] Update drv_flds_in namelist definition in CTSM
- [x] Add capability to set the dust emissions settings in drv_flds_in
- [ ] Update build-namelist tester to get the new capability working
- [x] Remove the old dust namelist items in CTSM
- [ ] Run standard testing and make sure everything is working

### Specific notes

  Fixes #2376
  Fixes #2150
  Fixes #2170
  Fixes #2524

  Namelist items removed from CTSM and into the drv_flds_in namelist in CMEPS

Testing performed, if any: will run standard testing, have run a few singleton tests so far

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -48,9 +48,12 @@ tag = cime6.0.217_httpsbranch03
 required = True
 
 [cmeps]
-tag = cmeps0.14.50
+#tag = cmeps0.14.50
+#branch = dust_emis_mod
+hash = d79e34f67df825f9265bdaaf3091f8995d5df1a2
 protocol = git
-repo_url = https://github.com/ESCOMP/CMEPS.git
+#repo_url = https://github.com/ESCOMP/CMEPS.git
+repo_url = https://github.com/ekluzek/CMEPS.git
 local_path = components/cmeps
 required = True
 
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -50,7 +50,7 @@ required = True
 [cmeps]
 #tag = cmeps0.14.50
 #branch = dust_emis_mod
-hash = d79e34f67df825f9265bdaaf3091f8995d5df1a2
+hash = e22c4eafd595964d88a1077961a809a60b794431
 protocol = git
 #repo_url = https://github.com/ESCOMP/CMEPS.git
 repo_url = https://github.com/ekluzek/CMEPS.git


ekluzek:
Closing because this was the wrong branch. #2538 is where this work is being done.

PR #2485, opened 2024-04-23
Plumber.5.2

### LINKED ISSUE #2484: PLUMBER2 surface data for CTSM5.2
### Brief summary of bug

PLUMBER2 usermods in #2482 were set up for 5.1 surface data

### General bug information

ctsm5.2.001

**Does this bug cause significantly incorrect results in the model's science?** No, PLUMBER2 sites are a new feature, WIP that are not being tested

**Configurations affected:** PLUMBER2 single point cases

**Error in log file states 
```
using /glade/u/home/wwieder/CTSM/tools/site_and_regional/subset_data_single_point/surfdata_1x1_PLUMBER2_ZM-Mon_hist_16pfts_Irrig_CMIP6_simyr2000_c231005.nc
 Opened existing file
 /glade/u/home/wwieder/CTSM/tools/site_and_regional/subset_data_single_point/sur
 fdata_1x1_PLUMBER2_ZM-Mon_hist_16pfts_Irrig_CMIP6_simyr2000_c231005.nc
         155
 check_var: variable xc is not on dataset
 surfrd_get_data lon_var = LONGXY lat_var =LATIXY
 ncd_inqvid: variable PCT_OCEAN is not on dataset
 ENDRUN:
 ERROR:
  ERROR: PCT_OCEAN NOT on surfdata file but required when running ctsm5.2 or new
 er;  you are advised to generate a new surfdata file using the mksurfdata_esmf
 tool ERROR in surfrdMod.F90 at line 421
```

wwieder:
### Description of changes
Fixes usermods to point to 5.2 datasets

### Specific notes
Still kind of hacky, ideally we can take this out of the custom shell commands for individual sites.
I think this can happen after @TeaganKing adds PLUMBER2SITE to cdeps?

Fixes #2484

updates for usermods

Testing performed, if any:
Manual test to make sure sites run as expected with 5.2 datasets


wwieder:
@ekluzek I wasn't clear if you thought this should go on b4b-dev or master.  If we merge it with other changes from @TeaganKing it may be easier?  I just wanted to create this branch and PR so we can keep making progress on PLUMBER2 work.

ekluzek:
I think this should go to b4b-dev, and I just rebased it to reflect that. Technically right now both are identical.

TeaganKing:
[This is the WIP PR](https://github.com/ESCOMP/CTSM/pull/2406/files) where some of these changes are being made. Maybe I can just make a separate PR for the `config_component.xml` and `cime_config` details in order to contribute that to this PR? I think the rest of the linked PR will take a decent amount of additional effort to finalize.

wwieder:
That could be nice to bring in the cime_config and config_component work here, as it will further clean up the usermods_dirs.  I'll let you manage what makes the most sense, @TeaganKing.  I mainly wanted to get something that you could start working with for the larger run_towers refactor you're doing.

TeaganKing:
I actually realized those changes I mentioned should all be under CDEPS instead of CTSM anyways. I'll update the [CDEPS PR](https://github.com/ESCOMP/CDEPS/pull/262) I made though.

TeaganKing:
I ran black on `cime_config/buildnml` and got a clean run with the second commit, so I'm not sure why that's still failing on black? Have you seen that before?

Screen Shot 2024-04-23 at 2 27 55 PM

TeaganKing:
In the `ctsm_pylib` environment, I ran `black --version` and saw: `black, 22.3.0 (compiled: no)`.  Black that's being run on GitHub is `black[colorama]==22.3.0`. I think this is the same version.

`python/pyproject.toml` contains the same information that is in my local copy-- could this be overwritten somewhere during the black check such that black fails on GitHub? This includes items such as line-length and target-version.

This seems like it should be a very simple fix...

TeaganKing:
I still haven't resolved the black issue, but need to head out soon so am leaving this unresolved at the moment. Where I'm at right now is that I found the reverting the following line to the original passes black (and does not include PLUMBER functionality that we need):
`if ("NEON" in clm_usrdat_name or "PLUMBER" in clm_usrdat_name) and clm_force_coldstart == "off":`

If I run black locally, it makes a whole lot of other changes that apparently aren't needed to pass the black check on push.

Adding a 'format ignore' (see commit titled 'trying a workaround' does not seem to resolve the issue, either.

TeaganKing:
Issues with black-check on push are finally resolved! I think that should be all of the changes for the `cime_config` and `config_component` part of this work.

TeaganKing:
And moving forward, is the plan to pull in the [CDEPS PR](https://github.com/ESCOMP/CDEPS/pull/262) and then bring this on to `b4b`? Or does this PR need additional work, @wwieder ?

slevis-lmwg (commented):
I understand this PR as waiting for https://github.com/ESCOMP/CDEPS/pull/262 (correct me if I'm wrong).

tools/site_and_regional/plumber2_usermods.py
```diff
@@ -49,7 +49,7 @@ def write_usermods(lat,lon,site,start_year,end_year,
         '\n' \
         # TODO, get working for CTSM5.1, remove this line as it's redundant after PLUMBER2SITE is added
         # Alternatively, we can take this out of default/user_nl_clm since doing it this way is works fine TODO for 5.2
-        'echo "fsurdat=\'/glade/u/home/wwieder/CTSM/tools/site_and_regional/subset_data_single_point/surfdata_1x1_PLUMBER2_'+site+'_hist_16pfts_Irrig_CMIP6_simyr2000_c231005.nc \' " >> user_nl_clm \n' \
+        'echo "fsurdat=\'\$DIN_LOC_ROOT/lnd/clm2/surfdata_esmf/PLUMBER2/surfdata_1x1_PLUMBER2_'+site+'_hist_2000_16pfts_c240326.nc\'  " >> user_nl_clm \n' \
```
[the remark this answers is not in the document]
wwieder: We can likely remove this all from plumber2_usermods.py script, if we add PLUMBER2SITE to cdeps this can be handled in user_nl_clm in the default usermods_dirs.

cime_config/buildnml
```diff
@@ -203,15 +203,16 @@ def buildnml(case, caseroot, compname):
...
+        if ("NEON" in clm_usrdat_name) or ("PLUMBER" in clm_usrdat_name):
+            if clm_force_coldstart == "off":
+                if ("_transient" in clm_nml_use_case) and (
+                    re.fullmatch(r"\w\w\w\w\.transient", casename) is None
+                    or clm_usrdat_name is "NEON.PRISM"
+                ):
+                    logger.warning(
+                        "WARNING: Do you have appropriate initial conditions for this simulation?"
+                        + " Check that the finidat file used in the lnd_in namelist is appropriately spunup for your case"
+                    )
```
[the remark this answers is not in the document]
slevis-lmwg: From looking at a different PR, I see that CLM_NML_USE_CASE has the "_transient" suffix for various compsets (Hist and SSP), so I'm retracting my comment as unnecessary.

TeaganKing:
@wwieder I'm thinking this PR should also include a PLUMBER `defaults` usermod directory that also includes the xmlchange for `CLM_USRDAT_NAME` to be set to `PLUMBER` for all the plumber cases?

cime_config/buildnml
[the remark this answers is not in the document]
ekluzek: Just to add to the conversation, but also point out something I'm concerned about. CLM_NML_USE_CASE is better to use than the compset long-name because you might change it after you create the case. And there is a standard for having transient use-cases end in _transient, and build-namelist checks for it -- so it is the safe thing to do.

**_The thing I'm concerned about:_**
But, this also had me look at the reg-ex above, and it looks wrong to me. It's checking for 4 word characters a "." and then the transient ending. And that looks to me like it wouldn't match: 2018-PD_transient. And it should probably be setup to just check for the _transient ending rather than add additional requirements. So does this actually work in practice?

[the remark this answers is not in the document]
TeaganKing: Hmm. The warning `WARNING: Do you have appropriate initial conditions for this simulation?` certainly shows up regularly in running transient cases, which means that this does get through the if statements. I think the NEON cases specifically are set up to be named `<4-letter-neonsite-name>.transient`-- so maybe this wouldn't work for non-neon cases, but works for neon cases? This might need to change for PLUMBER case names (eg, `AR-SLu`), though?

wwieder:
> @wwieder I'm thinking this PR should also include a PLUMBER `defaults` usermod directory that also includes the xmlchange for `CLM_USRDAT_NAME` to be set to `PLUMBER` for all the plumber cases?

I think this is marked with a TODO in #2485?  But maybe a quick conversation would be helpful?  It's been a while since I thought about this.

TeaganKing:
> > @wwieder I'm thinking this PR should also include a PLUMBER `defaults` usermod directory that also includes the xmlchange for `CLM_USRDAT_NAME` to be set to `PLUMBER` for all the plumber cases?
> 
> I think this is marked with a TODO in #2485? But maybe a quick conversation would be helpful? It's been a while since I thought about this.

Did you intend to tag another issue? I'll be in the office tomorrow, and it looks like you're onsite too-- let's chat sometime then?

TeaganKing:
> Sorry yes, [this one that was already merged?](https://github.com/ESCOMP/CTSM/pull/2155)

Oh, I see it now! Thanks!

tools/site_and_regional/plumber2_usermods.py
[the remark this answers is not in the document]
TeaganKing: Updating this in the commit titled "update usermods script and shell commands for fsurdat"

TeaganKing:
I also did uncomment the following line in the last commit given that this will come in after the CDEPS PR:
`./xmlchange PLUMBER2SITE='+site + '\n'`

TeaganKing:
The merge conflicts are fixed now, too. Some variables had already been renamed in b4b-dev to conform to snake case, and the plumber2_usermods.py file had been moved to the python directory.

TeaganKing:
Ready for review with the caveat of the dependency of CDEPS PR 262 (as the label on this PR suggests). I can also leave this as not ready for review if that's preferred given the dependency.

wwieder:
Thanks @TeaganKing .  @slevis-lmwg do you have time to review this?  I'm hosting a workshop next week.

cime_config/buildnml
```diff
@@ -203,15 +203,16 @@ def buildnml(case, caseroot, compname):
...
-                or clm_usrdat_name is "NEON.PRISM"
-            ):
-                logger.warning(
-                    "WARNING: Do you have appropriate initial conditions for this simulation?"
-                    + " Check that the finidat file used in the lnd_in namelist is appropriately spunup for your case"
-                )
+        if ("NEON" in clm_usrdat_name) or ("PLUMBER" in clm_usrdat_name):
+            if clm_force_coldstart == "off":
+                if ("_transient" in clm_nml_use_case) and (
+                    re.fullmatch(r"\w\w\w\w\.transient", casename) is None
```
slevis-lmwg, line 209: @TeaganKing and I realized from this string check that PLUMBER should be in a separate if-block with a different string check for it to work correctly for PLUMBER.

@TeaganKing and I also discussed the order that these three related PRs will get merged:
https://github.com/ESCOMP/CDEPS/pull/262
This PR
#2406 

I will preemptively approve this PR.

slevis-lmwg (approved):

cime_config/buildnml
TeaganKing: Thanks Sam! I restructured that bit, but am running into some more issues with black and will get back to this hopefully in a few hours!

TeaganKing: Ok, I got back early from lunch and fixed this! This file is now passing black checks and includes an updated PLUMBER-specific `fullmatch`.

TeaganKing:
- [ ] Last dependency is 262 . Adding this unchecked task to make sure this doesn't get merged in accidentally.

TeaganKing:
When running PLUMBER with this 5.2 dataset, it looks like there is some missing boundary data for PCT_OCEAN (see below). @slevis-lmwg or @wwieder , do we need to provide or point to these files alongside the 5.2 dataset?

```
Attempting to read surface boundary data .....
(GETFIL): attempting to find local file
surfdata_1x1_PLUMBER2_AU-Ctr_hist_16pfts_Irrig_CMIP6_simyr2000_c231005.nc
(GETFIL): using /glade/u/home/wwieder/CTSM/tools/site_and_regional/subset_data_single_point/surfdata_1x1_PLUMBER2_AU-Ctr_hist_16pfts_Irrig_CMIP6_simyr2000_c231005.nc
 Opened existing file
 /glade/u/home/wwieder/CTSM/tools/site_and_regional/subset_data_single_point/sur
 fdata_1x1_PLUMBER2_AU-Ctr_hist_16pfts_Irrig_CMIP6_simyr2000_c231005.nc
         158
 check_var: variable xc is not on dataset
 surfrd_get_data lon_var = LONGXY lat_var =LATIXY
 ncd_inqvid: variable PCT_OCEAN is not on dataset
 ENDRUN:
 ERROR:
  ERROR: PCT_OCEAN NOT on surfdata file but required when running ctsm5.2 or new
 er;  you are advised to generate a new surfdata file using the mksurfdata_esmf
 tool ERROR in surfrdMod.F90 at line 421
```

ekluzek:
@TeaganKing it looks like that dataset is an older ctsm5.1 version, so you'll need to use a ctsm5.2 from:

/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/surfdata_esmf/NEON/

And yes, one of the differences is that PCT_OCEAN is now required in ctsm5.2.

TeaganKing:
Thanks @ekluzek !

cime_config/buildnml
```diff
@@ -313,13 +315,22 @@ def buildnml(case, caseroot, compname):
...
-                    run_reftod,
-                )
+                if "PLUMBER2" in clm_usrdat_name:
+                    clm_startfile = "%s.%s%s.r.%s-%s.nc" % (
+                        run_refcase,
+                        comp,
+                        inst_string,
+                        run_refdate,
+                        start_tod,
+                    )
```
TeaganKing, line 325: Should RUN_REFTOD be set to START_TOD instead of using START_TOD here?

python/ctsm/site_and_regional/plumber2_usermods.py
```diff
@@ -51,29 +63,14 @@ def write_usermods(
...
-        + site
-        + "/CLM1PT_data/CTSM_DATM_"
-        + site
-        + "_"
-        + str(start_year)
-        + "-"
-        + str(end_year)
-        + '.nc " >> user_nl_datm_streams \n'
+        "\n"
         'echo "presaero.SSP3-7.0:year_first=' + str(start_year) + '" >> user_nl_datm_streams \n'
```
TeaganKing, line 74: These echo lines should be moved into CDEPS.

cime_config/buildnml
TeaganKing: @ekluzek Do you have any suggestions on this?

python/ctsm/site_and_regional/plumber2_usermods.py
TeaganKing: After some discussion with Will, I'm planning to leave these in usermods to avoid more complicated logic statements in CDEPS.

TeaganKing:
`[CDEPS #262](https://github.com/ESCOMP/CDEPS/pull/262)` is basically done, but just needs the merge button to be clicked. I was planning on adding that CDEPS tag to #2406 ; the one question I have for you, @ekluzek , is if the CDEPS tag should also be added to this PR? Besides that, this is ready for review (#2406 will take a little more work).

cime_config/buildnml
ekluzek: In this context RUN_REFTOD is the time of day you want to use for the restart file to startup from. And START_TOD is the time of day you want the new run to start from. I can't think of a reason you'd ever want these to be different from each other. So using them interchangeably in one sense is OK, but it's maybe less clear on what's going on? In any case I don't think there is much of an issue with either way...

TeaganKing: Ok thanks for clarifying!

ekluzek:
Does this one require the CDEPS tag for it to function? It looks like it does to me -- but I could be wrong. If it does I'd include the update here. If this comes in before #2406 that just means it's already in place for it. Since, they both do the same thing it likely won't even mark it as a conflict when you update #2406.

TeaganKing:
> Does this one require the CDEPS tag for it to function? It looks like it does to me -- but I could be wrong. If it does I'd include the update here. If this comes in before #2406 that just means it's already in place for it. Since, they both do the same thing it likely won't even mark it as a conflict when you update #2406.

Ok, I will do that once the CDEPS PR is merged then!

ekluzek:
> Besides that, this just needs to have conflicts resolved (which I am getting a bit confused by, but it looks like Will's branch needs to be rebased to point to b4b instead of master, then updated, and then conflicts in buildnml resolved?) and then will be ready for review (#2406 will take a little more work).

Since, b4b-dev just came to master, it should be easy to merge or rebase to b4b-dev at this point. And yes there will be some conflicts in buildnml to resolve. Let me know if you want someone to help going over them.

TeaganKing:
Ok, thanks @ekluzek for your help! The `cdeps` tag has been updated and this is ready for review!

TeaganKing:
Thanks Will. FYI I also rebased this on Friday to bring in updates from b4b-dev that were made before then.

cime_config/buildnml
```diff
@@ -233,6 +235,14 @@ def buildnml(case, caseroot, compname):
                     "WARNING: Do you have appropriate initial conditions for this simulation?"
                     + " Check that the finidat file used in the lnd_in namelist is appropriately spunup for your case"
                 )
+        if "PLUMBER2" in clm_usrdat_name and clm_force_coldstart == "off":
```
ekluzek, line 238: I'm thinking rather than copy and paste this whole section, that the if just be changed to do the same thing for PLUMBER2 and NEON.

cime_config/buildnml
```diff
@@ -340,13 +350,22 @@ def buildnml(case, caseroot, compname):
...
+                        run_refdate,
+                        start_tod,
+                    )
+                else:
+                    clm_startfile = "%s.%s%s.r.%s-%s.nc" % (
+                        run_refcase,
+                        comp,
+                        inst_string,
+                        run_refdate,
+                        run_reftod,
```
ekluzek, line 367: OK, the difference here is whether start_tod or run_reftod is being given. This could use a comment or a restructuring to make that obvious. Having something like

``` python
if "PLUMBER2" in clm_usrdat_name:
   tod = start_tod
else:
   tod = run_reftod

   clm_startfile = "%s.%s%s.r.%s-%s.nc" % (
                        run_refcase,
                        comp,
                        inst_string,
                        run_refdate,
                        tod,
   )
```

makes it obvious what the difference is for PLUMBER2.

cime_config/buildnml
ekluzek: Opening this back up, to just check on what both run_reftod and start_tod are going to be for PLUMBER2 cases? I take it at least start_tod is going to be non-zero -- or does that depend on the type of case? And will run_reftod be non-zero for the IC files for PLUMBER2 cases? It would be good to think this through to make sure we need this extra complexity in the code....

cime_config/config_component.xml
```diff
@@ -355,7 +356,7 @@
       cold start (finidat will be set to blanks).
       A value of on forces the model to spin up from a cold-start
       (arbitrary initial conditions). Setting this value in the xml file will take
-      precedence over any settings for finidat in the $CASEROOT/user_nl_ctsm file.</desc>
+      precedence over any settings for finidat in the $CASEROOT/user_nl_clm file.</desc>
```
ekluzek, line 359: Thanks for the fix!

ekluzek (changes requested):
This is great and a nice improvement that gets us so much further along with a clean implementation of PLUMBER that works with the newer datasets. Thanks for all the work here.

The bulk of this is straightforward work that's site specific so not something to review. There's a couple points where you catch some misspellings which is awesome.

The one thing I had was to simplify the buildnml code a bit to remove some duplication. This could also be done with an issue that's done later. But, I think it's straightforward enough to do now. We could go over this also if that's helpful.

cime_config/buildnml
TeaganKing: That is much clearer! Thanks for the suggestion.

cime_config/buildnml
TeaganKing: These do have different name formats that are used, eg `\w\w-\w\w\w\` for PLUMBER2 and `\w\w\w\w\` for NEON, so I'd slightly prefer to keep these separate.

If, with that additional bit of information, you still think this is necessary, I could do something like this (though I think it would be more confusing than helpful):
```
if ("NEON" in clm_usrdat_name) or ("PLUMBER2" in clm_usrdat_name) and clm_force_coldstart == "off":
            if "NEON" in clm_usrdat_name:
                 match = '\w\w\w\w\'
            if "PLUMBER2" in clm_usrdat_name:
                 match = '\w\w-\w\w\w\
            if ("_transient" in clm_nml_use_case) and (
                re.fullmatch(r"{}.transient".format(match), casename) is None
                or clm_usrdat_name is "NEON.PRISM"
            ):
                logger.warning(
                    "WARNING: Do you have appropriate initial conditions for this simulation?"
                    + " Check that the finidat file used in the lnd_in namelist is appropriately spunup for your case"
                )
```

TeaganKing:
@ekluzek thanks for reviewing. Let me know what you think about the above conversation; otherwise, your suggestions have been addressed!

cime_config/buildnml
TeaganKing: Oops, sorry I didn't see this one since it was still resolved! The various PLUMBER cases (ad, postad, transient) all have RUN_REFTOD set to 00000, and START_TOD set to some nonzero value (eg, 54000)-- hence the decision to use START_TOD. We could update the shell commands to change RUN_REFTOD instead of START_TOD; I'd need to more thoroughly check that RUN_REFTOD is okay to change without other side effects, though.

cime_config/buildnml
ekluzek: Ahh, I see a different match.

Because of the reduced duplication I personally prefer the code you propose. Reducing duplication is really important for code maintenance.  So I like your proposal. Are you warm enough to the proposal that you could bring it in? Maybe adding a comment to explain what's going on would help if you find it confusing?

Code readability and trying to not be confusing are important. I think the best way to do that for buildnml is to break it up into smaller subroutines and have python unit testing for them. That's obviously a longer term thing though.

cime_config/buildnml
TeaganKing: Ok, I added this to issue #2671

cime_config/buildnml
TeaganKing: Ok, I removed some of the duplication. I ended up leaving the line with match as one line rather than using the 'match' variable because the backslashes were causing some issues when it was interpreted. But, I think this is an improvement.

ekluzek (approved):
Cool. Awesome. Let's bring this in.

PR #2488, opened 2024-04-25
Add line about documentation in PR template

samsrabin:
### Description of changes

As discussed in the CTSM SE meeting today, this PR adds a line about documentation to the PR template.

### Specific notes

**Testing performed, if any:** None


ekluzek (approved):
Awesome, thank you.

Should this also be added to the ChangeLog template file (doc/.ChangeLog_template)? Although moving to having more things integrated into github is a better workflow. But, it may also be good to have it in both places...

I'd actually like to move to a workflow that manages the ChangeLog with github actions, based on a talk I saw at SEA ISS. But, that's a longer term thing...

samsrabin:
Good thinking @ekluzek! Now added; merging now.

PR #2523, opened 2024-05-03
Land Energy Balance Solve with variable stomatal conductance coupling

rgknox:
### Description of changes

This **draft** set of changes is an initial stab and breaking up the Land Energy Balance iterative solve, so that updates stomatal conductance and photosynthesis as a separate loop and with variable coupling strength.  The motivation behind this, is that a nearly similar answer can be achieved as the base method, but with fewer calls to photosynthesis.  Photosynthesis is overwhelmingly the slowest part of the process when FATES is active, and the slowest (but probably not overwhelmingly slowest) part of the process when the native vegetation model is used.

### Specific notes

The current implementation has a local flag that enables either the default or the variable control behavior.  But this could easily be tied to a namelist control (UPDATE: THIS IS BEING IMPLEMENTED).

The current implementation has been written to accommodate both FATES and native CLM vegetation structure, but was really only tested with FATES.

There are various calls to subroutines in this module that are designed to make calculations over a vector of patches, that are now only being called for a single patch. So the callinng of the subroutines would potentially be re-written if the general looping structure is something that we want...

The convergence criteria is still immature.  I started with a relative difference criteria on resistance.  But @gbonan has advocated for absolute criteria in the past and that may be a good way to go here instead.  Perhaps a sensitivity analysis on aboslute convergence criteria is in order.  Using relative criteria creates a situation where areas of high resistance generate large acceptable differences and I saw larger differences in the base in places with large resistance. When I swapped to use relative conductance criteria, the opposite was true.  The next step is to determine a reasonable absolute criteria.

Contributors other than yourself, if any:  many people provided feedback and this was presented at the LMWG, conversations with @gbonan guided the design

Any User Interface Changes (namelist or namelist defaults changes)?  Potentially changes to the namelist

Does this create a need to change or add documentation? Did you do so?  Yes, this would potentially require an update to the tech doc

Testing performed, if any:

Timing: this makes FATES quite a bit faster!  No data on making non-FATES faster

Regression testing TBD

### DIFF
--- a/src/biogeophys/CanopyFluxesMod.F90
+++ b/src/biogeophys/CanopyFluxesMod.F90
@@ -78,11 +78,9 @@ module CanopyFluxesMod
   logical,  public :: perchroot_alt = .false.  
   !
   ! !PRIVATE DATA MEMBERS:
-  logical, private :: use_undercanopy_stability = .false.      ! use undercanopy stability term or not
-  integer, private :: itmax_canopy_fluxes = -1  ! max # of iterations used in subroutine CanopyFluxes
+  logical, private :: use_undercanopy_stability = .false.         ! use undercanopy stability term or not
+  integer,dimension(2), private :: itmax_canopy_fluxes = /-1,-1/  ! max # of iterations used in subroutine CanopyFluxes
 
-  integer, parameter :: itmax_stoma_calcs = 50
-  
   character(len=*), parameter, private :: sourcefile = &
        __FILE__
   !------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rgknox:
Slides from the LMWG [here](https://docs.google.com/presentation/d/1nxCGj9oJbxL1YIP_eBXQG61t5xoAv0qMto3FVKRuLfM/edit?usp=sharing)

samsrabin:
From discussion today: If this comes in after the [pre-ctsm6 FATES SP freeze](https://github.com/ESCOMP/CTSM/milestone/38) but before the [CESM3 feature freeze](https://github.com/ESCOMP/CTSM/milestone/39), it should be default off.

glemieux:
Per CTSM SE discussion, @rgknox suggested enabling this through a namelist option if this impacts satellite phenology mode post sp mode freeze.

PR #2528, opened 2024-05-06
Fix subset_data (cth and cbh not iterable)

### LINKED ISSUE #2527: subset_data is broken
### Brief summary of bug

subset_data fails for a point when cbh and cth are not given and a dominant pft is defined.
@samsrabin suggested a python fix is pretty straight forward, but we need to also add testing around subset_data.
### General bug information

CTSM5.2.001

No

Subset data for single point cases

### Details of bug
Subset data fails if cbh and cth are not given.  This seems to have come in with the PLUMBER2 PR to b4bdev, which introduced information on canopy top and bottom heights.

```
  ./subset_data point --lat 38.9592 --lon 253.0102 --create-surface --outdir /glade/derecho/scratch/wwieder/nldas/gothic/ --site Gothic_evergreen --dompft 1
  File "./subset_data", line 37, in <module>
    main()
  File "/glade/u/home/wwieder/CTSM/tools/site_and_regional/../../python/ctsm/subset_data.py", line 713, in main
    subset_point(args, file_dict)
  File "/glade/u/home/wwieder/CTSM/tools/site_and_regional/../../python/ctsm/subset_data.py", line 604, in subset_point
    specify_fsurf_out=file_dict["fsurf_out"],
  File "/glade/u/home/wwieder/CTSM/tools/site_and_regional/../../python/ctsm/site_and_regional/single_point_case.py", line 503, in create_surfdata_at_point
    f_out = self.modify_surfdata_atpoint(f_tmp)
  File "/glade/u/home/wwieder/CTSM/tools/site_and_regional/../../python/ctsm/site_and_regional/single_point_case.py", line 418, in modify_surfdata_atpoint
    zip_pfts = zip(self.dom_pft, self.pct_pft, self.cth, self.cbh)
TypeError: zip argument #3 must support iteration

```

samsrabin:
### Description of changes

Ensures that `cth` and `cbh` are iterable before passing to `zip()`. Without this, most calls of `subset_data` (at least for single-point cases) will fail.

### Specific notes

**CTSM Issues Fixed:**
- Fixes #2527

**Does this create a need to change or add documentation? Did you do so?** Not needed

- Python unit tests were corrected to catch this error. They now pass after the fix.
- New unit tests added for new utility function `ensure_iterable()`.
- Python system tests pass.


slevis-lmwg (approved):
@samsrabin thank you for coming up with this elegant solution and for including unit tests for it. I do not see issues or have concerns if you have confirmed that it works.

wwieder:
Thanks for getting this in @samsrabin.  when I suggested the next b4bdev tag I wasn't meaning this one, but thanks for busting this out!

PR #2534, opened 2024-05-09
Adds a broadcast of the namelist value 'urbantvmapalgo'

### LINKED ISSUE #1912: List of mapalgo options for streams is different for MCT than NUOPC, currently we are using the MCT list
### Brief summary of bug

The list of options for the mapalgo namelist fields in CTSM is the list for MCT, rather than NUOPC. Since, we are moving to NUOPC we should update it to the NUOPC list and remove the MCT only options.

### General bug information

**CTSM version you are using:** ctsm5.1.dev114 (but comes with the introduction of NUOPC)

**Configurations affected:** with NUOPC

### Details of bug

The list of mapalgo streams options in the namelist are:

ndepmapalgo
lai_mapalgo
lightngmapalgo
popdensmapalgo
urbantvmapalgo

The list of stream options for mapalgo for MCT is:

```
    bilinear = bilinear interpolation
    nn       = nearest neighbor
    nnoni    = nearest neighbor on the "i" (longitude) axis
    nnonj    = nearest neighbor on the "j" (latitude) axis
    spval    = set to special value
    copy     = copy using the same indices
```

While the list for NUOPC is:

```
    bilinear = bilinear interpolation
    nn       = nearest neighbor
    consf  = first order conservative mapping
    consd = second order conservative mapping?
    none    = don't do any interpolation (same as copy for MCT?)
```

### LINKED ISSUE #2533: Urban time-varying data doesn't broadcast the mapping algorithm
### Brief summary of bug

This is a very simple bug that likely is never encountered in practice, but was hit while investigating an issue scaling CTSM out to large processor counts while using native (ultra high-resolution) grids for the map algorithm for this streams file.  The hope was that by switching to a native grid, we could avoid some of the communication patterns in the nearest neighbor mapping, but when testing that we noticed that the 'urbantvmapalgo' namelist variable is read, but not broadcast to other ranks.  As a result, other ranks got the default 'nn' method, and the main rank didn't, leading to a hang in communication on the MPI communicator.

### General bug information

ctsm5.2.003

No, but it causes a hang when changing the map algorithm from the default.

### Details of bug

Simply put, a namelist variable is read, but not broadcast to other tasks.  The result is that those other tasks try the default map algorithm vs the specified one.  This is likely immaterial to 99.99% of users, but it's an easy fix, hence the upcoming one-line PR.

Admittedly I haven't tested too many configurations, but the change is a logical one-line one that has been confirmed to work in at least two cases.

No output, since it leads to a hang as N-1 processors call an MPI_Allreduce (when set to 'nn') and the main rank does not.

### LINKED ISSUE #3345: Missing mpi_bcast for popdensmapalgo, ndepmapalgo, lightngmapalgo -- causes CTSM to hang in initialization
### Brief summary of bug

Same as #2533 for these other variables: popdensmapalgo, ndepmapalgo, lightngmapalgo

You can see this from code inspection of looking for mapalgo in the code.

### General bug information

**CTSM version you are using:** ctsm5.3.063-43-g1fe739da6

**Does this bug cause significantly incorrect results in the model's science?** No (it just hangs)

**Configurations affected:** Any case where you set these to non-default values

### Details of bug

I found this in adding a test for #2534 

The test I'm currently running in #2534 is:

SMS_D.f09_f09_mt232.I1850Clm50BgcCrop.derecho_intel.clm-obscureMapping

It just hangs in initialization and waits for the full JOB_WALLCLOCK_TIME to end before terminating.

briandobbins:
Adds a broadcast of the namelist value (urbantvmapalgo) after reading it in so it is consistent across all ranks when not using the default value ('nn').

### Description of changes

Simple one-line change; a namelist value wasn't being shared with ranks other than the main one reading the namelist.  This lead to hangs when trying a different algorithm.

### Specific notes

  Fixes #2533 (Done)
  Also fixes #3345 (Done)
  Also fixes #1912 (Done)

No

No

No

Testing performed, if any:
Limited testing, but it's a one-line logical change.  If more is needed, let me know.  I just did a simple test with and without the change, looking at the values on all ranks.


billsacks (approved):
Thanks for the fix @briandobbins !

ekluzek (approved):

ekluzek:
@briandobbins I'm rebasing this to come in on b4b-dev, so it can come in quicker.

wwieder:
Is this ready to merge?

samsrabin:
@briandobbins Would you mind rebasing this onto ctsm5.3.0 and force-pushing? It should go without conflicts. Not a big deal, but makes it clearer exactly what files are being changed.

samsrabin:
@briandobbins Rebase no longer needed; I'll start testing now.

briandobbins:
Well, I'm embarrassed to say I've only now gotten back to this (from last May!), but added one more change - enabling the 'redist' method for the 'urbantvmapalgo' variable in bld/namelist_files/namelist_definition_ctsm.xml.

Happy to hop on a quick call if anyone has questions or needs additional changes, but it looks like since the rebase wasn't needed, this is good to go?  Apologies for the ridiculously long delay!

ekluzek:
Thanks for the update @briandobbins. The problem you saw with redist is a more general problem that's important to correct throughout the code (#1912). So I'll do that on top of what you've done here and bring it in. That fix is something we really need to have corrected by the release, since it means you can't use many of these options for a list of other fields either.

We are all busy. We have lots of PR's that we can't bring in for many months as well...

ekluzek:
I created a branch of this that I'll bring in as a separate PR, on top of this one. But, I'll update, do some testing, and merge this in now, so I can use it for the decomp_init work.

PR #2545, opened 2024-05-15
Dust emissions control moved to cmeps

### LINKED ISSUE #2150: Add capability to flip between different dust emission parameterizations?

### LINKED ISSUE #2170: Update the drv_flds namelist definition file in CTSM to work with latest CAM/cmeps

### LINKED ISSUE #2376: Update CTSM with CMEPS update for controlling Dust emissions

### LINKED ISSUE #2524: Add option to NOT set the dust-emission drv_flds_in settings when coupled to CAM

### LINKED ISSUE #2666: 1x1_cidadinhoBR tests fail in build-namelist_test.pl
### Brief summary of bug

1x1_cidadinhoBR tests fail in `build-namelist_test.pl`.

### General bug information

**CTSM version you are using:** ctsm5.2.016

### Details of bug

Fine; that resolution isn't used by anything yet.

```
862/3314 < FAIL> <Test Id: 862> <Desc: -bgc bgc -crop -res 1x1_cidadinhoBR -use_case 1850_control -envxml_dir . clm4_5-bgc: lnd_in file exists>
1273/3314 < FAIL> <Test Id: 1273> <Desc: -bgc bgc -crop -res 1x1_cidadinhoBR -use_case 1850_control -envxml_dir . clm5_0-bgc: lnd_in file exists>
1684/3314 < FAIL> <Test Id: 1684> <Desc: -bgc bgc -crop -res 1x1_cidadinhoBR -use_case 1850_control -envxml_dir . clm5_1-bgc: lnd_in file exists>
2095/3314 < FAIL> <Test Id: 2095> <Desc: -bgc bgc -crop -res 1x1_cidadinhoBR -use_case 1850_control -envxml_dir . clm6_0-bgc: lnd_in file exists>
```

ekluzek:
### Description of changes

Remove the dust emissions namelist items from CTSM and use the namelist in the drv_flds_in for CMEPS.

This updates CTSM to use the namelist control in CMEPS (in https://github.com/ESCOMP/CMEPS/pull/429). So the CMEPS external needs to be updated, and the namelist control in CTSM changed to use CMEPS rather than the internal CTSM control settings and the CTSM ones removed.

The new XML variable:

LND_SETS_DUST_EMIS_DRV_FLDS

controls whether dust emission settings are set by CTSM or by CAM. Only one or the other can set them, and it's required so when CAM and CTSM are running together they need to know which one will select.

Add a unit test for the CMEPS code to make sure it's working correctly.

Fix the cidinahoBR test in the build-namelist unit tester. 

- [x] unit test
- [x] Add drv_flds_in namelist definition in CMEPS
- [x] Finalize CMEPS tag point to it here
- [x] Update drv_flds_in namelist definition in CTSM
- [x] Add capability to set the dust emissions settings in drv_flds_in
- [x] Have build-namelist respond to LND_SETS_DUST_EMIS_DRV_FLDS
- [x] Update build-namelist tester to get the new capability working
- [x] Remove the old dust namelist items in CTSM
- [x] Use the new cmeps shared code for namelist in FORTRAN
- [x] Run standard testing and make sure everything is working

### Specific notes

  Fixes #2376
  Fixes #2150
  Fixes #2524
  Fixes #2666

  Namelist items removed from CTSM and into the drv_flds_in namelist in CMEPS

Testing performed, if any: will run standard testing, have run a few singleton tests so far


ekluzek:
This replaces #2538 which got closed when I messed up the rebase to b4b-dev for it. This is actually the same branch, but I force pushed the rebase after I correctly applied it. Since the other was already auto closed I had to open a new one, even though it's the same branch.

ekluzek:
The CMEPS update is going to have to wait until after the cesm3_0_beta01 tag freeze for external updates.

ekluzek:
This is on hold because of the cesm3_0_beta01 freeze going on. But, can go to b4b-dev.

ekluzek:
I had some build problems using the updated CMEPS tag alone. But, replicating the ccs_config and CMEPS versions from #2590 I got the build to work.

bld/unit_testers/build-namelist_test.pl
```diff
@@ -924,303 +839,270 @@ sub cat_and_create_namelistinfile {
...
+                                     },
+     "soil_erod_none_w_Zender"  =>{ options=>"--envxml_dir .",
+                                     namelst=>"dust_emis_method='Zender_2003', " .
+                                     "zender_soil_erod_source='none'",
+                                     GLC_TWO_WAY_COUPLING=>"FALSE",
+                                     phys=>"clm5_1",
+                                     },
+     "soil_erod_bad_w_Zender"   =>{ options=>"--envxml_dir .",
+                                     namelst=>"dust_emis_method='Zender_2003', " .
+                                     "zender_soil_erod_source='zztop'",
```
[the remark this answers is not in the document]
slevis-lmwg: Also, is it ok that most (all?) dust-related namelst lines have a comma "," right after the method and end with a period "."? Looks inconsistent to me relative to other examples.

slevis-lmwg (commented):
Thank you @ekluzek 

Other than one concern that you will find in my review, nothing else jumped out at me. For me to understand a few things better, I would not mind a quick (30-minute?) meeting to go over this PR together if that's ok with you.

bld/unit_testers/build-namelist_test.pl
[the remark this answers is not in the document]
ekluzek: You are right about the duplication. Thanks for pointing that out. I think I might have added it on the previous branch that came in as well as this one. So I'll remove the second duplicate one.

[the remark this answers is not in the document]
ekluzek: Good question on the ending with the period. The syntax is correct, but as you point out it looks very arcane. The thing that's happening is that the "." concatenates two strings together. So the lines here...

``` perl
dust_emis_method='Zender_2003', " .
                                     "zender_soil_erod_source='none'",
```

Is really this...

``` perl
dust_emis_method='Zender_2003', zender_soil_erod_source='none'",
```

I did the concatenate to break up the long line. But, this might be a case that goes along with our recent discussion about long lines in code. With modern editors there isn't a reason to break lines up as much as in the past. The NAG compiler for FORTRAN code has a 132 character limit. So there's still reason to break up our FORTRAN code. And for readability I'd argue you should keep FORTRAN code shorter than that. But, perl and python don't have these limits. It's still an ingrained habit for me to shorten code lines, but this is a case where the concatenation makes the code unreaable. So I'll work with and removing that confusing use of concatenation syntax.

bld/CLMBuildNamelist.pm
```diff
@@ -4708,7 +4723,7 @@ sub write_output_files {
   $log->verbose_message("Writing clm namelist to $outfile");
 
   # Drydep, fire-emission or MEGAN namelist for driver
-  @groups = qw(drydep_inparm megan_emis_nl fire_emis_nl carma_inparm);
+  @groups = qw(drydep_inparm megan_emis_nl fire_emis_nl carma_inparm dust_emis_inparm);
```
glemieux, line 4726: I'm not very familiar with how `@groups` works, but I noticed that I couldn't find `dust_emis_inparm` (or `carma_inparm`) elsewhere in the CLMBuildNamelist.pm.  I'm assuming this is fine, but wanted to double check.

glemieux (approved):
Per our discussion I only reviewed the namelist defaults and CLMBuildNamelist.  I only had one question asking for a clarification.

bld/CLMBuildNamelist.pm
```diff
@@ -3960,50 +3960,65 @@ sub setup_logic_fire_emis {
...
+  # Only set dust emission settings -- if not connected to CAM
+  my $lnd_sets_dust = logical_to_fortran($envxml_ref->{'LND_SETS_DUST_EMIS_DRV_FLDS'});
+  if ( &value_is_true( $lnd_sets_dust)) {
+
+      # First get the dust emission method
+      my $var = "dust_emis_method";
+      add_default($opts,  $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var );
+      my $dust_emis_method = remove_leading_and_trailing_quotes( $nl->get_value($var) );
+
+      my @zender_files_in_lnd_opts = ( "stream_fldfilename_zendersoilerod", "stream_meshfile_zendersoilerod",
```
ekluzek, line 3975: Make sure the logic is such that this isn't set if dust_emis_method != Zender_2003

src/biogeochem/DustEmisFactory.F90
```diff
@@ -28,30 +28,28 @@ function create_dust_emissions(bounds, NLFilename) result(dust_emis)
         !---------------------------------------------------------------------------
         use DustEmisBase      , only : dust_emis_base_type
         use DustEmisZender2003, only : dust_emis_zender2003_type
-        use clm_varctl        , only : dust_emis_method
         use decompMod         , only : bounds_type
         use shr_kind_mod      , only : CL => shr_kind_cl
+        use shr_dust_emis_mod , only : is_dust_emis_zender, is_dust_emis_leung
```
slevis-lmwg, line 33: I suggest a short justification explaining why this is preferable over using the generic dust_emis_method variable.

bld/unit_testers/build-namelist_test.pl
```diff
@@ -924,303 +839,270 @@ sub cat_and_create_namelistinfile {
...
+                                     phys=>"clm5_1",
+                                     },
+     "Set_Dust_When_CAM_Sets"   =>{ options=>"--envxml_dir .",
+                                     namelst=>"dust_emis_method='Zender_2003'",
+                                     LND_SETS_DUST_EMIS_DRV_FLDS=>"FALSE",
+                                     phys=>"clm5_1",
+                                     },
+     "soil_erod_none_w_Zender"  =>{ options=>"--envxml_dir .",
+                                     namelst=>"dust_emis_method='Zender_2003', " .
+                                     "zender_soil_erod_source='none'",
```
ekluzek, line 1099:
```suggestion
                                     namelst=>"dust_emis_method='Zender_2003', zender_soil_erod_source='none'",
```

bld/CLMBuildNamelist.pm
ekluzek: This is correct for the perl code. This just lists what namelists you want to write out and you only want to list each one once for the perl code. The namelist is referenced in the new namelist_definition file for dust_emis.

src/biogeochem/DustEmisFactory.F90
ekluzek: Yep will do.

ekluzek: Adding some notes here. The design doc for the CMEPS code is here:
[
](https://docs.google.com/document/d/18nZ3LJF5W-YF9iBhqed6s_NWeKOvSSL2-k0Lye1nnLg)

It uses a OOP design in order to make sure that it can't be messed up by either CAM or CTSM (since it's used by both). Using the function calls is a way of ensuring internal variables aren't able to be changed by either. There's a few other ways it could be done, but since it's used by both I thought it important to make sure neither could mess with the internal settings. That's also the reason to have a unit tester for it, so that both models can rely on the behavior for it.

bld/CLMBuildNamelist.pm
ekluzek: I couldn't tell from the diff if the logic is done correctly, but looking at the code it looks fine. So resolving.

src/biogeochem/DustEmisFactory.F90
ekluzek: @slevis-lmwg I added a note in this regard at the function header. Let me know about any feedback.

src/biogeochem/DustEmisFactory.F90
```diff
@@ -25,33 +25,41 @@ function create_dust_emissions(bounds, NLFilename) result(dust_emis)
         !---------------------------------------------------------------------------
         ! Create a dust_emission base class objecct
         ! The method implemented depends on namelist input
+        !
+        ! DESIGN NOTES:   Erik Kluzek 07/15/2024
```
ekluzek, line 29: @adrifoster if you can I'd be interested in your feedback on this little design note section I added. Thanks...

bld/unit_testers/build-namelist_test.pl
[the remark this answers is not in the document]
ekluzek: OK, handled both these comments.

src/biogeochem/DustEmisFactory.F90
slevis-lmwg: Looks good. Thank you for doing that @ekluzek

src/biogeochem/DustEmisFactory.F90
adrifoster: Good note. I like this strategy better (at least in this instance) since as you say we need both CAM and CTSM to know about the dust emission type. It's better to have it chosen as a function that we can control elsewhere. CTSM doesn't need to know anything about *how* we know which dust emission it is, we just need to know is it this? is it that?

ekluzek: Awesome. Thanks for looking at it.

Notes like this might be something appropriate to do in other places when we bring in new code and we want to talk about the design. I think it is better kept here in the code for this case than under the doc/design directory.

ekluzek:
Testing on Izumi is as expected, with answers identical with the exception of changes in the field list. There are two new fields in the cpl files due to the cmeps update.

Similarily on Derecho 173 tests differ because of the field list. Most tests pass, but I am seeing the following fails right now (but could be because of Derecho problems):

ERI_D.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode		
ERP_D_Ld5.f10_f10_mg37.I1850Clm50BgcCropG.derecho_gnu.clm-glcMEC_changeFlags		
ERP_D_Ld9.ne30pg3_t232.I1850Clm51BgcCropG.derecho_intel.clm-clm51cam6LndTuningMode		
ERP_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode		
ERP_D_Ld9.ne30pg3_t232.IHistClm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode		
ERP_P256x2_D_Ld5.f19_g17_gris4.I1850Clm50BgcCropG.derecho_intel.clm-glcMEC_increase		
ERP_P64x2_D_Ld10.f10_f10_mg37.IHistClm50SpG.derecho_intel.clm-glcMEC_decrease		
ERS_D_Ld12.f10_f10_mg37.I1850Clm50BgcCropG.derecho_intel.clm-glcMEC_spunup_inc_dec_bgc		
ERS_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode		
ERS_Ly3_P64x2.f10_f10_mg37.IHistClm50BgcCropG.derecho_intel.clm-cropMonthOutput		
LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac		
SMS_Lm1.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode		
SMS_Lm37.f10_f10_mg37.I1850Clm50SpG.derecho_intel.clm-glcMEC_long

ekluzek:
ALL of the above tests with CISM failed because CMEPS expects to receive runoff from CISM. So I needed to update CISM. I was afraid that would change answers, but from one short test it looks like it might be OK.

I also need to look into the LILAC test.

ekluzek:
I figured out the fix for LILAC.

The CISM update is identical for the shorter tests, but the 3-year and 37 month test change answers. I could make it identical by having branches of submodules -- but that seems to be too much. So it seems the best way to go is to bring this to master, which should then be b4b since the submodule updates have already come in.

ekluzek:
Testing on izumi is b4b with ctsm5.2.015 and as expected. On Derecho only the following are unexpected:

ekluzek:
Looks like the FATES tests are due to this:

https://github.com/ESCOMP/CTSM/issues/2656

ekluzek:
aux_clm testing is as expected on both Derecho and Izumi. The following tests show differences in answers because of the issues talked about above...

ERP_P64x2_Lm13.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-monthly--clm-matrixcnOn_ignore_warnings
SMS.f10_f10_mg37.I2000Clm50BgcCrop.derecho_nvhpc.clm-crop
SMS_D.1x1_brazil.I2000Clm60FatesSpCruRsGs.derecho_gnu.clm-FatesColdDryDepSatPhen
SMS_D.1x1_brazil.I2000Clm60FatesSpCruRsGs.derecho_gnu.clm-FatesColdMeganSatPhen

The "nvhpc" test is because of a missing baseline for ctsm5.2.016.

ekluzek:
About a week for sprint 8.

Issue #1928, opened 2022-12-14
"Easy wins" for crop calendars


samsrabin:
One thing that's emerging from my analysis and writeup of #1863—e.g., [my AGU poster](https://agu2022fallmeeting-agu.ipostersessions.com/Default.aspx?s=EA-57-89-17-FF-79-D0-1C-4D-22-E4-2B-3E-DA-C9-9E)—is that there are a few crops where using the GGCMI prescribed calendars makes a really big difference in global yield, mainly due to changes in growing season length. This is especially evident for sugarcane, cotton, and rice, depending on how "maturity" is defined (see #1914). For sugarcane and cotton, we see a huge performance improvement. For rice, it appears to cause decreased performance—a pretty big overestimation—but I would argue it's "wrong for the right reasons."

This has got me thinking about how the existing crop calendar system can learn from the GGCMI prescribed calendars without fully relying on them.

Tagging @danicalombardozzi for eventual discussion.

# Ideas

## Sowing dates

CLM default sowing is implemented as a set of hemisphere- and crop-specific sowing windows, within which planting occurs if and when some historical and/or current-season temperature criteria are met. These windows allow for some interannual variation in sowing date, as well as some long-term adaptation to warming temperatures. The GGCMI sowing data are just individual static dates, which of course allow for neither.

What I'm thinking is that CLM could use the GGCMI sowing date maps as the center of new sowing windows, with the sowing window radius being the same as it was before. (This would require some testing to ensure correct handling of sowing windows that cross into the new year.)

## Harvest requirements

The simplest implementation of this would be to use the harvest requirement files generated using the GGCMI calendars over some baseline period, as I'm already doing.

One could also imagine some scaling of harvest requirements up or down according to recent climate—e.g., with 20-year mean annual temperature.

## Mitigating immature harvests

Cotton and rice yields are really strongly affected by slight changes in how "maturity" is defined—i.e., how one chooses to include or exclude certain harvests as successful nonzero yields. (See my poster linked above.) An apparently okay performance for global rice yield under a strict 100% maturity threshold is only possible because huge parts of China never have a successful yield! They end up being harvested due to reaching maximum growing season length. Because relaxing the maturity threshold to 90% results in many more successful harvests (again, see poster), this implies that allowing a slightly longer growing season might do the same.

I propose that maximum growing season length (a global parameter) be set according to the maximum seen in the GGCMI static calendars plus, say, 10%.

[...]

PR #2560, opened 2024-05-28
Allow prescribed gddmaturity to vary based on recent climate (v2)

samsrabin:
### Description of changes
In the existing code, when using the default CLM crop calendars, `gddmaturity` (maturity requirement; the number of growing degree-days it takes for a crop to mature and be harvested) can vary within some bounds based on climate over the past 20 years. Specifically, CLM looks at some annual "warm period" which differs between the northern and southern hemispheres. This variation allows "farmers" to adapt the "cultivars" (varieties) they grow based on recent climate.

When _prescribing_ maturity requirement, though—with input file `stream_fldFileName_cultivar_gdds`—it's currently not allowed to vary over time. Instead, it's locked in to what's provided in the input file: the mean GDD accumulation in the GGCMI growing period over the 1980-2009 growing seasons.

We want this variation to allow "farmers" to adapt the "varieties" they grow based on whether the recent climate has been warmer or cooler than that baseline 1980-2009 period.

This PR represents the first step toward adding that ability. I'm breaking this work up into separate PRs to keep them manageable:
1. (This PR) Define "recent climate" based on the same "warm period" currently used by default CLM.
1.1. Add more namelist options and checks.
2. Define "recent climate" based on the GGCMI growing seasons.
3. Restrict adapted `gddmaturity` values to a realistic range.

### Specific notes

- Contributes to #1928

**Are answers expected to change (and if so in what way)?** Not unless enabling this new, experimental feature.

- Adds input file `stream_fldFileName_gdd20_baseline` (default `''`). Pointing this at a file automatically enables adaptation of prescribed maturity requirements; there's no separate boolean namelist variable to worry about. Providing this file but not prescribed maturity requirements will cause an error in the build namelist step.

**Does this create a need to change or add documentation? Did you do so?** Documentation will need to be updated when all the PRs in my plan are merged and this feature is considered ready for use. But at this point, things are preliminary enough that I'd have to rewrite whatever docs I added in this PR. So no: I'm adding no documentation here.

- [x] Ensure `clm_short` passes bit-for-bit
- [x] Ensure a transient run with `stream_fldFileName_gdd20_baseline` provided works as we expect (`gddmaturity` changes over time)
- [x] Ensure `aux_clm` passes bit-for-bit


ekluzek:
@samsrabin this PR says that it may be "subject to force pushes". In the branch protection rules there is a button that currently doesn't allow force pushes -- should I switch that to allow them?

samsrabin:
Oh thanks for that heads up, Erik. Let's see if I end up needing to do any—if not, we won't need to worry about it.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -143,6 +143,9 @@ module CNPhenologyMod
   logical,  public :: generate_crop_gdds = .false. ! If true, harvest the day before next sowing
   logical,  public :: use_mxmat = .true.           ! If true, ignore crop maximum growing season length
 
+  ! For use with adapt_cropcal_rx_cultivar_gdds .true.
+  real(r8), parameter :: min_gdd20_baseline = 0._r8  ! If gdd20_baseline_patch is ≤ this, do not consider baseline.
```
slevis-lmwg, line 147: @samsrabin was considering adding this as a namelist option. I recommended:
- adding to the params file OR
- not adding at all

I think @samsrabin prefers the latter because he will converge on some value that is unlikely to ever need updating.

slevis-lmwg (approved):
@samsrabin and I looked over this PR.
@samsrabin took notes and will make changes accordingly.
I think my comments were minor enough, plus this PR is getting merged to a branch (not main), so I'm comfortable approving preemptively.

PR #2577, opened 2024-06-03
Update gitfleximod to latest that fixes some particular problems.

### LINKED ISSUE #2591: Start using "submodule" rather than "external" in the code (with the git fleximod update)
With manage_externals we referred to other components maintained as part of CTSM as "externals". However, now with git fleximod externals isn't accurate as they are now managed using "git submodule". So using the term "submodule" is more accurate. We should start using it when we refer to components that are part of the CTSM checkout and we should replace the word external in the code documentation with submodule to reflect the new more accurate language at this point.

ekluzek:
### Description of changes

Just update the gitfleximod subtree to v0.7.7. This fixes an issue in identifying that a tag is out of sync in a submodule. An issue about recursion on a submodule. Allows untracked files, and fixes a MS Window Memory error.

### Specific notes

  Fixes #2591

Does this create a need to change or add documentation? Did you do so? No But, did update documentation

Testing performed, if any: 

 Tested git-fleximod issues in https://github.com/ESMCI/git-fleximod/issues/38 and made sure it works on Izumi
 Also tested two other git-fleximod issues I had and they work as well.


ekluzek:
Since, this isn't critical to come in, we likely might rebase to b4b-dev and bring it in there after b4b-dev is updated to master.

ekluzek:
We talked about this in CTSM SE this morning. We figure this should come to master and be part of the cesm3_0_beta01b tag (that I'm confirming with CSEG that is what we plan to do).

We also figure we'll wait until 01a is done and we think we have a "final" version of git-fleximod for the beta01 tag. So we'll update the git-fleximod version again with that version, before this comes in.

jedwards4b:
@ekluzek Please update this with a new pull of fleximod main.  The latest version is git-fleximod 0.7.7

slevis-lmwg (approved):
@ekluzek the documentation updates look good, thank you for catching those!

@ekluzek and I agreed that I could approve this preemptively, while Erik is running tests.

PR #2581, opened 2024-06-05
add surface water outflow parameter

swensosc:
### Description of changes
change hard coded constant to a parameter on the parameter file

### Specific notes
netcdf file was modified by adding the following:
h2osfc_outflow_scalar = 0.002
double h2osfc_outflow_scalar ;
		h2osfc_outflow_scalar:long_name = "h2osfc outflow scalar" ;
		h2osfc_outflow_scalar:units = "unitless"

when use_hillslope = .true.  surface water outflow will change.  The original constant was 1e-4, the new parameter value is currently 2e-3

Testing performed, if any:

### DIFF
--- a/src/biogeophys/SurfaceWaterMod.F90
+++ b/src/biogeophys/SurfaceWaterMod.F90
@@ -41,6 +41,7 @@ module SurfaceWaterMod
   type, private :: params_type
      real(r8) :: pc              ! Threshold probability for surface water (unitless)
      real(r8) :: mu              ! Connectivity exponent for surface water (unitless)
+     real(r8) :: h2osfc_outflow_scalar    ! H2osfc outflow scalar (unitless)
   end type params_type
   type(params_type), private ::  params_inst
 
@@ -68,7 +69,9 @@ contains
     call readNcdioScalar(ncid, 'pc', subname, params_inst%pc)
     ! Connectivity exponent for surface water (unitless)
     call readNcdioScalar(ncid, 'mu', subname, params_inst%mu)
-
+    ! H2osfc outflow scalar (unitless)
+    call readNcdioScalar(ncid, 'h2osfc_outflow_scalar', subname, params_inst%h2osfc_outflow_scalar)
+    
   end subroutine readParams
 
   !-----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/biogeophys/SurfaceWaterMod.F90
```diff
@@ -486,7 +489,8 @@ subroutine QflxH2osfcSurf(bounds, num_hydrologyc, filter_hydrologyc, &
           k_wet=1.0e-4_r8 * sin((rpi/180._r8) * topo_slope(c))
           if (col%is_hillslope_column(c)) then
              ! require a minimum value to ensure non-zero outflow
-             k_wet = 1e-4_r8 * max(col%hill_slope(c),min_hill_slope)
+             !k_wet = 1e-4 * max(col%hill_slope(c),min_hill_slope)
```
ekluzek, line 492: Remove the commented out line.

ekluzek (changes requested):
This is great. Always better to have parameters on the parameter file rather than hardcoded in.

I asked for a commented out line to be deleted. And we'll need to have a new parameter file to point to for this in bld/namelist_files/namelist_defaults_ctsm.xml. So that's required.

The other suggestion I have though is that we rebase this to b4b-dev rather than master and bring it in with a parameter file with the same value. Then as an answer changing tag we'll update the parameter file as a separate thing. One reason to do that is that it's easier to bring something in on b4b-dev, and we could do that right away, whereas master is frozen right now until cesm3_0_beta01 is done, and that will be near the end of the month. How does that sound?

src/biogeophys/SurfaceWaterMod.F90
ekluzek: Perfect

swensosc:
that sounds good.  I have created a parameter file:
/scratch/cluster/swensosc/ctsm60_params.c240208_kwet_params.nc with the new
value, and with the old: /scratch/cluster/swensosc/
ctsm60_params.c240208_kwet_params_1e-4.nc
I compared using the old value via the parameter file, and I think it had
roundoff differences.

ekluzek:
We should just do this on b4b-dev.

ekluzek:
Pinging @samsrabin is this something you can still do?

samsrabin:
Yep!

PR #2585, opened 2024-06-06
Allow "manual" resets of all accumulator types

### LINKED ISSUE #2584: Reset accumulators to initval instead of 0
Accumulation fields (`accumulMod`) with type `timeavg` or `runaccum` are hard-coded to be set to 0 when reset. This should be changed to instead use `initval`.

This isn't a bug per se, because currently no such accumulators are initialized to anything other than 0. The only accumulators initialized to something other than 0 are type `runmean`, which can't (yet) be reset.

samsrabin:
## Description of changes

There are three types of accumulators specified in `accumulMod`: `timeavg`, `runmean`, and `runaccum`. `timeavg` is reset (value set to 0) every N steps, where N is the length of the period over which the field is being averaged. `runaccum` can be reset "manually" whenever the value passed in to the "update" subroutine is -99999.

`runmean`, in contrast, has no way to reset. This ends up being a problem for me with the crop calendar work, so I wanted to add that ability.

This PR adds the ability for `runmean` to be manually reset. It does the same for `timeavg`, and it makes it so that `runaccum` doesn't use the -99999 method anymore. See Design Note below for more details.

## Specific notes

**Contributors other than yourself, if any:** @ekluzek 

**CTSM Issues Fixed:**
- Fixes #2584

- Existing `accumulMod` unit tests pass.
- Added new `accumulMod` unit tests, which also pass.
- `aux_clm` OK on Izumi. On Derecho, there were various namelist and true differences, plus a run submission failure. For various reasons I'm suspicious of the `ctsm5.2.005` baseline on Derecho, so I'm creating my own and will re-run `aux_clm` against that.

---------------------------------------

#### General Contraints

The `timeavg` method doesn't make sense for `runmean`, so I was thinking I would use the method from `runaccum`. But as I thought about it more, there were some things I didn't like:
- All variables we use `runaccum` for at the moment are by definition zero or positive, but theoretically some new variable might be added for which -99999 could be a valid value. I call this the "overloading" problem: Usually the update value is a real value, but sometimes it's a flag saying the accumulator should be reset.
- If a patch is inactive when a reset is requested for `timeavg` or `runaccum`, it will not be reset. I call this the "missed reset" problem. See also: #2587
- `runaccum` can't both reset and update itself in the same call of `update_accum_field_runaccum()`. If a reset is requested, the value for that timestep is lost. I call this the "missed update" problem. See also: #2586

Whatever I do to solve the first issue, I want it to also be used for `runmean` so there aren't multiple reset methods floating around. With regard to the other two issues, I want `runmean`'s reset implementation to behave how I need it to, but I don't want to change `runaccum` because fixing the "missed update" problem will be answer-changing. (We should probably fix that and the "missed reset" problem, but that will require more of a discussion—I'd like my current PR to be bit-for-bit.)

Finally, even though it's not necessary for my current work, I wanted to add a way for `timeavg` fields to be manually reset.

For my purposes, I want to be able to reset (all levels of) a given point. However, there may also be situations where one might want to reset (all levels of) *all* points, one level of all points, or one level of one point.

### Solutions
One way to tell `update_accum_field()` that some point-levels should be reset, while avoiding the "overloading problem," could be to pass in an optional logical array. However, that would get messy. It would require a lot of boilerplate code to be added to the two `update_accum_field_[sm]l()` procedures and to the `update_accum_field_(timeavg,runmean,runaccum)` subroutines.

Instead, I decided to give the `accum_field` type a new logical member. Because my crop calendar work requires the ability to reset *only* certain patches, I'm implementing this as an array of the same size as `val`—i.e., number of points (patches, columns, etc.) by number of levels. This new array is to be called `reset`.

`reset` will be changed using a new subroutine, `markreset_accum_field()`. This will optionally accept two indices indicating which points and/or levels, if any, should have `reset` set to `.true.`. If neither optional index is provided, `reset` will be set to `.true.` for all point-levels.

To avoid the "missed reset" problem, I could have made it so that all points with `reset == .true.` are reset whether they're active or not. However, that could lead to spurious values of the accumulator if the point becomes active again and its accumulator queried (with `extract_accum_field()`) before a new (first) value is accumulated (with `update_accum_field()`). So instead, if an inactive point has `reset == .true.`, its value is not reset and its `reset` remains `.true.`. If and when the point becomes active again, its value will be reset and its `reset` will be set to `.false.`. That's why I named the above subroutine `markreset_accum_field`: It *marks* point-levels as needing to be reset, but doesn't actually reset them.

Avoiding the "missed update" problem is simple enough—don't put the "reset" and "update" code blocks on either side of an if/else.

Note, again, that I will not be solving the "missed reset" and "missed update" problems where they already exist (`timeavg`/`runaccum` and `runaccum`, respectively).

#### Algorithm or Pseudo code for main components

screenshot_0934
screenshot_0935
screenshot_0936

### DIFF
--- a/src/biogeochem/CNDVType.F90
+++ b/src/biogeochem/CNDVType.F90
@@ -496,7 +496,7 @@ contains
     end do
     if (month==1 .and. day==1 .and. secs==int(dtime)) then
        ! Reset annually
-       call markreset_accum_field('AGDD')
+       call markreset_accum_field('AGDDTW')
     end if
     call update_accum_field  ('AGDDTW', rbufslp, nstep)
     call extract_accum_field ('AGDDTW', this%agddtw_patch, nstep)
--- a/src/biogeochem/CNDVType.F90
+++ b/src/biogeochem/CNDVType.F90
@@ -489,6 +489,9 @@ contains
     ! Accumulate and extract AGDDTW (gdd base twmax, which is 23 deg C
     ! for boreal woody patches)
     
+    ! SSR 2024-06-07: Don't wrap this do-loop in an "if it's not time to reset." Behavior would
+    ! be identical for now, but if "missed update" behavior is fixed (see ESCOMP/CTSM#2585), you
+    ! would end up updating AGDDTW with uninitialized values.
     do p = begp,endp
        rbufslp(p) = max(0._r8, &
             (t_a10_patch(p) - SHR_CONST_TKFRZ - dgv_ecophyscon%twmax(ndllf_dcd_brl_tree)) &
@@ -503,6 +506,9 @@ contains
 
     ! Accumulate and extract AGDD
 
+    ! SSR 2024-06-07: Don't wrap this do-loop in an "if it's not time to reset." Behavior would
+    ! be identical for now, but if "missed update" behavior is fixed (see ESCOMP/CTSM#2585), you
+    ! would end up updating AGDD with uninitialized values.
     do p = begp,endp
        rbufslp(p) = max(0.0_r8, &
             (t_ref2m_patch(p) - (SHR_CONST_TKFRZ + 5.0_r8)) * dtime/SHR_CONST_CDAY)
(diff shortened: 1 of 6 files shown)


slevis-lmwg (approved):
@samsrabin and I went over this PR together and Sam R. took notes for some minor changes and comment additions, plus an important calculation fix.

We agreed that I could approve preemptively.

samsrabin:
@slevis-lmwg All tasks we discussed are done, as is my baseline generation. Submitting aux_clm now.

PR #2623, opened 2024-06-26
Change a NIWO test to HARV and update expected failures

slevis-lmwg:
### Description of changes
Motivated by problems with the NIWO test in #640 (this [comment](https://github.com/ESCOMP/CTSM/pull/640#issuecomment-2143101361))

@wwieder recommended switching the NIWO test to HARV.

@ekluzek recommended opening a PR to give this change more visibility. I implemented the change in #640 as well.

### Specific notes

@wwieder 

No, but one NIWO test has been changed to a HARV test.

Testing performed, if any:
I ran the tests
```
SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Bgc.derecho_gnu.clm-default--clm-NEON-HARV
SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_intel.clm-FatesFireLightningPopDens--clm-NEON-FATES-NIWO
```
- the former to show that the new test passes
- the latter because this test does not fail, and I updated the expected failures
- the latter also appears as derecho_gnu, but only in the expected fails (not in testlist), so i'm removing

### DIFF
--- a/cime_config/testdefs/ExpectedTestFails.xml
+++ b/cime_config/testdefs/ExpectedTestFails.xml
@@ -76,13 +76,6 @@
     </phase>
   </test>
   
-  <test name="SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_gnu.clm-FatesFireLightningPopDens--clm-NEON-FATES-NIWO">
-    <phase name="SHAREDLIB_BUILD">
-      <status>FAIL</status>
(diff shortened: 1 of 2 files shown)


slevis-lmwg:
@ekluzek could you take a look and approve if it makes sense to you?

samsrabin (approved):

PR #2678, opened 2024-08-08
Add check in build-namelist_test.pl

### LINKED ISSUE #2673: Catch use of namelist attribute in failure testing in perl build-namelist tester
Until #585 happens a little bit of error checking added to the failure and warning testing in ./build-namelist_test.pl could really help us with adding new tests. These tests set a hash, with options, namelst, phys, and XML variable names. If any of those are misspelled (with the misspelled namelist attribute namelst being particularly easy thing to miss) -- the test won't function correctly. But, just adding error checking to ensure the expected things are set would catch this for developers and make adding a new test much easier.

Also change namelst to namelist would also help as well.

mvdebolskiy:
### Description of changes

Added a check in `failtest`, `warntest`, `finidat_files`, so when a new test or finidat file are added and one of the subkeys is misspelled you will get an error and tester will die.

### Specific notes

@ekluzek 

Resolves #2673


mvdebolskiy:
You can test this by misspelling a subkey for one of the existing ones.

ekluzek (approved):
This looks great to me. I'm going to bring this in as it will help with other namelist work we will be working on.

PR #2703, opened 2024-08-15
Add a namelist warning when hillslope hydrology is used with initial interpolation.

### LINKED ISSUE #1806: Get init_interp working correctly with hillslope initial conditions files
I'm pretty sure that init_interp won't currently work correctly with initial conditions files from a hillslope run. (I'm also not positive it's working ideally for interpolation from a non-hillslope run to a hillslope run, but it seems likely that it's at least doing something reasonable in that situation.) One need – maybe the main one – is to label each column with appropriate metadata so that, in the interpolation, we can copy data from the source lowland column to the destination lowland column, etc. This can be done with discrete labels giving the type of each column and/or a continuous (real-valued) variable that lets us find the closest match. Currently, all vegetated columns are given the same type index and init_interp doesn't have a way to handle multiple columns in the vegetated landunit. (In contrast, for some other landunits, like crop and glacier, we have mechanisms to ensure that the appropriate source column from a landunit is associated with a given destination column.)

A good first-order check of whether this is working correctly is whether interpolating an initial conditions file onto an identical configuration will result in a replica of the original initial conditions file (this is what the LII test checks). Currently I believe this would not be the case. But eventually - at least if hillslope becomes default - we also want this to be doing something reasonable when interpolating from one resolution / hillslope configuration to another.

@swensosc

mvdebolskiy:
### Description of changes

Added warning to `CLMBuildNamelist.pm` and a warning test in `build-namelist_test.pl`.

### Specific notes

The warning is added after logic for initial conditions is done.

 Fixes #1806

No.

You have to run `./xmlchange --append CLM_BLDNML_OPTS=-ignore_warnings` if you want to have `use_init_interp` with `use_hillslope` both true.

Testing performed, if any:

ran build-namelist_test.pl and tested a new case with above namelist changes


ekluzek (approved):
This should be rebased to b4b-dev, but otherwise this is great.

samsrabin:
Converting to draft, pending the rebase.

samsrabin (changes requested):
This is good to go, pending your merge of mvdebolskiy/CTSM#6.

samsrabin (approved):

PR #2500, opened 2024-04-30
CTSM5.3.0: New raw pft/lai/glc-behavior/soilcolor/fire datasets

slevis-lmwg:
## Description of changes

Updating gen_mksurfdata_namelist.xml, initially with temporary paths to the new raw data.

New surface datasets because of updates in PFT/LAI/soil-color datasets, as well as two fire datasets (AG fire, pearland and the glacier behavior dataset. Also bring in an updated "PPE informed, hand-tuned" parameter file. And turn on Li2024 fire method. Also updates to allow creating transient landuse timeseries files going back to 1700. Fix an important bug on soil fields that was there since ctsm5.2.0. This results in mkaurfdata_esmf giving identical answers with a change in number of processors. Add in creation of ne0np4.POLARCAP.ne30x4 surface datasets.

Asides: Remove VIC fields from surface datasets and testing. Add version to the surface datasets. Update the fire emissions factor dataset for the fire emissions testing in CTSM. Remove the --hires_pft option from mksurfdata_esmf as we don't have the datasets for it.

TODOs:
- [x] 1700 fsurdat for TRENDY2024 spinup and
- [x] 1700-2023 landuse file for TRENDY2024 historical
- [x] Run ctsm5.2 spinup and historical as the control and
- [x] Compare ctsm5.3 (aka TRENDY2024) to ctsm5.2
- [x] Bring in new dataset from Bill (see below)
- [x] @samsrabin @lifang0209: new peatland and AG fire peak month datasets to @slevis-lmwg (will move to streams later on in the ctsm6.0.0 cycle)
- [x] @adrifoster: Table new population stream file for now. Adrianna explains, "it's formatted oddly and would be a pain to deal with"
- [x] Update gen_mksurfdata_namelist.xml accordingly
- [x] Bring in new code changes from Bill (See below)
- [x] Before making any ssp landuse/fsurdat files, confirm that @lawrencepj1 has updated the raw ssp data
- [x] Remove the hires_pft option (will add back in in ctsm6.1 with hires datasets)
- [x] Remove the -vic option and special handling in Makefile, as well as remove vic tests (see #2597)
- [x] Existing "make all" seems to have worked after a few corrections; took 6 hrs to run but did not include ultra high res in this iteration
- [x] Get new PtVg (noAnthro) raw data file from Peter
- [x] Move all new files and update temporary file paths to their final locations, then ./rimport:
  - [x] new param file(s), new GlacierRegion, fire, NOT population
  - [x] raw lake and urban files 1700-1849
[...]

PR #2737, opened 2024-08-30
mksurfdata_esmf: Refactor mksoiltex

samsrabin:
### Description of changes

Refactors `mksoiltex()` subroutine to make it easier to understand and avoid duplicated code.

Also:
- Checks for negative values in top sand layer (was already happening for other soil components)
- Changes `gen_mksurfdata_jobscript_single.py` to use user's `ACCOUNT` environment variable as default.

### Specific notes

As of eacff1077:
- [x] Builds without error
- [x] Runs without error
- [x] f19 outputs are ~~bit-for-bit identical~~ acceptably different with previous version (organic_o option 1)—roundoff-level diffs due to refactoring.
- ~~f19 outputs are bit-for-bit identical with previous version (disabled organic_o option 2)~~ Removing that code
- [ ] **(IN PROGRESS)** Check that `make all` generates all datasets without error.

### Other remaining tasks
- [x] Rename `is_neg_4` to be more meaningful. **Changed to `is_sand_dune` with d5f389a97**
- [x] Remove organic_o option 2. **Done with 99acaf648**
- [x] Check sand and `organic_o` for negative values; see [this comment](https://github.com/ESCOMP/CTSM/pull/2737#discussion_r2023587435).

### DIFF
--- a/tools/mksurfdata_esmf/src/mksoiltexMod.F90
+++ b/tools/mksurfdata_esmf/src/mksoiltexMod.F90
@@ -35,6 +35,63 @@ module mksoiltexMod
 contains
 !=================================================================================
 
+  subroutine mksoiltex_i_to_o(no, lookup_index, n_scid, nlay, name, val_neg_4, val_neg_other, data_i, data_o)
+    !
+    ! Arguments
+    integer, intent(in) :: no
+    integer, intent(in) :: lookup_index
+    integer, intent(in) :: n_scid
+    integer, intent(in) :: nlay
+    character(*), intent(in) :: name
+    real(r4), intent(in) :: val_neg_4  ! Fallback value if read-in value is -4
+    real(r4), intent(in) :: val_neg_other  ! Fallback value if read-in value is negative but not -4
+    real(r4), intent(in) :: data_i(:,:,:)
+    real(r4), intent(out) :: data_o(:,:)
+    !
+    ! Locals
+    integer :: l
+    logical :: is_neg_4
+
+    ! Fill first layer of output array with first positive value on SCID dim of input array
+    data_o(no,1) = data_i(1,1,lookup_index)
+    if (data_o(no,1) < 0._r4) then
+       do l = 2,n_scid
+          if (data_i(1,l,lookup_index) >= 0._r4) then
+             data_o(no,1) = data_i(1,l,lookup_index)
+             exit
+          end if
+       end do
+    end if
+
+    ! Handle cases where no positive value was found
+    if (data_o(no,1) < 0._r4) then
+       is_neg_4 = int(data_o(no,1)) == -4  ! TODO: Rename. Maybe indicates sand dune?
+       if (is_neg_4) then
+          data_o(no,:) = val_neg_4
+       else
+          data_o(no,:) = val_neg_other
+       end if
+    end if
+
+    ! Ensure negative values are handled
+    ! TODO: Remove as unnecessary?
+    if (data_o(no,1) < 0._r4 .and. trim(name) /= "sand") then
+       write(6,*)'ERROR: at no, lookup_index = ',no,lookup_index
+       call shr_sys_abort('could not find a value >= 0 for '//name)
+    end if
+
+    ! Top soil layer is filled above. Here, we fill the other layers.
+    do l = 2,nlay
+       data_o(no,l) = data_i(l,1,lookup_index)
+
+       ! If a layer is negative, fill it with the previous layer's value
+       if (data_o(no,l) < 0._r4) then
+          data_o(no,l) = data_o(no,l-1)
+       end if
+    end do
+
+  end subroutine mksoiltex_i_to_o
+
   subroutine mksoiltex(file_mesh_i, file_mapunit_i, file_lookup_i, mesh_o, pioid_o, rc)
     !
     ! make %sand, %clay, organic carbon content, coarse fragments, bulk density,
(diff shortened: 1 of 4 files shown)


samsrabin:
Output surface dataset at 1.9x2.5 resolution as of 9a2759f is bit-for-bit identical with version from ctsm5.2.027. However, my implementation of `organic_o` option 2—which is hard-coded off, as before—produces differences when forced on. I wonder if it would make sense at this point to just remove that code?

slevis-lmwg:
@samsrabin I think it's ok to remove "option 2" at this point. There's record of it if we ever decide to resuscitate it.

Also, is this PR affected by the latest updates that occurred [here?](https://github.com/ESCOMP/CTSM/pull/2500#issuecomment-2334861748)

samsrabin:
Thankfully no, this touches a different part of the code.

slevis-lmwg (approved):
@samsrabin thanks again for this refactor. I like what you've done and thanks for keeping the comments indicating, for example, what -4 means and so on.

I think I looked this over previously and didn't formally review/approve it at the time. Based on our earlier conversations here, I see that you have a few unchecked checkboxes, but I'm happy to approve preemptively.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -318,180 +407,26 @@ subroutine mksoiltex(file_mesh_i, file_mapunit_i, file_lookup_i, mesh_o, pioid_o
...
-             do l = 2,n_scid
-                if (float(sand_i(1,l,lookup_index)) >= 0._r4) then
-                   sand_o(no,1) = float(sand_i(1,l,lookup_index))
-                   exit
-                end if
-             end do
-          end if
-          if (sand_o(no,1) < 0._r4) then
-             if (int(sand_o(no,1)) == -4) then
-                sand_o(no,:) = 99._r4
```
slevis-lmwg, line 336: @samsrabin I haven't looked at this in a long while, but judging from `sand_o + clay_o = 100` when these two start equal to -4, I will agree with your assessment that the flag -4 indicates sand dunes regardless of which variable we look at.

samsrabin:
Will 5.4 have updated datasets that require `mksoiltex`? If so, yes, I can get it done for 5.4 (although I may need reminding). If not, it can be in a later milestone.

ekluzek:
> Will 5.4 have updated datasets that require `mksoiltex`? If so, yes, I can get it done for 5.4 (although I may need reminding). If not, it can be in a later milestone.

Yes, mksoiltex is always used when surface datasets are made (it's just not on the landuse.timeseries datasets). As we are updating both this would be good to bring in. And since this is mostly done it would just be something good to finish off. 

Would you like some help in testing this? I could merge to the latest and just try it out to see what happens with it, if that would help...

samsrabin:
@ekluzek Nah, I'll work on it. Merging in ctsm5.3.036 gave no conflicts, so that's good.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -35,6 +35,81 @@ module mksoiltexMod
...
+       end if
+    end if
+
+    ! Ensure negative values are handled
+    ! TODO: Remove as unnecessary?
+    ! TODO: Also handle organic_o? Not handled in original
+    if (data_o(no,1) < 0._r4 .and. trim(name) /= "sand") then
+       write(6,*)'ERROR: at no, lookup_index = ',no,lookup_index
+       call shr_sys_abort('could not find a value >= 0 for '//name)
+    end if
```
samsrabin, line 99: @slevis-lmwg @ekluzek The last question I have pertains to this "Ensure negative values are handled" block. In the original code, sand and `organic_o` weren't checked for negative values. Should they be?

ekluzek: Checking that data read in is in the range it's expected to be in, is good standard error checking that's worth including. Strange things can happen to input data files and if they have bad values, you want to know as soon as possible. This is something that I think we should always have in place. Otherwise, it's possible you could run with a bad input file, and not know about any problems until far down the road when you've run simulations and done analysis. These are much more expensive type of problems to have, where it silently does the wrong thing.

So I would remove the "Remove as unnecessary?" comment. And also add handling for organic_o.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -514,10 +428,11 @@ subroutine mksoiltex(file_mesh_i, file_mapunit_i, file_lookup_i, mesh_o, pioid_o
           ! (calculated from orgc_i, cfrag_i, and bulk_i) to organic_o. This
```
ekluzek, line 428: Change above line:

``` fortran
! Alternative approach considered but not selected: Regrid organic_i
```

to

```
! ----------------------------------------------------------------------
! organic_o: Alternative approach considered and previously available
! ----------------------------------------------------------------------
```

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -514,10 +428,11 @@ subroutine mksoiltex(file_mesh_i, file_mapunit_i, file_lookup_i, mesh_o, pioid_o
...
           ! approach first calculates organic_i and then regrids to organic_o
           ! rather than regridding all the terms first and then calculating
-          ! organic_o. Commented out code above belongs to that option.
+          ! organic_o. That would be organic_o_option 2, last available in
+          ! commit d5f389a97.
```
ekluzek, line 432:
```suggestion
          ! organic_o. This option was last available in commit d5f389a97.
```

ekluzek (approved):
OK, cool. I think this is a great simplification with refactoring and code-reuse that's really great to bring in. I do have some suggestions on changes (mostly in the language for comments). But, I'm marking as approve, so you don't have to wait on me. And I'm good with you doing what you think is appropriate for my advice.

I think the important thing to include is the error checking for negative values. It's good to have this check in case of a bad or corrupted datafile, so things work as expected and you don't have a silent problem that's only found way down the road.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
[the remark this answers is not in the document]
samsrabin: The whole hash probably isn't necessary, as `git` gives helpful output when trying to do something with an ambiguous hash.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
samsrabin: Resolved with 0e9217875.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
samsrabin: I like the original text more, because it tells the reader what to look for when they check out that commit.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -328,180 +403,19 @@ subroutine mksoiltex(file_mesh_i, file_mapunit_i, file_lookup_i, mesh_o, pioid_o
...
+               "cfrag", 1._r4, 0._r4, float(cfrag_i(:,:,:)), cfrag_o)
+          call mksoiltex_i_to_o(no, lookup_index, n_scid, nlay, n_mapunits, &
+               "bulk", 1.5_r4, 1.5_r4, bulk_i, bulk_o)  ! TODO: 1.5 ok for sand dunes and -7?
+          call mksoiltex_i_to_o(no, lookup_index, n_scid, nlay, n_mapunits, &
+               "phaq", 7._r4, 7._r4, phaq_i, phaq_o)
+          call mksoiltex_i_to_o(no, lookup_index, n_scid, nlay, n_mapunits, &
+               "orgc", 1._r4, 0._r4, orgc_i, orgc_o)
 
           ! ---------------------------------------------------------------
           ! organic_o OPTION 1, as we plan to calculate organic in the CTSM
```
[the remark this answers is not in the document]
samsrabin: Resolved with 0e9217875.

tools/mksurfdata_esmf/src/mksoiltexMod.F90
```diff
@@ -35,6 +35,81 @@ module mksoiltexMod
 contains
 !=================================================================================
 
+  subroutine mksoiltex_i_to_o(no, lookup_index, n_scid, nlay, n_mapunits, name, val_neg_4, val_neg_other, data_i, data_o, data_modifier, organic_o)
+    !
```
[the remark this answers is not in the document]
samsrabin: Resolved with aec3576ab.

samsrabin:
@ekluzek Bill helped me figure out that the `organic_o` diffs were due to optimization changes in its assignment loop when the error check got added. Since they're roundoff-level (for single precision, anyway), I've left that check in. So I think this is done.

Are there any additional tests you think I should perform (in addition to `aux_clm`)? As a reminder, all I've done is check for diffs between this branch and ctsm5.3.036 for `--res 1.9x2.5 --start-year 1850`.

ekluzek:
Awesome. That makes sense on the optimization difference. And yes that's fine.

The last test I suggest is to just use the makefile to have it make all of the standard resolutions. And just make sure all of them were created. So this is just a smoke test. You'll need to do this on scratch in order to have the space for all the files.

You've already done some comparisons of the files, so I see no need to check it for any others. But it would be good to know that the new error checks don't trigger for any of the standard resolutions.

PR #2750, opened 2024-09-10
correct surface data for 5.3-NEON

### LINKED ISSUE #2749: Correct paths for surface datasets, esp NEON
### Brief summary of bug

It seems like with the addition of 5.3 surface datasets the paths to old surface datasets have now broken. 
This should be corrected early in the 5.3 tags.

### General bug information

ctsm5.2.028

no

currently running NEON cases (created with usermods_dirs) that point to the following path
 `/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/surfdata_esmf/NEON/surfdata_1x1_NEON_HARV_hist_2000_78pfts_c240206.nc`

This has now changed to the 5.3 surface data, which has a different creation date:
`/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/surfdata_esmf/NEON/surfdata_1x1_NEON_HARV_hist_2000_78pfts_c240908.nc`

Hopefully this is just a short term issue for people using older tags and usermods_dirs.

### Details of bug

Have not looked to see how other cases may be impacted here, but it's a good way to encourage people to use the 5.3 tag once it's created.

wwieder:
### Specific notes
Fixes #2749,   could be merged on #2500 or added to b4b-dev

### DIFF
--- a/cime_config/usermods_dirs/NEON/defaults/user_nl_clm
+++ b/cime_config/usermods_dirs/NEON/defaults/user_nl_clm
@@ -19,7 +19,7 @@
 !----------------------------------------------------------------------------------
 
 flanduse_timeseries = ' '   ! This isn't needed for a non transient case, but will be once we start using transient compsets
-fsurdat = "$DIN_LOC_ROOT/lnd/clm2/surfdata_esmf/NEON/surfdata_1x1_NEON_${NEONSITE}_hist_2000_78pfts_c240206.nc"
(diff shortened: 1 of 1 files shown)


slevis-lmwg:
I'm sorry about this confusion @wwieder and thank you for bringing to our attention.

@ekluzek and I decided to resolve this as follows:
- [x] I moved the original files back to where they used to be
- [x] I moved the new files to a ctsm5.3.0 subdirectory under the same path

I think this means that you may close this PR as "Won't merge" or similar.

wwieder:
Wont fix, but I seems like we need a fix for the 5.3 tag that points to the right place in usermods?

slevis-lmwg:
True... @ekluzek will make the update here before he merges his PR to #2500:
https://github.com/slevis-lmwg/ctsm/pull/10

Issue #2755, opened 2024-09-12
Choice of pressure in the CLMU building energy model


cathyxinchangli:
### Brief summary of bug

In the CLMU building energy model (BEM), standard pressure is assumed when calculating indoor air density. It may be more realistic and accurate to assume bottom layer of atmosphere pressure, which is used by CLMU to calculate canyon air properties, since indoor and outdoor pressure differences are usually small. 

### General bug information

**CTSM version you are using:** ctsm5.2.001

**Does this bug cause significantly incorrect results in the model's science?** Not much (see below for details)

### Details of bug

I changed the pressure used in the air density calculation in BEM from standard pressure to bottom layer of atmosphere pressure of the corresponding grid cell, and compared results from land-only simulations of the year 2000 before and after the change. The change leads to a small reduction in urban 2m air temperature (up to -0.08 K) and a small increase in urban 2m relative humidity (up to 0.25%) in places of higher altitude. This is likely due to a reduction in urban air-conditioning energy flux (up to -0.9 W/m^2 of urban area) and urban heating energy flux (up to -3 W/m^2 of urban area) caused by reduced indoor air density in those areas. Plots are available here: 
[bugfix_pressure_4vars_diff_global_2000mean_urbanarea.pdf](https://github.com/user-attachments/files/16974093/bugfix_pressure_4vars_diff_global_2000mean_urbanarea.pdf)

### Important details of your setup / configuration so we can reproduce the bug

The code change was done in [this commit](https://github.com/cathyxinchangli/CTSM/commit/9687d03aa5fbe0c0dad046ba4d48c63331f0592d) on my CTSM fork.

Below are details of the simulations I ran to produce the results above:
- compset: 2000_DATM%GSWP3v1_CLM51%SP_SICE_SOCN_MOSART_CISM2%NOEVOLVE_SWAV  
- resolution: f09_g17  
- machine: Derecho

samsrabin:
Bugfix and urban, so should be allowed now that we're in the chill.

olyson:
Thanks for this @cathyxinchangli .  Do you have time to create a branch off of CTSM main and issue a PR?  If not, I'm happy to do it.

cathyxinchangli:
Hi @olyson! I'd be happy to, but it seems my account doesn't have the authorization to do that. However, I can create a PR from the branch on my forked repository. Would that be okay? Or if I am allowed to be given the authorization then I can create a branch then!

olyson:
[...]

PR #2758, opened 2024-09-12
clm5.3.024: Change the choice of pressure in CLMU building energy model (BEM)

cathyxinchangli:
### Description of changes

In the CLMU BEM, the choice of pressure for calculating indoor air density is changed from standard pressure to bottom layer of atmosphere pressure of the corresponding grid cell. 

### Specific notes

**CTSM Issues Fixed (include github issue #):** #2755 

**Are answers expected to change (and if so in what way)?**  
Yes, some urban variables such as urban 2m air temperature, 2m relative humidity, urban air-conditioning and heating energy fluxes will change slightly. See #2755 for details.

**Does this create a need to change or add documentation? Did you do so?**  
I have updated the inline documentation in the code where it's needed. I personally don't think it requires changing or adding documentation elsewhere, but if so, I would be happy to help.

The original code change was done based on ctsm5.2.001, with which I performed one land-only simulation for the year 2000 under DEBUG mode, and the simulation ran as expected. Details are available in #2755. I did not run additional tests after syncing the branch to the latest CTSM main.


src/biogeophys/UrbBuildTempOleson2015Mod.F90
```diff
@@ -310,6 +312,7 @@ subroutine BuildingTemperature (bounds, num_urbanl, filter_urbanl, num_nolakec,
     ctype             => col%itype                         , & ! Input:  [integer (:)]  column type
     zi                => col%zi                            , & ! Input:  [real(r8) (:,:)]  interface level below a "z" level (m)
     z                 => col%z                             , & ! Input:  [real(r8) (:,:)]  layer thickness (m)
+    forc_pbot         => atm2lnd_inst%forc_pbot_not_downscaled_grc, & ! Input: [real(r8) (:)]  atmospheric pressure (Pa)
```
olyson, line 315: I'm thinking it would be good to use the downscaled version of this (forc_pbot_downscaled_grc).  I think the downscaled and not downscaled versions are the same except for glacier columns but it would be good to verify this.  This would make it consistent with other parts of the code that use the downscaled version (e.g., CanopyFluxesMod.F90).

olyson (changes requested):

src/biogeophys/UrbBuildTempOleson2015Mod.F90
cathyxinchangli: Thanks for bringing this up, @olyson! I chose to use `forc_pbot_not_downscaled_grc` because it was used in UrbanFluxesMod.F90 in QSat() function calls. I see now that this is different from CanopyFluxesMod.F90. Would you suggest that I follow CanopyFluxesMod.F90 here, and perhaps update UrbanFluxesMod.F90 to be consistent too?

olyson: Oh, I see, good point.  Let's just use the not_downscaled version then since all of the other forcing quantities in UrbanFluxesMod.F90 are not_downscaled versions also, e.g., forc_t, forc_th, forc_q, etc.

cathyxinchangli:
Thank you for the suggestions, @olyson! I have updated the code. Please let me know if you would like me to run additional tests or help with anything else!

olyson (approved):
Looks good thanks.

samsrabin:
@olyson Does anything other than `aux_clm` testing need to be run for this?

slevis-lmwg:
`./run_sys_tests -s aux_clm -c ctsm5.3.023 -g ctsm5.3.024`
IN PROGRESS on derecho
IN PROGRESS on izumi

slevis-lmwg:
@cathyxinchangli @olyson
A preemptive confirmation question, while the aux_clm tests are in progress:
I see DIFFs coming up in the test-suite `/glade/derecho/scratch/slevis/tests_0210-111015de`
and will assume that they are expected, unless I hear otherwise.

olyson:
> @cathyxinchangli @olyson A preemptive confirmation question, while the aux_clm tests are in progress: I see DIFFs relative to the baseline coming up in the test-suite `/glade/derecho/scratch/slevis/tests_0210-111015de` and will assume that they are expected, unless I hear otherwise.

Yes, we expect DIFFs.

PR #2772, opened 2024-09-18
mesh_plotter improvements

samsrabin:
### Description of changes

- Adds `--dpi` option (default 300)
- Existing output file without `--overwrite` will now actually fail
- More reasonable cell-center marker sizes, with an option to disable them

### Specific notes

**Any User Interface Changes (namelist or namelist defaults changes)?** Yes:
- Adds `--dpi` option (default 300)
- Adds `--no-center-coords` option; useful for when the red X's are uselessly small or way too big

- [x] Options tested manually
- [x] New Python system/unit tests added
- [x] Python unit testing passes
- [x] Python system testing passes


adrifoster (approved):
This looks great to me. Some nice pythonic additions and improvements to code robustness.

Issue #2732, opened 2024-08-28
Tuning of Leung_2024 dust emissions for clm6_0_cam7.0


ekluzek:
@dmleung has tested some F cases with ctsm5.2.019 and found some issues that will need some tuning to give good results. He previously did some tuning for clm5_0_cam6.0 for the ctsm5.2 surface datasets, and will need to do more for cam6_0_cam7.0. This is part of the normal tuning exercise that needs to be done for all of CESM3.0.

@dmleung is working on this now and will get us some changes to bring in as soon as possible. Hopefully, this won't need to be done multiple times as the model should be stabilizing as we move forward. However, as other tuning changes happen in both CAM and CTSM it may need to be done more than once. Mostly this is work that @dmleung will do and ask us to bring in when ready. Code changes for this are likely to be small.

dmleung:
I tried to use the ctsm5.2.019/cam6_3_137 tag to run an FLTHIST case for year 1991 and saw some very high annual mean dust AOD values (>1) over Argentina (Patagonia) and Australia. I tried to look into this issue and understand what were changed to cause these blowups, since we did not see them in older tests. The older test used the ctsm5.2.006/cam6_3_160 tag to run a FHIST case for year 1991.

I am showing the side-by-side comparison of the two cases/tests. See the two CAM history variables below, upper is dust AOD and lower is dust emission (plotted in log scale).

<img width="1478" alt="Screenshot 2024-08-30 at 4 46 52 PM" src="https://github.com/user-attachments/assets/5fb0d748-8482-44df-b7e4-b57c93869458">

-----

Since CTSM dust emission is dependent on meteorological and land-surface fields, I looked into the differences between these fields in the two model tags. My primary conclusion is that the main cause is the enhanced friction velocity (or ustar) over the two countries that made dust emissions explode. See the two CTSM variables below, upper is FV from CTSM and lower is the modified friction velocity used in dust emission module. FV has some visible increases over places like Argentina, Australia, and the US. Increases are also seen over other places like boreal forests, but it is less relevant to dust simulations. Meanwhile, FV has some decreases over northwestern China, causing further low bias of dust over China.

<img width="1476" alt="Screenshot 2024-08-30 at 4 52 29 PM" src="https://github.com/user-attachments/assets/1bc058fd-0276-4bd5-833c-d5bca39ae19a">

-----

[...]

PR #2803, opened 2024-09-30
Tuning Leung_2023 dust emissions for clm6_0_cam7.0

### LINKED ISSUE #2826: Failing warning test in build-namelist tester
### Brief summary of bug

Hillslope with use_init_interp test fails

`1218/3997 < FAIL> <Test Id: 1218> <Desc: hillslope with init_interp>`

### General bug information

**CTSM version you are using:** ctsm5.3.004

**Configurations affected:** Only the build-namelist test results
### Details of bug

The problem is just that the test needs to have hillslope_file added to the test, so it's an easy fix.

dmleung:
### Description of changes

This is a PR for tuning Leung_2023 dust emissions in CESM3. Code changes in this PR were implemented based on a branch using ctsm5.3.002. As described in issue #2732 we used a sandbox based on ctsm5.2.019/cam6_3_137 to run a low-top F case (FLTHIST) with an SE dycore (ne30pg3 grid) and clm6_0_cam_7.0 physics in the 1990s for testing. The surfdata was updated to surfdata_ne30np4.pg3_hist_1850_78pfts_c240908.nc. Tuning was needed since we saw some high biases in dust over semiarid regions given all the updates in CTSM and CAM. The biggest changes in CTSM that affects dust is an increase in friction velocity (ustar; fv in CTSM) over vegetated, semiarid regions, mainly due to a switch in the roughness scheme from 'ZengWang2007' to 'Meier2022'. Since dust emission is very sensitive to ustar, the dust emission scheme magnifies this increase and caused strong high biases over semiarid regions, including Australia and Patagonia (see plots in issue #2732). To enhance the robustness of Leung_2023 and reduce the likelihood to see huge changes in answers in the future, we tried to limit the sensitivity of dust emissions to ustar. We also tried different methods (see specific notes below) to reduce dust emissions from semiarid regions given the high biases over there. The edited .F90 files are:
src/biogeochem/DustEmisLeung2023.F90
src/biogeochem/SoilStateInitTimeConstMod.F90

I only changed a few parameters for now. I hope this is an easy PR. These global parameters could be used for regional dust tuning (see specific notes), so @tilmes suggested to change some of the parameters into namelist variables for easier CESM tuning when CAM people do it. 
This is a draft for CAM people to really start working on their coupled simulations, so we can work on the namelist edits a little later (in this PR or we open a separate PR). @tilmes will start aerosol simulations and CESM3 tests based on this PR. For a global low-top F case one can use a dust_emis_fact somehwere around 2.8 in user_nl_cam for aerosol simulations.

### Specific notes

I document the changes here and the rationale behind for @tilmes and @jfkok's reference
frag_expt_thr = 2.5_r8
We cap the exponent of ustar to 2.5 at most such that dust emission is less sensitive to ustar (dustemis ∝ ustar^4.5 or less). It was previously 5, which means dustemis ∝ ustar^7 or less. The exponent (frag_expt) tends to be large (>5) for high moisture regions and tend to be low for major deserts (<1), so capping frag_expt at a small frag_expt_thr value like 2.5 tends to suppress emissions from semiarid regions.

vai_mbl_thr = 0.5_r8 
This means only patches with patch-level LAI+SAI < 0.5 are allowed to emit dust. Furthermore, the emission flux scales with (1 - tlai_lu(l)/vai_mbl_thr), so using a smaller vai_mbl_thr suppresses dust emissions more from semiarid regions more than from hyperarid regions. vai_mbl_thr was previously set as 1 by me for Leung_2023 (allowing more semiarid dust) and was 0.3 by Natalie Mahowald for Zender_2003 (allowing dust from major deserts only) in most previous CTSM and CLM tags. 

real(r8), parameter :: a = 0.5_r8
ThresholdSoilMoistZender2003 = a * ( 0.17_r8 + 0.14_r8 * (clay*0.01_r8) )
The soil moisture threshold represents a minimum soil moisture above which soil moisture starts to limit dust emissions, as a function of clay content. Charlie Zender suggested that the threshold could be tuned. Given a new surface dataset with new soil moisture, we use a smaller a = 0.5 to suppress dust emissions from semiarid regions. It was previously a = 1.

@ekluzek @tilmes @dlawrenncar 

  First step of work in #2732
  Fixes #2826

Yes. I design the code changes here to suppress dust emissions from semiarid regions to more appropriate levels.

Maybe adding them soon

I think we need documentation for the scheme and these parameter changes. I will discuss this with @ekluzek.

Testing performed, if any:
We did a test using FLTHIST with ne30pg3_ne30pg3_mg17 grid. 
Will also do system testings later.


dmleung:
Also, this is a tuning for CESM to adequately represent the spatial distribution of dust and its interactions with clouds and radiation etc to get some reasonable radiative budgets. The impacts of these tuning changes on temporal variability, including day-to-day variability, seasonality, and interdecadal variability, will require further assessments later.

ekluzek:
@dmleung as you note we'll probably want these tuning parameters on the namelist. Or they could go into the parameter file. There's been a preference for moving them there. I think that's the solution we want in the end.

We can save this version as a branch tag. And then make a b4b-dev tag with values on the parameter file that gives identical results. And then the final version that comes to master would be the one with these tuned settings. But, once on the namelist or parameter file -- i'll be easy to tune them

ekluzek:
Actually now that I read the full introduction I see that @dmleung and I say essentially the same thing. @dmleung just gives the expediency in some form for getting this in right away.

ekluzek (approved):

src/biogeophys/SoilStateInitTimeConstMod.F90
```diff
@@ -765,12 +765,14 @@ real(r8) function ThresholdSoilMoistZender2003( clay )
   !------------------------------------------------------------------------------
       real(r8), intent(IN) :: clay ! Fraction of clay in the soil (%)
 
+      real(r8), parameter :: a = 0.5_r8   ! Tuning factor for the soil moisture effect. This parameter could also be added into ThresholdSoilMoistKok2014. Simone Tilmes suggested to change this parameter into a namelist variable for easier CESM tuning. dmleung 30 Sep 2024.
```
ekluzek, line 768: @dmleung note since this function is used by both Leung and Zender, it changes answers for both.

So I'm going to modify the code slightly so that it will be identical for Zender, by passing in the "a" parameter with it hardcoded to 1 for Zender. Let me know if you have other thoughts on that.

dmleung: Good point Erik, I forgot about this. I could also do it the other way, but it's good that you can also take care of it. At least I will also change this in my next commit.

dmleung:
@ekluzek I made a first attempt to change the part of the code for the tuning parameter for soil moisture effect so that we don't accidentally change answers for Zender's dust emissions. Not sure if that is something you want. I will let you modify that part from now on.

dmleung:
As mentioned in issue #2732, I was working on bringing in a tuning parameter (for now named dust_veg_drag_fact) for scaling the vegetation drag partition effect. The reason is that CTSM has higher vegetation roughness (and hence ustar) after updating the roughness scheme to Meier2022. Since winds should decrease with increasing roughness while Okin's formulation only accounts for vegetation cover/density, we use the factor to account for the enhanced vegetation roughness. We currently suggest a tuning factor of dust_veg_drag_fact of ~0.7–0.8 when using Meier2022. This change directly offsets the ustar increase over sparsely vegetated semiarid regions. So, we no longer need to scale up moisture effect to cancel the increased ustar, which created compensating errors in different variables.

For dust_veg_drag_fact = 0.7, for now we set frag_expt_thr = 3, dust_moist_fact = 0.9, and vai_mbl_thr = 0.6. I use dust_emis_fact = 2.2. See #2732 for more discussions and some plots. I brought in this change in the commit [6c35216](https://github.com/ESCOMP/CTSM/pull/2803/commits/6c352166d3e7a4f0909ba24d8f338a55ecf98622).

ekluzek:
I ran testing on Derecho and Izumi and the only unexpected fails were baseline differences (for clm4_5 and clm5_0) as discussed above, and a unit test fail because of the same. @dmleung put in a change to handle this so I'll update to it and resend the testing.

src/biogeophys/SoilStateInitTimeConstMod.F90
```diff
@@ -770,7 +781,7 @@ real(r8) function ThresholdSoilMoistZender2003( clay )
          call endrun( 'Clay fraction is out of bounds (0 to 100)')
          return
       end if
-      ThresholdSoilMoistZender2003 = 0.17_r8 + 0.14_r8 * clay * 0.01_r8
```
ekluzek, line 773: @dmleung is this change important for something? It looks to me that it shouldn't matter...

dmleung: No it doesn't matter. I just deleted the moisture tuning parameter and tidied up the code by grouping the %clay and the 0.01 inside a bracket.

dmleung:
I am settling at the following parameters:
no tuning for soil moisture effect (dust_moist_fact = 1), dust_veg_drag_fact = 0.7, frag_expt_thr = 2.5, vai_mbl_thr = 0.6. user_nl_cam uses dust_emis_fact = 2.2 for a FLTHIST case.
I think this gives a good balance between 4 things: not having too high dust emissions from vegetated regions, not having a high dust bias over the Sahara, not having really high dust emission exponents (and insane dust fluctuations) over vegetated regions, and not suppressing too much high-latitude dust.
The remaining (longstanding) issue is that we still see an underestimation of dust over NW China, given that the ustar there is too low and soil moisture likely too high (even higher than India). The use of Meier2022 further reduces the roughness and ustar over NW China, worsening the already existing low dust bias. Even though the climatological global mean dust AOD should be 0.03, I don't want to generate high dust biases from other regions to compensate the underestimated Chinese emissions, just to make a globally correct dust level. So, for now I tune a global mean dust AOD of about 0.27, and see how we can fix the Chinese emissions in the future.
It is an option to suppress dust from Australia and Patagonia further, but we want a reasonable amount of global dust and some dust in the S Hemisphere.
Dust from the Middle East, the US, and the high-latitude Canada look pretty good to me.
I will stop the dust tuning for now, and maybe open a new PR in a few months when more CESM3 updates are in place (TSS people will tell me what to do). @ekluzek will merge this PR when ready. I will probably stop tuning dust at this point and move on to other tasks...!

image

Issue #2814, opened 2024-10-07
matrixcnOn testmod shouldn't include default testmod


samsrabin:
### Brief summary of bug

It looks like the `matrixcnOn` (and thus also `matrixcnOn_ignore_warnings`) test have `../default` in their `include_user_mods`. This makes it so that there is WAY too much output being saved. Most of the tests with those testmods are meant to have monthly output, but instead they're getting files saved daily or more frequently.

This is probably causing the "job exceeded walltime" error for test `ERS_Ly5_P128x1.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput--clm-matrixcnOn_ignore_warnings`.

### General bug information

**CTSM version you are using:**  ctsm5.3.002

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** Any tests with the `matrixcnOn` (and thus also `matrixcnOn_ignore_warnings`) testmods.

### Details of bug

I'm going to fix this by deleting `cime_config/testdefs/testmods_dirs/clm/matrixcnOn/include_user_mods`. That will cause a number of tests to change answers, so this will seemingly be a non-b4b change.

PR #2815, opened 2024-10-07
Remove include_user_mods from matrixcnOn testmod

samsrabin:
Should avoid unnecessary levels of history writing.

### Specific notes

**CTSM Issues Fixed:**
- Resolves ESCOMP/CTSM#2814.

**Are answers expected to change (and if so in what way)?** Only for our tests.

**Testing performed, if any:** None yet.


cime_config/testdefs/testlist_clm.xml
```diff
@@ -1333,7 +1333,7 @@
       <machine name="derecho" compiler="intel" category="aux_clm"/>
     </machines>
     <options>
-      <option name="wallclock">01:40:00</option>
+      <option name="wallclock">01:40:01</option>
```
samsrabin, line 1336: In case this comes in after the b4b-dev merge (which it probably will) that changes this test to 2:30:00.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -3332,7 +3332,7 @@
       <option name="tput_tolerance">0.1</option>
     </options>
   </test>
-  <test name="PFS_Ld10_PS" grid="f19_g17" compset="I2000Clm50BgcCrop" testmods="clm/matrixcnOn">
+  <test name="PFS_Ld10_PS" grid="f19_g17" compset="I2000Clm50BgcCrop" testmods="clm/default--clm/matrixcnOn">
```
samsrabin, line 3335: This was the only test with `matrixcnOn*` that didn't have another testmod as well.

slevis-lmwg: Are you adding visibility to what's already happening in `include_user_mods` or is there other significance to introducing "default" explicitly here?

slevis-lmwg: Oh, now I see that you removed "default" from the include_user_mods.

cime_config/testdefs/testmods_dirs/clm/matrixcnOn/include_user_mods
```diff
@@ -1 +0,0 @@
-../default
```
slevis-lmwg, line 1: Remove the file entirely?

cime_config/testdefs/testlist_clm.xml
```diff
@@ -11,6 +11,7 @@
         aux_cime_baselines: CESM cime baselines (subset of aux_clm tests)
         hillslope: Experimental test list used for the hillslope option of the model
         rxcropmaturity: Short tests to be run during development related to prescribed crop calendars
+        matrixcn: Tests exercising the matrix-CN capability
```
slevis-lmwg, line 14: I assume that we prefer the matrixcn tests to still run by default. So, clarification... Does the newly introduced matrixcn test-suite run as part of aux_clm? Or do you envision submitting it separately?

slevis-lmwg (commented):
@samsrabin thank you for this work. I included one or two separate questions to my review, and here is another:
Did I understand correctly that the motivation for this PR is the reduction of history output when running matrixcn tests? If so, I missed which of your changes accomplishes that.

cime_config/testdefs/testlist_clm.xml
samsrabin: This is for separate testing when developing CNmatrix stuff. For example, I'm running it right now to test this PR branch.

cime_config/testdefs/testmods_dirs/clm/matrixcnOn/include_user_mods
samsrabin: It is deleted; the Github view is just a bit confusing.

cime_config/testdefs/testlist_clm.xml
slevis-lmwg: Got it, I should have noticed that they all have aux_clm also listed.

samsrabin:
> @samsrabin Did I understand correctly that the motivation for this PR is the reduction of history output when running matrixcn tests? If so, I missed which of your changes accomplishes that.

Removing the `include_user_mods` file from the test means that the `default` outputs no longer overwrite outputs specified by earlier test mods. For example, in `LCISO_Lm13.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-ciso_monthly--clm-matrixcnOn_ignore_warnings`, it should have been saving monthly (the `ciso_monthly` test mod) but was actually being saved at much higher frequency (because `matrixcnOn_ignore_warnings` was after `ciso_monthly` in the test name). That test goes from 30 to 3 GB after this change.

cime_config/testdefs/testmods_dirs/clm/matrixcnOn/include_user_mods
slevis-lmwg: Yes, I should have also noticed "This file was deleted" before clicking to see the diffs...

slevis-lmwg:
> > @samsrabin Did I understand correctly that the motivation for this PR is the reduction of history output when running matrixcn tests? If so, I missed which of your changes accomplishes that.
> 
> Removing the `include_user_mods` file from the test means that the `default` outputs no longer overwrite outputs specified by earlier test mods. For example, in `LCISO_Lm13.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-ciso_monthly--clm-matrixcnOn_ignore_warnings`, it should have been saving monthly (the `ciso_monthly` test mod) but was actually being saved at much higher frequency (because `matrixcnOn_ignore_warnings` was after `ciso_monthly` in the test name). That test goes from 30 to 3 GB after this change.

Thank you, I had not realized this difference in behavior from moving "default" to the test name.

slevis-lmwg (approved):
Approving.

samsrabin:
> > > @samsrabin Did I understand correctly that the motivation for this PR is the reduction of history output when running matrixcn tests? If so, I missed which of your changes accomplishes that.
> > 
> > 
> > Removing the `include_user_mods` file from the test means that the `default` outputs no longer overwrite outputs specified by earlier test mods. For example, in `LCISO_Lm13.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-ciso_monthly--clm-matrixcnOn_ignore_warnings`, it should have been saving monthly (the `ciso_monthly` test mod) but was actually being saved at much higher frequency (because `matrixcnOn_ignore_warnings` was after `ciso_monthly` in the test name). That test goes from 30 to 3 GB after this change.
> 
> Thank you, I had not realized this difference in behavior from moving "default" to the test name.

Well technically it's not moving `default` to the test name that does it; it's removing the `include_user_mods` file. Every test mod in a test name gets its `user_nl_clm` appended, in order, to create the one used by the test. So if `default`'s `user_nl_clm` comes after `ciso_monthly`'s, then if both test mods set something, `default`'s will be the one used.

PR #2820, opened 2024-10-08
Update ChangeLog and ChangeSum for hillslopes-off-fsurdat (ctsm5.3.004)

samsrabin:
### Description of changes

Forgot to do this in #2434 before merging it. Once this PR is merged, I'll make the ctsm5.3.004 tag.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,57 @@
 ===============================================================
+Tag name: ctsm5.3.004
+Originator(s): samrabin (Sam Rabin)
+Date: Mon Oct  7 21:25:20 MDT 2024
+One-line Summary: Move hillslope data off surface datasets
+
+Purpose and description of changes
+----------------------------------
+
+Moves hillslope data off surface datasets onto its own separate hillslope_file. This makes it so we don't need to generate new surface datasets specifically for hillslope testing whenever the surface datasets are updated.
+
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


olyson (approved):
Looks good to me.

doc/ChangeLog
```diff
@@ -1,4 +1,57 @@
...
+  regular tests (aux_clm: https://github.com/ESCOMP/CTSM/wiki/System-Testing-Guide#pre-merge-system-testing):
+
+    derecho ----- OK
+    izumi ------- OK
+
+
+Answer changes
+--------------
+
+Changes answers relative to baseline: Not in real runs, only our testing.
```
slevis-lmwg, line 44: @samsrabin I know this is merged, but could you elaborate briefly for my benefit?
Code configurations: I assume hillslope active. Other?
Platforms/compilers: I assume ALL.
Nature of change: ?
Definition of "real" runs: ?
Hypothesis for why in testing and not in real runs: ?

slevis-lmwg (commented):
I'm posting this after the fact to ask some questions for my own benefit. Ideally I think these should typically be answered in the ChangeLog whenever there are diffs from the baseline.

doc/ChangeLog
samsrabin: Sorry, yes, only with hillslope active but on both Derecho and Izumi. It's only in our testing because I had to make changes to the test setup for compatibility; this included changing mesh file and hillslope data for some tests/testmods.

PR #2445, opened 2024-03-28
Make separate instantaneous and non-inst. history tapes; for the latter set time = mid of time_bounds

slevis-lmwg:
### Description of changes
I started from issue #1059 and PR #2019 and am restarting here. From closed PR #2019 I preserved a few things:
- If avgflag /= 'I', time_bounds is present and time = mid of time_bounds
- Some long_name changes
- Treating avgflag as a tape (not field) trait for 'I' and 'L' tapes

In this PR we also intend to split all history tapes into instantaneous and non-instantaneous. The CAM group accomplished this in PR https://github.com/ESCOMP/CAM/pull/903 and we plan to maintain consistency in the appearance of CLM and CAM history files.

### Specific notes

Contributors other than yourself, if any:
@ekluzek @billsacks 

CTSM Issues Fixed (include github issue #):
#1059 

Are answers expected to change (and if so in what way)? No

Any User Interface Changes (namelist or namelist defaults changes)?
We intend to split all history tapes into instantaneous and non-instantaneous.

Issue #2542, opened 2024-05-13
Going to a CISM active test for a processor change PEM, causes answers to change...


ekluzek:
### Brief summary of bug

With ctsm5.2.0 we discovered we didn't have enough testing that corresponded to CESM or CAM testing. CESM testing is always done with CISM active, so I changed some tests in #2501 from I1850Clm60BgcCrop to I1850Clm60BgcCropG. However, 

### General bug information

**CTSM version you are using:** ctsm5.2.004-31-ga09d22376

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** With CISM active

### Details of bug

PEM_D_Ld9.ne30pg3_t232.I1850Clm60BgcCrop.derecho_intel.clm-clm60cam6LndTuningMode passes, however
PEM_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam6LndTuningMode fails in the comparison of different processors...

FAIL PEM_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam6LndTuningMode COMPARE_base_modpes

### Important details of your setup / configuration so we can reproduce the bug

In the test list there are PEM and ERP tests for glc* testmods that have a comment that says this

> cism is not answer preserving across processor changes, but short test length should be ok

Those tests range from 5 days to 10 days. But, many are f10, and the highest resolution is f19 which runs 5 days.

ekluzek:
Still fails for 3 days, which is about the shortest I think we should try...

ekluzek:
I talked to @Katetc about this after the CSEG meeting. She also said that the issue is a traditional global-sum issue in MPI which is solved in other places and as such should be relatively easy to fix.

In confirming the timeline on this she sent me an email, which says that they will work on this relatively soon.

samsrabin:
On ctsm5.2.005, I'm getting a failure in the same step for 
```
PEM_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam6LndTuningMode
```
Should this be marked as an expected fail? I see that a slightly different test named
```
PEM_D_Ld3.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam6LndTuningMode
```
is present in the expected fail list (and points to this issue), but that's not actually in the test list.

ekluzek:
[...]

PR #2052, opened 2023-07-11
ctsm5.3.021: Standardize time metadata (we will mark this as the release tag for ctsm5.3)

samsrabin:
### Description of changes
Standardizes a dimension name of output variable `time_bounds` (from `hist_interval` to `nbnd`), as well as attributes for that plus `mcdate`, `mcsec`, `mdcur`, and `mscur`.

### Specific notes

**Contributors other than yourself:** @billsacks, @ekluzek 

**CTSM Issues Fixed:** 
Resolves #1693.
Fixes #2923 
Resolves https://github.com/ESCOMP/MOSART/pull/66
Resolves https://github.com/ESCOMP/RTM/pull/35

**Are answers expected to change (and if so in what way)?** No.

**Any User Interface Changes (namelist or namelist defaults changes)?** No.

#### Testing performed

New metadata in a history file from a short test:
```
netcdf standardize-time-metadata_rtm.clm2.h0.1850-01-01-00000 {
dimensions:
...
    nbnd = 2 ;
...
variables:
...
    int mcdate(time) ;
        ...
        mcdate:calendar = "noleap" ;
    int mcsec(time) ;
        ...
        mcsec:calendar = "noleap" ;
    int mdcur(time) ;
        ...
        mdcur:calendar = "noleap" ;
    int mscur(time) ;
        ...
        mscur:calendar = "noleap" ;
...
    double time_bounds(time, nbnd) ;
        time_bounds:long_name = "time interval endpoints" ;
        time_bounds:units = "days since 1850-01-01 00:00:00" ;
        time_bounds:calendar = "noleap" ;

```

All `aux_clm` tests pass bit-for-bit against `ctsm5.1.dev131`, except for expected failures.

PR #2838, opened 2024-10-18
Change history time to be equal to the middle of the time bounds

slevis-lmwg:
### Description of changes
This PR subsets the scope of issue #1059 and PR #2445 as a result of the October 2024 conversation in #2445.
This PR changes history time to be equal to the middle of the time bounds. 
This PR does not put instantaneous fields on their own separate history files.

I will also bring submodule changes from https://github.com/ESCOMP/MOSART/pull/106 (was https://github.com/ESCOMP/MOSART/pull/69) and https://github.com/ESCOMP/RTM/pull/39.

### Specific notes

No.

Maybe. No.

Testing performed, if any:
Plan to run aux_clm, mosart, rtm test-suites.


slevis-lmwg:
I submitted this manual test to confirm that the committed modifications work as intended:
`./create_test SMS_Lm1.f10_f10_mg37.I1850Clm60BgcCropCmip6waccm.derecho_gnu.clm-basic -c /glade/campaign/cgd/tss/ctsm_baselines/ctsm5.3.009`
Check on Friday.

slevis-lmwg:
The previous test completed its 1 month and the monthly output looked good, but there were annual history files that I could not tell. So I started another test (default is Ly1, but I changed to Ly2) and I added  hist_avgflag_pertape(6) = 'I' to see what happens:
`SMS_Ly2_Mmpi-serial.1x1_brazil.IHistClm60BgcQianRs.derecho_intel.clm-output_bgc_highfreq`

slevis-lmwg:
I updated the submodules to point to https://github.com/ESCOMP/MOSART/pull/106 and https://github.com/ESCOMP/RTM/pull/39 and submitted the three corresponding test-suites:
```
./run_sys_tests -s mosart -c mosart1.1.02-ctsm5.2.029 --skip-generate
./run_sys_tests -s rtm -c rtm1_0_80-ctsm5.2.029 --skip-generate
./run_sys_tests -s aux_clm -c ctsm5.3.009 --skip-generate
```

slevis-lmwg:
@ekluzek and I agreed on the order that the "hist" PRs would get merged. The order as shown in [Upcoming Tags ](https://github.com/orgs/ESCOMP/projects/46/views/1)is 
#2838 
#2084 
#2052
...and we will follow the same order in mosart/rtm. Before I merge the mosart/rtm "hist" PRs, @ekluzek will merge the work in the "simple bfb" mosart/rtm cards.

slevis-lmwg:
@ekluzek review and approval of this PR should take 5 minutes, as it looks the same as the corresponding RTM and MOSART PRs that you reviewed/approved. Thanks :-)

slevis-lmwg:
TODOs left for me:
- Update to ctsm5.3.012
- Update .gitmodules
- Run aux_clm, mosart, rtm
- Merge and make tag

slevis-lmwg:
derecho testing IN PROGRESS
./run_sys_tests -s aux_clm -c ctsm5.3.012 -g ctsm5.3.013
./run_sys_tests -s rtm -c rtm1_0_82-ctsm5.3.009 -g rtm1_0_82-ctsm5.3.013
./run_sys_tests -s mosart -c mosart1.1.04-ctsm5.3.009 -g mosart1.1.04-ctsm5.3.013

izumi testing

slevis-lmwg:
The diffs of the two tests above seem potentially related to the earlier update to mosart1.1.02 (https://github.com/ESCOMP/MOSART/pull/94), but Adrianna's test passed just fine pointing to mosart1.1.02. So I will try the two tests pointing to mosart1.1.02. If that passes, then mosart1.1.03.

slevis-lmwg:
I brainstormed for a bit with @billsacks and Bill pointed out/suggested:
- these two are the only long tests with active cism
- the changes appear in the coupler due to the changes in mosart and not due to changes in ctsm
- that he would not expect these diffs (as I also didn't), so he would recommend going through a methodical way of testing, making baselines, and updating the code and the baselines to confirm whether I still get these diffs

slevis-lmwg:
@billsacks I mentioned to you a vague memory I had of an issue that could relate to these diffs, and it is this one:
#2542

billsacks:
> I mentioned to you a vague memory I had of an issue that could relate to these diffs, and it is this one:
> #2542

Ah, yes. But I don't think that should be the issue with these tests, right?

slevis-lmwg:
Right, I don't think so. I think I have now found that the problem starts with the introduction of mosart1.1.03. I see this in a new test today and in my testing from yesterday (somehow I missed the sign when I looked originally). First I will confirm beyond doubt and then I will try bisecting mosart1.1.03 to find the culprit.

src/main/histFileMod.F90
```diff
@@ -3397,18 +3422,23 @@ subroutine htape_timeconst(t, mode)
...
        call ncd_defvar(nfid(t) , 'mscur' , ncd_int, 1, dim1id , varid, &
-          long_name = 'current seconds of current day')
+          long_name = long_name)
        call ncd_defvar(nfid(t) , 'nstep' , ncd_int, 1, dim1id , varid, &
           long_name = 'time step')
 
        dim2id(1) = hist_interval_dimid;  dim2id(2) = time_dimid
-       call ncd_defvar(nfid(t), 'time_bounds', ncd_double, 2, dim2id, varid, &
-          long_name = 'history time interval endpoints')
+       if (tape(t)%hlist(1)%avgflag /= 'I') then  ! NOT instantaneous fields tape
```
samsrabin, line 3438: Might be prudent to introduce a variable like `is_inst_tape` instead of repeating this conditional everywhere. That way, once you start segregating instantaneous/non variables onto their own tapes, you only have to change it in one place.

slevis-lmwg:
> then I will try bisecting mosart1.1.03 to find the culprit.

Looking at https://github.com/ESCOMP/MOSART/pull/70, I have converged on two commits:
We removed two lines: 77494596d594df6c5843747944abf1612c8d5494
Instead of removing the two lines, we added if-statemt that is .false.: 692d1839489d6979c1a241e0f638189186405cfc

I have confirmed that the two lines that we removed caused the diffs.
@ekluzek I will check with you how to resolve this.

slevis-lmwg:
Trying a case with cism NOT active and the if-statement still commented out (as in my last test)
`./create_test SMS_D.f10_f10_mg37.I2000Clm60Bgc.derecho_intel -c /glade/campaign/cgd/tss/ctsm_baselines/ctsm5.3.012`

src/main/histFileMod.F90
slevis-lmwg: Thank you @samsrabin 
I linked your comment to #2445 and will address it no later than at that time.

slevis-lmwg:
Rerunning this test to generate a baseline, but stuck in the SHAREDLIB_BUILD phase, so I will kill it and try again next week. Besides, I will need to generate a ctsm5.3.014 baseline, so rerunning right now is redundant:
`./create_test RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput -c /glade/campaign/cgd/tss/ctsm_baselines/ctsm5.3.012 -g /glade/campaign/cgd/tss/ctsm_baselines/ctsm5.3.013_slevis_ok`

samsrabin:
I'm wondering if the "treat a file as instantaneous if its first variable is" might be premature to bring in here instead of #2445. Specifically, I think the `time_bounds` variable should for now be saved no matter what. Not having it messed up the RXCROPMATURITY test, and though I was able to work around it, others might not be.

slevis-lmwg:
> [...] Specifically, I think the `time_bounds` variable should for now be saved no matter what. Not having it messed up the RXCROPMATURITY test, and though I was able to work around it, others might not be.

@olyson what do you think about @samsrabin's comment? Most concerning to me would be any vulnerability in the land diagnostic package.

olyson:
I don't think the land diagnostics package uses time_bounds.  But I may not understand the issue here.  Seems like we could do a short simulation using this branch once stable and see if there are any problems?  Also should check ILAMB.  So maybe an I2000 case for a test.

samsrabin:
And actually, this comment goes for the "exact middle" vs. "end of" change as well. It seems arbitrary (and against the "principle of least astonishment") that the first variable in the `hist_fincl` list should affect this. My understanding was that this would be changed to "exact middle" for all history files for now, and we would just accept that being wrong for instantaneous variables.

samsrabin:
A bonus from what I'm proposing: Always (a) including `time_bounds` and (b) setting `time` to the exact middle means that people always have what they need for postprocessing either instantaneous variables (just look at the second value in `time_bounds`) or averaged/etc. variables (either both values in `time_bounds` or the value in `time`).

ekluzek:
time_bounds is an expected part of the CF convention, so I do endorse using it for anything with time. That's likely why some tools might expect it to be there.

For instantaneous it should likely be the time bounds of the time-step that was output. So the previous time-step time first to the ending time-step time for the endpoint. You could have both being the same ending time-step, but that doesn't show that it is a model with a finite time-step.

Here's information on the CF Convention attributes. Look up "bounds"...

https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#attribute-appendix

samsrabin:
@ekluzek That link contains the following text, which implies that instantaneous files should actually _not_ have `time_bounds`:

> It is often the case that data values are not representative of single points in time and/or space, but rather of intervals or multidimensional cells. This convention defines a bounds attribute to specify the extent of intervals or cells.

ekluzek:
Good point @samsrabin. That convinces me we should remove it for I fields then. That would then be in line with the convention.

We don't similarly report on the grid cell bounds either (which could be done for 2D grids, but would be harder for unstructured grids), so we shouldn't for Instantaneous time fields either.

billsacks:
Earlier discussion led to the conclusion that we should *not* have time_bounds in instantaneous files (though CAM's mistakenly still does): https://github.com/ESCOMP/CAM/issues/1166

slevis-lmwg:
I had a quick meeting with @samsrabin 40 minutes ago:
I proposed and he agreed to an alternate version of the if-statement that rightly concerned him. The alternate version eliminates the risk of wrongly labeling a tape "instantaneous" just because the first field is instantaneous.

Oh, good, looks as though we're removing time_bounds from instantaneous tapes, as originally planned :-)

Ok, so I'm testing the alternate if-statement with aux_clm right now and then I will push it to the PR.

src/main/histFileMod.F90
slevis-lmwg: The alternate if-statement [discussed below ](https://github.com/ESCOMP/CTSM/pull/2838#issuecomment-2492286937)resolves this.

In the context of treating instantaneous tapes differently with this PR and #2445, @samsrabin asked whether we loop through all of a tape's fields to identify instantaneous tapes at runtime. The answer is no, and we agreed that #2445 will handle this by separating all instantaneous fields onto separate tapes and, therefore, no need to worry about doing this now.

slevis-lmwg:
@samsrabin back to this test
`/glade/derecho/scratch/slevis/tests_1121-124652de/RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput.GC.1121-124652de_int.gddgen`
I get a new error. The problem and solution are not obvious to me from a quick look, but I'm happy to look together if you want.

slevis-lmwg:
@samsrabin I'm thinking please hold regarding my last post for now, because in my aux_clm tests with the two subsequent PRs, which included this one, the test passed.

slevis-lmwg:
Note to @ekluzek 
- If you're up for approving this one, then I would be ready to move forward with the merge when it's time.
- You have seen it before; it should look very familiar, other than an addition from Sam Rabin pertaining to cropcal_module.py.

Note to self:
Wait for Erik's upcoming mosart/rtm tags before proceeding with this merge.

slevis-lmwg:
Updated to rtm1_0_84 and mosart1.1.06 and started on derecho:
```
PASS ./build-namelist_test.pl
PASS lint and black
PASS ./run_ctsm_py_tests -u and -s
OK ./run_sys_tests -s rtm -c rtm1_0_84-ctsm5.3.016 -g rtm1_0_84-ctsm5.3.018
IN PROGRESS ./run_sys_tests -s aux_clm -c ctsm5.3.017 -g ctsm5.3.018
WAIT ./run_sys_tests -s mosart -c mosart1.1.07-ctsm5.3.017 -g mosart.1.1.07-ctsm5.3.018
```
Mosart test-suite waiting for 07 tag to come in with this PR: https://github.com/ESCOMP/MOSART/pull/112

**Notes**
How I check for aux_clm results in tests_0109-170044de:
```
./cs.status.fails | grep -v '17: DI'
grep NORM */*cprnc.out | grep -v time
```
- The first command confirms that all FAILs are baseline diffs, as expected.
- The second line confirms that all baseline diffs are in the time variable, as expected.

ekluzek (approved):
@slevis-lmwg and I went over this. I had some comments and questions that we worked through. The testing is done so this can be made shortly.

The one suggestion I had was to point out in the ChangeLog that the mosart tag used for the Izumi baseline included the bug-fix that will be made in mosart1.1.07 that will come in the next tag ctsm5.3.019.

Since, testing and baselines had been done, and there are things that have to be done by hand on Izumi we didn't want to have to redo any of the testing.

Issue #2294, opened 2023-12-15
use of mct_mod is deprecated 


jedwards4b:
And must be removed prior to CESM3.
https://github.com/ESCOMP/CTSM/blob/master/lilac/src/lilac_mod.F90#L11

jedwards4b:
Also components/clm/src/unit_test_stubs/csm_share/CMakeLists.txt
components/clm/src/unit_test_stubs/csm_share/mct_mod_stub.F90

ekluzek:
Discussed in meeting, this is important to do. Really want this done before CESM3, and it's in our planning now for that. Right now we don't have someone to work on this.

jedwards4b:
@ekluzek As I understand it it is no longer needed in ctsm, or mosart and we have removed it from csm_share.  If you do not remove it from slim prior to cesm3 then I propose that slim is not released with cesm3.   What other requirement is there for mct?

wwieder:
I think this issue is about LILAC, not SLIM, but maybe I'm wrong.  
Maybe it's worth hearing from @dlawrenncar what the plan for LILAC should be moving forward?

jedwards4b:
@wwieder Thanks for the clarification - it looks like lilac is a single processor application - am I reading that correctly?

billsacks:
> It looks like MCT is being used for CPL7-data-models for LILAC to fill the roll of a DATM for the testing. ESMF is widely used in LILAC, and the CTSM cap for LILAC is an ESMF component, so it shouldn't be too hard to do this, but it's not trivial either. I think when LILAC was developed CDEPS wasn't mature enough to use it yet. Now we just need to switch the use of CPL7/DATM to CDEPS/DATM.

Thanks a lot for your investigation of this @ekluzek . I think this is mostly right, but a minor correction is that I think the use of MCT / data model functionality is to fill in fields that might be missing from the actual atmosphere. Currently this is used for prescribing aerosols. I agree that the right path forward is to change lilac_atmaero.F90 to use CDEPS streams.

(LILAC also has a fake atmosphere driver, lilac/atm_driver/atm_driver.F90, which serves a similar purpose as DATM for testing, though much simpler. But I don't think that relies on MCT: it just uses hard-coded data.)

dlawrenncar:
I am hesitant to keep adding to SE burdens, but having a regional model
capability, coupled to CTSM is an important feature.  That could be through
LILAC or it could be if WRF was coupled into CESM infrastructure ... or it
[...]

PR #2853, opened 2024-10-30
Update submodules to ones that are based on cesm3_0_alpha04a

### LINKED ISSUE #2375: Stop building esmf_wrf_timemgr for unit tests; instead use the actual ESMF library
In making the changes for #2374 , I noticed that the unit test build still builds the esmf_wrf_timemgr to satisfy the ESMF dependency of clm_time_manager.F90. At some point we should probably stop doing that and instead have the unit tests link against the actual ESMF library.

See
https://github.com/ESCOMP/CTSM/blob/c030c2342dac53d30ed419aca262db8fa7213ee6/src/CMakeLists.txt#L21

and some related code in that file.

I briefly started down this path but couldn't see how to get the value of ESMFMKFILE for this machine/compiler in the CMake build, so I gave up for now.

Note that we could probably also remove this block of code: I assume that was put in before we needed to include ESMF stuff in the unit test build, but now we do because of the unit tests of clm_time_manager:
https://github.com/ESCOMP/CTSM/blob/c030c2342dac53d30ed419aca262db8fa7213ee6/src/CMakeLists.txt#L57-L73

### LINKED ISSUE #2640: Update externals to cesm3_0_alpha02a versions
Update externals to those in cesm3_0_alpha02a for: cime, ccs_config, cmeps, share, mizuRoute, and cdeps

Right now this is:

cime6.1.0
ccs_config_cesm1.0.0
share1.1.2
cmeps1.0.2
cdeps1.0.43
cesm-coupling.n02_v2.1.3 (mizuroute)

#2590 updates cism, rtm, and mosart in addition to these (and updates cmeps and ccs_config as well). Between the two that should give us submodules for cesm3_0_alpha02a (and cesm3_0_beta02, as the other planned updates are only for CAM and MOM).

This will go on the b4b-dev branch.

### LINKED ISSUE #2654: Failing tests for izumi_nag with mpi-serial in cesm3_0_alpha02a
### Brief summary of bug

@fischer-ncar found the following two tests failing in cesm3_0_alpha02a (which uses ctsm5.2.009 with ccs_config_cesm1.0.0 and cime6.1.0)

ERS_D_Ld5_Mmpi-serial.1x1_vancouverCAN.I1PtClm50SpRs.izumi_nag.clm-CLM1PTStartDate
SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA

### General bug information

**CTSM version you are using:** ctsm5.2.009

**Configurations affected:** mpi-serial with nag on Izumi

### Details of bug

cesm3_0_alpha02a uses ctsm5.2.009 with:
ccs_config_cesm1.0.0
MPIserial_2.5.0
cime6.1.0

Note, these two tests pass in the ctsm5.2.009 baselines and tests since then

I also replicated the result that Chris saw in cesm3_0_alpha02a

mpi-serial.bld.log:

```
gmake --output-sync -f /scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/Tools/Makefile  -C /scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h/nag/mpi-serial/debug/nothreads/mpi-serial  CIME_MODEL=cesm  SMP=FALSE CASEROOT="/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h" CASETOOLS="/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/Tools" CIMEROOT="/scratch/cluster/erik/cesm3_0_alpha02a/cime" SRCROOT="/scratch/cluster/erik/cesm3_0_alpha02a" COMP_INTERFACE="nuopc" COMPILER="nag" DEBUG="TRUE" EXEROOT="/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/bld" RUNDIR="/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/run" INCROOT="/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/bld/lib/include" LIBROOT="/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/bld/lib" MACH="izumi" MPILIB="mpi-serial" NINST_VALUE="c1a1l1" OS="LINUX" PIO_VERSION=2 SHAREDLIBROOT="/scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h" BUILD_THREADED="FALSE" USE_ESMF_LIB="TRUE" USE_MOAB="FALSE" COMP_ATM="datm" COMP_ICE="sice" COMP_GLC="sglc" COMP_LND="clm" COMP_OCN="socn" COMP_ROF="srof" COMP_WAV="swav" USE_TRILINOS="FALSE" USE_ALBANY="FALSE" USE_PETSC="FALSE"  COMP_NAME=mpi-serial /scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h/nag/mpi-serial/debug/nothreads/mpi-serial/Makefile.conf
gmake: Entering directory '/scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h/nag/mpi-serial/debug/nothreads/mpi-serial'
SHAREDLIBROOT |/scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h| SHAREDPATH |nag/mpi-serial/debug/nothreads|
/scratch/cluster/erik/SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA.20240717_091501_ir4w0h/Tools/Makefile:635: recipe for target '/scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h/nag/mpi-serial/debug/nothreads/mpi-serial/Makefile.conf' failed
gmake: Leaving directory '/scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h/nag/mpi-serial/debug/nothreads/mpi-serial'
/bin/sh: line 1: /scratch/cluster/erik/cesm3_0_alpha02a/libraries/mpi-serial/configure: No such file or directory
gmake: *** [/scratch/cluster/erik/sharedlibroot.20240717_091501_ir4w0h/nag/mpi-serial/debug/nothreads/mpi-serial/Makefile.conf] Error 127
ERROR: /bin/sh: line 1: /scratch/cluster/erik/cesm3_0_alpha02a/libraries/mpi-serial/configure: No such file or directory
```

### LINKED ISSUE #2871: LIMITATION: ERP tests no longer support the "month" length option
### Brief summary of bug

From @fischer-ncar 

> Hi Erik,
> 
> Jim made some changes to the testing scripts which caused ERP tests to not support _Lm anymore.  This is a low priority, but when you get the time, can you change
> 
> ERP_P64x2_Lm36.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-clm50cropIrrigMonth_interp
> to
> ERP_P64x2_Ly3.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-clm50cropIrrigMonth_interp
> 
> Thanks
> Chris

### General bug information

**CTSM version you are using:** cesm3_0_beta04

**Configurations affected:** ERP tests

### Details of bug

Does this affect other ER tests as well?

ekluzek:
### Description of changes

Update the submodules to something close to cesm3_0_alpha04a. I needed to update cime and ccs_config beyond to get the PF Unit testing working.

### Specific notes

Remove mct from submodules
Add mpi-serial to submodules
Update the PF unit testing to use the full ESMF library (which will enable wider testing), this also required bringing in NetCDF and PIO libraries which we shouldn't actively use but may allow us to do fewer stub modules for I/O.

  Fixes #2640
  Fixes #2375
  Fixes #2654
  Fixes #2871
  Finishes resolving #2294

  I'm actually not sure yet, I think possibly compsets with active CISM might

Testing performed, if any: will do regular and ctsm_sci
   So far done PF UNIT testing and testing of two simple cases
   I haven't tested this for LILIC and I wonder if it will fail


ekluzek:
@billsacks and @jedwards4b could you review this for the cmake changes I made for the PF unit testing? I learned more about cmake as a result of getting this to work, but I'd like to have it reviewed by the two of you with more knowledge/skill in using cmake.

jedwards4b (approved):
I would remove the lines in CMakeLists.txt that you have commented out, but otherwise LGTM.

src/CMakeLists.txt
```diff
@@ -3,22 +3,25 @@ cmake_minimum_required(VERSION 2.8)
...
+
+project(clm_tests Fortran C)
 
 include(CIME_utils)
 
 set(CLM_ROOT "..")
 
-# This definition is needed to avoid having ESMF depend on mpi
-add_definitions(-DHIDE_MPI)
+#find_package(ESMF 8.2.0 REQUIRED )
```
billsacks, line 16: I see that this is commented out. How do you get away without needing ESMF, now that we no longer have the esmf wrf timemgr?

src/CMakeLists.txt
```diff
@@ -53,43 +56,41 @@ foreach (sourcefile ${share_sources})
...
-  string(REGEX MATCH "shr_cal_mod.F90" match_found ${sourcefile})
-  if(match_found)
-    list(REMOVE_ITEM share_sources ${sourcefile})
-  endif()
-endforeach()
+if (DEFINED PIO)
+  set(PIO_PATH ${PIO})
+else()
+  set(PIO_PATH $ENV{PIO})
+endif()
```
billsacks, line 63: Here and some related stuff below: Do we truly need PIO? I thought we had been taking pains to avoid dependence on PIO in the unit test build.

At a glance it looks like this relies on having PIO pre-installed on the system, as opposed to building it from source which I believe is still an option in the CESM build.

billsacks (commented):
Thanks for getting this working @ekluzek! A couple of questions here....

ekluzek:
OK, my suspicion about LILAC was correct, running the LILAC test I get a fail:

```
    Case dir: /glade/work/erik/ctsm_worktrees/external_updates/cime/scripts/LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac.20241030_152038_a3kydm
    Errors were:
        Building test for LILACSMOKE in directory /glade/work/erik/ctsm_worktrees/external_updates/cime/scripts/LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac.20241030_152038_a3kydm
          File "/glade/work/erik/ctsm_worktrees/external_updates/cime/scripts/LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac.20241030_152038_a3kydm/./case.build", line 267, in <module>
            _main_func(__doc__)
          File "/glade/work/erik/ctsm_worktrees/external_updates/cime/scripts/LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac.20241030_152038_a3kydm/./case.build", line 226, in _main_func
          File "/glade/work/erik/ctsm_worktrees/external_updates/cime/CIME/utils.py", line 2272, in find_system_test
            mod = import_module(path)
          File "/glade/u/apps/derecho/23.09/opt/._view/yazo4iwystz7p2hxu5ukzrw3xa24ksen/lib/python3.10/importlib/__init__.py", line 126, in import_module
            return _bootstrap._gcd_import(name[level:], package, level)
          File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
          File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
          File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
          File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
          File "<frozen importlib._bootstrap_external>", line 883, in exec_module
          File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
          File "/glade/work/erik/ctsm_worktrees/external_updates/cime_config/SystemTests/lilacsmoke.py", line 28, in <module>
            from CIME.utils import run_cmd, run_cmd_no_fail, symlink_force, new_lid, safe_copy, append_testlog
        ImportError: cannot import name 'append_testlog' from 'CIME.utils' (/glade/work/erik/ctsm_worktrees/external_updates/cime/CIME/utils.py)

Waiting for tests to finish
FAIL LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac (phase SHAREDLIB_BUILD)
    Case dir: /glade/work/erik/ctsm_worktrees/external_updates/cime/scripts/LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac.20241030_152038_a3kydm
Due to presence of batch system, create_test will exit before tests are complete.
To force create_test to wait for full completion, use --wait
test-scheduler took 7.439677476882935 seconds
```

src/CMakeLists.txt
ekluzek: Good question. ESMF comes in by including the share code which adds this there. The problem I had was that having this explicitly here conflicted with the

`
add_library(ESMF)`

in the share code cmake. So I needed to remove it here.

I'll add a comment about this to point this out...

billsacks: Sounds good - thanks for clarifying. Yes, a comment about that would be great - thanks!

src/CMakeLists.txt
ekluzek: The need for PIO and NETCDF is because ESMF requires it. It wouldn't link without having PIO and NetCDF libraries available. It's maybe possible to use a different ESMF version that doesn't have I/O -- but that would mean having a different ESMF than the main CESM build uses. And I don't think we want that.

You are right that we don't actually want I/O in our PF unit tests (and I confirmed this with the group last week). But, not having to do stubs for modules that have I/O might actually make it easier to add new PF unit tests.

So I think this is OK, and maybe a benefit. Does that sound acceptable to you?

ekluzek: Resolving this.

src/CMakeLists.txt
ekluzek: Added some comments, so resolving.

ekluzek:
I've fixed the LILAC problem between a minor update, and something I need to add to cime:

https://github.com/ESMCI/cime/issues/4703

Now, I'm wondering what will happen with run_neon?

src/CMakeLists.txt
billsacks: Thanks for the explanation about the need for PIO. This makes sense now.

My main concern is that this will make it somewhat harder for people to build the unit tests on their own systems. Now they will need to maintain PIO as a separately-built library rather than letting it be built as part of the CESM build process. But maybe that's the direction things are (or should be) moving anyway, so I guess this is okay. My understanding is that currently you have a choice: you can either (1) Build PIO on your system, then build ESMF in a way that uses this external PIO, then build CESM in a way that links against the same external PIO; or (2) Do *not* bother building PIO on your system, then build ESMF in a way that uses its own internal version of PIO (which is something close to the latest PIO release), then build CESM in a way that uses its own internal version of PIO. (2) is simpler and is what I've done up until now on my local system, but can theoretically cause problems if the PIO versions in ESMF and CESM are incompatible in some way.

ESMF provides a way to query the relevant build options, including libraries that should be included in the link like pio, via ESMFMKFILE, but I'm not sure if there's a way to easily ingest this into a CMake script.

That's all to say: I don't love that it will now be a little more complex for people to maintain their own systems for the sake of unit testing, and especially that supported ways of building the full CESM aren't supported for unit testing (which, if anything, should be simpler rather than more complex). But the reasons for this make sense, and arguably we shouldn't be supporting the use of internal PIO in the ESMF and CESM builds anyway, so I'm okay with this.

ekluzek: That makes sense @billsacks. I did tell @adrifoster that she'll need to build ESMF, PIO, and NetCDF on her system. She already has ESMF and NetCDF, but will need to make sure PIO is also in the mix now. Although now that you mention it -- she must have PIO somewhere -- so maybe was using PIO that's part of ESMF? In her case she also wanted to get regular CESM cases to work, and there you do need all those things anyway. So it's only a problem if you want to do unit-testing easily -- but do NOT need to run CESM cases. But, you are right it's good to reduce complexity when it isn't necessary.

So you are right it isn't great, but seems like the reasonable way forward right now. Thanks for your comments and review.

billsacks: @ekluzek and @adrifoster - yes, I think @adrifoster probably had things set up similarly to me, which hasn't required any separate / pre-built PIO, instead using the default internal PIO both for ESMF and for CESM.

@ekluzek , you say, "In her case she also wanted to get regular CESM cases to work, and there you do need all those things anyway." -- but I think that's not actually true, because I think it's still possible to use the internal PIO for regular CESM cases, whereas starting with this PR that won't be possible with the unit testing.

Again, I think that's probably okay, since it seems like the way we're moving anyway, but it does add one more bit of complexity for setting up and maintaining a system for CTSM development.

One possible way around this would be to make all of this PIO addition conditional: If PIO isn't defined in your machine setup, I think it's safe to assume that you're using internal PIO in both ESMF and CESM on your system, and you should be able to bypass the PIO stuff you added here - at least until we actually need PIO in the CTSM unit tests for their own sake, beyond just ESMF. I'm not sure it's worth doing that, but I'm raising it as an option, e.g., if @adrifoster finds it's challenging to set up or maintain the PIO build.

ekluzek:
It looks like mpi-serial works up to the following set of submodules:

ccs_config_cesm1.0.0
cime6.0.249
share1.0.19
cmeps0.14.17
(Note, cdeps and pio are the latest versions used in this PR, not the previous ctsm5.3.009 ones)

Beyond that it starts breaking. The set I should use together to get working are the cesm3_0_alpha02a externals:

ccs_config_cesm1.0.0
cime6.1.0
share1.1.2
cmeps1.0.2

cesm3_0_alpha03b has the next set of important updates for ccs_config and cime. The next set of updates are the recent ones that were important in getting PFunit tests to work.


NOTE: The top set is still showing changes to answers on Izumi, so the change in answers must be between the ctsm5.3.009 externals and the ones listed above.

jedwards4b:
Hi Erik,

I found that the configure script was not included in the mpi-serial distribution, I have added it in https://github.com/ESMCI/mpi-serial/pull/30.   I think that this will allow you to update to the latest tags.

ekluzek:
Excellent, thanks @jedwards4b!

Yeah, that is one of the things I saw, I thought it might be generated as part of the build process. But, obviously not. Thanks for figuring that out. I"ll try with that branch and make sure it works.

ekluzek:
I'm rerunning all of the tests that failed with the new branch. But, a few of them have built and pass now with the mpi-serial update. So it's looking good so far.

src/CMakeLists.txt
billsacks: I wanted to add a bit more information about linking ESMF related to the earlier discussion about the current hard-coding of linking with PIO: Based on some discussion with Dan Rosen on the ESMF team, it sounds like it should be possible to get ESMF's required link libraries if you use ESMF's provided FindESMF.cmake: that provides INTERFACE_LINK_LIBRARIES, which I think are the extra link flags (paths, libs and options) needed when linking ESMF. (And if that doesn't provide what's needed, then everything in the esmf.mk file is available in CMake after calling FindESMF.cmake, but currently that requires some parsing / manipulation of the settings there.) Generally you would use a CMake Find module like this to actually find the package; in this case we already know where to find ESMF, so it's a little funny to use the Find module, but you can still use it to get various variables like this.

I think that, in the interest of moving things along, it's fine to stick with what you have for now, but this is a note as something that could clean this up and make it more machine-portable long-term.

ekluzek:
The mpi-serial update gets more mpi-serial tests to pass the following 17:

ERS_D_Ld7_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropRs.izumi_intel.clm-decStart1851_noinitial			
ERS_Lm20_Mmpi-serial.1x1_smallvilleIA.I1850Clm50BgcCrop.izumi_gnu.clm-cropMonthlyNoinitial			
ERS_Lm40_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.izumi_gnu.clm-cropMonthlyNoinitial			
ERS_Lm54_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.izumi_intel.clm-cropMonthlyNoinitial			
ERS_Ly20_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.izumi_intel.clm-cropMonthlyNoinitial			
ERS_Ly20_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.izumi_intel.clm-cropMonthlyNoinitial--clm-matrixcnOn			
ERS_Ly3_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianRs.izumi_gnu.clm-cropMonthOutput			
ERS_Ly5_Mmpi-serial.1x1_smallvilleIA.I1850Clm50BgcCrop.izumi_gnu.clm-ciso_monthly			
ERS_Ly5_Mmpi-serial.1x1_smallvilleIA.I1850Clm50BgcCrop.izumi_gnu.clm-ciso_monthly--clm-matrixcnOn			
ERS_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianRs.izumi_intel.clm-cropMonthOutput			
ERS_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianRs.izumi_intel.clm-cropMonthOutput--clm-matrixcnOn_ignore_warnings			
SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_gnu.clm-ptsRLA			
SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_gnu.clm-ptsROA			
SMS_D_Ly6_Mmpi-serial.1x1_smallvilleIA.IHistClm45BgcCropQianRs.izumi_intel.clm-cropMonthOutput			
SMS_Ld5_Mmpi-serial.1x1_brazil.IHistClm60Bgc.izumi_gnu.clm-mimics			
SMS_Ly3_Mmpi-serial.1x1_numaIA.I2000Clm50BgcDvCropQianRs.izumi_gnu.clm-ignor_warn_cropMonthOutputColdStart			
SMS_Ly5_Mmpi-serial.1x1_brazil.IHistClm50BgcQianRs.izumi_intel.clm-newton_krylov_spinup	

While the following 8 are failing:

ERS_D_Ld5_Mmpi-serial.1x1_vancouverCAN.I1PtClm50SpRs.izumi_nag.clm-CLM1PTStartDate	(NLCOMP RUN)		
ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesRs.izumi_nag.clm-FatesCold	(NLCOMP RUN)		
ERS_Lm13.f10_f10_mg37.I1850Clm60Bgc.izumi_intel.clm-monthly_matrixcn_fast_spinup	(NLCOMP RUN)		
SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_nag.clm-ptsRLA	(NLCOMP RUN)		
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60FatesRs.izumi_nag.clm-FatesCold	(NLCOMP RUN)		
SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60SpRs.izumi_nag.clm-default--clm-NEON-TOOL	(NLCOMP RUN)		
SMS_Ly5_Mmpi-serial.1x1_smallvilleIA.IHistClm60BgcCropQianRs.izumi_gnu.clm-gregorian_cropMonthOutput	(NLCOMP MODEL_BUILD)		
SSPMATRIXCN_Ly5_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.izumi_intel.clm-ciso_monthly	(SHAREDLIB_BUILD NLCOMP SUBMIT)	

There's both DEBUG and production tests in both lists, as well as iintel and gnu compilers in both as well as both combinations of compiler and DEBUG or production for both of those as well. The nag compiler fails consistently though, and we only do nag with DEBUG on.

The fails have to do with using the shared NetCDF library as so...

`/scratch/cluster/erik/tests_ctsm539erikb4bacl/SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60SpRs.izumi_nag.clm-default--clm-NEON-TOOL.GC.ctsm539erikb4bacl_nag/bld/cesm.exe: error while loading shared libraries: libnetcdf.so.13: cannot open shared object file: No such file or directory`

ekluzek:
Jim provided an update to mpi-serial that fixes some of this. We also found that Izumi has some nodes with different versions of autoconf so I've opened a ticket to straighten that out. There's also a change to ccs_config that needs to come in as well.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -1524,13 +1524,13 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
-  <test name="ERP_P64x2_Lm25" grid="f10_f10_mg37" compset="I2000Clm60BgcCrop" testmods="clm/monthly">
+  <test name="ERP_P64x2_Ly36" grid="f10_f10_mg37" compset="I2000Clm60BgcCrop" testmods="clm/monthly">
```
ekluzek, line 1527: Need to move this back to previous length, as it increases wallclock time for the test and it runs into the limit even with increasing it by an hour.

So will use _Ld762

jedwards4b: Why did you change it to LY36 instead of Ly3???

ekluzek: Oh, no wonder is was taking forever! Good catch.

The 36 was unintentional, I must have been thinking in terms of months in 3 years...

billsacks: @ekluzek - I haven't looked at these specific tests in a while, but keep in mind that some tests were deliberately set up to restart mid-year. So it's important to consider whether you want _Lm36 or _Ly3, since they give different behavior in terms of when they restart. (Maybe you already have considered this.)

ekluzek:
OK, good news. The aux_clm testing passes as expected on both Izumi and Derecho, and the ctsm_sci test on Derecho similarily.

cime_config/testdefs/testlist_clm.xml
ekluzek: @billsacks thanks for that point. I kept some of the ER tests to the same length, so they'll be fine. But, you have a good point about not restarting on the year boundary, since that's already the standard default. I might adjust a couple in regard to that.

ekluzek: OK, there's five tests this applies to:

``` diff
-  <test name="ERS_Lm54_Mmpi-serial" grid="1x1_numaIA" compset="I2000Clm50BgcCrop" testmods="clm/ciso_monthly_matrixcn_spinup">
+  <test name="ERS_Ly4_Mmpi-serial" grid="1x1_numaIA" compset="I2000Clm50BgcCrop" testmods="clm/ciso_monthly_matrixcn_spinup">
-  <test name="ERP_P64x2_Lm36" grid="f10_f10_mg37" compset="I2000Clm50BgcCrop" testmods="clm/clm50cropIrrigMonth_interp">
+  <test name="ERP_P64x2_Ly3" grid="f10_f10_mg37" compset="I2000Clm50BgcCrop" testmods="clm/clm50cropIrrigMonth_interp">
-  <test name="ERS_P128x1_Lm25" grid="f10_f10_mg37" compset="I2000Clm60Fates" testmods="clm/FatesColdNoComp">
+  <test name="ERS_P128x1_Ly3" grid="f10_f10_mg37" compset="I2000Clm60Fates" testmods="clm/FatesColdNoComp">
-  <test name="ERS_Lm25" grid="1x1_smallvilleIA" compset="IHistClm50BgcCropQianRs" testmods="clm/smallville_dynlakes_monthly">
+  <test name="ERS_Ly3" grid="1x1_smallvilleIA" compset="IHistClm50BgcCropQianRs" testmods="clm/smallville_dynlakes_monthly">
-  <test name="ERS_Lm25" grid="1x1_smallvilleIA" compset="IHistClm50BgcCropQianRs" testmods="clm/smallville_dynurban_monthly">
+  <test name="ERS_Ly3" grid="1x1_smallvilleIA" compset="IHistClm50BgcCropQianRs" testmods="clm/smallville_dynurban_monthly">
```

So I'll change _Lm54 to _Ld1640, Lm25 to Ld761 (762 for 2000), and Lm36 to Ld1096

PR #2870, opened 2024-11-08
Fix handling of finidat with cold starts

### LINKED ISSUE #2856: FATES will never start from specified finidat
### Brief summary of bug

When trying to run I2000Clm60FatesSpCruRsGs specifying `finidat=/path/to/finidat.nc`.

### General bug information

`namelist_defaults_overall.xml` sets `clm_start_type` to `cold`:
```perl=
<clm_start_type                use_fates=".true.">cold</clm_start_type>
```
which means that `CLMBuildNamelist.pm` in any fates run that is not branch or hybrid, you will hit this warning:
```perl
  my $var = "finidat";
  my $finidat = $nl->get_value($var);
  $nl_flags->{'excess_ice_on_finidat'} = "unknown";
  if ( $nl_flags->{'clm_start_type'} =~ /cold/ ) {
    if (defined $finidat ) {
      $log->warning("setting $var (either explicitly in your user_nl_clm or by doing a hybrid or branch RUN_TYPE)\n is incomptable with using a cold start" .
              " (by setting CLM_FORCE_COLDSTART=on)." );
      $log->warning("Overridding input $var file with one specifying that this is a cold start from arbitrary initial conditions." );
      my $group = $definition->get_group_name($var);
      $nl->set_variable_value($group, $var, "' '" );
    }
...
```

That makes it impossible to run not only that compset, but the coupled model with CLM[^_]*FATES ( you want it f.e. when you don't want to spinup the model every time you have added new things to the atmosphere)

[ctsm5.3.009](https://github.com/ESCOMP/CTSM/releases/tag/ctsm5.3.009)

Yes, if you pass -ignore-warnings to CLMBuildNamelist.pm, you will get acutally coldstart instead of finidat. And you will not see the warning for that, since when `-ignore-warnings` is not set, you don't see the second warning at all.

Any Fates configuration.

### Details of bug

Might be related to https://github.com/NGEET/fates/issues/1234

`finidat='/not/an/empty/string'`

[Fill this in with anything relevant that you haven't already noted; if there is nothing to add, delete this section.]

samsrabin:
### Description of changes
This PR changes things so that:
1. Cold-start FATES runs with `finidat` specified will not require `-ignore_warnings`, and the supplied `finidat` will actually be obeyed.
2. Cold-start non-FATES runs will not be allowed, even with `-ignore_warnings`.

### Specific notes

**Contributors other than yourself, if any:** @mvdebolskiy 

**CTSM Issues Fixed:**
- Resolves #2856

**Are answers expected to change (and if so in what way)?** Yes: In cold-start FATES runs where the user was supplying `finidat`, that setting will actually be obeyed.

**Testing performed, if any:** Added checks to `build-namelist_test.pl`, which now pass.


samsrabin:
@ekluzek I've converted this to a draft while I figure out what the deal is with those extra commits up top.

samsrabin:
> Should this come to master though or b4b-dev?

@ekluzek I lean toward master, because although it doesn't change answers in our tests, it will change answers for anyone like Matvey who had tried to do cold-start FATES runs with `finidat` by setting `-ignore_warnings`. This is because their `finidat` setting will now actually be respected, rather than being overwritten to `' '`.

ekluzek:
Ahhh, good point master would give it more visibility. And since this makes a standard thing impossible for FATES it's good to give it more visibility and that it's fixed.

You will need to make a full ChangeLog and tag for it. But, I see your point where this warrants that even though it's small. I added it to upcoming tags and you might as well do this after b4b-dev comes in. This also brings it in sooner than the 2 week timeline for b4b-dev right now.

PR #2875, opened 2024-11-11
run_sys_tests: Print test list in --verbose/--debug

samsrabin:
### Description of changes
See title.

### Specific notes

**CTSM Issues Fixed:** None

**Testing performed, if any:** Tried it with both `-v` and `--debug`; looks good.
```shell
./run_sys_tests -s fire --skip-compare --skip-generate -v --dry-run
```

```
...
INFO: Tests:
INFO:    SMS_D_Ld65.f10_f10_mg37.I2000Clm45BgcCropQianRs.derecho_intel.clm-FireLi2014Qian
INFO:    SMS_D_Ld65.f10_f10_mg37.I2000Clm50BgcCru.derecho_gnu.clm-FireLi2016Cru
INFO:    SMS_D_Ld65.f10_f10_mg37.I2000Clm60BgcCrop.derecho_intel.clm-FireLi2024CruJra
...
```

- [x] Run Python testing.

### DIFF
--- a/python/ctsm/crop_calendars/cropcal_figs_module.py
+++ b/python/ctsm/crop_calendars/cropcal_figs_module.py
@@ -1,6 +1,7 @@
 """
 Functions for making crop calendar figures
 """
+# pylint: disable=abstract-class-instantiated
 
 import numpy as np
 
--- a/python/ctsm/crop_calendars/generate_gdds_functions.py
+++ b/python/ctsm/crop_calendars/generate_gdds_functions.py
@@ -1,7 +1,7 @@
 """
 Functions to support generate_gdds.py
 """
-# pylint: disable=too-many-lines,too-many-statements
+# pylint: disable=too-many-lines,too-many-statements,abstract-class-instantiated
 import warnings
 import os
 import glob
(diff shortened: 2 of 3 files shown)


samsrabin:
Could improve by skipping the new `for` loop if logging level isn't sufficient to print anything.

ekluzek (commented):
Hmmm, this is interesting. In a sense this is redundant with cime/scripts/query_testlists which just prints testlist for the options you ask it for.

However, I think I would find this useful as it's nice to be able to just work in run_sys_tests. The main problem I see with doing this here is duplicating something that's handled elsewhere -- but at the same time this is also small. And it's just telling you for the options you have given what tests will be run. So I'm thinking it would be useful in practice.

ekluzek (approved):
@adrifoster thinks this would be good to have. And in practice I know I'll be using it. Longer term it makes me want to work on CESM testing in the wider CESM community, but that shouldn't get in the way of doing something that would be helpful now, especially this as something that is simple.

samsrabin:
Thanks, @ekluzek!

PR #2877, opened 2024-11-12
Fix cleanup in build-namelist_test.pl

### LINKED ISSUE #2876: build-namelist_test.pl fails to clean up
### Brief summary of bug

The cleanup at the end of `build-namelist_test.pl` fails with the following message:
```
Successfully ran all testing for build-namelist

Cleanup files created
/bin/rm: cannot remove 'temp_file.txt': No such file or directory
```

This can probably be avoided by adding the `-f` flag to all calls of `rm` during cleanup.

### General bug information

**CTSM version you are using:** ctsm5.3.010

samsrabin:
### Description of changes
This PR is intended to fix the failure of `build-namelist_test.pl` to clean up after itself.

Unfortunately, at the moment it doesn't.

### Specific notes

CTSM Issues Fixed:
- Will resolve #2876

**Testing performed, if any:** Running `build-namelist_test.pl`.

### DIFF
--- a/bld/unit_testers/build-namelist_test.pl
+++ b/bld/unit_testers/build-namelist_test.pl
@@ -412,7 +412,7 @@ foreach my $site ( "ABBY", "BLAN", "CPER", "DEJU", "GRSM", "HEAL", "KONA", "LENO
          }
       }
    }
-   system( "/bin/rm $namelistfile" );
+   system( "/bin/rm -f $namelistfile" );
    &cleanup();
 }
 print "\n===============================================================================\n";
@@ -466,7 +466,7 @@ foreach my $site (
    if ( defined($opts{'generate'}) ) {
       $cfiles->copyfiles( "$options", $mode );
    }
-   system( "/bin/rm $namelistfile" );
+   system( "/bin/rm -f $namelistfile" );
    &cleanup();
 }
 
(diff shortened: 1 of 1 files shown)


samsrabin:
I think the issue has to do with all these calls
```perl
eval{ system( "$bldnml -envxml_dir . $options > $tempfile 2>&1 " ); };
```
which are set up with this at the top:
```perl
my $bldnml = "../build-namelist -verbose -csmdata $inputdata_rootdir -configuration clm -structure standard -glc_nec 10 -no-note";
if ( $opts{'test'} ) {
   $bldnml .= " -test";
}

my $tempfile = "temp_file.txt";
```

Somehow all the options are getting put into the name of the tempfile?

samsrabin:
The files are created here, in the `copyfiles` subroutine:
https://github.com/ESCOMP/CTSM/blob/931b4250d4e9ad9cf5ae3c703e62a1dab0897763/bld/unit_testers/NMLTest/CompFiles.pm#L134-L136

wwieder:
Simple update here, but ultimately we want to get rid of the .pl, as this is something that's harder than @samsrabin appreciated here.  Will need @ekluzek to fix this as a quality of life improvement.

PR #2893, opened 2024-11-25
Merge b4bdev 20241125

### LINKED ISSUE #17: albgrd and albgri history fields depend on decomposition, for urban points
**Bill Sacks < sacks@ucar.edu > - 2013-08-31 07:26:34 -0600**
**Bugzilla Id:** [1806](http://bugz.cgd.ucar.edu/show_bug.cgi?id=1806)
**Bugzilla CC:**  muszala@ucar.edu,  oleson@ucar.edu,  rfisher@ucar.edu, 

I believe that this problem only affects history fields, and doesn't actually affect the operation of the model. This is a restatement of the initial bug in bug 1310 (most of the comments in that bug report really relate to a different bug).

For urban points, albgrd and albgri depend on the decomposition - either the number of tasks or the number of threads per task.

I believe that what is going on is the following (copied from bug 1310; I haven't checked carefully whether the behavior has changed slightly from this): 

(1) In UrbanMod.F90: UrbanAlbedo: A count is made of the number of urban
landunits with coszen > 0 (num_solar); note that this count is just of the
number of landunits that this processor is responsible for; thus, this is where
the # PE-dependence comes in, I think.

(2) Later in that subroutine, a bunch of calculations are done if num_solar > 0
-- i.e., if this PE is responsible for at least one urban landunit with coszen
> 0. Note that many of these calculations are done for all landunits, even ones
for which coszen = 0. This introduces the possibility for different results
depending on the decomposition.

(3) The particular difference that I am seeing is in albgrd & albgri. These are
initialized to 0 at the start of the subroutine, and so remain 0 on any PE for
which num_solar = 0. However, for PEs with num_solar > 0, landunits that have
coszen = 0 end up getting albgrd = albgri = 1. This is because the calculation
of albgrd & albgri depends on the values of the sref_* variables, which are
initialized to 1 (and stay at 1 for any landunit for which coszen = 0).

I have confirmed this by comparing the following from clm4_5_20: SMS_Lm1_P180x1.f19_g16.ICLM45BGC.yellowstone_intel vs SMS_Lm1_P360x1.f19_g16.ICLM45BGC.yellowstone_intel - both with ALBGRD, ALBGRI, ALBD and ALBI added to history output.

One thing I am confused about is whether this problem occurs from albgrd and albgri, but not for albd and albi - since the latter seem to just be copies of the former in UrbanMod.

### LINKED ISSUE #747: In UrbanAlbedo: Create a filter for coszen>0 and operate over that
UrbanAlbedo currently calculates a variable `num_solar`, and then there is a big block of code that is only done if `num_solar > 0`: 
https://github.com/ESCOMP/ctsm/blob/2632be6af65ad7d1c0d23ae501c76b3cf68d0d6c/src/biogeophys/UrbanAlbedoMod.F90#L264-L429

This logic causes a dependence of some diagnostic fields on decomposition (#17). Moreover, while the current logic avoids doing any work if num_solar == 0, it is less efficient than it could be in the (probably common) case where num_solar > 0 but < num_urban.

So for the sake of both fixing this dependence on decomposition and performance, we should introduce a filter in this routine (which is also passed down to the routines it calls) that holds just the points with coszen>0. Then we should operate over that filter, and remove the `num_solar > 0` conditional.

Blocks #17 .

ekluzek:
### Description of changes

- Update git-fleximod
- Fix equation number in tech note
- Implement urbanl and urbanc coszen filters
- Make ALBGRD and ALBGRI active by default
- Some adjustments to control pylint for the crop calendar python code
- Run_sys_tests: Print test list in --verbose/--debug

### Specific notes

Contributors other than yourself, if any: @jedwards4b @olyson @adrifoster @seleneonowe @samsrabin 

  Fixes #17 albgrd and albgri history fields depend on decomposition, for urban points
  Fixes #747 In UrbanAlbedo: Create a filter for coszen>0 and operate over that

- New options in run_sys_tests
- ALBGRD/ALBGRI now on h0 history file by default

Testing performed, if any: will run regular testing
  So far ran PFUNIT tests, build-namelist tests, and python tests, running aux_clm on Derecho


slevis-lmwg (approved):
Approving to unblock Erik.

ekluzek:
Testing on Derecho and Izumi is passing as expected. I'm still seeing the mpi-serial nag tests failing at runtime as we did before, even though I had seen it working for cases yesterday. So I'm moving forward with the tag.

PR #2898, opened 2024-12-04
change the order of passing the "use_fates_luh" flag

rgknox:
### Description of changes

The flag that specifies using LUH datasets is passed earlier in the call sequence so that it can be used to guide memory allocations.  If landuse is turned on, particularly in fates nocomp mode, then the space required for patch allocations changes.  Therefore this flag has to be passed earlier, before these allocations are decided on.

### Specific notes

This PR does not need to be synchronized with, but does need to precede: https://github.com/NGEET/fates/pull/1226

Are answers expected to change (and if so in what way)? none, b4b

Does this create a need to change or add documentation? Did you do so? no need

Testing performed, if any:

TBD, but some tests with the FATES test suite so far

### DIFF
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -305,6 +305,7 @@ module CLMFatesInterfaceMod
      integer                                        :: pass_nocomp
      integer                                        :: pass_sp
      integer                                        :: pass_masterproc
+     integer                                        :: pass_luh
      logical                                        :: verbose_output
      type(fates_param_reader_ctsm_impl)             :: var_reader
      
@@ -348,6 +349,14 @@ module CLMFatesInterfaceMod
         end if
         call set_fates_ctrlparms('masterproc',ival=pass_masterproc)
 
+        ! FATES landuse modes
+        if(use_fates_luh) then
+           pass_luh = 1
+        else
+           pass_luh = 0
+        end if
+        call set_fates_ctrlparms('use_luh2',ival=pass_luh)
+        
      end if
 
 
@@ -540,16 +549,12 @@ module CLMFatesInterfaceMod
 
         ! FATES landuse modes
         if(use_fates_luh) then
-           pass_use_luh = 1
            pass_num_luh_states = num_landuse_state_vars
            pass_num_luh_transitions = num_landuse_transition_vars
         else
-           pass_use_luh = 0
            pass_num_luh_states = 0
            pass_num_luh_transitions = 0
         end if
-
-        call set_fates_ctrlparms('use_luh2',ival=pass_use_luh)
         call set_fates_ctrlparms('num_luh2_states',ival=pass_num_luh_states)
         call set_fates_ctrlparms('num_luh2_transitions',ival=pass_num_luh_transitions)
 
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -301,11 +301,11 @@ module CLMFatesInterfaceMod
      integer,intent(in)                             :: surf_numpft
      integer,intent(in)                             :: surf_numcft
      integer,intent(out)                            :: maxsoil_patches
-     integer                                        :: pass_biogeog
-     integer                                        :: pass_nocomp
-     integer                                        :: pass_sp
+     integer                                        :: pass_use_fixed_biogeog
+     integer                                        :: pass_use_nocomp
+     integer                                        :: pass_use_sp
      integer                                        :: pass_masterproc
-     integer                                        :: pass_luh
+     integer                                        :: pass_use_luh2
      logical                                        :: verbose_output
      type(fates_param_reader_ctsm_impl)             :: var_reader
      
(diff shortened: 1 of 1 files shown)


src/utils/clmfates_interfaceMod.F90
```diff
@@ -305,6 +305,7 @@ subroutine CLMFatesGlobals1(surf_numpft,surf_numcft,maxsoil_patches)
      integer                                        :: pass_nocomp
      integer                                        :: pass_sp
      integer                                        :: pass_masterproc
+     integer                                        :: pass_luh
```
samsrabin, line 308: Could the existing `pass_use_luh` not be used for consistency in terms of naming (i.e., with `pass_use_potentialveg`)? If not, `pass_use_luh` seems to not be used anymore, so it should be deleted.

samsrabin (changes requested):

src/utils/clmfates_interfaceMod.F90
rgknox: I previously removed the "use" just to be less  verbose, nothing more. I'm fine with adding it back in.
@ekluzek, the pass_use_potentialveg is a separate flag from pass_use_luh, I think @samsrabin was pointing it out as an example of convention. 
I'm happy to expand this PR though if you would like to update the name of that flag to something less "overlappy" with native ctsm.  @ckoven, could you weigh in here on potential alternative names for [use_fates_potentialveg](https://github.com/ESCOMP/CTSM/blob/master/src/main/controlMod.F90#L241).

samsrabin:
If tests reveal this to be b4b, then it should probably be merged into `b4b-dev` instead of `master`, no?

ekluzek:
Absolutely merge to b4b if it is so. It's just that FATES updates almost always bring in answer changes. I don't remember the last one that was...

rgknox:
I just discussed this PR with @ckoven.  There is a good chance we do not need this PR anymore, so don't sink any time into it. I will close when I can confirm it is unnecessary.

rgknox:
Please disregard my last message.  This PR does have useful content.

samsrabin (approved):

src/utils/clmfates_interfaceMod.F90
ckoven: Hi All -- My preference is to use "Potential Veg" or similar, rather than "Spinup". The reason is that potential vegetation is a more accurate description of what the relevant code does. Spinup is more of a workflow concept; a user might use potential vegetation apart from spinup, and might do spinup without using potential vegetation mode. It also might make it sound to a user like potential veg has something to do with AD spinup, which it doesn't.  In terms of the conceptual overlap of how the phrase is used in non-FATES CTSM, I think that it makes sense that both vegetation models might use the same vocabulary for something that is conceptually similar?  I guess I don't see why that is a problem. It seems sort of analogous to how both models have stomatal conductance switches that share names. But I guess another possibility would be to call it something like "All Primary Land Mode"? But I feel like that is more verbose.

rgknox: I'm fine with sticking with use_fates_potentialveg

ekluzek: Based on what Charlie says above I'd prefer moving to something like say...

use_fates_all_primary_land

just 4 characters longer.

The usage is different enough to be sure to distinguish them. For Medlyn stomatal conductance the FATES implementation is slightly different than CTSM-BGC, but it using the same paper for the scientific reference. In. this case I'd say it's much different. In FATES you could use a potential-veg surface dataset for example and then there becomes the confusion of if you HAVE to do that for this mode in FATES, and what the impact of each is on simulations. I suppose the potential vegetation surface dataset is similar in that it's constructed without human disturbance, so it's all primary land. For CTSM we also currently ONLY use the potential vegetation surface dataset for a paleo setup turning off human influences. So I tend to think of the potential vegetation surface dataset as the Paleo dataset. And the FATES all primary land mode is doing something very distinct from a paleo mode.

ckoven: Hi @ekluzek and all, apologies, but I've sort of lost track of what the discussion here is really about, I feel like the `potentialveg` terminology question is somewhat decoupled from the specific line of code under discussion here or even this PR? Is the proposal to change every instance of the phrase `potentialveg` across FATES, CTSM, and ELM code? Or is it more limited, and if so, what exactly is it?

ekluzek: Yes, let's move this discussion outside of this PR. And let this PR come in pretty much as is. Then if we decide on a terminology change it can be done later and won't hold this one up.

So I'll open a discussion about this. In terms of scope I'm thinking mainly FATES. But, the FATES use of it in CTSM would also need to change. I'm not really thinking about ELM, but maybe it would for consistency. And if the phrase is used extensively the change might not be worth it. I was thinking it's use in FATES is new and as such figured it might be limited.

samsrabin:
Not doing this on b4b-dev because FATES team wants it in CTSM master sooner than today + 2 weeks.

glemieux:
Preliminary regression testing on `derecho` underway.

glemieux:
Initial `aux_clm` regression tests against `ctsm5.3.020` should B4B results, with the exception of the following that I think are due to other issues:

```
    FAIL RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput RUN time=1357
    FAIL SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.derecho_nvhpc.clm-crop RUN time=30
```
results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr2898-aux_clm`

Running prelim fates tests next.

PR #2869, opened 2024-11-07
Update temperature cost function for symbiotic Nfix in FUN

wwieder:
### Description of changes
Updates temperature cost function of Nfix using lab measurements from [Bytnerowicz et al. (2022)](https://doi.org/10.1038/s41477-021-01090-x), see Supplementary Table 3.  

### Specific notes
Replaces the Houlton temperature cost function that's currently used in CNFUN.
- requires updated parameter file with min, optimum, and max temperatures for tropical and extra-tropical PFTs. `/glade/work/wwieder/modified_inputs/ctsm60_params_nfix.c241101.nc`

Updated temperature function (below in blue) compared to the default  temperature function from Houlton et al (orange) to calculate C costs of N fix in FUN (second figure).

image
image

**Contributors other than yourself, if any:**
Tom Bytnerowicz, @tbytnero, t.bytnerowicz@ufl.edu

**CTSM Issues Fixed (include github issue #):**
None

**Are answers expected to change (and if so in what way)?**
Reduces temperate N fix  and vegetation C pools.  Also modifies temporal changes over historical and SSP scenarios.

Diagnostics from CRU-JRA forced runs at f19 [posted here](https://webext.cgd.ucar.edu/I20TR/ctsm530_f19_nfix_hist2/lnd/ctsm530_f19_nfix_hist2_2004_2023-ctsm530_f19_PPE_hist_2004_2023/setsIndex.html)

Issue on [LMWG-dev linked here](https://github.com/NCAR/LMWG_dev/issues/72)
Discussion on Nfix started on [LMWG-dev](https://github.com/NCAR/LMWG_dev/discussions/73)
 
**Any User Interface Changes (namelist or namelist defaults changes)?**
TODO add namelist option for Bytnerowicz vs. Houlton Nfix cost function

**Does this create a need to change or add documentation? Did you do so?**
Needs to be added to documentation, not currently done

**Testing performed, if any:**
Ran scientific tests in point simulations (CLM5.1 code base) and global 2 degree spinup, historical and SSP2-4.5 case.

PR #2917, opened 2024-12-20
Merge b4bdev 20241220

slevis-lmwg:
For details, see the ChangeLog for tag ctsm5.3.017.

For reference, this resolves #2869.


slevis-lmwg:
I saw Joseph's message that izumi had been restored to service, so I submitted aux_clm this morning. It seemed as though tests were going to work, but a large number "failed to initialize" in the SHAREDLIB_BUILD phase. I tried resubmitting three of these, and I get "Read-only file system" errors.

I will try again on Monday. If I see the same behavior, I will return to the original plan of merging this PR without izumi testing.

src/main/clm_varctl.F90
```diff
@@ -227,6 +227,8 @@ module clm_varctl
   !
   real(r8), public :: nfix_timeconst = -1.2345_r8 
 
+  character(len=25), public :: nfix_method  ! choice of nfix parameterization
```
ekluzek, line 230: Since, this is only needed as part of CNFUNMod.F90, it should be a local module variable there, with a ReadNML method that's in there as well. This reduces unneeded dependencies,, makes modules more independent, more interoperable, and easier to extract or add in.

src/biogeochem/CNFUNMod.F90
```diff
@@ -1608,6 +1619,44 @@ real(r8) function fun_cost_fix(fixer,a_fix,b_fix,c_fix,big_cost,crootfr,s_fix, t
...
+! Function result.
+!--------------------------------------------------------------------------
+! real(r8) , intent(out) :: cost_of_n   !!! cost of fixing N (kgC/kgN)
+!--------------------------------------------------------------------------
+  integer,  intent(in) :: fixer     ! flag indicating if plant is a fixer
+                                    ! 1=yes, otherwise no.
+  real(r8), intent(in) :: nfix_tmin ! As in Bytnerowicz et al. (2022)
+  real(r8), intent(in) :: nfix_topt ! As in Bytnerowicz et al. (2022)
+  real(r8), intent(in) :: nfix_tmax ! As in Bytnerowicz et al. (2022)
+  real(r8), intent(in) :: big_cost  ! an arbitrary large cost (gC/gN)
```
ekluzek, line 1643: There's a couple details here that I wonder about, that probably don't matter much. I'm not sure why big_cost is passed in if it's fixed? And there is a 1._r8 that doesn't seem to be needed in the question below. Also -1*s_fix should just be -s_fix. The multiply by the integer -1 doesn't add anything, and just adds complexity.

src/biogeochem/CNFUNMod.F90
```diff
@@ -1608,6 +1619,44 @@ real(r8) function fun_cost_fix(fixer,a_fix,b_fix,c_fix,big_cost,crootfr,s_fix, t
   end if    ! ends up with the fixer or non-fixer decision
   
   end function fun_cost_fix
+
+
+!=========================================================================================
+  real(r8) function fun_cost_fix_Bytnerowicz_noAcc(fixer,nfix_tmin,nfix_topt,nfix_tmax,big_cost,crootfr,s_fix, tc_soisno)
```
ekluzek, line 1625: I love that this is nicely isolated in it's own clean subroutine. This is also something that could have unit or functional testing around it.

src/biogeochem/CNFUNMod.F90
```diff
@@ -1041,42 +1044,50 @@ subroutine CNFUN(bounds,num_soilc, filter_soilc,num_soilp&
...
+
                if(pftcon%c3psn(patch%itype(p)).eq.1)then
                  fixer=1
                else
                  fixer=0
                endif
-               costNit(j,icostFix)     = fun_cost_fix(fixer,a_fix(ivt(p)),b_fix(ivt(p))&
-               ,c_fix(ivt(p)) ,big_cost,crootfr(p,j),s_fix(ivt(p)),tc_soisno(c,j))
+
+               select case (nfix_method)
```
ekluzek, line 1065: I like the namelist control from the user perspective, but am a bit concerned about the complexity this adds at run time to a pretty deeply nested loop over both patch and soil level. A string comparison is much bigger deal than an integer comparison for example.

One way to improe on this tath we also do elsewhere is to have the namelist a string just as this is, but to convert it to an integer control flag at initialization for use at runtime.

The even better way would be to have fun_cost_fix be a small OOP object with a base class for either chosen at initialization and then at runtime it just does the correct one without an if statement. And the resulting code looks cleaner as well.

slevis-lmwg:
I am getting new errors on izumi today, rather than the error that I reported yesterday. I have submitted aux_clm again to see what happens. Meanwhile Ithough I will move on with this merge, because I get the sense that izumi is unreliable.

src/biogeochem/CNFUNMod.F90
```diff
@@ -495,9 +494,10 @@ subroutine CNFUN(bounds,num_soilc, filter_soilc,num_soilp&
   !  fixers, 2 for non fixers. This will become redundant with the
   !   'fixer' parameter if it works. 
   
+  character(len=32) :: subname = 'CNFUN'
```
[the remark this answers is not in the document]
slevis-lmwg: Good point, thanks.
Done.

src/main/pftconMod.F90
```diff
@@ -268,6 +268,9 @@ module pftconMod
      real(r8), allocatable :: b_fix         (:)   ! A BNF parameter
      real(r8), allocatable :: c_fix         (:)   ! A BNF parameter
      real(r8), allocatable :: s_fix         (:)   ! A BNF parameter
+     real(r8), allocatable :: nfix_tmin     (:)   ! A BNF parameter
```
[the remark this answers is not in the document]
slevis-lmwg: Done

src/main/clm_varctl.F90
slevis-lmwg: Ok, done.

src/biogeochem/CNFUNMod.F90
slevis-lmwg: I'm simplifying the equation by removing the two 1s that, I agree, seem redundant. Unless they represent some constant or coefficient that could later assume a different value. If it turns out that way, we can revisit later.

slevis-lmwg: I will not change how big_cost is handled for now, but I think I see that it could be set as a parameter at the module level and not need passing.

src/biogeochem/CNFUNMod.F90
slevis-lmwg: True. I will assume that testing is outside the scope of the current PR.

src/biogeochem/CNFUNMod.F90
slevis-lmwg: I definitely prefer the cleaner approach. I think of it as a refactor of what we have now, so I will consider it outside the scope of this PR, so that we can proceed with the merge.

slevis-lmwg:
Submitting aux_clm after the code revisions.

slevis-lmwg:
aux_clm
- on derecho OK same results as before
- on izumi NOT OK but this computer seems to have serious problems; e.g. all 13 gnu tests:
```
    FAIL ERS_Ld1200_Mmpi-serial.1x1_numaIA.I2000Clm50BgcCropQianRs.izumi_gnu.clm-cropMonthlyNoinitial GENERATE exception
    FAIL ERS_Ld600_Mmpi-serial.1x1_smallvilleIA.I1850Clm50BgcCrop.izumi_gnu.clm-cropMonthlyNoinitial GENERATE exception
    FAIL ERS_Ly3_Mmpi-serial.1x1_smallvilleIA.IHistClm50BgcCropQianRs.izumi_gnu.clm-cropMonthOutput GENERATE exception
    FAIL ERS_Ly5_Mmpi-serial.1x1_smallvilleIA.I1850Clm50BgcCrop.izumi_gnu.clm-ciso_monthly--clm-matrixcnOn GENERATE exception
    FAIL ERS_Ly5_Mmpi-serial.1x1_smallvilleIA.I1850Clm50BgcCrop.izumi_gnu.clm-ciso_monthly GENERATE exception
    FAIL SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.izumi_gnu.clm-crop GENERATE exception
    FAIL SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_gnu.clm-ptsRLA GENERATE exception
    FAIL SMS_D_Ld1_Mmpi-serial.f45_f45_mg37.I2000Clm50SpRs.izumi_gnu.clm-ptsROA GENERATE exception
    FAIL SMS.f10_f10_mg37.I2000Clm50BgcCrop.izumi_gnu.clm-crop GENERATE exception
    FAIL SMS_Ld5_Mmpi-serial.1x1_brazil.IHistClm60Bgc.izumi_gnu.clm-mimics GENERATE exception
    FAIL SMS_Ln9.f10_f10_mg37.I1850Clm45Bgc.izumi_gnu.clm-clm45cam4LndTuningModeZDustSoilErod GENERATE exception
    FAIL SMS_Ly3_Mmpi-serial.1x1_numaIA.I2000Clm50BgcDvCropQianRs.izumi_gnu.clm-ignor_warn_cropMonthOutputColdStart GENERATE exception
    FAIL SMS_Ly5_Mmpi-serial.1x1_smallvilleIA.IHistClm60BgcCropQianRs.izumi_gnu.clm-gregorian_cropMonthOutput GENERATE exception
```
Similarly with most nag and intel tests, plus a few tests say "SHAREDLIB_BUILD failed to initialize"

adrifoster (approved):

PR #2927, opened 2025-01-08
Reduce log noise

### LINKED ISSUE #2717: cleanup cesm log
In file https://github.com/ESCOMP/CTSM/blob/master/src/soilbiogeochem/CNSoilMatrixMod.F90#L67
output to the log should be restricted to the root task.  We don't need to know this from every task:

 CN Soil matrix solution is off

### LINKED ISSUE #2736: Reduce log message noise
As discussed at [CSEG meeting 2024-06-18](https://docs.google.com/document/d/186U6-dt_wWZZGU9NzYQ5zNlMnpx9XX6oweuTXzQY-oo/edit#heading=h.gjot13e7ggkb), reducing unnecessary messaging is a CSEG goal for CESM3.

Things to consider for each message:
- Is it needed at all?
- Can it be changed to only show if the XML `DEBUG` setting is true? (By wrapping in `#ifndef NDEBUG`/`#endif`)
- If it's currently in cesm.log, should it be moved to lnd.log?
- If every process is printing it, can it be set so that just the primary process does?

Messages to consider:
- [x] `clm: leaving fates model` (lnd.log, **every process**)—especially strange considering there's no "entering" message.  _Get rid of this_
- [x] `CN Soil matrix solution is off` (**cesm.log**, **every process**) _get rid of this_ (also filed at #2717)
- ~~`pes participating in computation for CLM` and table (lnd.log),~~ _maybe we keep this?_ **yes**
- [x] `Init2: calling FireInit` (**cesm.log**, **every process**)

samsrabin:
### Description of changes

Deletes a few annoying and unnecessary messages. Changes others to appear only on masterproc.

### Specific notes

- Resolves #2717
- Resolves #2736

- [x] `aux_clm` passes


ekluzek (approved):
Excellent log reducing changes to bring in. Thanks for doing this.

Issue #2932, opened 2025-01-13
'Meier 2022' z0 parameterization causes errors with FATES (especially in SP mode.) 


rosiealice:
### Brief summary of bug
The implementation of the Meier 2022 paameterization for the z0param_method causes errors in FATES that are especially large in SP mode.

### General bug information

**CTSM version you are using:** [output of `git describe`]
ctsm5.3.018

**Does this bug cause significantly incorrect results in the model's science?** Yes

**Configurations affected:** [Fill this in if known.]
FATES, FATES-SP

### Details of bug
The Meier 2022  implementation of the z0 param method, specifically this line
https://github.com/ESCOMP/CTSM/blob/fb89f3359052e4fbba4ff1ccc89e82416a6e544d/src/biogeophys/CanopyFluxesMod.F90#L901

overwrites the _DISPLA_PATCH_ values, that are previously calculated as a FATES variable. These values are wrong in FATES-SP mode, because the code uses the '_HTOP_PATCH_' variable to recalculate '_DISPLA_' , but in FATES-SP mode, this _HTOP_ value is the _driver_ variable for SP mode and is not an _output_ of FATES. Instead, we should be using _HTOP_HIST_PATCH_ for this calculation, which contains the output from FATES assigned the the correct PFT patches. 

There are also several instances where _itype_ is used to define a patch property, which cannot be used when FATES is on. 

https://github.com/ESCOMP/CTSM/blob/fb89f3359052e4fbba4ff1ccc89e82416a6e544d/src/biogeophys/CanopyFluxesMod.F90#L906

https://github.com/ESCOMP/CTSM/blob/fb89f3359052e4fbba4ff1ccc89e82416a6e544d/src/biogeophys/CanopyFluxesMod.F90#L903

This means that the displacement height in FATES-SP, for example, is way to low in the tropics. 

![image](https://github.com/user-attachments/assets/a248f0ad-9833-4b8d-a28e-10a6415b0fd5)

and this seems to lead to a larger boundary layer resistance. 

![image](https://github.com/user-attachments/assets/9bcacf76-0096-47b8-bbc0-75a5010a9a2e)

A solution is to use the alternative parameterization (ZengWang2007). I am not familiar with the motivation for including Meier2022 in the first place though. Should we reimplement it in FATES?  @swensosc @RonnyMeier can you advise? 

I currently suspect that since whenever this is was made the default it will have affected all CLM-FATES simulations, including @adrifoster's calibration runs.  

[...]

Issue #2930, opened 2025-01-10
RXCROPMATURITYSKIPGEN test can fail for file length of fsurdat beyond 255 


ekluzek:
### Brief summary of bug

RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput

fails because the fsurdat file it creates for the case is too long. The filename it creates include the testname which is long, a filename which is fairly long, the test-id, and then the path to scratch. This failed for me with a testid that was using the temporary branch tag name which is longer than a standard CTSM tag name.

One way to remove this possibility is to increase the character length for files. We are using 255 right now because there was a limitation in the past. And currently we have the code with lots of len=256 for filenames.

### General bug information

**CTSM version you are using:** branch_tags/tmp-241219.n01.ctsm5.3.016-23-g1224d97a7

**Does this bug cause significantly incorrect results in the model's science?** no
**Configurations affected:** Tests with too long of a filename for fsurdfat

Fails in the build-namelist step

### Important output or errors that show the problem

Dies in build-namelist with:

```
2025-01-10 10:23:17: ERROR: Command /glade/work/erik/ctsm_worktrees/newbranch/bld/build-namelist failed rc=255
out=
err=ERROR : CLM build-namelist::CLMBuildNamelist::process_namelist_infile() : Invalid namelist variable in '-infile' /glade/derecho/scratch/erik/tests_tmp241219n1ctsm5316erikb4bacl/RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput.GC.tmp241219n1ctsm5316erikb4bacl_int.gddgen/Buildconf/clmconf/namelist.
 ERROR: in validate_variable_value (package Build::Namelist): Variable name fsurdat has a string element that is too long: '/glade/derecho/scratch/erik/tests_tmp241219n1ctsm5316erikb4bacl/RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput.GC.tmp241219n1ctsm5316erikb4bacl_int.gddgen/surfdata_10x15_hist_1850_78pfts_c240908.all_crops_everywhere.nc'
```

ekluzek:
We've brought in changes that allow filename lengths to be 512, so we shouldn't see this anymore. So closing.

[...]

Issue #1733, opened 2022-04-30
`nvhpc` compiler tests are failing on cheyenne/derecho


ekluzek:
### Brief summary of bug

MPI tests with DEBUG on are failing at runtime with the nvhpc compiler on cheyenne.
This continues in ctsm5.1.dev155-38-g5c8f17b1a (derecho1 branch) on derecho

### General bug information

**CTSM version you are using:** ctsm5.1.dev082 in cesm2_3_alpha08d

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** tests with nvhpc and DEBUG on

### Details of bug

These tests fail:

SMS_D.f19_g17.IHistClm50Bgc.cheyenne_nvhpc.clm-decStart		
SMS_D.f45_f45_mg37.I2000Clm50FatesRs.cheyenne_nvhpc.clm-FatesColdDef		
SMS_D_Ld1.f10_f10_mg37.I1850Clm50Sp.cheyenne_nvhpc.clm-default		
SMS_D_Ld1_P25x1.5x5_amazon.I2000Clm50SpRs.cheyenne_nvhpc.clm-default

While DEBUG off tests PASS:

SMS.f19_g17.IHistClm50Bgc.cheyenne_nvhpc.clm-decStart	
SMS_Ld1.f10_f10_mg37.I1850Clm50Sp.cheyenne_nvhpc.clm-default			

As well as mpi-serial tests:

SMS_D_Ld1_Mmpi-serial.1x1_brazil.I2000Clm50SpRs.cheyenne_nvhpc.clm-default			
SMS_D_Ld1_Mmpi-serial.5x5_amazon.I2000Clm50SpRs.cheyenne_nvhpc.clm-default			
SMS_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesRs.cheyenne_nvhpc.clm-FatesColdDef			
SMS_D_Mmpi-serial.1x1_brazil.IHistClm50BgcQianRs.cheyenne_nvhpc.clm-default					
SMS_Mmpi-serial.1x1_brazil.IHistClm50BgcQianRs.cheyenne_nvhpc.clm-default

### Important output or errors that show the problem

For the smallest case: SMS_D_Ld1_P25x1.5x5_amazon.I2000Clm50SpRs.cheyenne_nvhpc.clm-default

The only log file available is the cesm.log file as follows.

cesm.log file:

```
 (t_initf)       profile_single_file=       F
[...]

PR #2934, opened 2025-01-14
ctsm5.3.022: Update FATES namelist and build options to avoid `Meier2022`

glemieux:
### Description of changes
Avoid running `fates` with `z0param_method=Meier2022` in light of #2932.

### Specific notes

This also merges #2936 and #2898

Contributors other than yourself, if any: @rgknox 

   Does the first step in #2932

Are answers expected to change (and if so in what way)? Yes, for fates compsets with `Clm60`

Any User Interface Changes (namelist or namelist defaults changes)? Yes, changes default for roughness for FATES and disallows one option when FATES is used

Testing performed, if any: will do regular and fates tests suites


glemieux:
@ekluzek, in general does the order of options in the namelist_defaults_ctsm.xml matter?  I.e. is it important the `"use_fates"` preceed `"phys"`?

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -540,7 +540,9 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- ================================================================== -->
 
 <z0param_method>ZengWang2007</z0param_method>
-<z0param_method phys="clm6_0">Meier2022</z0param_method>
+<!-- <z0param_method phys="clm6_0">Meier2022</z0param_method> -->
+<z0param_method use_fates=".true."  phys="clm6_0" >ZengWang2007</z0param_method>
```
glemieux, line 544: @ekluzek, in general does the order of options in the namelist_defaults_ctsm.xml matter? I.e. is it important the "use_fates" preceed "phys"?

ekluzek: No, the order does NOT matter in terms of getting the right result.

I think the file may follow some conventions on the order, which can make it easier to read. I've done vertical alignment for that sort of thing as well, but we are trying to get away from that sort of thing.

bld/CLMBuildNamelist.pm
```diff
@@ -2125,10 +2125,13 @@ sub setup_logic_roughness_methods {
...
   }
   my $phys = $physv->as_string();
-  if ( $phys eq "clm4_5" || $phys eq "clm5_0" ) {
-    if ( $var eq "Meier2022" ) {
+  if ( $var eq "Meier2022" ) {
+    if ( $phys eq "clm4_5" || $phys eq "clm5_0" ) {
       $log->fatal_error("z0param_method = $var and phys = $phys, but this method has been tested only with clm6_0 and later versions; to use with earlier versions, disable this error, and add Meier2022 parameters to the corresponding params file");
     }
+    if ( &value_is_true($nl_flags->{'use_fates'}) ) {
+      $log->fatal_error("z0param_method = $var and use_fates currently are not compatible.  Please update the z0param_method to ZengWang2007")
```
ekluzek, line 2133: Add a comment that points to the issue to get the details. Probably add the issue number to the error message as well.

ekluzek: Should we also disallow this in the FORTRAN code? I vote yes, because it results in bad answers.

ekluzek (changes requested):
Thanks for working on this @glemieux and getting this in quickly. We'll talk Thursday on where this should go in upcoming tags.

I ask for a couple changes that are easy.

bld/CLMBuildNamelist.pm
glemieux: Oh I see, you mean to update the case code here?

https://github.com/ESCOMP/CTSM/blob/fb89f3359052e4fbba4ff1ccc89e82416a6e544d/src/biogeophys/CanopyFluxesMod.F90#L899-L908

ekluzek: It should be something done at initialization rather than runtime. So earlier on.

bld/namelist_files/namelist_defaults_ctsm.xml
rgknox: do we need this line?
> <z0param_method use_fates=".true."  phys="clm6_0" >ZengWang2007</z0param_method>

The default was already set to ZengWang2007.  It seems like we covered our bases by applying the use_fates tag here: 

> <z0param_method use_fates=".false." phys="clm6_0" >Meier2022</z0param_method>

ekluzek: Yes, thanks @rgknox that's a bit cleaner/clearer way to do it. Let the default be set as ZengWang for everything and only overridden for non-FATES and clm6_0 physics.

bld/CLMBuildNamelist.pm
glemieux: @ekluzek I've added an `endrun` with what I think might be the earliest, pre-existing `use_fates` check during initialization.  Does 43407e1 fit what you were thinking?

glemieux: > Add a comment that points to the issue to get the details. Probably add the issue number to the error message as well.

Addressed per cb361eb

bld/namelist_files/namelist_defaults_ctsm.xml
glemieux: @ekluzek @rgknox this is what I thought too, but I just tested removing this line and the build fails saying that Meier2022 is still being set somehow:
```
ERROR: Command /home/glemieux/Repos/ESCOMP/ctsm/bld/build-namelist failed rc=255
out=
err=ERROR : CLM build-namelist::CLMBuildNamelist::setup_logic_roughness_methods() : z0param_method = Meier2022 and use_fates currently are not compatible.  Please update the z0param_method to ZengWang2007.  See issue #2932 for more information.
```

bld/CLMBuildNamelist.pm
```diff
@@ -2125,10 +2125,14 @@ sub setup_logic_roughness_methods {
     $log->fatal_error("$var is incorrect entry for the namelist variable z0param_method; expected Meier2022 or ZengWang2007");
   }
   my $phys = $physv->as_string();
```
ekluzek, line 2127: Add use_fates to the add_default call above for z0param_method

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: Try it again with adding use_fates to the add_default call for z0param_method, as I suggest above.

bld/CLMBuildNamelist.pm
ekluzek: This will work. But, a few thoughts. clm_initialize is a pretty high level module. I think it would be good to have it one level lower. Putting specific logic for things like this a level lower is a little more modular.

I looked through the code and have two suggestions. One would be to add this to control_mod.F90 where both use_fates and z0param_method is set. The other would be in init in clmfates_interfaceMod.

bld/CLMBuildNamelist.pm
glemieux: That did the trick.  Thanks for the reminder.

bld/CLMBuildNamelist.pm
glemieux: Thanks for the suggestions @ekluzek.  The fates compatibility section in `control_init` looks good.  I've reverted the previous commit and made the update via ccb55da.

glemieux:
Regression testing on `derecho` is underway.

glemieux:
`aux_clm` testing against `ctsm5.3.020` baseline are showing DIFFs only in tests with `Clm60` in the compsets as expected.

Testing on izumi underway.

glemieux:
@ekluzek what do you think about me bringing in #2936 with this PR?  It would create more diffs since we'd need to update the fates tag in `.gitmodules` for the fates-side PR associated with #2936 (ngeet/fates#1273).  But both the ctsm-side changes are pretty minimal and ctsm tests would still be b4b.

ekluzek:
Perfectly fine @glemieux and up to you. Since this is a tag on master there is no need to make it b4b. So do what you need to do.

glemieux:
Regression testing `aux_clm` on `izumi` and `derecho` is underway.

glemieux:
Regression testing on `derecho` is  complete.  All expected tests are passing with one exception.  I discovered ngeet/fates#1324 during testing and have updated the expected failure test list accordingly.  Also I ran into #2930 again, so I reran the suite omitting the `--testid-base` option for the system tests.  

Expected fates tests are b4b with the exception of tests that use `Clm60Fates` based compsets, which is as expected due to using `ZengWang2007` instead of `Meier2022` per this PR.  An additional three tests are showing DIFFs for the `FATES_GDD` variable only.  @rgknox  

```
FAIL ERS_D_Ld20.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdTwoStream BASELINE ctsm5.3.021: DIFF
FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdFixedBiogeo BASELINE ctsm5.3.021: DIFF
FAIL SMS_D_Lm6_P256x1.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesCold BASELINE ctsm5.3.021: DIFF
```

All expected clm tests are b4b except for `SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.derecho_nvhpc.clm-crop`.  This is failing during `RUN` with the following in the Teststatus:

```
153  ---------------------------------------------------
154 2025-02-03 00:19:44: ERROR: Model did not complete - see /glade/u/home/glemieux/scratch/ctsm-tests/tests_0202-192217de/SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.derecho_nvhpc.clm-crop.GC.0202-192217de_nvh/    run/med.log.7645521.desched1.250203-001918
155 
156  ---------------------------------------------------
```

I haven't been able to check `izumi` results due to what I'm assuming are continued issues seen from this past week.

Results : `/glade/u/home/glemieux/scratch/ctsm-tests/tests_0202-192217de`

slevis-lmwg:
#1733 discusses this test SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.derecho_nvhpc.clm-crop

glemieux:
Per discussion with @rgknox at the standup this morning, it is likely that NGEET/fates#1306 is the cause of the `FATES_GDD` differences seen above.  To confirm, I ran one of the above tests using the fates tag associated with that PR ,`fates-sci.1.80.9_api.37.0.0-ctsm5.3.018`, and compared it against the previous baseline.  The results are consistent and can be found below.

Results: `/glade/u/home/glemieux/ctsm/tests_0203-114755de`

As such, testing on `derecho` is complete.  Just waiting on izumi to come back up to check tests there.

glemieux:
Per discussion with @rgknox, I've removed `FATES_C13_DISC_SZPF` from `AllVars` as this is a known low priority issue (see ngeet/fates#700).

Rerunning tests on izumi as it seems some tests were mangled during last Friday's issues.

glemieux:
Rerunning izumi tests again as a number of them failed due to input data still not being available yesterday evening.

glemieux:
Resubmitting failed `aux_clm` cases via `--rerun-existing-failures`  seem to have mostly worked.  Nearly all expected tests are B4B.  The only DIFFs are for fates testmods due either to the z0 method update or due to a fates update to the temperature average (i.e. the `FATES_GDD` diffs mentioned above).

This PR is ready to merge.

ekluzek (approved):

Issue #2933, opened 2025-01-14
setting patch%itype to invalid for FATES patches


rgknox:
In #2392 we found a regression where FATES runs were allowed to erroneously perform operations with the patch%itype structure.  I think we've allowed the allocation of this array in fates runs because of the anticipation of allowing non-fates modules (that would need access to itype) to run along side fates.

Setting values of patch%itype for fates patches back to the uninitialized value would help prevent issues that were experienced in #2392, because calling that array will result in out of bounds errors.

You can see we played with the idea of doing this here:

https://github.com/ESCOMP/CTSM/blob/ctsm5.3.018/src/utils/clmfates_interfaceMod.F90#L933

ekluzek:
I love this idea @rgknox. This will help us find problems when cases are run in DEBUG mode. I'm hoping we could bring this in soonish, but I'm also afraid that this will expose existing problems that might take some time to work through. But, that would also be a benefit of doing this. I've assigned you and @glemieux to it -- but please change as appropriate. I also put it as something to do by the answer changing freeze, but hope this can come in ASAP.

Since, this is JUST changing the value over FATES columns what you have here should be fine for working towards our desire to have both FATES and non-FATES columns running at the same time. So I don't think your second sentence applies for what you've proposed.

rgknox:
Created this PR to address itype, just a strawman but does the job: https://github.com/ESCOMP/CTSM/pull/2935
Should we combine this with #2934 ?

wwieder:
Given the urgency of this PR, are you OK with the suggested order on upcoming tags?

rgknox:
@wwieder This PR does not seem as urgent as Greg's PR that turns of Meier turbulence with fates, this one is more about preventing future heartache.  That said, I do think its complete and should help us catch errors, but no rush here.

adrifoster:
@mvdebolskiy this is what I was thinking of

Issue #1784, opened 2022-06-17
Get dry-deposition/MEGAN working with FATES-SP mode


ekluzek:
The most straightforward part of getting FATES to work with dry deposition is to get it to work in FATES-SP mode. This is something that the code is very close to being able to do right now.

This is a particular part of #1044

ekluzek:
Open this up as I need to review the code and make sure all of the data that is used by MEGAN and drydep is set when FATES is on. I've reviewed some of it and it's set, but I should make sure everything is set.

After I do that Louisa will run some simulations and make sure FATES-SP is mode is worked as expected for MEGAN and drydep.

rgknox:
Removed FATES SP DryDep from the aux_clm tests.  When dryep and FATES SP is re-enabled, here is the commit where the test was removed (to make it easier to re-instate: https://github.com/ESCOMP/CTSM/pull/2935/commits/7e4e6b243694e6b18598933a8096facb541a661e)

Here is the commit where we change the namelist build script logic to be a fatal error:
https://github.com/ESCOMP/CTSM/pull/2935/commits/3f6eb018b224e09a8cdfb2662b907a7795f8d1f0

ekluzek:
The changes on NorESM should allow us to get this working for both FATES-SP and regular FATES, so it seems fine to let it be broken for right now. And that's the fastest way forward. So this all sounds good. And the commits look good to me.

PR #2935, opened 2025-01-15
Initializing itype as spval for fates patches

rgknox:
### Description of changes

On FATES patches, there should be no notion of pft associated with the patch, and therefore patch%itype should always be invalid.  Even for FATES-SP, the patch should be associated with the FATES pft, which is not associated with itype. In this set of changes, itype is set to spval on fates patches. This will help prevent bugs in the future, because its use in a fates context should trigger errors, particularly in debug mode.

### Specific notes

Addresses part of: #2932

@rosiealice @ekluzek @glemieux others

Fixes: #2933

Testing performed, if any:

One off tests to see if the model would run with Zeng and Wang turbulent transport.  It does. Ran into machine issues testing with Meier2022, which should crash.

### DIFF
--- a/src/biogeophys/SurfaceAlbedoMod.F90
+++ b/src/biogeophys/SurfaceAlbedoMod.F90
@@ -864,23 +864,6 @@ contains
           end if
     end do
 
-    ! Weight reflectance/transmittance by lai and sai
-       ! Only perform on vegetated patches where coszen > 0
-
-    do fp = 1,num_vegsol
-       p = filter_vegsol(fp)
-       wl(p) = elai(p) / max( elai(p)+esai(p), mpe )
-       ws(p) = esai(p) / max( elai(p)+esai(p), mpe )
-    end do
-
-    do ib = 1, numrad
-       do fp = 1,num_vegsol
-          p = filter_vegsol(fp)
-          rho(p,ib) = max( rhol(patch%itype(p),ib)*wl(p) + rhos(patch%itype(p),ib)*ws(p), mpe )
-          tau(p,ib) = max( taul(patch%itype(p),ib)*wl(p) + taus(patch%itype(p),ib)*ws(p), mpe )
-       end do
-    end do
-
     ! Diagnose number of canopy layers for radiative transfer, in increments of dincmax.
     ! Add to number of layers so long as cumulative leaf+stem area does not exceed total
     ! leaf+stem area. Then add any remaining leaf+stem area to next layer and exit the loop.
@@ -1051,7 +1034,24 @@ contains
             fcansno(bounds%begp:bounds%endp), surfalb_inst)
 
     else
-
+       
+       ! Weight reflectance/transmittance by lai and sai
+       ! Only perform on vegetated patches where coszen > 0
+       
+       do fp = 1,num_vegsol
+          p = filter_vegsol(fp)
+          wl(p) = elai(p) / max( elai(p)+esai(p), mpe )
+          ws(p) = esai(p) / max( elai(p)+esai(p), mpe )
+       end do
+       
+       do ib = 1, numrad
+          do fp = 1,num_vegsol
+             p = filter_vegsol(fp)
+             rho(p,ib) = max( rhol(patch%itype(p),ib)*wl(p) + rhos(patch%itype(p),ib)*ws(p), mpe )
+             tau(p,ib) = max( taul(patch%itype(p),ib)*wl(p) + taus(patch%itype(p),ib)*ws(p), mpe )
+          end do
+       end do
+       
        call TwoStream (bounds, filter_vegsol, num_vegsol, &
             coszen_patch(bounds%begp:bounds%endp), &
             rho(bounds%begp:bounds%endp, :), &
--- a/src/main/clm_driver.F90
+++ b/src/main/clm_driver.F90
@@ -18,6 +18,7 @@ module clm_driver
   use clm_time_manager       , only : get_nstep, is_beg_curr_day, is_beg_curr_year
   use clm_time_manager       , only : get_prev_date, is_first_step
   use clm_varpar             , only : nlevsno, nlevgrnd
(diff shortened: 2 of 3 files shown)


rgknox:
combine this with #2934 ?

ekluzek:
@rgknox I suggest you leave this outside of #2934 for now. I'm wondering what will happen with this when the fates and aux_clm test lists are run. I'm curious if you'll see new issues when you run DEBUG level tests with FATES. If so I'm thinking it might be a bit to resolve. But, I might be wrong...

src/biogeophys/SurfaceAlbedoMod.F90
```diff
@@ -1051,7 +1034,24 @@ subroutine SurfaceAlbedo(bounds,nc,  &
             fcansno(bounds%begp:bounds%endp), surfalb_inst)
 
     else
-
+
```
ekluzek, line 1037: Ahh, so this is another place where this caught this pattern that was being done for FATES, and now you have it only done outside of FATES.

rosiealice:
Good idea @rgknox :) I hope @ekluzek is not right about needing to fix lots of things but maybe we should keep them separate just incase :/

src/main/clm_driver.F90
```diff
@@ -704,16 +705,23 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
...
-       leafn_patch = bgc_vegetation_inst%get_leafn_patch(bounds_clump)
-
        allocate(froot_carbon(bounds_clump%begp:bounds_clump%endp))
        allocate(croot_carbon(bounds_clump%begp:bounds_clump%endp))
-       froot_carbon = bgc_vegetation_inst%get_froot_carbon_patch( &
-            bounds_clump, canopystate_inst%tlai_patch(bounds_clump%begp:bounds_clump%endp))
-       croot_carbon = bgc_vegetation_inst%get_croot_carbon_patch( &
-            bounds_clump, canopystate_inst%tlai_patch(bounds_clump%begp:bounds_clump%endp))
 
+       if(use_fates)then
```
ekluzek, line 711: I'd add a comment to this about why this is being done. Also try using nan rather than spval.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -930,7 +930,7 @@ subroutine init(this, bounds_proc, flandusepftdat)
             c = this%f2hmap(nc)%fcolumn(s)
             pi = col%patchi(c)+1
             pf = col%patchf(c)
-!            patch%itype(pi:pf) = ispval
+            patch%itype(pi:pf) = ispval
```
ekluzek, line 933: Add a note about why this is being done, and what it will do.

ekluzek (changes requested):
Awesome, to see a start at this so quickly. And for it to already have caught something. Things were being set and then overridden by FATES, so it won't change answers, but it's good to know that the change caught that.

I'm just requesting adding some comments to describe what's going on and why.

src/utils/clmfates_interfaceMod.F90
rgknox: actually, the comments on 925-927 explain what is going on already.

rgknox: let me update the text a little though, its not as clear as it could be

src/biogeophys/SurfaceAlbedoMod.F90
rgknox: yes

src/main/clm_driver.F90
rgknox: added comments

src/utils/clmfates_interfaceMod.F90
rgknox: updated text, clearer now

src/main/clm_driver.F90
ekluzek: The way to get nan within a scope is to add this use statement:

``` fortran
use shr_infnan_mod  , only : nan => shr_infnan_nan, assignment(=)
```

rgknox: changed spval to nan

ekluzek: Perfect!

ekluzek:
@rgknox this looks to me like it could come in right now. As next in line.

We plan to have @slevis-lmwg do the integrator step from..

https://github.com/ESCOMP/CTSM/wiki/Protocols-on-updating-FATES-within-CTSM#ctsmfates-developer

rgknox:
That works for me @ekluzek , I believe it is feature complete. I'll work with you and @slevis-lmwg  to start testing

ekluzek (approved):

ekluzek:
@rgknox testing showed that we have FATES-SP tests that rely on using itype. I remember that now that he mentions it, but didn't think of it beforehand. This is a a good example of this change catching this problem. So it's actually a good thing, and will serve as an example of what needs to be done when the assumption that you can use itype on the CTSM side for FATES. But, it will take a little bit to fix.

rgknox:
I updated the code to address another issue. We had a test in our aux_clm list that was testing FATES SP and DryDep.  DryDep code relies on an association between patch and pft via itype.  This test has been removed from the testlist, and a fatal_error call has been added to the namelist perl script if FATES is run with DryDep.  I also linked the following issue in the test list:  https://github.com/ESCOMP/CTSM/issues/1784

ekluzek (approved):

samsrabin:
Testing for this is finished. A few of Ryan's Izumi tests failed, but I reran them and they worked. I linked any files missing from Ryan's ctsm5.3.046 baseline to my ctsm5.3.046.ssr baseline.

samsrabin:
Now just waiting on ChangeLog/ChangeSum to merge.

Issue #2739, opened 2024-09-03
run_neon not working as expected for AD simulations


wwieder:
### Brief summary of bug

running a NEON site in AD mode with the run_neon command fails after the base case is created. 

`./run_neon --neon-sites TOOL --run-type ad`

The base case is created and build correctly.  But the following error message Error message after the base case builds and the SITE.ad case is created is below:
```
...
Check case OK
submit_jobs case.run
ERROR: Do not know about batch job case.run
```

This is easily addressed by running `./case_setup` in the AD case, but it's not really working as designed.

### General bug information

**CTSM version you are using:** [output of `git describe`]
ctsm5.2.023

**Does this bug cause significantly incorrect results in the model's science?** [Yes / No]
No

**Configurations affected:** [Fill this in if known.]
run_neon cases.  I'll try with postAD and running transient cases from postAD options too & update this issue.

Maybe this is something that can be addressed by #2671 or #2406?  is this something you're seeing in your run_tower testing, @TeaganKing?

I confirmed the same is NOT true with a transient case out of the box.
`./run_neon --neon-sites BART`, this transient case runs as intended.

TeaganKing:
Hey Will,

I think this will be addressed by the `case.setup` in the Plumber2 Implementation PR: https://github.com/ESCOMP/CTSM/pull/2406/files#diff-8643a6b1f4d44a68a01d1daa1e60ff693ce02b559148bf686df829a92d7d0ce1 , line 390.

Issue #2944, opened 2025-01-27
Make sure NEON or PLUMBER options are only in their own extension of the tower_site base class


ekluzek:
We want to make sure that the NEON and PLUMBER specific options are only in their own extension of the tower_site base class.

ekluzek:
#2969 helps with this as it removes the need for an extension of the build_base_case subroutine extension for NEON and PLUMBER2.

PR #2963, opened 2025-02-13
run_tower fixes and improvements

### LINKED ISSUE #2884: run_neon --overwrite is ignored
### Brief summary of bug

`run_neon` `--overwrite` option is ignored. I also think the error message needs to be reworked.

### General bug information

**CTSM version you are using:** ctsm5.3.012

### Details of bug

The following patch fixes it:
```diff
diff --git a/python/ctsm/site_and_regional/neon_site.py b/python/ctsm/site_and_regional/neon_site.py
index 70414d9e5..ced4c4c64 100755
--- a/python/ctsm/site_and_regional/neon_site.py
+++ b/python/ctsm/site_and_regional/neon_site.py
@@ -47,7 +47,8 @@ class NeonSite(TowerSite):
             user_mods_dirs = [
                 os.path.join(self.cesmroot, "cime_config", "usermods_dirs", "NEON", self.name)
             ]
-        case_path = super().build_base_case(cesmroot, output_root, res, compset, user_mods_dirs)
+        case_path = super().build_base_case(cesmroot, output_root, res, compset, user_mods_dirs,
+                                            overwrite=overwrite)

         return case_path
```

Tested on Derecho.

```
# Interrupt this after "base case created"
python tools/site_and_regional/run_neon --neon-sites ABBY --setup-only

# Then try again with --overwrite:
python tools/site_and_regional/run_neon --neon-sites ABBY --setup-only --overwrite
```

```
---- building a base case -------
ERROR: Existing base case is a historical type and should not be
                        --rerun with the --overwrite option
```

Bit about `--rerun` seems irrelevant, since I don't see that checked anywhere in the call stack.

### LINKED ISSUE #2885: run_neon --setup-only is ignored
### Brief summary of bug

`run_neon` `--setup-only` option is ignored.

### General bug information

**CTSM version you are using:** ctsm5.3.012

### Details of bug

The following patch fixes it:
```diff
diff --git a/python/ctsm/site_and_regional/neon_site.py b/python/ctsm/site_and_regional/neon_site.py
index 70414d9e5..ced4c4c64 100755
--- a/python/ctsm/site_and_regional/neon_site.py
+++ b/python/ctsm/site_and_regional/neon_site.py
@@ -47,7 +47,8 @@ class NeonSite(TowerSite):
             user_mods_dirs = [
                 os.path.join(self.cesmroot, "cime_config", "usermods_dirs", "NEON", self.name)
             ]
-        case_path = super().build_base_case(cesmroot, output_root, res, compset, user_mods_dirs)
+        case_path = super().build_base_case(cesmroot, output_root, res, compset, user_mods_dirs,
+                                            setup_only=setup_only)

         return case_path
```
Tested on Derecho.

```
# Running this continues through the build step, but --setup-only means it shouldn't
python tools/site_and_regional/run_neon --neon-sites ABBY --setup-only
```

```
---- building a base case -------
/glade/work/samrabin/ctsm_anping_lecture/ABBY
Removing the existing case at: /glade/work/samrabin/ctsm_anping_lecture/ABBY
---- creating a base case -------
---- base case created ------
---- base case setup ------
---- base case build ------
--- This may take a while and you may see WARNING messages ---
WARNING: No .input_data_list files found in dir 'Buildconf'
...
```

### LINKED ISSUE #2946: Python sys tests much longer with added tower testing...
#2406 makes the python system testing much longer. It's still less than an hour but a good fraction of it though. This is something that would help our python development and testing if we make it shorter. The reason for the length is really #2885. So fixing it should resolve this.

I'm also wondering about breaking up python system tests to another set of "long_sys" tests, that maybe we don't run as often? And actually if we ran that test list as part of aux_clm hour long testing would be fine.

samsrabin:
### Description of changes
Resolves some bugs and improves some messaging. Also, two new options:
- `--no-inputdata-check` option implies `--setup-only` but skips the check/download of input data. This is used instead of `--setup-only` in `run_tower` system testing for a speedup of ~20%.
- `--xmlchange` option allows user to specify `xmlchange` settings to apply. E.g., `--xmlchange CLM_CO2_TYPE=constant,CCSM_CO2_PPMV=850`.

### Specific notes

**CTSM Issues Fixed:**
- Resolves #2884 
- Resolves #2885
- Resolves #2946

**Testing performed, if any:** Site/regional Python unit and system tests.
- [x] Full Python unit and system testing
- [x] `./run_neon --neon-sites TOOL --run-type ad` should complete without issues (see #2739)
- [x] `ctsm_pymods` suite

### Remaining work
- [x] Add system test to check that `--setup-only` is obeyed
- [x] Add system test to check that `--overwrite` is obeyed
- [x] Add system test to check that `--xmlchange` is obeyed


samsrabin:
In my testing of ctsm5.3.024, test_sys_run_tower.py takes about 871 seconds (n=2). This PR, as of now, speeds that up by >90% (~82 seconds, n=3). I will mark that PR as closing this issue.

samsrabin:
With the new tests, this now takes about 5 minutes (300 seconds). It would be shorter if I didn't feel like I needed to duplicate tests between both NEON and PLUMBER sites. @ekluzek and @TeaganKing, would it be crazy for me to try and get rid of the separate `NeonSite` and `Plumber2Site` classes, and instead just implement those differences within `TowerSite`? It seems simple to do, but I feel like maybe I'm missing something.

TeaganKing:
Hi @samsrabin , We were hoping to implement the NeonSite and Plumber2Site classes separately in order to have clearer parts of the code that were shared versus specific to each type of tower. We also wanted to set this up so that we could eventually create more generic tower site classes parallel to those types of classes. Thus, while it's feasible to merge them back together, I think it's preferable to keep these classes as child classes that are separate from the TowerSite class.

samsrabin:
@TeaganKing Yup, makes sense—that's the context I was missing. Thanks!

TeaganKing:
Thanks Sam! I'll review this hopefully later this afternoon or maybe Tuesday!

samsrabin:
(Couldn't stop thinking about simplifying, but I recognize that's a bigger task, so I started a new branch [here](https://github.com/samsrabin/CTSM/commits/simplify-tower-types/). Will submit a PR for that only after this one is approved, because it branches from this one.)

python/ctsm/site_and_regional/neon_site.py
```diff
@@ -68,6 +76,8 @@ def run_case(
         no_batch=False,
         rerun=False,
         experiment=False,
+        no_input_data_check=False,
```
TeaganKing, line 79: These are great options to add! Thanks for implementing/clarifying the no input data one!

python/ctsm/site_and_regional/tower_arg_parse.py
```diff
@@ -104,6 +104,19 @@ def get_parser(args, description, valid_neon_sites, valid_plumber_sites):
         default=False,
     )
 
+    parser.add_argument(
+        "--no-input-data-check",
+        "--no-check-input-data",
```
TeaganKing, line 109: Good idea to have this work for either order in case users get confused by which one it is!

python/ctsm/site_and_regional/tower_site.py
```diff
@@ -128,6 +128,7 @@ def build_base_case(
             abort("Input output_root directory does NOT exist: " + str(output_root))
 
         case_path = os.path.join(output_root, self.name)
+        print(case_path)
```
TeaganKing, line 131: Can we get rid of this print statement?

python/ctsm/test/test_sys_run_tower.py
```diff
@@ -115,6 +120,187 @@ def test_plumber_site(self):
...
+            "BART",
+            "--no-input-data-check",
+            "--experiment",
+            "TEST",
+            "--xmlchange",
+            "CLM_CO2_TYPE=constant,CCSM_CO2_PPMV=1987",
+            "--output-root",
+            self._tempdir,
+        ]
+        print(sys.argv)
```
TeaganKing, line 141: Can we also remove these print statements (and below?)

TeaganKing (approved):
Requesting a few minor changes, but this looks great! Thanks so much for putting this together!

python/ctsm/site_and_regional/tower_arg_parse.py
samsrabin: Haha, well I did it because I got confused myself!

python/ctsm/test/test_sys_run_tower.py
samsrabin: Done with 491507a22.

python/ctsm/site_and_regional/tower_site.py
samsrabin: Done with 491507a22.

ekluzek:
> Hi @samsrabin , We were hoping to implement the NeonSite and Plumber2Site classes separately in order to have clearer parts of the code that were shared versus specific to each type of tower. We also wanted to set this up so that we could eventually create more generic tower site classes parallel to those types of classes. Thus, while it's feasible to merge them back together, I think it's preferable to keep these classes as child classes that are separate from the TowerSite class.

Just FYI adding to the conversation a bit.

Yes in #2944 we talk about the desire to make the design more fully Object Oriented and to have all of the type specific stuff in the extensions. This removes having logic in the lower class, and allows freedom in the extensions to be added or removed easily. Logic is one of the things that easy to get wrong so can be buggy and requires maintenance. So using the OO programming model helps eliminate that source of problems.

PR #2809, opened 2024-10-03
Update docs infrastructure

samsrabin:
### Description of changes
- Removes dependence on outdated version of `sphinx_rtd_theme`
- Enables embedding of Markdown docs in .rst files
- Might enable us to finally ditch Docker and have Github workflows for checking and publishing our docs 🤞...

### Specific notes

**Contributors other than yourself, if any:** None

**CTSM Issues Fixed (include github issue #):** tk

**Are answers expected to change (and if so in what way)?** The way users switch between documentation versions will be different. 

**Any User Interface Changes (namelist or namelist defaults changes)?** No

**Testing performed, if any:** Builds locally with no errors or warnings. Hopefully the version switching isn't broken—it looks broken locally, but I don't know if that's just an issue of it being local.

Issue #2625, opened 2024-06-27
Move ctsm_pylib to something beyond 3.7


samsrabin:
It's [unsupported](https://devguide.python.org/versions/)! For the release we need to update our conda environment and python version to one that is supported, and hopefully will be supported for most of the release timeframe.

samsrabin:
Blocking #2979.

samsrabin:
Also:
- [ ] Update Python versions in workflows introduced in #2964.

PR #2964, opened 2025-02-14
Add GitHub workflows to check black, pylint, git-fleximod

### LINKED ISSUE #1580: Add an action to test python code through different python versions

samsrabin:
### Description of changes
- Adds workflows to test git-fleximod and pylint.
- `black` and `pylint` workflows only run when relevant files are changed
- `black` and `pylint` workflows use `ctsm_pylib` conda environment

### Specific notes

**CTSM Issues Fixed:**
Because I'm assuming we only want to support people running/linting our tools in `ctsm_pylib` and not other Python versions:
- Closes #1580

**Testing performed, if any:** None needed except what GitHub does.


.gitmodules
```diff
@@ -31,7 +31,7 @@ url = https://github.com/NGEET/fates
 fxtag = sci.1.80.11_api.37.0.0
 fxrequired = AlwaysRequired
 # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
-fxDONOTUSEurl = https://github.com/NCAR/fates-release
+fxDONOTUSEurl = https://github.com/NGEET/fates
```
samsrabin, line 34: @ekluzek Could you sync https://github.com/NCAR/fates-release with the upstream repo? That would then allow us to have `url` and `fxDONOTUSEurl` match, allowing the `fleximod_test` workflow to pass.

It would probably be good for all of us CTSM maintainers to have maintain permissions there too, for ease of keeping it up-to-date.

wwieder: @ekluzek likes this change, as the NCAR/FATES-release repo is not being maintained.

samsrabin:
@ekluzek Reminding you to review, as discussed. Thanks!

samsrabin:
Blocked by #2625. Once that's done:
- [ ] Update Python versions tested
- [ ] Use our conda env instead of requirements.txt.

samsrabin:
#2982 is merged, so this is unblocked.

samsrabin:
@ekluzek This is complete and ready for review. Thanks!

samsrabin:
Blocking #2809.

.github/workflows/fleximod_test.yaml
```diff
@@ -0,0 +1,31 @@
...
+on: [push, pull_request]
+
+jobs:
+  fleximod-test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # oldest supported and latest supported
+        python-version: ["3.7", "3.x"]
```
ekluzek, line 15: Cool, so 3.x get's translated into the very latest python 3. version? That's really cool here.

.github/workflows/fleximod_test.yaml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - id: checkout-CTSM
+        uses: actions/checkout@v4
+      - id: run-fleximod
+        run: |
+          $GITHUB_WORKSPACE/bin/git-fleximod update
+          echo
+          echo "Update complete, checking status"
+          echo
+          $GITHUB_WORKSPACE/bin/git-fleximod test
```
ekluzek, line 25: This looks like the right test to do to me.

.github/workflows/fleximod_test.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - id: check-cleanliness
+        run: |
+          echo
+          echo "Checking if git fleximod matches expected externals"
```
ekluzek, line 29: Ahh, nice. So this will make sure that the submodule directory was updated in addition to the .gitmodule file.

.github/workflows/formatting_python.yml
```diff
@@ -0,0 +1,44 @@
...
+          activate-environment: ctsm_pylib
+          environment-file: python/conda_env_ctsm_py.yml
+          channels: conda-forge
+          auto-activate-base: false
+
+      # Run pylint check
+      - name: Run pylint
+        run: |
+          cd python
+          conda run -n ctsm_pylib make lint
```
ekluzek, line 36: Ahh, nice. This makes the workflow nicely aligned with the Makefile that we use.

I wanted to get the workflow to use the make directly, but couldn't get that to go. I think it's this technique you used here to run "make lint" and later "make black" beneath "conda run" that get's it to go here. make probably needs to be run under the conda subshell.

.gitmodules
```diff
@@ -31,7 +31,7 @@ url = https://github.com/NGEET/fates
 fxtag = sci.1.82.3_api.39.0.0
 fxrequired = AlwaysRequired
 # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
-fxDONOTUSEurl = https://github.com/NCAR/fates-release
+fxDONOTUSEurl = https://github.com/NGEET/fates
```
ekluzek, line 34: Yep, this is the thing needed to get "git-fleximod test" to work.

ekluzek (approved):
I have one thing I suggest you do. But, other than that this is great. You found a solution for a problem that I didn't, so the resulting workflow is better than before.

The one suggestion is to possibly keep the two step create and then initialize in py_env_create. So we should talk if you disagree.

So I've marked as approve so you don't have to wait on anything from me. We could even talk about the two step thing later.

ekluzek:
This also gets me thinking about adding some more tests in a future PR that could be done at some frequency. The python unit tests for example. Maybe we add some longer tests like them to happen less often then with each push, but maybe even something like nightly.

.github/workflows/fleximod_test.yaml
samsrabin: Yep! That's something I took straight from the CESM implementation of this workflow. Most of the workflow is like that, actually.

.github/workflows/formatting_python.yml
samsrabin: Yeah, that's exactly it.

samsrabin:
Running tests on a schedule is definitely a good idea in general. I actually have one in the docs revamp PR that checks once per week that `ctsm_pylib` is able to be built, because it's possible at some point that some Python module version gets taken off the conda/mamba channels, resulting in the solver failing. I added that because it's something that would happen outside of this repo that would cause the failure. We might want to run the Python unit tests weekly for the same reason, although we'd need to resolve #2984 (at least the unit test bit) first.

Issue #2452, opened 2024-04-03
Run mksurfdata_esmf with the new vegetation raw data


slevis-lmwg:
@lawrencepj1 has a new set of raw datasets for vegetation that we need to test with mksurfdata_esmf

In `/glade/campaign/cesm/development/lmwg/landuse_source_data/CTSM5.2RawData` there are 10 directories:
```
globalctsm52CCISENSBASELINEDeg025_240402  globalctsm52SSP126Deg025_231228
globalctsm52histLUH2Deg025_231228         globalctsm52SSP245Deg025_231228
globalctsm52histMKSRFDeg025_231228        globalctsm52SSP370Deg025_231228
globalctsm52histTRENDY2023Deg025_231228   globalctsm52SSP460Deg025_231228
globalctsm52SSP119Deg025_231228           globalctsm52SSP585Deg025_231228
```
In the above,
MKSRF contains the lai and soilcolor raw data
no need to test the contents of CCISENSBASELINE and TRENDY

- [x] @slevis-lmwg @lawrencepj1 @olyson @wwieder tested successfully and with satisfactory results (https://github.com/NCAR/LMWG_dev/issues/59) using the above raw datasets, but we have decided instead to go with the TRENDY2023 and soon TRENDY2024 data.
- [x] @slevis-lmwg test mksurfdata_esmf with f09.
- [x] @slevis-lmwg test with f09 offline simulations compared to current ctsm5.2 datasets (control):
1850 spinup and historical (https://github.com/NCAR/LMWG_dev/issues/59)
SSP245 (didn't do)
- [x] Make all fsurdat/landuse files; name them based on the branch tag used to create them (WON'T DO as part of this issue).

slevis-lmwg:
@wwieder pinging you to make you aware of this discussion with @lawrencepj1 and @ekluzek

slevis-lmwg:
I'm adding my part of the work to the Sprint board.
(Mentioning this because the issue is assigned to me and @lawrencepj1.)

lawrencepj1:
Thanks @slevis-lmwg @ekluzek and @wwieder

The new datasets have improved cropping distributions as well as the capability for reforestation and bioenergy with carbon capture and storage (although the carbon capture will need further work). The updated crop distributions can be seen in the attached power point slides.

https://docs.google.com/presentation/d/1886QjkvUfYFAZRB-LPykir2We568ah1-/edit?pli=1#slide=id.p1

Cheers
Peter

[...]

PR #3023, opened 2025-03-20
Get pFUnit-based unit tests working on my Mac

### LINKED ISSUE #2453: FUNIT test not working on izumi
### Brief summary of bug

The FUNIT test is no longer working on izumi. This doesn't seem to be related to a code change, but rather to a change on izumi itself, as previous baselines which passed now also fail.

### General bug information

Fails with these errors:
```
2024-04-04 13:47:39: Compiler is intel
/scratch/cluster/afoster/tests_0404-115752iz/FUNITCTSM_P1x1.f10_f10_mg37.I2000Clm50Sp.izumi_intel.GC.0404-115752iz_int/bld/env_mach_specific.xml already exists, delete to replace
Setting resource.RLIMIT_STACK to -1 from (-1, -1)
Using PFUNIT_PATH: /fs/cgd/csm/tools/pFUnit/pFUnit4.7.0_izumi_Intel20.0.1_noMPI_noOpenMP
Setting NETCDF environment variable: /usr/local/netcdf-c-4.7.4-f-4.5.2-intel-cluster-20.0.1
==================================================
Running make for __command_line_test__/__command_line_test__.
==================================================
ERROR: Command: 'make -j 8' failed with error 'make: *** No targets specified and no makefile found.  Stop.' from dir '/scratch/cluster/afoster/tests_0404-115752iz/FUNITCTSM_P1x1.f10_f10_mg37.I2000Clm50Sp.izumi_intel.GC.0404-115752iz_int/bld/__command_line_test__/__command_line_test__'

 ---------------------------------------------------
2024-04-04 13:47:39: ERROR: RUN FAIL for FUNIT
 ---------------------------------------------------
```

**CTSM version you are using:** ctsm5.1.dev175

**Configurations affected:** FUNIT test on izumi

billsacks:
### Description of changes

With recent versions of CTSM, the unit test build wasn't working on my Mac, and @adrifoster was having the same problem. The problem was that the unit test build wasn't finding the esmf library. I don't understand how this was working on derecho; my best guess is that the modules are adding something needed to the link line that isn't added on a simpler system that doesn't use modules.

This PR explicitly finds ESMF and adds necessary includes and link line flags. This borrows from what's done for the share unit test build, though differs somewhat from that. This also removes some things that are unneeded. It also removes `shr_pio_mod.F90` from the unit test build. I believe that that change (credit to @adrifoster ) allows the unit test build to work in either of two configurations: either with a separately-built PIO library (which is now the recommended configuration for CESM) or without that and instead building ESMF with its internal PIO library (which is the default build for ESMF and is simpler to set up). Some of the other diffs in this PR were also made with that goal of supporting both of these configurations. Thus, PIO isn't referenced directly in the unit test build; instead, if it's needed, it will be picked up via ESMF's link libraries (in the line `link_libraries(${ESMF_INTERFACE_LINK_LIBRARIES})`; note that that line adds some duplicate things to the link line, which results in some warnings, but this doesn't seem to cause any issues).

I have tested this on my Mac (with the current version of https://github.com/billsacks/mac_cime_configuration - `37262a9bca308179e50524d39c9e45e1b205bb18`) and on derecho. Unit tests build in both places. (The first commit on this branch worked on my Mac and on derecho, but not on Adrianna's Mac, where she didn't have a separate PIO library. The second commit worked on my Mac and Adrianna's Mac but not on derecho. The third - and latest - commit works on my Mac and derecho; I'm hopeful it will work for Adrianna, too.)

(I had some related discussion with @ekluzek a few months ago: https://github.com/ESCOMP/CTSM/pull/2853#discussion_r1823145082 . He mentioned needing to remove the find of ESMF from CTSM's unit test build, but things are working for me now that it's back.)

### Specific notes

  Fixes #2453

Testing performed, if any:
- pFUnit-based unit testing on derecho (using the instructions in `README.unit_testing`)
- pFUnit-based unit testing on my Mac

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,8 +25,13 @@ find_package(ESMF REQUIRED)
 # This adds include directories needed for ESMF
 set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${ESMF_F90COMPILEPATHS} ")
 # This (which is *not* done in the share CMakeLists.txt) adds all directories and
-# libraries needed when linking ESMF, including any dependencies of ESMF (but it does
-# *not* include the "-lesmf" itself).
+# libraries needed when linking ESMF, including any dependencies of ESMF. (But note that
+# this does *not* include the "-lesmf" itself). In particular, note that this includes any
+# link flags needed to link against PIO, which is needed on some systems (including
+# derecho); bringing in these PIO-related link flags via this ESMF mechanism allows us to
+# avoid explicitly including PIO as a link library, which wouldn't work on systems where
+# there is no separate PIO library and instead ESMF is built with its internal PIO
+# library.
 link_libraries(${ESMF_INTERFACE_LINK_LIBRARIES})
 
 # Add source directories from other share code (csm_share, etc.). This should be
@@ -70,10 +75,10 @@ foreach (sourcefile ${share_sources})
   endif()
 
   # Remove shr_pio_mod from share_sources. This is needed to avoid an explicit dependency
-  # on PIO. It appears that this removal is needed on some systems but not on others: the
-  # unit test build works without this removal on a Mac with a pre-built PIO library, but
-  # I think didn't work on a Mac without a pre-built PIO (where ESMF was built with its
-  # internal PIO).
+  # on PIO. This removal is needed on some systems but not on others: the unit test build
+  # works without this removal on a Mac with a pre-built PIO library, but failed (with
+  # error message, "Cannot open module file 'pio.mod'") on a Mac without a pre-built PIO
+  # (where ESMF was built with its internal PIO).
   string(REGEX MATCH "shr_pio_mod.F90" match_found ${sourcefile})
   if(match_found)
     list(REMOVE_ITEM share_sources ${sourcefile})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -17,21 +17,6 @@ set(CLM_ROOT "..")
 # NetCDF is required -- because PIO and NetCDF are required by the standard default ESMF libraries
 find_package(NetCDF 4.7.4 REQUIRED Fortran)
 
-# The following - for finding PIO - is copied from the share CMakeLists.txt
-if (DEFINED ENV{PIO_ROOT})
-  message("PIO_ROOT is $ENV{PIO_ROOT}")
-else()
-  if (DEFINED PIO)
-    set(PIO_PATH ${PIO})
-  else()
-    set(PIO_PATH $ENV{PIO})
-  endif()
-  find_package(PIO REQUIRED COMPONENT C Fortran PATH ${PIO_PATH})
-endif()
-# The following line differs a bit from what's in the share CMakeLists.txt
-include_directories(${PIO_Fortran_INCLUDE_DIR})
-link_libraries(${PIO_Fortran_LIBRARIES})
-
 # The following - for finding ESMF - is copied from the share CMakeLists.txt
 if (DEFINED ENV{ESMF_ROOT})
   list(APPEND CMAKE_MODULE_PATH $ENV{ESMF_ROOT}/cmake)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -14,10 +14,21 @@ include(CIME_utils)
 set(CLM_ROOT "..")
 
 # find needed external packages
-# This is where ESMF could be asked for, but it's already included in the share build brought in below
 # NetCDF is required -- because PIO and NetCDF are required by the standard default ESMF libraries
 find_package(NetCDF 4.7.4 REQUIRED Fortran)
 
+# The following - for finding ESMF - is copied from the share CMakeLists.txt
```
ekluzek, line 20: Excellent, I was expecting to need to have this when I initially got this working on Derecho. It always seemed to conflict with elsewhere were ESMF was defined. And I couldn't get it to work with something like this there. In that PR none of us where perfectly satisfied, but since it worked, we went with it. I'm glad to see the proper treatment in place now.

And by the way, this fixes #2452 which is awesome!

src/CMakeLists.txt
```diff
@@ -45,29 +56,32 @@ add_subdirectory(${CLM_ROOT}/src/self_tests clm_self_tests)
...
 foreach (sourcefile ${share_sources})
+  # Remove shr_mpi_mod from share_sources.
+  # This is needed because we want to use the mock shr_mpi_mod in place of the real one
   string(REGEX MATCH "shr_mpi_mod.F90" match_found ${sourcefile})
   if(match_found)
     list(REMOVE_ITEM share_sources ${sourcefile})
   endif()
-endforeach()
 
-# Bring in PIO, jsut because it's needed for the default ESMF library and included in other submodules like share and cmeps
```
ekluzek, line 61: So does the ESMF library used already have PIO built with it -- so this wasn't needed?

src/CMakeLists.txt
```diff
@@ -45,29 +56,32 @@ add_subdirectory(${CLM_ROOT}/src/self_tests clm_self_tests)
...
-if (DEFINED PIO)
-  set(PIO_PATH ${PIO})
-else()
-  set(PIO_PATH $ENV{PIO})
-endif()
+  # Remove shr_pio_mod from share_sources. This is needed to avoid an explicit dependency
```
ekluzek, line 72: Hmmm, since the following isn't required -- I wonder about removing it? It does seem like it's reasonable to expect that the builds of ESMF used for PFunit testing have PIO built in -- right?

It's good for us to use code in shr_pio_mod when we can, so I'd rather have it included, unless it's really necessary to remove.

src/CMakeLists.txt
```diff
@@ -77,24 +91,13 @@ add_dependencies(clm csm_share esmf)
 # We need to look for header files here, in order to pick up shr_assert.h
 include_directories(${CLM_ROOT}/share/include)
 
-
```
ekluzek, line 80: Happy, to see the explicit link with PIO go. But, I think we should have some notes about needing to use an ESMF version that includes PIO pre-built as part of it.

src/CMakeLists.txt
ekluzek: I also think that we should state that we expect PIO to be pre-built into the ESMF version used here. I mention this below. But, this might need a discussion.

ekluzek (changes requested):
@billsacks excellent, so glad you were able to spend some time to figure this out. I have a some small requests and a few questions. It might be best to discuss, but let me know.

I ran this on Derecho and Izumi and it works for me! I did have to delete the previous unit_tests.temp build directory.

I was going to try it on my Mac as well, but I need to do the setup to get the CESM build to work so that's a bigger thing to do. So will likely do that later...

src/CMakeLists.txt
ekluzek: It works both ways.

src/CMakeLists.txt
ekluzek: In talking this over we are leaving this in, because it would complicate the build for supporting ESMF with internal PIO. Longer term we should maybe require having ESMF built with pre-built PIO, but this complicates the CESM build. And as I/O is something we try to avoid in unit testing it makes sense to just do this.

ekluzek (approved):

adrifoster (approved):
Thanks for figuring this out @billsacks!

I just tested it on my mac and it works as is.

I also tried removing the block to remove `shr_pio_mod` (as @ekluzek asks) and it DID NOT work:

```
/Users/afoster/Documents/ncar/CTSM/share/src/shr_pio_mod.F90:2:7:

    2 |   use pio, only : iosystem_desc_t
      |       1
Fatal Error: Cannot open module file 'pio.mod' for reading at (1): No such file or directory
```
this is the error that prompted us to remove this block.

PR #3036, opened 2025-03-27
Make snow resetting dependent on glacier region

billsacks:
### Description of changes

For now this is a hack, hard-coded to only do snow resetting over Greenland. If we want this behavior long-term, we should introduce a new glacier region behavior namelist flag controlling this behavior; this would require changes in glcBehaviorMod.F90 in addition to other changes to add the new namelist flag. (The changes in this commit outside of glcBehaviorMod.F90 could remain as they currently are.)

I am opening this as a draft PR to share these changes, because @gustavo-marques wants them for some test runs, and others have asked for a change like this in the past. If there's some desire to have this long-term, then we should add a namelist flag to replace the hard-coded behavior. If there's no desire to have it long-term, then we can close it.

### Specific notes

Testing performed, if any:
- SMS_D_Ld3_P8x1.f10_f10_mg37.I2000Clm50BgcCropQianRs.green_gnu.clm-default: passes and bit-for-bit with baseline
- SMS_D_Ld3_P8x1.f10_f10_mg37.I2000Clm50BgcCropQianRs.green_gnu.clm-default with `reset_snow = .true.` and `reset_snow_glc = .true.`: passes; manual inspection of SNOW_DEPTH suggests this new code is working as intended (over Greenland, looks like baseline with these reset_flags, elsewhere looks like baseline without these reset flags)

### DIFF
--- a/src/biogeophys/HydrologyNoDrainageMod.F90
+++ b/src/biogeophys/HydrologyNoDrainageMod.F90
@@ -29,6 +29,7 @@ Module HydrologyNoDrainageMod
   use WaterStateBulkType    , only : waterstatebulk_type
   use WaterDiagnosticBulkType    , only : waterdiagnosticbulk_type
   use CanopyStateType   , only : canopystate_type
+  use glcBehaviorMod    , only : glc_behavior_type
   use LandunitType      , only : lun                
   use ColumnType        , only : col                
   use TopoMod, only : topo_type
@@ -141,7 +142,7 @@ contains
        atm2lnd_inst, soilstate_inst, energyflux_inst, temperature_inst, &
        water_inst, &
        soilhydrology_inst, saturated_excess_runoff_inst, infiltration_excess_runoff_inst, &
-       aerosol_inst, canopystate_inst, scf_method, soil_water_retention_curve, topo_inst)
+       aerosol_inst, canopystate_inst, scf_method, soil_water_retention_curve, topo_inst, glc_behavior)
     !
     ! !DESCRIPTION:
     ! This is the main subroutine to execute the calculation of soil/snow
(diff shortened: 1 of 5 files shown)


dlawrenncar:
Do we have any sense about whether or not the runoff fluxes from nearby regions, say the Canadian archipelago, which may have snow capping and which also dump water into the critical region are an issue?  Looking at [maps](https://webext.cgd.ucar.edu/BLT1850/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.138/lnd/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.138_15_34-b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.133_21_40/set2/set2_ANN_H2OSNO.gif), it doesn't seem like there is as much snow in those regions, but also snow depth is [high](https://webext.cgd.ucar.edu/BLT1850/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.138/lnd/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.138_15_34-b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.133_21_40/set3/set3_landf_Canadian_Arctic.gif) for the Canadian archipelago as an average, which must mean I think that there are some snow capped regions.   But, it is a LOT less snow than [Greenland](https://webext.cgd.ucar.edu/BLT1850/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.138/lnd/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.138_15_34-b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.133_21_40/set3/set3_landf_Greenland.gif), so maybe it would have a much smaller effect (?)

billsacks:
At the request of Cecile, I just rebased this to ctsm5.3.023, since that's the version she's been using in recent coupled runs.

billsacks:
This didn't solve the Lab Sea freezing problem (see https://github.com/NCAR/cesm_dev/issues/64). For the sake of testing the Lab Sea freezing problem, I could make an alternative branch that resets snow north of some latitude (say, 60°N) rather than doing this based on glacier region... at least, I could do a hacky version of that (coming up with a long-term version would require more discussion about the desired namelist interface). Let me know if that's desired.

I am fine with this PR being closed as unmerged, unless anyone sees some reason why this would be a useful feature to have long-term, given that it has now been requested a couple of times. If putting this in place, it would need some additional work to make this namelist-controllable.

PR #3056, opened 2025-04-08
QOL improvements in mksurfdata_esmf makefile

samsrabin:
### Description of changes

Adds two quality-of-life improvements in the Makefile for `mksurfdata_esmf`:
- Error quickly if tool hasn't yet been built.
- Look for project number in `$PROJECT` env variable before checking `~/.cesm_proj`, standardizing on the CIME behavior.

### Specific notes

**CTSM Issues Fixed:** None

- Tested as part of #2737.
- Re-tested at `/glade/derecho/scratch/samrabin/tests_0515-112337de/`


tools/mksurfdata_esmf/Makefile
```diff
@@ -1,5 +1,7 @@
 # -*- mode:Makefile -*-
 #
+# Running with this Makefile requires that you build the code first.
```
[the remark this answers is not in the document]
ekluzek: See below where I suggest adding the target...

tools/mksurfdata_esmf/Makefile
```diff
@@ -28,14 +30,23 @@
 # Set up special characters
 null  :=
 
+# Check that code has been built
+BUILT = $(shell ls -d tool_bld | wc -l)
```
ekluzek, line 34: I think we should use Makefile type logic to check that the executable is there, rather than this format. The Makefile logic is going to be more robust. The way this will work is that it will only run the build if the executable doesn't exist. Users will have to know to run the build when they change code. We could add more smarts to have it rebuild when the source changes, but this probably fine for now.

``` make
# This should go to the settings section
MKSURFBLD_DIR = tool_bld
MKSURFDATA_EXE = $(MKSURFBLD_DIR)/mksurfdata_esmf

# This should go to the target section
$(MKSURFDATA_EXE):
\tgen_mksurfdata_build
```

NOTE: the \t means the tab character, which is critical in makefiles.

ekluzek (changes requested):
These are nice improvements that help catch helpful issues for users.

I suggest using Make logic to run the build, rather than just report on it not existing. It's actually simpler and likely more robust as it's built into Make.

tools/mksurfdata_esmf/Makefile
samsrabin: This results in the following error:
```
./gen_mksurfdata_build
Build directory exists so will skip the configure and cmake steps...
Use existing ../pio_iotype.txt file
./gen_mksurfdata_build: line 150: ./.env_mach_specific.sh: No such file or directory
Error sourcing the env_mach_specific.sh file
make: *** [Makefile:118: tool_bld/mksurfdata_esmf] Error 1
```

samsrabin: Ope, never mind.

samsrabin: Resolved with 9b3698b52.

samsrabin: @ekluzek That build doesn't seem to trigger when I do e.g. `make all`.

samsrabin: I think the problem is that the Makefile isn't smart enough to know that the various rules using the `MKSURFDATA` command do actually depend on the executable.

ekluzek (approved):

PR #3072, opened 2025-04-16
User control over snow thermal conductivity scheme over glaciers

### LINKED ISSUE #3071: Option to use Jordan1991 over glacier land units
Biases in coupled model Greenland snow melt can be resolved with reverting the `snow_thermal_cond_method = 'Jordan1991'`, but it really reduces snow melt over vegetated columns and will likely bring back our dead arctic vegetation issue.  Can we use Jordan over glacier land units and Sturm over all other land units?  

@billsacks, @olyson, @ekluzek would any of you be willing to help with this?  At first we're looking for simple source mods that can be used for a coupled model test.

### LINKED ISSUE #3080: Lake snow thermal conductivity hard coded to Jordan
@billsacks noticed this in #3071 

> It appears in two places for lakes:
> 
> https://github.com/ESCOMP/CTSM/blob/be65fad30a741fc594bca9e1221ffe987e7bf271/src/biogeophys/LakeFluxesMod.F90#L483
> 
> https://github.com/ESCOMP/CTSM/blob/be65fad30a741fc594bca9e1221ffe987e7bf271/src/biogeophys/LakeTemperatureMod.F90#L1184
> 
> (Since I saw that the SoilTemperature code was run over non-lake points, I was curious what happened over lakes, so searched for some of the relevant constants in order to find this code.) 

 _Originally posted by @billsacks in [#3071](https://github.com/ESCOMP/CTSM/issues/3071#issuecomment-2810519516)_

wwieder:
### Description of changes
Test runs with F and B cases for CESM3 show excessive melt and low albedo over Greenland.  Reverting to the Jordan1991 snow thermal conductivity scheme seems to help address these biases, but also appears to slow snowmelt across the Arctic and prevent vegetation growth.   This PR will allow different schemes over glacier and non-glacier land units (Jordan and Sturm, respectively).  

### Specific notes
Current PR is intended for testing only. If this is something we want to merge it's likely better to allow namelist control over schemes used oven glacier and non-glacier land units?

@olyson 

- Fixes #3071 
- Fixes #3080

- Yes, should increase albedo and snowpack over glacier regions

Testing performed, if any:
- running I case to evaluate changes

Additional tasks:
- [x] Fix lake bug Bill identified below
- [x] add namelist option for method used over glacier land units


ekluzek (approved):
This makes sense as a starting hack.

billsacks (approved):
Looks great - thanks for taking care of this @wwieder !

(For future reference there is a `col%lun_itype` variable for convenience that prevents you from needing to get `col%landunit` and then `lun%itype`, but what you have is equivalent and totally fine, too.)

billsacks:
And yes, I agree with this approach:

> it's likely better to allow namelist control over schemes used oven glacier and non-glacier land units?

i.e., having separate namelist variables for those two.

olyson (approved):
Looks good thanks.

wwieder:
@billsacks or @ekluzek can you review this PR to see what I'm missing? 

I've added namelist controls for `snow_thermal_cond_glacier_method` and `snow_thermal_cond_lake_method`.

My intent is for this first PR to go onto b4b_dev.  Then we can have an answer changing tags to turn: 
- Jordan on over glacier land units (assuming this looks good to the ice sheet modelers) and 
- Sturm on for lakes (to be consistent with the broader CLM6 physics scheme)

ekluzek (approved):
@wwieder good job. I think these are the changes that we should bring in now.

There's some refactoring that if this wasn't something needed ASAP that I would ask for.  And I'll create an issue for those things. 

Briefly, for modularity and keeping data limited to where it's needed I'd want to move the namelist items to the modules that use them. This allows science code to be replaced and NOT affect other unrelated CTSM code, as well as restrict where it can get messed up. So move snow_thermal_cond_method and snow_thermal_cond_glacier method from controlMod to a ReadNML subroutine in SoilTemperatureMod. And move snow_thermal_cond_lake_method to a ReadNML subroutine in LakeTemperatureMod.F90.

The other problem is that now we have Jordan and Sturm code and strings copied in various places around the code, and we really want to have a "Snow_Thermal_Conducitvity" class that handles it. An OO class would allow for the two options to be chosen without adding logic to handle them both throughout the code. And it prevents the maintenance of magic strings and copied code all around. So we would have the assurance that we've caught all the places that needed to change as it's only a single place, instead of these three places. We also don't want to propagate a habit of more cut and paste of this code if it's needed in new places in the code as well in the future. We also don't really want the endrun error check to be done for every timestep for every column, as that should be done once at initialization. And finally the Sturm constants are all single precision and for reproducibility we should update them to double, but that will likely change answers.

ekluzek:
@wwieder do you want to run the aux_clm testing for this, or have one of us do it?

wwieder:
Thanks for your suggestions Erik, I think they may be slightly above my skill level at this point.
FWIW I really wanted to be able to create a "Snow_Thermal_Conducitvity" class, but was kind of at a loss for how to implement this (I'm OK to replicate existing code, but struggle to refactor it into something better, sorry).

I'll also let you (or @samsrabin or @olyson) do the aux_clm testing if that's OK? Note this PR should likely be rebases to b4b dev, not main.  Let me know if anyone needs to be added as a collaborator on my end.

ekluzek:
@wwieder I'm not expecting anything to be done now. But, wanted to mention things we should do in the future. And yes, they probably wouldn't be things that you'd be involved in doing.

I'm going to turn the refactoring I suggested into an issue that can be tackled later.

billsacks:
I also took a quick look. I agree that this looks good for now. Thanks a lot for getting this in place, @wwieder !

Adding to @ekluzek 's comments about future changes, the piece of this that stands out to me is the string comparison inside the double loop. If this were changed to an OO approach, that would presumably solve this, but I think there could be some simpler solutions that address this by extracting these calculations to their own subroutines (which wouldn't need to be OO) and pulling the conditional outside of the loop. (Instead, the loop would be inside these new subroutines.) In addition, the string variable could be converted to an integer inside the code and that integer used in comparisons... probably more important if the conditional stayed inside the loop, though I think that's something we try to do in general, e.g., both for performance and so typos lead to compile-time rather than run-time errors.

As with @ekluzek 's comments, these can wait until later, though it would be worth taking a quick look at the timing of the relevant timing section before and after this change to make sure that the current implementation doesn't have too much impact on performance.

ekluzek:
We talked about making sure that _glacier_ was used in the namelist name (I had incorrectly remembered it as having _ice_ in the name), but it has that so this is good to go as is. We sometimes have glc_ or glacier_ used as a prefix, but as we are inconsistent with that usage, just including _glacier_ in the name is fine.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -484,6 +484,14 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
...
 <snow_thermal_cond_method phys="clm6_0">Sturm1997</snow_thermal_cond_method>
 
+<!-- TODO change clm6 to Jordan with answer changing tag for glacier mass ballance -->
+<snow_thermal_cond_glc_method>Jordan1991</snow_thermal_cond_glc_method>
+<snow_thermal_cond_glc_method phys="clm6_0">Sturm1997</snow_thermal_cond_glc_method>
+
+<!-- TODO change clm6 to Sturm with answer changing tag -->
+<snow_thermal_cond_lake_method>Jordan1991</snow_thermal_cond_lake_method>
+<snow_thermal_cond_lake_method phys="clm6_0">Jordan1991</snow_thermal_cond_lake_method>
+
```
ekluzek, line 494: Oh, there is a missing part. This needs to go into the perl script. I forgot about that part...

wwieder: Thanks @ekluzek I think I've added what needs to be included now?  Please review.

ekluzek:
This is passing testing other than the st_archive issue, so I'm merging it in now.

Issue #2851, opened 2024-10-29
Update to CMIP7 LULCC forcings


wwieder:
It looks like a preliminary dataset for LULCC is [available for CMIP7](https://wcrp-cmip.org/cmip7-task-teams/forcings/) (scroll down the table with links)

Specifically, is the [v3.0 dataset](https://aims2.llnl.gov/search?project=input4MIPs&versionType=all&activeFacets=%7B%22source_id%22%3A%22UofMD-landState-3-0%22%7D) any different that what we're already using from TRENDY?
This [github page ](https://github.com/PCMDI/input4MIPs_CVs/issues/136)could be helpful for tracking changes / updates to the dataset.

If this dataset is different from TRENDY, we should migrate to the latest v3 data so we're testing the model with LULCC data that's hopefully consistent with the final CMIP7 data.

@lawrencepj1 can you have a look at this dataset to see if it's different that what we've already integrated into the 5.3 tag?

samsrabin:
See also: [This brief slide deck](https://wcrp-cmip.org/wp-content/uploads/2024/10/S1_03_LChini.pdf) with some info from Louise Chini.

lawrencepj1:
Hi Everyone
Just for reference this is actually "CMIP6Plus" based on the CMIP Forcings meeting this morning. It is a test version that will be updated if needed to CMIP7 in 2025.

https://wcrp-cmip.org/wp-content/uploads/2024/10/2024-10-29_Forcings-drop-in_session.pdf

That said I am very happy to start going through the LUH3 data and getting CTSM53 raw PFT data generated from the new data.
Thanks
Peter

olyson:
Since we now have dynamic urban capability, I wonder if we should include urban in this version.  I'm not sure what was done with urban previously in CMIP6, maybe it was assigned to some other landunit?

lawrencepj1:
Hi @olyson 

Urban is transient in the LUH data and the values from LUH are explicitly put into the raw CLM5 and CTSM53 PFT data. Mksurfdata however does not use the raw PCT_URBAN data. There is a lot more information required for CTSM than provided by LUH so it would not be useful as a source for final surface or landuse timeseries data.

Cheers
Peter

olyson:
[...]

PR #3083, opened 2025-04-23
draft to start updating to CMIP7 data for CTSM5.4

wwieder:
### Description of changes
updating to latest LULCC and urban data for CMIP7

### Specific notes
Data processed by @lawrencepj1 and @olyson 

Fixes #[#2851](https://github.com/ESCOMP/CTSM/issues/2851)

purpose here is just to make ne30 and f09 grids to evaluate the new surface datasets before updating gen_mksurfdata_namelist.xml and creating the 'final' CTSM5.4 datasets


wwieder:
@ekluzek and @olyson can you look to see I updated the xml file correctly?  If it looks OK I'll generate the 1 degree surface datasets.

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
```diff
@@ -253,10 +253,10 @@ version of the raw dataset will probably go away.
     <!-- Lake data from 2018 to 2023 is copied from 2017 -->
     <!-- Urban data from 2016 to 2023 is copied from 2015 -->
```
olyson, line 254: Should remove line 254 soon, since urban data is not longer copied from 2015 in this version.

olyson (approved):
Looks ok to me.  See my comment about line 254

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
wwieder: This is done. Creating surface data now.

ekluzek: Yes, this is OK for this first go through.

We really need to remove the Deracho specific paths here, but it looks like that was there beforehand. This is something we'll want to fix in a future version. And we'll want to move these directories over to the inputdata location.

So good to go for the first pass!

wwieder: agreed, I thought after we evaluate these results we can move raw data to input data and then create the full suite of surface data in one clean go.  This draft PR was mainly to facilitate getting eyes on the xml file quickly so I could create the datasets.

olyson:
I think something went wrong with the PCT_URBAN part of the surface dataset generation for /glade/derecho/scratch/wwieder/mksrf_5.4/surfdata_0.9x1.25_hist_1850_78pfts_c250423.nc.  I'm looking into it.

olyson:
The mksrf_furban_mesh in the namelist needs to be changed from /glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/mappingdata/grids/UNSTRUCTgrid_3x3min_nomask_cdf5_c200129.nc to /glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/mappingdata/grids/UNSTRUCTgrid_0.25x0.25_nomask_cdf5_c200129.nc

The GaoOneill data was at 3x3min, but the CMIP7 data is at 0.25x0.25.

When you regenerate the 1850 dataset can you generate a landuse timeseries as well?  I'd like to look at the transient urban before we commit to doing spinup simulations with the 1850 dataset.

olyson:
The new 1850 surface dataset and landuse timeseries (1850-2023) that you generated (https://github.com/ESCOMP/CTSM/pull/3089#pullrequestreview-2792710063) look reasonable with respect to urban.  I would like to make one change however, that is, eliminate the TBD class prior to 1900.  Even though the percentages are very small, that type really shouldn't exist prior to 1900 as the first skyscrapers weren't built until the 1880s.  So I will generate a new set of raw urban files and then we can re-create the datasets, perhaps after Peter has a look at the datasets also...?

wwieder:
Sounds good. Let me know when it's ready

olyson:
> Sounds good. Let me know when it's ready

I've made the change to TBD and generated a new set of raw urban files, looks good.  They are in the same location with a date of 250423, e.g.,

/glade/campaign/cgd/tss/people/oleson/Dynamic_Urban_Data/CMIP7/urban_properties_CMIP7_ThreeClass_2023_c250423.nc

PR #3089, opened 2025-04-24
Correct urban Mesh file

wwieder:
Correct urban mesh file for CMIP7 data


wwieder:
@olyson and @lawrencepj1 please see surface data and land use time series (the later will migrate when ready)
`/glade/derecho/scratch/wwieder/mksrf_5.4`

datasets were generated here:
`/glade/derecho/scratch/wwieder/ctsm/tools/mksurfdata_esmf`

olyson:
The new urban mesh file looks like it is working properly.  The 1850 surface dataset and landuse timeseries datasets look like I would expect them to.  No more urban in Antarctica!

olyson (commented):
This doesn't seem like it has to have the full path since the file is located in the default path for inputdata.

wwieder:
> This doesn't seem like it has to have the full path since the file is located in the default path for inputdata.

corrected now

PR #3106, opened 2025-05-04
Update externals to cesm3_0_alpha06c

ekluzek:
### Description of changes

Update externals to cesm3_0_alpha06c. Also bring in the revert for the st_archive issue.

### Specific notes

   Brings in the fix for #3198

Testing performed, if any: regular ctsm_sci

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -68,7 +68,7 @@ fxDONOTUSEurl = https://github.com/ESCOMP/mizuRoute
 [submodule "ccs_config"]
 path = ccs_config
 url = https://github.com/ESMCI/ccs_config_cesm.git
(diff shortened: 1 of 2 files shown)


ekluzek:
The main purpose here is to have a set of submodule updates that runs successfully without answer changes, to document what those changes are. Showing this is b4b, allows me to build on this and do the next step, that won't be. And then another step can be done that will show submodule updates beyond that, that don't change answers from the second step.

ekluzek:
aux_clm on izumi and Derecho are almost done. I'm also running ctsm_sci on Derecho as this is a submodule update. It's almost done as well.

PR #3140, opened 2025-05-20
Fix error when docker-image-common.yml runs on new tag

### LINKED ISSUE #3139: docker-image-common workflow fails on tag run
### Brief summary of bug

docker-image-common.yml fails when run due to tag creation. This happened for @ekluzek with ctsm5.3.045. When he later ran it manually, it worked. Based on that as well as the workflow stdout/stderr, I think the issue is that there are two tags being created, and this isn't handled correctly.

### General bug information

**CTSM version you are using:** ctsm5.3.044, ctsm5.3.045

See examples at runs for [ctsm5.3.044](https://github.com/ESCOMP/CTSM/actions/runs/15050417089) and [ctsm5.3.045](https://github.com/ESCOMP/CTSM/actions/runs/15146489708).

samsrabin:
### Description of changes
- Delete an unused and possibly bad output
- Try another way to get one image tag

### Specific notes

**CTSM Issues Fixed:**
- Resolved #3139

**Testing performed, if any:** Tested in a separate repository and it works!

### DIFF
--- a/.github/workflows/docker-image-build.yml
+++ b/.github/workflows/docker-image-build.yml
@@ -8,14 +8,14 @@ on:
       - 'doc/ctsm-docs_container/**'
       - '!doc/ctsm-docs_container/README.md'
       - '.github/workflows/docker-image-ctsm-docs-build.yml'
-      - '.github/workflows/docker-image-build-common.yml'
+      - '.github/workflows/docker-image-common.yml'
 
   pull_request:
     paths:
       - 'doc/ctsm-docs_container/**'
       - '!doc/ctsm-docs_container/README.md'
       - '.github/workflows/docker-image-ctsm-docs-build.yml'
-      - '.github/workflows/docker-image-build-common.yml'
+      - '.github/workflows/docker-image-common.yml'
 
   workflow_dispatch:
 
--- a/.github/workflows/docker-image-common.yml
+++ b/.github/workflows/docker-image-common.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Set image tag for docs build
         id: set-image-tag
         run: |
-          echo "IMAGE_TAG=$(echo '${{ steps.meta.outputs.tags }}' | cut -d',' -f1)" >> $GITHUB_ENV
+          echo "IMAGE_TAG=$(echo '${{ steps.meta.outputs.tags }}' | head -n 1 | cut -d',' -f1)" >> $GITHUB_ENV
       - name: Build docs using container
         id: build-docs
         run: |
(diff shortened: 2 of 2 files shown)


.github/workflows/docker-image-common.yml
```diff
@@ -86,7 +83,7 @@ jobs:
       - name: Set image tag for docs build
         id: set-image-tag
         run: |
-          echo "IMAGE_TAG=$(echo '${{ steps.meta.outputs.tags }}' | cut -d',' -f1)" >> $GITHUB_ENV
+          echo "IMAGE_TAG=$(echo '${{ steps.meta.outputs.tags }}' | head -n 1 | cut -d',' -f1)" >> $GITHUB_ENV
```
samsrabin, line 86: I think the problem was that the delimiter in `steps.meta.outputs.tags` was newline instead of comma. I added `| head -n 1` to handle that. I'm not sure why I had the comma-handling bit (`cut ...`), so I kept that too.

.github/workflows/docker-image-build.yml
```diff
@@ -8,14 +8,14 @@ on:
       - 'doc/ctsm-docs_container/**'
       - '!doc/ctsm-docs_container/README.md'
       - '.github/workflows/docker-image-ctsm-docs-build.yml'
-      - '.github/workflows/docker-image-build-common.yml'
+      - '.github/workflows/docker-image-common.yml'
```
samsrabin, line 11: Had the wrong path here.

.github/workflows/docker-image-build.yml
```diff
@@ -8,14 +8,14 @@ on:
       - 'doc/ctsm-docs_container/**'
```
ekluzek, line 8: I think we might want to have a 

``` yaml
branches:
   - master
   - b4b-dev
```

Above this so it only runs on push to master and b4b-dev? Which would mean it wouldn't run when tags are made, which is probably what we want, because the tag almost always will be identical to the PR merged to master. I wasn't sure we want it on b4b-dev either, but because changes like this will go on b4b-dev, you probably do want it to run.

ekluzek (approved):
This looks right to me as well. The pathname fix is obviously required. And I looked at the change by hand, and think it makes sense to me, in removing the newline from the tagnames.

The one thing I wonder about though -- do we WANT the workflow to run on both push to master and the tag made immediately afterwards? And do we want it to run on push to b4b-dev? I'll let you think on that and decide, so approving.

.github/workflows/docker-image-build.yml
samsrabin: I think it makes sense for this workflow to run whenever the listed files are changed, regardless of what branch the changes are on. Remember, this is not the **publish** workflow; this only checks that the image can be built. Also note that, because this workflow runs on every pull request, the "run on push" only ever comes into play when someone pushes changes in those files **to a branch without doing a PR**.

The "run when tags are made" behavior will be addressed in a forthcoming PR.

PR #3143, opened 2025-05-20
Don't run workflows on plain tag creation

### LINKED ISSUE #3142: Workflows run on new tags even w/o file changes
### Brief summary of bug

Some workflows currently run on the push of a new tag, even if there are no changes associated with the tag. This is duplicative, because the workflows will have already run before the tag was created.

### General bug information

**CTSM version you are using:** ctsm5.4.045

From the Actions tab:

Image

samsrabin:
### Description of changes

Some workflows currently run on the push of a new tag, even if there are no changes associated with the tag. This is duplicative, because the workflows will have already run before the tag was created (either due to `push:` or `pull_request:` conditions).

This workflow makes it so that the push of a new tag will not trigger workflows unless it comes with actual changes to files.

### Specific notes

**CTSM Issues Fixed:**
- Resolves #3142

**Testing performed, if any:** Tested in a separate repo and it seems to work as expected.

### DIFF
--- a/.github/workflows/docker-image-build.yml
+++ b/.github/workflows/docker-image-build.yml
@@ -4,6 +4,7 @@ name: Test building ctsm-docs Docker image and using it to build the docs
 # Configures this workflow to run every time a change in the Docker container setup is pushed or included in a PR
 on:
   push:
+    branches: ['*']
     paths:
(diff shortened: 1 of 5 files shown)


ekluzek (approved):
Yep, looks good.

PR #3146, opened 2025-05-23
ctsm5.3.048: Automatically publish docs to this repo

samsrabin:
### Description of changes

- Enables automatic publication of documentation to this repo's GitHub Pages, making ctsm-docs repo obsolete.
- Updates documentation documentation.

### Specific notes

**Contributors other than yourself, if any:** None

**CTSM Issues Fixed:**
- Resolves #2839
- Resolves #2541
- Resolves #3121

**Are answers expected to change (and if so in what way)?** No

**Any User Interface Changes (namelist or namelist defaults changes)?** Not to the model, no.

**Does this create a need to change or add documentation? Did you do so?** Yes; yes.

**Testing performed, if any:** See checks.

### Remaining work
- [x] Update documentation about building multiple versions
- [x] Replace all references to `ctsm-docs` repo
- [x] Add all `include::`ed files as triggers for docs workflows
- [x] `build-docs-omnibus-test` is expected to fail until `versions_list.py` is updated to point to clm5.0 release branch

**Waiting on b4b-dev merge**
- [x] Merge in latest master tag
- [x] Point this PR at master instead of b4b-dev

### After merging
- Remove redirect from https://escomp.github.io/ctsm to https://escomp.github.io/ctsm-docs
- Replace ctsm-docs and docs-deploy-testing Pages with just a custom 404 page pointing to this repo's Pages
- Add issue about improving `testing.sh` and breaking its workflow up into different jobs.

PR #3153, opened 2025-05-24
ctsm5.3.049: Preferentially use Podman for ctsm-docs container

samsrabin:
### Description of changes

Updates doc-builder to a version that prefers to use `podman` instead of `docker`. Also updates documentation to instruct docs-writers to use Podman instead of Docker, along with some other improvements.

- [x] Try brew-then-Desktop method for Mac
- [x] Automatically `podman machine start` if `Cannot connect to Podman` in stderr
- [x] Get multi-arch support working
- [x] Check https://www.redhat.com/en/blog/run-containers-mac-podman for tips
- [x] Merge in latest master tag after #3146 comes in.
- [x] Update all docs links to point to (case-sensitive!) https://escomp.github.io/CTSM
- [x] Add Windows instructions for Podman

### Specific notes

**CTSM Issues Fixed:** None

**Does this create a need to change or add documentation? Did you do so?** Yes; yes

**Testing performed, if any:** Documentation builds as expected.


ekluzek (approved):
@samsrabin and I went over this, and the changes make sense. Some references to docker are changed to container or podman. The github runners need to use the docker Linux command line interface which is already valid under the license, since it's just using DockerEngine which is open-source under Apache 2.0. So the use of dockerfile is acceptable also for example.

I searched for docker in this version, and thought the uses of it made sense, since they are mostly in the .github workflow section.

I did notice that LILAC apparently uses docker, so there might be changes to the instructions for LILAC that we need to make. But, I'll open an issue for that.

This is a nice update that helps us to feel confident that we aren't inappropriately using something that we need to buy a license for.

PR #3125, opened 2025-05-10
ctsm5.3.051: Update submodules to cesm3_0_beta06 + MEGAN namelist (answer change)

ekluzek:
### Description of changes

Update submodules to cesm3_0_beta06

This starts from #3111 which brings in the answer changes for derecho_intel, by updating the compiler to use the intel-oneapi backend.

Contributors other than yourself, if any:

CTSM Issues Fixed (include github issue #):
   Fixes #2710
   Fixes #2476
   Fixes #3135
   Fixes #3108
   Address some things in #3156

Are answers expected to change (and if so in what way)? Yes
  derecho_intel and derecho_nvhpc

Any User Interface Changes (namelist or namelist defaults changes)? No

Does this create a need to change or add documentation? Did you do so? No No

Testing performed, if any: Running regular testing ctsm_sci and fates test lists

PR #3211, opened 2025-05-31
ctsm5.3.055: Remove broken FTorch submodule

### LINKED ISSUE #3210: Fix build-and-deploy failure from ctsm5.3.053
And figure out why this wasn't caught by omnibus script!!

https://github.com/ESCOMP/CTSM/actions/runs/15365441594/job/43237738393

### LINKED ISSUE #3214: Build fails with FTorch checked out
### Brief summary of bug

The CTSM build fails if FTorch is checked out.

### General bug information

**CTSM version you are using:** ctsm5.3.051

### Details of bug

There's this in `libraries/FTorch/buildlib`:
```python
torch_dir = case.get_value("TORCH_DIR")
torch_library = os.path.join(torch_dir, "lib", "libtorch_cpu.so")
```

ctsm5.3.051 brought this in (#3125), but didn't bring in the CMEPS update that introduced `TORCH_DIR`. The above code thus fails in `os.path.join`.

ctsm5.3.051 brought in cmeps1.0.48, which based on the "cmeps1.0.47" in the title of the [CMEPS PR bringing in FTorch](https://github.com/ESCOMP/CMEPS/pull/565) should have been enough. However, that PR actually came in on cmeps1.0.51.

Do `bin/git-fleximod update -o` and try to build the model. Tested with `SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.izumi_nag.clm-crop`.

From `bld/FTorch.bldlog`:
```
expected str, bytes or os.PathLike object, not NoneType
```

samsrabin:
### Description of changes

As of ctsm5.3.051 (#3125), `bin/git-fleximod checkout -o` [results in an unclean state](https://github.com/ESCOMP/CTSM/issues/3210#issuecomment-2925808249), which [prevents the documentation from deploying](https://github.com/ESCOMP/CTSM/actions/runs/15365441594/job/43237738393). This PR resolves the issue. It also edits a couple of GitHub workflows to catch such problems in the future.

Additionally, ctsm5.3.051 did not bring in the version of CMEPS required for FTorch to build properly—see #3214. Fixing it is going to be a big task, so for now we're just removing it. Issue to add it back: #3228

### Specific notes

**CTSM Issues Fixed:**
- Resolves #3210
- Resolves #3214

**Testing performed, if any:** GitHub workflows, plus:
- [x] `SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.izumi_nag.clm-crop` to check build

### DIFF
--- a/.github/workflows/docs-omnibus.yml
+++ b/.github/workflows/docs-omnibus.yml
@@ -51,4 +51,4 @@ jobs:
       # TODO: Split testing.sh tests into their own steps in this job
       - name: Text Sphinx builds with omnibus script
         run: |
-          cd doc && conda run -n ctsm_pylib ./testing.sh
+          cd doc && ./testing.sh
--- a/.github/workflows/fleximod_test.yaml
+++ b/.github/workflows/fleximod_test.yaml
@@ -32,4 +32,4 @@ jobs:
           echo
           echo "Checking if git fleximod matches expected externals"
           echo
-          git diff --exit-code
+          git add . && git diff --quiet && git diff --cached --quiet
--- a/.github/workflows/fleximod_test.yaml
+++ b/.github/workflows/fleximod_test.yaml
@@ -32,4 +32,4 @@ jobs:
           echo
           echo "Checking if git fleximod matches expected externals"
           echo
-          git add . && git diff --quiet && git diff --cached --quiet
+          git add . && git diff --exit-code && git diff --cached --exit-code
--- a/.github/workflows/fleximod_test.yaml
+++ b/.github/workflows/fleximod_test.yaml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
       - id: run-fleximod
         run: |
-          $GITHUB_WORKSPACE/bin/git-fleximod update
+          $GITHUB_WORKSPACE/bin/git-fleximod update -o
           echo
           echo "Update complete, checking status"
           echo
--- a/doc/testing.sh
+++ b/doc/testing.sh
@@ -38,7 +38,7 @@ rm -rf _build_container
 # Check that Makefile method works
 echo "~~~~~ Check that Makefile method works"
 rm -rf _build
-make SPHINXOPTS="-W --keep-going" BUILDDIR=${PWD}/_build html
+conda run -n ctsm_pylib make SPHINXOPTS="-W --keep-going" BUILDDIR=${PWD}/_build html
 
 # Check that -b works
 echo "~~~~~ Check that -b works (Podman)"
--- a/doc/testing.sh
+++ b/doc/testing.sh
@@ -2,6 +2,10 @@
 set -e
 set -x
 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "${SCRIPT_DIR}"
+
+../bin/git-fleximod update -o
 rm -rf _publish*
 
 # Build all docs using container
--- a/.gitmodules
+++ b/.gitmodules
@@ -128,13 +128,3 @@ fxtag = v2.2.6
 fxrequired = ToplevelOptional
 # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
 fxDONOTUSEurl = https://github.com/ESMCI/doc-builder
-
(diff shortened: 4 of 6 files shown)


ekluzek (dismissed):
This is great and good to come in now. This was a great catch by @samsrabin and good to have in place.

ekluzek:
Oh, shouldn't this be going to b4b-dev?

samsrabin:
Thanks @ekluzek! Per discussion just now we'll put this on master. I'll wait until #3212 before doing this, just because I tested that one without this. I don't think there'll be side effects of this but who knows.

samsrabin:
#3214 means this is more complicated now. I'm passing it off to @ekluzek to fix, because I think updating CMEPS to cmeps1.0.51 is going to require other submodule updates.

wwieder:
Hey all, I think this is good work, but I don't really understand why it's suddenly at the top of the project board (not where on our sprint planning board)?   It seems like this is ready to go, and OK to move it.  But I'm kind of surprised it was prioritized when it's not really a required focus for CESM3 / CLM6 release.

wwieder:
Reading this a little more carefully it seems like this is fixing submodules needed for documentation, which is required for CLM6. I guess I was confused by the FTorch capabilities (which don't seem critical for CESM3).

samsrabin:
@wwieder Yeah, Erik thought it would be easy to bring in, so he did (ctsm5.3.051) but later I realized it was broken. The decision from this morning is we're just going to remove it. Changing the title of this PR to reflect that.

.github/workflows/fleximod_test.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
       - id: run-fleximod
         run: |
-          $GITHUB_WORKSPACE/bin/git-fleximod update
+          $GITHUB_WORKSPACE/bin/git-fleximod update -o
```
ekluzek, line 25: Yes, this test should checkout all the optional submodules.

.github/workflows/docs-omnibus.yml
```diff
@@ -51,4 +51,4 @@ jobs:
       # TODO: Split testing.sh tests into their own steps in this job
       - name: Text Sphinx builds with omnibus script
         run: |
-          cd doc && conda run -n ctsm_pylib ./testing.sh
+          cd doc && ./testing.sh
```
ekluzek, line 54: Looks like this was moved to just the part of testing.sh that actually needs it. That makes sense as a change.

doc/testing.sh
```diff
@@ -2,6 +2,10 @@
 set -e
 set -x
 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
ekluzek, line 5: Cool, adding the infrastructure to enable running the script anywhere.

doc/testing.sh
```diff
@@ -2,6 +2,10 @@
...
+cd "${SCRIPT_DIR}"
+
+../bin/git-fleximod update -o
```
ekluzek, line 8: This makes sure that the optional submodules are checked out. Since, doc-builder is needed it's good to make sure it's there.

The caveat to this though is that it will do this as part of running testing.sh could potentially update other submodules and the user could be unaware of that. Since this script is mostly run by git-workflows that might be fine. But, as a developer I don't like running a script that ends up modifying my sandbox. As such I'd prefer if this step were outside the script. But, the script could/should make sure that doc-builder is there and updated appropriately.

ekluzek: Yeah, it looks like the github workflow checks out doc-builder explicitly already. So my preference would be for this to just check that doc-builder is in place, and die if it's not.

samsrabin: All submodules need to be checked out because it's possible we will eventually include files from there in our docs, e.g. via [`literalinclude`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude). Want me to add a comment explaining this?

samsrabin: Oh sorry I get what you were saying now. Yes, let's plan to make that change as part of resolving #3213.

samsrabin: Now [referenced in that issue](https://github.com/ESCOMP/CTSM/issues/3213#issuecomment-2945811430).

PR #3206, opened 2025-05-30
alpha-ctsm5.4.CMIP7.02.ctsm5.3.055: Update default settings

slevis-lmwg:
### Description of changes
This PR starts from the alpha-ctsm5.4.CMIP7 branch at tag `alpha-ctsm5.4.CMIP7.01.ctsm5.3.052`.

As a sanity check, I ran aux_clm on derecho on the tag `alpha-ctsm5.4.CMIP7.01.ctsm5.3.052`. See "Testing" a few lines down for the results.

Definition of done appears in [this Upcoming Tags card](https://github.com/orgs/ESCOMP/projects/46?pane=issue&itemId=112950697). (My branch name references issue3137 that we ended up removing from the definition of done.)

### Specific notes

Contributors other than yourself, if any:
@ekluzek @wwieder 

CTSM Issues Fixed (include github issue #):
Resolves #3116 
Resolves #3005

Are answers expected to change (and if so in what way)? Yes, greater than roundoff, same climate:
1) Even before the changes coming in with this PR, this branch changed the raw datasets in mksurfdata_esmf, so the MKSURFDATAESMF test shows DIFF from baseline (read "Testing" a few lines down).
2) a3ce6a7a703401c75af6c42b91961500080a6f10 changes the default snow thermal conductivity schemes over lakes and glaciers in clm6 cases
3) a75e4888baf30fba9c25f0b6bc0ba606e86100ac introduces a new paramfile to clm6 cases
4) b1890ac5ebbf3fd72c807b54abf7495c5f657591 introduces new f09 & ne30 fsurdat/landuse files
5) 3a8c43262c425882f2ee10686684158ff2349477 introduces new f09 & ne30 finidat files for 1850 and 2000 clm6 cases

Any User Interface Changes (namelist or namelist defaults changes)?
Namelist defaults changes

Does this create a need to change or add documentation? Did you do so?
Probably. Not at this time.

Testing performed, if any:
1) OK aux_clm on derecho for tag `alpha-ctsm5.4.CMIP7.01.ctsm5.3.052`
`FAIL MKSURFDATAESMF_P128x1.f10_f10_mg37.I1850Clm50BgcCrop.derecho_intel` DIFF from baseline I think is expected because this branch has changed the raw datasets used in mksurfdata_esmf.
[...]

PR #3204, opened 2025-05-30
Fire update May 2025

lifang0209:
### Description of changes

- Reduces fire count 99.6% above 2500 meters elevation based on GFED5 elevation bins of burned area, to eliminate overestimation of burning on Tibetan Plateau.
- Updates deforestation fire parameterization based on ctsm5.4 LU and GFED5 burned area.
- Fixes/improves comments

For now, also makes it so that crop fires don't happen if cropland is < 10% of the gridcell. However, this was only a kludge to avoid some mathematical weirdness with interpolation of agricultural peak burning month using average. This will be replaced by a more appropriate interpolation method with #3188, after which maybe remove 10% cropland threshold. However, that will make cropland burned area 10-30 Mha worse.

### Specific notes

**Contributors other than yourself, if any:** None

**CTSM Issues Fixed:** None

**Are answers expected to change (and if so in what way)?** Yes. @lifang0209, can you please add a comment explaining the results before and after?

**Any User Interface Changes (namelist or namelist defaults changes)?**
Namelist defaults:
- `defo_fire_precip_thresh_bet` for CRU-JRA
- `cli_scale` for CRU-JRA and GSWP3
Parameter file:
- `rswf_max(3)`
- `rswf_min(3)`
- `fsr_pft` and `fd_pft`, although these changes cancel each other out

**Does this create a need to change or add documentation? Did you do so?** No

**Testing performed, if any:**
@lifang0209, please add a comment explaining scientific testing.

Before merging:
- [x] `rimport` new parameter file.
- [x] `aux_clm`, expecting diffs only in `li202[14]*` fire methods. (2021 because of paramfile changes to `rswf_min`/`max`; 2024 due to those changes and others.)

PR #3224, opened 2025-06-04
Include abm = 13 values when determining dominant abm in mksurfdata_esmf

### LINKED ISSUE #2663: mksurfdata_jobscript_single.sh fails with the wrong message
### Summary of bug

When mksurfdata_jobscript_single.sh fails and gives errors, it finishes with this deceptive message in the *.o* file:
`Successfully ran resolution`

Presumably
- the same will happen with mksurfdata_jobscript_multi.sh and
- the problem is in this line
`if [ $? != 0 ]; then echo "Error running for namelist  surfdata_0.9x1.25_hist_1700_78pfts_c240725.namelist"; exit -4; fi`
and, if so, the same problem likely applies to the similar line upstream from this one.

IDEAS:
- Gemini suggests that "exit" followed by a negative is non-standard. What if I removed the "-4" from the exit commands? If that works, would it also work with "exit $?"
- I need to introduce an error that will trigger this error check. Possibly leave a raw dataset blank in the namelist?
- THIS ONE: The solution is as simple as changing the message from `Successfully ran resolution` to `Successfully started fsurdat/landuse generation; see .log file for progress`

### General bug information

ctsm5.2.015-6-gdc9996fa3

**Configurations affected:** mksurfdata_esmf only

### LINKED ISSUE #3188: mksurfdata_esmf should interpolate "abm" value 13 among the dominant values to work correctly
> I'm preparing to open a new issue that @lifang0209 brought up. I'm starting the conversation here because it's related. @lifang0209 and everyone, please correct and/or comment on the below:
> 
> @lifang0209 points out that
> - abm is "agricultural fire peak month" and should not be spatially averaged from the raw data to the fsurdat file
> - Instead we should pick the dominant value when interpolating
> - We are not doing this correctly in mksurfdata_esmf and this is the issue that I plan to open
> - And my question: Could this be done correctly if abm were changed to a stream, which is why I'm bringing up the topic here
> 
> Also @lifang0209 has a fix in the fire model to address this until we correct mksurfdata_esmf. In an email @lifang0209 wrote:
> "I checked the f1.9 dataset, and it’s not correct. I’ve added an IF statement in the fire model to assume no crop fire if crop coverage is less than 10%. It's consistent with observations as crop fires are rare in these grids and it also helps resolve the issue." 

 _Originally posted by @slevis-lmwg in [#2617](https://github.com/ESCOMP/CTSM/issues/2617#issuecomment-2917821489)_

slevis-lmwg:
### Description of changes
mksurfdata_esmf interpolates abm (ag. fire peak month) from the raw dataset to fsurdat with a "dominant value" algorithm. However, from what I can tell, the dominant value gets selected only from valid month values (1 through 12) which leads to valid month values appearing in areas of limited crop coverage and crop fires overestimated by approximately 10–20 Mha/year in these regions.

I propose a very simple change in the code. I will not have a chance to try it out before derecho returns to service. My current concern is that abm = 13 is a netcdf Fillvalue in the raw dataset. If that causes trouble, I may need to replace the file's Fillvalue.

- [x] May be more appropriate to rebase to the alpha-ctsm5.4.CMIP7 branch. Currently I'm having trouble with this. It may become easier when I merge #3206 to alpha-ctsm5.4.CMIP7.

### Specific notes

@lifang0209 @samsrabin 

Resolves #3188 
Resolves #2663

Changes fsurdat variable abm

No

No

Testing TODO:
- [x] Generate and compare f19 (and ne30?) fsurdat files with and without the code modifications

### DIFF
--- a/tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
+++ b/tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
@@ -22,6 +22,7 @@ module mkagfirepkmonthMod
 
   integer , parameter :: min_valid = 1  ! month value for January
   integer , parameter :: max_valid = 13  ! value for no agricultural fire
+  integer , parameter :: unsetmon = 14  ! value for no data
 
   type(ESMF_DynamicMask) :: dynamicMask
 
@@ -119,6 +120,10 @@ contains
     call mkpio_get_rawdata(pioid_i, 'abm', mesh_i, idata_i, rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) return
     call ESMF_VMLogMemInfo("After mkpio_getrawdata in "//trim(subname))
+    ! Update idata_i to unsetmon where mask_i == 0, i.e. over ocean
+    do ni = 1, ns_i
+       if (mask_i(ni) == 0) idata_i(ni) = unsetmon
+    end do
 
      ! Create ESMF fields that will be used below
     field_i = ESMF_FieldCreate(mesh_i, ESMF_TYPEKIND_R4, meshloc=ESMF_MESHLOC_ELEMENT, rc=rc)
@@ -167,8 +172,8 @@ contains
 
     ! Check validity of output data
     if (min_bad(agfirepkmon_o, min_valid, 'agfirepkmon') .or. &
-        max_bad(agfirepkmon_o, max_valid, 'agfirepkmon')) then
-        if (root_task) write(ndiag, '(a)') trim(subname)//" error in agfirepkmon_o value range: min_valid and max_valid equal ", min_valid, max_valid
+        max_bad(agfirepkmon_o, unsetmon , 'agfirepkmon')) then
+        if (root_task) write(ndiag, '(a)') trim(subname)//" error in agfirepkmon_o value range: expect min_valid to unsetmon which equal ", min_valid, unsetmon
         call shr_sys_abort()
      end if
 
@@ -183,7 +188,7 @@ contains
 
     ! Output diagnostics comparing global area of each peak month on input and output grids
     call output_diagnostics_index(mesh_i, mesh_o, mask_i, frac_o, &
-         min_valid, max_valid, idata_i, agfirepkmon_o, 'peak fire month', ndiag, rc)
+         min_valid, unsetmon, idata_i, agfirepkmon_o, 'peak fire month', ndiag, rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) call shr_sys_abort()
     
     ! Release memory
@@ -245,7 +250,7 @@ contains
              maxindex = maxloc(wts_o(:)) 
              dynamicMaskList(no)%dstElement = real(maxindex(1), kind=r4)
           else
-             call shr_sys_abort(subname//" error: hasdata needs to be true")
+             dynamicMaskList(no)%dstElement = real(unsetmon, kind=r4)
           end if
        end do
     end if


lifang0209:
@slevis-lmwg  Hi Sam, you mentioned that abm = 13 is a netcdf Fillvalue in the raw dataset.", I don't recall setting 13 as a FillValue in the abm raw data.  If you are sure it is, then simply removing the FillValue setting (or assign another FillValue as you suggested if a FillValue must be specified when making surface data) could solve the dominant value selection issue. For example, if the coarse grid (e.g. 2 deg) has more than half of the 0.5-deg with an abm value of 13, then the dominant method should get abm=13 for the 2-deg grid cell.

samsrabin:
@lifang0209 I guess your proposed handling of 13 makes sense too—if more than half of the contributing gridcells have no crop fire, then the interpolated gridcell shouldn't either. It would be better, though, if we could scale the fraction of cropland that gets burned based on fraction of contributing gridcells with crop fire.

samsrabin:
Ohh, sorry @slevis-lmwg, I misunderstood—it looks like there is already code to get the dominant month, and you're trying to fix it! But then the [issue](https://github.com/ESCOMP/CTSM/issues/3188) title—'mksurfdata_esmf needs to interpolate "abm" with dominant instead of average'—is confusing.

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
```diff
@@ -21,8 +21,8 @@ module mkagfirepkmonthMod
   public  :: mkagfirepkmon       ! Set agricultural fire peak month
 
   integer , parameter :: min_valid = 1
-  integer , parameter :: max_valid = 12
-  integer , parameter :: unsetmon = 13
+  integer , parameter :: max_valid = 13
+  integer , parameter :: unsetmon = 0
```
slevis-lmwg: From meeting with @samsrabin 
1) unsetmon probably unnecessary now, plus error check below will fail with unsetmon = 0
2) add comments in this module to explain things
3) add unit test in PFunit (@samsrabin adding separate comment about this)

samsrabin (changes requested):
It seems like this does address the issue. However, it'd be good to have unit testing of this function to make sure it's behaving properly. I will plan to at least start the file structure needed for unit testing, but after @slevis-lmwg adds comments and improves variable names throughout the subroutine so that I know what's what.

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
```diff
@@ -20,9 +20,9 @@ module mkagfirepkmonthMod
 
   public  :: mkagfirepkmon       ! Set agricultural fire peak month
 
-  integer , parameter :: min_valid = 1
-  integer , parameter :: max_valid = 12
-  integer , parameter :: unsetmon = 13
+  integer , parameter :: min_valid = 1  ! month value for January
+  integer , parameter :: max_valid = 13  ! value for no agricultural fire
+  integer , parameter :: unsetmon = 14  ! maximum value expected in the output file if raw dataset includes values outside the 1 through 13 range
```
slevis-lmwg: @samsrabin I thought I'd try `unsetmon = 14` instead of removing, so as to reduce code changes to a minimum.

slevis-lmwg: Please let me know if you see other pieces of the code where I may try to include comments, keeping in mind that I am not familiar with the ESMF calls and how they work.

samsrabin: I think in this case keeping code changes at a minimum would be detrimental to future work, because it really seems like `unsetmon` should never be needed. Could you try removing it and see if it causes problems or diffs?

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
```diff
@@ -220,6 +225,7 @@ subroutine get_dominant_indices(dynamicMaskList, dynamicSrcMaskValue, dynamicDst
     rc = ESMF_SUCCESS
 
     if (associated(dynamicMaskList)) then
+       ! TODO: Can parts of this loop and/or its subloops be broken out into separate functions/subroutines to facilitate testing?
```
slevis-lmwg: The subroutine get_dominant_indices gets called in a way not familiar to me in this line:
`call ESMF_DynamicMaskSetR4R8R4(dynamicMask, dynamicMaskRoutine=get_dominant_indices,  &
handleAllElements=.true., rc=rc)`
So I could see the subroutine cascading to other subroutines or functions but not the subroutine itself broken up into multiple calls, if that makes sense.

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
```diff
@@ -205,10 +210,10 @@ end subroutine mkagfirepkmon
   subroutine get_dominant_indices(dynamicMaskList, dynamicSrcMaskValue, dynamicDstMaskValue, rc)
 
     ! input/output arguments
-    type(ESMF_DynamicMaskElementR4R8R4) , pointer              :: dynamicMaskList(:)
-    real(ESMF_KIND_R4)                  , intent(in), optional :: dynamicSrcMaskValue
-    real(ESMF_KIND_R4)                  , intent(in), optional :: dynamicDstMaskValue
-    integer                             , intent(out)          :: rc
+    type(ESMF_DynamicMaskElementR4R8R4) , pointer              :: dynamicMaskList(:)  ! TODO: What is this?
```
slevis-lmwg: This gets into ESMF knowledge that I do not have. I came up with best guesses for most your questions, but this one in particular I left blank.

slevis-lmwg:
I'm generating a file without the fix and one with the fix:
```
/glade/work/slevis/git/b4b-dev/tools/mksurfdata_esmf/surfdata_1.9x2.5_SSP3-7.0_2015_16pfts_c250612.nc
/glade/work/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf/surfdata_1.9x2.5_hist_2015_16pfts_c250612.nc
```
to test this PR's code changes.

UPDATE1: No change in the results, so next I will remove FillValue as explained at the top:
`ncatted -h -a _FillValue,,d,, mksrf_abm_0.5x0.5_simyr2000.c240821.nc`
I kept a copy of the original file as mksrf_abm_0.5x0.5_simyr2000.c240821_BKUP.nc rather than changing the date stamp. The new file simply doesn't have a _FillValue, while the old file set _FillValue to 13.

UPDATE2: Still no change... I will need to troubleshoot... See diagnostic output in `/glade/work/slevis/git/mksurfdata_toolchain/tools/mksurfdata_esmf`
but why did write statements not work?

UPDATE3: See latest test using unsetmon = 14 over ocean

slevis-lmwg:
Testing on derecho
PASS python -u and -s
PASS black and lint

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
```diff
@@ -116,6 +120,10 @@ subroutine mkagfirepkmon(file_mesh_i, file_data_i, mesh_o, pioid_o, rc)
     call mkpio_get_rawdata(pioid_i, 'abm', mesh_i, idata_i, rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) return
     call ESMF_VMLogMemInfo("After mkpio_getrawdata in "//trim(subname))
+    ! Update idata_i to unsetmon where mask_i == 0, i.e. over ocean
+    do ni = 1, ns_i
+       if (mask_i(ni) == 0) idata_i(ni) = unsetmon
```
slevis-lmwg, line 125: @samsrabin @lifang0209 
I brought back unsetmon as 14 to distinguish ocean points and thus remove them from the determination of the dominant abm. I think it makes more sense than allowing the lack of peak month over oceans to skew the result.

slevis-lmwg: I tested the change and it works. Please let me know if you have concerns. Otherwise I will merge to the 5.4 branch in the next day or so.

slevis-lmwg: Fang, you can see the new fsurdat here:
/glade/campaign/cesm/cesmdata/cseg/inputdata/lnd/clm2/surfdata_esmf/ctsm5.4.0/surfdata_1.9x2.5_hist_2000_16pfts_c250617.nc

samsrabin: Sounds good to me!

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
slevis-lmwg: Resolving in preparation for merging PR.

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
slevis-lmwg: Resolving in preparation for merging PR.

tools/mksurfdata_esmf/src/mkagfirepkmonthMod.F90
slevis-lmwg: Resolving conversation in preparation for merging PR.

slevis-lmwg:
I made a new branch tag with this merge:
`alpha-ctsm5.4.CMIP7.03.ctsm5.3.055`

lifang0209:
@slevis-lmwg Could you tell me the path of the new 1.9x2.5 surface data file on the derecho generated using your revised code? I’d like to check it and ensure everything looks correct.

slevis-lmwg:
@lifang0209 
derecho and casper seem down right now, but I made this new file while working on this PR:
`/glade/campaign/cesm/cesmdata/cseg/inputdata/lnd/clm2/surfdata_esmf/ctsm5.4.0/surfdata_1.9x2.5_hist_2000_16pfts_c250617.nc`
I also made new fsurdat files for all resolutions in /glade/derecho/scratch/slevis/temp/ctsm5.4/.../tools/mksurfdata_esmf (I don't remember the exact path but it's something like that).

lifang0209:
@slevis-lmwg Thanks, Sam! I checked your 1.9x2.5 NC file, and the results look good. I can now remove the cropland coverage ≥10% condition from [PR #3204].
Since the valid ABM areas (1–12) in your file are smaller than those with cropland ≥10%, I need to recalibrate cropfire_a1. Do you have a  1850 1.9x2.5 78-PFTs surface dataset? I’d like to run a case to recalibrate it using the inverse method.

slevis-lmwg:
> [...] Do you have a 1850 1.9x2.5 78-PFTs surface dataset? I’d like to run a case to recalibrate it using the inverse method.

This should be the file that you need: `/glade/derecho/scratch/slevis/temp_work/ctsm5.4/mk_ctsm54_datasets/tools/mksurfdata_esmf/surfdata_1.9x2.5_hist_1850_78pfts_c250618.nc`

lifang0209:
@slevis-lmwg @samsrabin Using the dominant abm is definitely more reasonable than the earlier approach of averaging abm in regridding. 
However, there is still an issue. In a 1.9x2.5 grid, your method results in invalid abm (e.g., 13 or 14) in grid cells where cropland coverage < 50%, leading some grid cells being unreasonably classified as having no crop fires.
A more reasonable approach when regrid abm from 0.5 deg raw data to coarser grids would be to assign the dominant valid abm value (1, 2, ...,12) in a grid cell if cropland coverage exceeds 10% (or if proportion of valid abm values in the grid cell > 10%), and assigned an invalid abm otherwise.

lifang0209:
@slevis-lmwg @samsrabin Using the dominant abm is definitely more reasonable than the earlier approach of averaging abm in regridding. 
However, there is still an issue. In a 1.9x2.5 grid, your method results in invalid abm (e.g., 13 or 14) in grid cells where cropland coverage < 50%, leading some grid cells being unreasonably classified as having no crop fires.
A more reasonable approach when regrid abm from 0.5 deg raw data to coarser grids would be to assign the dominant valid abm value (1, 2, ...,12) in a grid cell if cropland coverage exceeds 10% (or if proportion of valid abm values in the grid cell > 10%), and assigned an invalid abm otherwise

samsrabin:
@lifang0209  Can you share the script you're using to test this? That will help us solve it once and for all.

slevis-lmwg:
@lifang0209 let me clarify what the abm regridding algorithm **did before** and **does now** and let's try to come up with a solution.
1) Before, we determined dominant abm from values 1 to 12 (min_valid to max_valid). If we found no dominant value, then we set abm to unsetmon = 13. This led to an underestimation of 13, because we used it only when no dominant existed.
2) Now, we use the same algorithm with the following modifications:
- We set abm = 14 in the raw data where LANDMASK = 0 (ocean) in the raw data. We do this in the code, keeping the file unchanged (other than removing the _FillValue attribute from value 13).
- We set min_valid = 1, max_valid = 13, unsetmon = 14. If we find no dominant value from 1 to 13, we set abm to 14.

I have two requests that I can think of:
- I do not want to modify this algorithm unless we find a mistake in it. 
- I do not want to introduce tuning assumptions to mksurfdata_esmf.

slevis-lmwg:
However, I would be open to using an updated raw dataset. Currently the raw dataset has values from 1 to 13, and the code uses 14 over ocean grid cells. You could update the file to set non-croplands (including oceans) to 14. Then the dominant algorithm would find fewer 13s than it does now, so values from 1 to 12 would increase again.

lifang0209:
@slevis-lmwg Changing non-cropland from 13 to 14 in the raw data does not resolve the issue.

fig4
Fig. 1 2001-2020 CTSM5.4 fractional coverage of cropland, and  GFED5 observed crop fires at 1.9x2.5 grids.
 
As shown in Fig. 1. areas with cropland exceeding 50% at coarse grid resolution are small, and crop fires generally occur in grids where cropland coverage exceeds 10%.

fig3
Fig. 2. distribution of valid abm (peak month of crop fires, 1 to 12)  in the CTSM5.4 surface file at 1.9x2.5 grids using the old (average) and new (dominant value) methods, compared with abm raw data at 0.5 deg grids.  

The issue primarily arises in regions with low cropland coverage, and its impact is not small.
 The old one got too many valid abm values in these regions and so around 30% more global crop burned area, because it derive abm by averaging valid abm in the coarse grids (13 is set to FillValue). 
In contrast, the new one get too few valid abm values in these regions because invalid abm dominate, resulting in around 30% less global crop burned area. 

To get reasonable abm and avoid crop fire simulations sensitive to resolution, the methods I could think of:
(1) Assign the dominant valid abm value (1, 2, ...,12) in a grid cell if cropland coverage exceeds 10% (or if proportion of valid abm values in the grid cell > 10%), and assign an invalid abm otherwise.
(2) Use abm which is dominant in valid abm values by setting 13 as Fillvalue, and add an if statement in fire module code to limit crop fires in grids with cropland coverage > 10%.
(3) Provide correct 1.9x2.5 abm data. and apply those values for high-resolution grids within the coarse grid cells.
Option (3) may be simplest.

Please let me know your thoughts or if any part requires further clarification.

lifang0209:
In the 0.5 deg raw data, non-croplands (including oceans) are set to 13, so option (2) works if we set the FillValue attribute from value 13 in mksurfdata_esmf and add a condition statement in the fire code. No need to add setting of 14 for oceans in mksurfdata_esmf. Right?

slevis-lmwg:
@lifang0209 I shared my calendar with you. I think we will resolve this more easily with a conversation because I (or both of us) may be misunderstanding. Feel free to check my availability and send me an invite.

slevis-lmwg:
Also, in your abm plots please show value 13, because it represents "no crop fire" rather than "invalid value" from my understanding and, if so, it is important to see this value's distribution.

slevis-lmwg:
@samsrabin has expressed interest in joining our meeting. 
@lifang0209 please let me know if it's easier if I set up the meeting, in which case tell me some convenient times for you and whether you have access to google meet or prefer zoom.

lifang0209:
@slevis-lmwg Tue, July 1, 5:30–6:30 PM, or Wed, July 2, from 4–5 PM work for me. Which do you prefer? I'm flexible with either Zoom or Google Meet.

slevis-lmwg:
I selected July 2nd, 4-5 pm Mountain Time, with Google Meet.

lifang0209:
@slevis-lmwg Hi Sam, the raw 0.5° abm plots (with the distribution of 13 shown in gray):
fig3

In the raw data, both the ocean and areas without crop fires on land are set to 13.
The new raw data (abm05-250701)  I generated today has a smaller area with valid abm values (1 to 12) because (1) I set 13 for 0.5° grids where cropland coverage in the 1.9°x2.5° grid is less than 10%, and (2) cropland coverage <10% regions are smaller in CTSM5.4 (used for the new abm) compared to CTSM5.2 (used for the old abm).

Now, with the new raw abm data (/glade/u/home/fangle/mksrf_abm_0.5x0.5_simyr2000.c250701.nc), we can retain the high resolution of abm for 0.5° or 1° simulations, avoid regridding issues in low cropland coverage regions, and resolve crop fire simulation sensitivity to resolution.

Remaining tasks:
(1) Make minor modifications in your mksurfdata_esmf file using the new raw abm data: remove the 14 setting and set 13 as FillValue.
(2) I recalibrate cropfire_a1 based on the 1850 1.9x2.5 78-PFTs surface dataset you generate from (1)

This is the best and simplest solution I could think of. Would it be better if we complete the remaining tasks before our meeting?

slevis-lmwg:
Thank you @lifang0209 
The first remaining task reverts the abm regridding algorithm to the original, which does not seem to me as what we want. I will refrain from working on this until we meet and try to clarify everything.

lifang0209:
It's not reverting to the original—dominant value is used rather than the average.
OK, let's meet tomorrow.

slevis-lmwg:
From meeting with @lifang0209 @samsrabin:
@slevis-lmwg to move this conversation to a new issue.

PR #3231, opened 2025-06-05
Clean up docs workflows

### LINKED ISSUE #3160: Need to add the release-clm5.0 branch to your fork for the docs build tests to work
### Brief summary of bug

With the new docs build updates, you need to have the release-clm5.0 branch on your fork for the github workflows to run.

### General bug information

**CTSM version you are using:** ctsm5.3.049-32-g0181deefd

**Configurations affected:** PR's on forks going to master

### Details of bug

See:

https://github.com/ekluzek/CTSM/actions/runs/15285357762

What you need to do to fix this is just add the branch to your personal fork. So for me for example:

``` shell
git checkout -b release-clm5.0 escomp/release-clm5.0
git push ekluzek release-clm5.0
```

@samsrabin

### LINKED ISSUE #3213: Break up "omnibus docs test"
Right now [`doc/testing.sh`](https://github.com/ESCOMP/CTSM/blob/443480112c50d87a2241a54f531eb88278af3abf/doc/testing.sh) is scheduled to run in its entirety whenever the docs are updated, via [`docs-omnibus.yml`](https://github.com/ESCOMP/CTSM/blob/443480112c50d87a2241a54f531eb88278af3abf/.github/workflows/docs-omnibus.yml). This should be broken up for clarity and to minimize the number of GitHub workflow runs that need to happen.

As a first step, each part of the script itself should be broken up into component scripts. This will enable us to keep `testing.sh` around (it's useful for local testing) while also breaking it up into different workflow jobs.

Second, we should exclude changes to `testing.sh` from triggering docs-related workflows.

Third, the "Build all docs using container" test should be added to `docs.yml`. It should run regardless of whether the existing `build-docs` job passes, because theoretically it's possible for either to pass while the other fails.

Finally, `docs-omnibus.yml` should be broken up into two workflows with the following tests from `testing.sh`:
1. `docs-updated-doc-builder.yml`
2. `docs-updated_ctsm_pylib.yml`

These would run only when their respective parts are changed. Both would contain these three parts, which constitute one test (job):
- "Build all docs using container"
- "Build all docs using ctsm_pylib"
- "Make sure container version is identical to no-container version"

They would also both contain a job for "Check that Makefile method works," since it depends on both docs-builder and `ctsm_pylib`.

`docs-updated-doc-builder.yml` would additionally contain the "Check that -r -v works" test. The "Check that -b works" test would be excluded from new GitHub workflows as it's already part of `docs.yml`.

samsrabin:
### Description of changes

Breaks up the "omnibus docs test" GitHub Workflow to minimize the number of workflow runs. Also solves problem where people were getting test failures on their forks.

### Specific notes

**CTSM Issues Fixed:** None
- Resolves #3160
- Resolves #3213

- Local running of `testing.sh`
- Actions on this PR


.github/workflows/docker-image-common.yml
```diff
@@ -76,14 +76,16 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
-      # Try building our docs using the new container
-      - name: Checkout doc-builder external
+      # Check out all submodules because we might :literalinclude: something from one
```
ekluzek, line 79: This is a nice insight. We will likely want to include things from CDEPS for sure, but possibly cime, or CMEPS as well. Thanks for realizing that and bringing this in.

ekluzek: It does end up going into a bunch of places. So I wondered about having a workflow for just it. But, it's also too small to worry about that. So we should leave it as is. It would only be helpful if we were to move away from git-fleximod, which I don't expect doing for a long time.

.github/workflows/docs-omnibus.yml
```diff
@@ -5,39 +5,33 @@ on:
...
-      - 'src/README.unit_testing'
-      - 'tools/README'
+      - 'doc/testing/*'
+      - 'doc/Makefile'
 
   workflow_dispatch:
 
 jobs:
   build-docs-omnibus-test:
+    # Don't run on forks
```
ekluzek: Ahh, I'm glad you have this in place. That will be helpful for outside people doing PR's.

.github/workflows/docs-update-ctsm_pylib.yml
```diff
@@ -0,0 +1,30 @@
+name: Docs tests to run when ctsm_pylib is updated
```
ekluzek, line 1: Hmmm. I'm wondering if this needs it's own workflow file? These requirements could be added elsewhere.

Or is this to isolate this workflow from others to remove complexity in one place?

.github/workflows/docs-build-and-deploy.yml
```diff
@@ -6,12 +6,14 @@ on:
     branches: ['master', 'release-clm5.0']
     paths:
       - 'doc/**'
+      - '!doc/testing/*'
```
ekluzek: I'm glad you are adding a test directory for the documentation build. I only wonder if it's name should be "test" to correspond with the directory names we have under src (such as src/main/test) for PF unit testing as well as the python testing directory under python/ctsm/test? I think that would be my preference.

We could also change the directory name for those to "testing", but I'm unsure if that is completely safe to do...

.github/workflows/docs-update-dependency-common.yml
```diff
@@ -0,0 +1,67 @@
+name: Jobs shared by docs workflows that run when a dependency is updated
+
+on:
+  workflow_call:
+    inputs:
+      conda_env_file:
```
ekluzek, line 9: So the conda input isn't required when a docker image is used to build documentation?

ekluzek: I think this could use a few comments to say what it's doing.

.github/workflows/docs-update-dependency-common.yml
```diff
@@ -0,0 +1,67 @@
...
+        type: string
+        default: ""
+      conda_env_name:
+        required: false
+        type: string
+        default: ""
+    secrets: {}
+
+jobs:
+  compare-docbuilder-vs-ctsmpylib:
```
ekluzek, line 20: Add a comment that this is going to compare the build using doc-build to using ctsm_pylib. Does it check that they are identical or just that both work?

.github/workflows/docs-update-dependency-common.yml
```diff
@@ -0,0 +1,67 @@
...
+          environment-file: ${{ inputs.conda_env_file }}
+          channels: conda-forge
+          auto-activate-base: false
+
+      - name: Compare docs built with container vs. ctsm_pylib
+        run: |
+          cd doc/testing
+          ./test_container_eq_ctsm_pylib.sh
+
+  makefile-method:
```
ekluzek, line 53: OK, a third method is to use the Makefile. A few comments around that would be helpful.

ekluzek: That also makes me wonder if we need to support all three of these methods though: doc-builder, ctsm_pylib, and make? Can any of them be dropped? What is the use-case for each of them?

Adding the testing though should mean that we'll be able to support all of them more easily though. As this testing will be running to make sure they all work all the time.

.github/workflows/docs-update-doc-builder.yml
```diff
@@ -0,0 +1,50 @@
+name: Docs tests to run when doc-builder is updated
+
```
ekluzek, line 2: Ahh, linking this to above .github/workflows/docs-update-ctsm_pylib.yml, this is to isolate the dependency part for each type of thing that needs to be covered, from the actual testing.

I then only wonder if there's more parts at the end that could be put into the common part?

.github/workflows/docs-update-ctsm_pylib.yml
ekluzek: Looking below at the other file doing something similar, it does look like that's the case. This is to put the part about dependencies for a particular thing in one place, and then have the tests run be common to each. This way looking at the list of files you get a good idea of the actions that happen and what triggers them.

This does seem helpful to me. I'm just wondering if more things can go into the common part? But, that's probably something to do later at this stage anyway...

doc/testing/test_build_docs_-b.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
ekluzek: I'm not sure about the "-" dash in the filename. It should be fine, but could it get confused with command arguments?

So would...

test_build_docs_b_option.sh

be better? Or something along those lines?

doc/testing/test_build_docs_-r-v.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -e
```
ekluzek: Same thing here, on the filename as above.

doc/testing/test_build_docs_-b.sh
ekluzek: Actually this also leads to another philosophical question on bash testing. Which is should bash tests be in a single file, or in a subroutine in one file?

doc/testing/test_container_eq_ctsm_pylib.sh
```diff
@@ -0,0 +1,35 @@
...
+echo "~~~~~ Build all docs using ctsm_pylib"
+rm -rf _build _publish
+d2="$PWD/_publish_nocontainer"
+conda run -n ctsm_pylib --no-capture-output ./build_docs_to_publish -r _build --site-root "$PWD/_publish"  --conf-py-path doc-builder/test/conf.py --static-path ../_static --templates-path ../_templates
+# VERSION LINKS WILL NOT RESOLVE IN _publish_nocontainer
+cp -a _publish "${d2}"
+
+# Make sure container version is identical to no-container version
+echo "~~~~~ Make sure container version is identical to no-container version"
+diff -qr "${d1}" "${d2}"
```
ekluzek: You'll need to trap for the error in case they differ.

doc/testing/test_container_eq_ctsm_pylib.sh
```diff
@@ -0,0 +1,35 @@
...
+echo "Yep!"
```
ekluzek: In several places in CTSM and CESM we use the term "Successful", so suggest using something like that here.

doc/testing/testing.sh
```diff
@@ -0,0 +1,33 @@
...
+rm -rf _build
+conda run --no-capture-output -n ctsm_pylib ./test_makefile_method.sh
+
+# Check that -b works
+cd "${SCRIPT_DIR}/"
+rm -rf _build
+./test_build_docs_-b.sh docker
+
+# Check that doc-builder tests pass
+# Don't run if on a GitHub runner; failing 🤷. Trust that doc-builder does this test.
```
ekluzek: Ahh, I'm glad you figured out a way to check for a GitHub runner, so we can turn off things not expected to work.

doc/testing/test_container_eq_ctsm_pylib.sh
ekluzek: Actually, that will need to be done in a lot more places. You should probably run some tests where you make a change so that you know it should fail, and make sure it does.

doc/testing/test_build_docs_-b.sh
ekluzek: And I think if we are going to have more bash testing, we should probably develop some philosophy around how to do it. I'll start a discussion on that. I know there are some frameworks for it, and looked into it a little bit with the testing I did in cime_config/testdefs/testmods_dirs/clm/FatesSetupParamBuild.

I think looking there would be useful also. Some of what we have there should likely be used here. So we probably want a directory for bash unit testing utilities, and make useful things for testing put in that place.

The functions in shell_commands there are candidates for this:

``` shell
fatal_error() {
log_msg_if_debug () {
check_error () {
check_if_runable () {
```

Similarly these functions in run_shell_commands_tests:
``` shell
conda_for_host() {
handle_error() {
expect_fail() {
expect_nofail() {
test_run_shell_commands() {
```

The test structure in the last one brings us closer to how the python unit testing is done, which is helpful in getting people up to speed on how to do this type of thing for bash. 

A lot of this might be beyond the scope of this PR, but we should start thinking about this. So I'll turn this into a discussion.

.github/workflows/docs-update-dependency-common.yml
ekluzek: From below I see that it checks that they are identical, which I'm impressed with. Glad you could do that!

.github/workflows/docs-update-dependency-common.yml
samsrabin: I do want to drop the Makefile, but it will require a significant rewrite of doc-builder.

.github/workflows/docker-image-common.yml
samsrabin: Unfortunately since it's only ever a _step_ in a job, it can't be written once and used everywhere. Only _jobs_ are able to do that.

doc/testing/test_container_eq_ctsm_pylib.sh
samsrabin: The `set -e` at the top of all my scripts means that any non-0 exit code will cause the entire script to error. Since `diff` returns 1 when there are diffs, it would error here.

.github/workflows/docs-update-dependency-common.yml
samsrabin: Added with 4a7d411. Also actually added defaults here.

.github/workflows/docs-update-dependency-common.yml
samsrabin: Descriptive name added with 67394de.

doc/testing/test_container_eq_ctsm_pylib.sh
ekluzek: Ahh, OK.

That's probably a good way to go for most things at this stage. So I have two recommendations then.

- [x] First that you add a comment above that line in the scripts that use the "set -e"
- Second, longer term (so in a future PR), we will want to move to the standard unit testing form where a fail of a expected comparison doesn't abort, but prints error information and continues.

I'll add the second to the discussion about bash testing.

.github/workflows/docs-update-dependency-common.yml
ekluzek: OK, good to know. That highlights the need to have testing like this for all three. And then the makefile testing part could be removed when that option is removed.

.github/workflows/docs-update-dependency-common.yml
ekluzek: Perfect.

ekluzek (approved):
Marking as approve now. I still have a few recommendations to do, but they are mostly in the space of adding comments.

I was concerned about the lack of error trapping, but the use of "set -e" will suffice for now. Longer term we should move to using error trapping for at least assert type testing. This allows for nice messaging on the problem and for the test script to continue afterwards.

.github/workflows/docs-build-and-deploy.yml
samsrabin: Changed to `doc/test/` with e168ad3.

doc/testing/test_container_eq_ctsm_pylib.sh
samsrabin: Done with 1cdcafa.

.github/workflows/docs-update-doc-builder.yml
samsrabin: The `test_build_docs_-r-v.sh` script doesn't need to run when `ctsm_pylib` is updated, so I didn't include it in the common workflow.

doc/testing/test_build_docs_-r-v.sh
samsrabin: I don't think the hyphens will cause confusion, and really they're not something we expect people to manually call anyway. So I'll leave the filenames as-is.

doc/testing/test_build_docs_-b.sh
samsrabin: I don't think the hyphens will cause confusion, and really they're not something we expect people to manually call anyway. So I'll leave the filenames as-is.

As far as the bash testing philosophy discussion—yes, it'd be good to get on the same page about that. Let's discuss on #3277.

doc/testing/test_container_eq_ctsm_pylib.sh
samsrabin: Done with ab546d5.

PR #3238, opened 2025-06-06
Add GitHub workflow for Python unit tests

### LINKED ISSUE #2984: Python unit tests aren't portable
There are some things preventing us from running the Python unit tests anywhere other than the NCAR clusters (or at least machines not specified in CIME's `config_machines.xml`). Aside from being a continuity risk, this means that we can't run them as GitHub workflows.

Unit tests failing on my Mac:
- `test_append_user_nl_step2`: PermissionError at `os.mknod(ufile)` (#2985)
- `test_check_args_outsurfdat_provided`: `inputdata directory does not exist` (#2985)
- `test_derecho_mpirun`: `ERROR: Undefined env var 'LMOD_ROOT'`. This one probably just needs to be skipped if not on Derecho, but it should throw a warning. (#2987)
- `test_inputdata_setup_files_basic`: `inputdata directory does not exist` (#2985)
- `test_run_phase`: PermissionError in `SSPCaseFake.create_clone()` at `os.mknod(os.path.join(newcase, "user_nl_clm"))` (#2985)

### LINKED ISSUE #2985: Fix need/checks for inputdata path in subset_data and Python testing
The following Python unit tests fail on non-NCAR machines with `inputdata directory does not exist`:
- `test_check_args_outsurfdat_provided`
- `test_inputdata_setup_files_basic`

That error is being thrown in `setup_files()` here:
https://github.com/ESCOMP/CTSM/blob/bfc5c3179771a4fabf4ac48975189be80a5eb908/python/ctsm/subset_data.py#L564-L571

However, those tests don't really need that directory at all.
- [`test_check_args_outsurfdat_provided`](https://github.com/ESCOMP/CTSM/blob/bfc5c3179771a4fabf4ac48975189be80a5eb908/python/ctsm/test/test_unit_subset_data.py#L103-L116) doesn't need the inputdata dir at all. `setup_files()` should be rewritten so that check is optional.
- [`test_inputdata_setup_files_basic`](https://github.com/ESCOMP/CTSM/blob/bfc5c3179771a4fabf4ac48975189be80a5eb908/python/ctsm/test/test_unit_subset_data.py#L43-L63) checks that the inputdata dir is `/glade/campaign/cesm/cesmdata/cseg/inputdata`, which is just hard-coding things so that this test will fail outside Derecho and Casper. Why? I also don't really get the point of the other two assertions there.

### LINKED ISSUE #2986: Avoid use of os.mknod() in Python testing for portability
It doesn't work everywhere; e.g., on my Mac. Instead, use the safer `os.makedirs()` for making empty directories or `pathlib.Path.touch()` for empty files.

Unit tests failing on my Mac because of this:
- `test_append_user_nl_step2`: PermissionError at `os.mknod(ufile)`
- `test_run_phase`: PermissionError in `SSPCaseFake.create_clone()` at `os.mknod(os.path.join(newcase, "user_nl_clm"))`

samsrabin:
### Description of changes

Adds a GitHub workflow job that runs our Python unit tests whenever a relevant file is changed. Required some changes to unit tests to ensure portability, including one being changed to a system test—we'll deal with making that portable later.

### Specific notes

**CTSM Issues Fixed:**
- Resolves #2985
- Resolves #2986
- Resolves #2984

**Testing performed, if any:** As of bd695ac5:
- [x] Python unit tests pass on my Mac
- [x] New GitHub Workflow to run Python unit tests passes
- [x] Python unit testing on Derecho passes
- [x] Python system testing on Derecho passes
- [x] `clm_pymods` suite passes: `/glade/derecho/scratch/samrabin/tests_0626-162704de/`


rgknox:
totally awesome feature @samsrabin

python/ctsm/unit_testing.py
```diff
@@ -1,8 +1,23 @@
...
+    args_to_add = [
+        "--machine",
+        machine,
+        "--number-of-nodes",
+        str(nodes),
+        "--tasks-per-node",
+        str(tasks),
+    ]
+    for item in args_to_add:
+        sys.argv.append(item)
```
samsrabin: - [x] Simplify this: Why loop? Just do `sys.argv += args_to_add`.

samsrabin: Done with a976348a3.

PR #3240, opened 2025-06-09
tips-for-working-with-rst.md: Add common errors, cheatsheet links.

samsrabin:
### Description of changes

Adds some useful links and info to the "Working with reStructuredText" page in the User's Guide.

### Specific notes

**Any User Interface Changes (namelist or namelist defaults changes)?** None

**Does this create a need to change or add documentation? Did you do so?** Yes?

**Testing performed, if any:** Local build.


doc/source/users_guide/working-with-documentation/tips-for-working-with-rst.md
```diff
@@ -80,3 +86,80 @@ Make sure to include at least one empty line after the comment text.
...
+### "WARNING: Inline literal start-string without end-string"
+
+An "inline literal" is when you want to mix code into a normal line of text (as opposed to in its own code block) ``like this``. This is accomplished with double-backticks:
+```reStructuredText
+An "inline literal" is when you want to mix code into a normal line of
+text (as opposed to in its own code block) ``like this``.
+```
+(A backtick is what you get if you press the key to the left of 1 on a standard US English keyboard.)
+
+If you have a double-backtick on a line, reStructuredText will say, "They want to start an inline literal here," then look for another double-backtick to end the literal. The "WARNING: Inline literal start-string without end-string" means it can't find one on that line.
```
adrifoster: maybe 'reStructuredText will think`... rather than "say" ', just to avoid someone thinking this is an actual feedback?

adrifoster (changes requested):
This looks nice and helpful. Just one minor suggestion which you can take or leave.

doc/source/users_guide/working-with-documentation/tips-for-working-with-rst.md
samsrabin: Done with 5df8e8f.

samsrabin:
@adrifoster Done!

PR #3248, opened 2025-06-12
ctsm5.3.058: Fix clm6 compset aliases

### LINKED ISSUE #3244: ISSP CLM60 Crujra compset aliases have longnames that use CLM50, not CLM60
### Brief summary of bug

./query_config --compsets clm | grep 'ISSP' yields

   ISSP585Clm50BgcCrop  : SSP585_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP585Clm60BgcCropCrujra : SSP585_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP126Clm50BgcCrop  : SSP126_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP119Clm50BgcCrop  : SSP119_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP245Clm50BgcCrop  : SSP245_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP245Clm60BgcCropCrujra : SSP245_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP370Clm50BgcCrop  : SSP370_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP370Clm60BgcCropCrujra : SSP370_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP434Clm50BgcCrop  : SSP434_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP460Clm50BgcCrop  : SSP460_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP534Clm50BgcCrop  : SSP534_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP585Clm60BgcCrop  : SSP585_DATM%GSWP3v1_CLM60%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV

Note that the Crujra Clm60 compset aliases use CLM50, not CLM60

### General bug information

**CTSM version you are using:** ctsm5.3.055

**Does this bug cause significantly incorrect results in the model's science?** Yes, because presumably you will get Clm50 physics instead of Clm60 physics if you use the compset alias

**Configurations affected:** ISSP crujra compsets

### LINKED ISSUE #3254: aux_clm has no Clm60 ISSP tests
I noticed this during testing of #3248 (ctsm5.3.058), where I expected to have diffs in baseline comparison for `ISSP` `Clm60` compsets. Turns out the only `ISSP` compsets we have in `aux_clm` are for `Clm50`. It might be a good idea to convert some/all of the existing tests from 5 to 6, and/or add some new ones.

It is worth noting that there _are_ `ISSP` `Clm60` compsets in the `ctsm_sci` test suite.

samsrabin:
### Description of changes

The following clm60 compset aliases were actually returning long names with CLM50 physics:
```
   ISSP245Clm60BgcCropCrujra : SSP245_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP370Clm60BgcCropCrujra : SSP370_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
   ISSP585Clm60BgcCropCrujra : SSP585_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
```

This PR fixes them. It also adds a GitHub workflow to prevent this from happening again, at least for clm6.

### Specific notes

**CTSM Issues Fixed:**
- Resolves #3244
- Resolves #3254 by switching three ISSP Clm50 tests to Clm60.

**Are answers expected to change (and if so in what way)?** Yes; the affected compsets will now correctly use CLM60 physics instead of CLM50 physics.

**Testing performed, if any:** None yet.

### DIFF
--- a/.github/workflows/check-clm6-aliases.sh
+++ b/.github/workflows/check-clm6-aliases.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-# Check that clm6* compset aliases return CLM5* longnames
+# Check that clm6* compset aliases return CLM6* longnames
 
 # Change to top level of clone
 cd "$(git rev-parse --show-toplevel)"
@@ -13,10 +13,11 @@ cime/scripts/query_config --compsets 1>/dev/null
 OLD_IFS=$IFS
 IFS='\n'
 set +e
-bad_compsets="$(cime/scripts/query_config --compsets | sort | uniq | grep Clm6 | grep CLM5)"
+# Relies on case sensitivity here: Alias should have Clm6 and longname should have CLM6
+bad_compsets="$(cime/scripts/query_config --compsets | sort | uniq | grep Clm6 | grep -v CLM6)"
 set -e
 if [[ "${bad_compsets}" != "" ]]; then
-    echo "One or more compsets with Clm6 alias but CLM5 longname:" >&2
+    echo "One or more compsets with Clm6 alias but not CLM6 longname:" >&2
     echo $bad_compsets  >&2
     exit 1
 fi
--- a/.github/workflows/check-clm6-aliases.yml
+++ b/.github/workflows/check-clm6-aliases.yml
@@ -1,4 +1,4 @@
-name: Check that clm6* compset aliases return CLM5* longnames
+name: Check that clm6* compset aliases return CLM6* longnames
 # Only check files in our repo that AREN'T in submodules
 # Use a Python command to check each file because xmllint isn't available on GH runners
 
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -15,7 +15,7 @@ The following clm60 compset aliases were actually returning long names with CLM5
    ISSP585Clm60BgcCropCrujra : SSP585_DATM%CRUJRA2024_CLM50%BGC-CROP_SICE_SOCN_MOSART_SGLC_SWAV
 ```
 
-This tag fixes them. It also adds a GitHub workflow to prevent this from happening again, at least for clm6 vs. clm5.
+This tag fixes them. It also adds a GitHub workflow to prevent this from happening again, at least for clm6.
 
 
 Significant changes to scientifically-supported configurations
@@ -49,6 +49,7 @@ Notes of particular relevance for developers:
 Changes to tests or testing:
 - Three ISSP tests in aux_clm changed from Clm50 to their Clm60 equivalents.
 - Other ISSP Clm50 tests are untouched.
+- Adds a GitHub workflow to check that Clm6 compset aliases return CLM6 longnames.
 
 
 Testing summary:


.github/workflows/check-clm6-aliases.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+# Check that clm6* compset aliases return CLM5* longnames
```
ekluzek: Automating this with a script is a great idea!

I think we might as well check other things as well though.  We might as well check this at least for all the CLM physics types for example. This would just mean making the check for bad compset as a function, that's called for particular cases.

ekluzek: You could also do this for things like:

- Veg type: Sp/Bgc/BgcCrop/Fates
- Config: Nwp
- Datm: Crujra/GSWP3/Cru/Qian/Spinup/Nldas
- Paleo: NoAnthro
- Stubs: Rs/Gs/As
- CISM: G/Ga/Gag
- ROF: Rtm/Miz
- Period: 2000/1850/Pt1/Hist/SSP

This could also be something that's expanded later. But, I think a lot of this could be added now and wouldn't take that much time. And this automated checking should ideally be in place before we make other changes to compsets which we do plan on.

There's also a couple obscure things that we should think about the support level for:

- Dv -- this should be removed
- Vic -- this should be removed
- RtmFl -- we need this for RTM flood mode testing
- Cmip6/Cmip6Waccm -- we will need modes like this for Cmip7

.github/workflows/check-clm6-aliases.yml
```diff
@@ -0,0 +1,40 @@
+name: Check that clm6* compset aliases return CLM5* longnames
```
ekluzek: Awesome, and a matching workflow that runs the checker.

.github/workflows/check-clm6-aliases.yml
```diff
@@ -0,0 +1,40 @@
...
+# Only check files in our repo that AREN'T in submodules
+# Use a Python command to check each file because xmllint isn't available on GH runners
+
+on:
+  push:
```
ekluzek, line 6: Do we need it both on push and pr? We rarely do a direct push to ESCOMP branches, so I think the latter is sufficient.

.github/workflows/check-clm6-aliases.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Checkout submodules
+        run: |
+          bin/git-fleximod update
+          
+      - name: Install xmllint for CIME
```
ekluzek, line 34: This seems to be unnecessary here, as nothing is done with xmllint. Although using xmllint for checking is a good thing to do as well.

samsrabin: I was getting a CIME error saying that `xmllint` needs to be on the path; this fixed it.

.github/workflows/check-clm6-aliases.yml
samsrabin: The push ensures that it runs when people push to their forks.

ekluzek: OK. So that could show up if I push to a branch on my fork that's not part of a PR. OK, that could be helpful. There are times when I have a branch that's not part of a PR.

.github/workflows/check-clm6-aliases.sh
samsrabin: Issue filed as #3249; I'll resolve this comment.

samsrabin:
@ekluzek This is ready for review! I think it's best to just get a basic check in now to prioritize the fix, and that can be expanded as described in the new issue #3249.

ekluzek (approved):

PR #3251, opened 2025-06-12
Add SystemTests for FATES unit/functional testing

samsrabin:
### Description of changes

Adds FUNITFATES and FUNCTIONALFATES SystemTests, analogous to the FUNITCTSM test. These are now included in the `aux_clm`, `fates`, and the new `unit_tests` test suites.

### Specific notes

**Contributors other than yourself, if any:** @adrifoster 

**CTSM Issues Fixed:** None

**Testing performed, if any:** `./run_sys_tests -s unit_tests --skip-compare --skip-generate` passes. Check `run_*_tests.py.log` in the test case directories.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,7 +28,7 @@
 [submodule "fates"]
 path = src/fates
 url = https://github.com/adrifoster/fates
-fxtag = 5374f3924350a96ac0492d6c1ff93f7737bb97c1
+fxtag = f35b05f03a1f9ed976bf75af0543fc324647927f
 fxrequired = AlwaysRequired
 # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
 fxDONOTUSEurl = https://github.com/NGEET/fates
--- a/.gitmodules
+++ b/.gitmodules
@@ -27,8 +27,8 @@
 #
 [submodule "fates"]
 path = src/fates
-url = https://github.com/adrifoster/fates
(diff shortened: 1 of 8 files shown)


samsrabin:
Now unblocked as NGEET/fates#1423 came in with ctsm5.3.070.

ekluzek:
@samsrabin I'm wondering the status of this? I'm assuming you just don't have time to work on this right now with other priorities?

Part of why I ask is that I did some work to add FATES source directories to the standard FUNIT testing. And with that I noticed that I could likely add the standard unit tests there. That would be simpler than this, which may or may not be a good thing. I'm wondering about doing the simpler solution either short term until this gets in, or possibly there might be reasons for both.

This might be something to talk about with the group -- but I wanted to hear from you about where this work is at first.

Thanks in advance...

samsrabin:
@ekluzek I think this might actually be ready to come in. I prefer this approach because it explicitly calls the FATES tools, so it tests those as well as the tests themselves. It also lets us distinguish failures in FATES tests from failures in CTSM tests.

ekluzek:
> @ekluzek I think this might actually be ready to come in. I prefer this approach because it explicitly calls the FATES tools, so it tests those as well as the tests themselves. It also lets us distinguish failures in FATES tests from failures in CTSM tests.

Awesome. Sounds great. Bringing this in sooner rather than later sounds good to me. Personally I think this could come into b4b-dev, and that might facilitate it coming in sooner...

PR #3264, opened 2025-06-18
Fix plumber2_surf_wrapper

### LINKED ISSUE #3262: plumber2_surf_wrapper: Some files not produced
### Brief summary of bug

`plumber2_surf_wrapper` is supposed to create a surface dataset file for each site in `tools/site_and_regional/PLUMBER2_sites.csv`. However, some are silently skipped.

### General bug information

**CTSM version you are using:** Commit [761bb47](https://github.com/ESCOMP/CTSM/pull/3259/commits/761bb47b2a717c4086a3300201cd4a8a342bcfab) on the branch at PR #3259. (Based on ctsm5.3.056, but that tag would throw a different error without the fixes on that commit.)

Run

```bash
cd tools/site_and_regional
rm -rf subset_data_single_point
./plumber2_surf_wrapper
ls subset_data_single_point | wc -l
```

That last line counts the number of files in `subset_data_single_point/`. There should be 170, but there are only 62.

samsrabin:
### Description of changes

- Include `--lon-type 180` in `subset_data` command to avoid ambiguity.
- If PLUMBER2 sites CSV file has -999 for a pft number, do not set that to zero—instead, just don't include a second PFT.
- Respect whether user requested a 16- or 78-PFT file.
- Add `--overwrite` option.
- Add `--plumber2-sites-csv` option to specify CSV file path.
- Add argument parser unit testing.
- Add system testing.
- Error on `subset_data` failure.

### Specific notes

**CTSM Issues Fixed:** Resolves #3262

**Testing performed, if any:** As of c3258057f:
- Manual testing with commands given in #3262
- [x] The 62 files originally being produced (at `761bb47`) aren't affected.
- [x] Python unit tests
- [x] Python system tests
- [x] `clm_pymods` suite: `/glade/derecho/scratch/samrabin/tests_0620-185305de/`

### DIFF
--- a/python/ctsm/site_and_regional/single_point_case.py
+++ b/python/ctsm/site_and_regional/single_point_case.py
@@ -200,7 +200,8 @@ class SinglePointCase(BaseCase):
                 If any dom_pft is less than 1.
             Error (ArgumentTypeError):
                 If mixed land units are chosen.
-                dom_pft values are both in range of (0 - MAX_NAT_PFT) and (MAX_NAT_PFT+1 - MAX_PFT_MANAGEDCROPS).
+                dom_pft values are both in range of
+                (0 - MAX_NAT_PFT) and (MAX_NAT_PFT+1 - MAX_PFT_MANAGEDCROPS).
 
 
         """


python/ctsm/site_and_regional/plumber2_shared.py
```diff
@@ -0,0 +1,25 @@
...
+PLUMBER2_SITES_CSV = os.path.realpath(
+    os.path.join(
+        os.path.dirname(__file__),
+        os.pardir,
+        os.pardir,
+        os.pardir,
+        "tools",
+        "site_and_regional",
+        "PLUMBER2_sites.csv",
+    )
```
ekluzek: It's cosmetic, but you could use os.path.normpath or os.path.realpath to get a shorter directory name that's easier to read for humans.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
```diff
@@ -23,15 +23,21 @@
 import argparse
 import logging
 import os
-import subprocess
+import sys
 import tqdm
 
-import pandas as pd
+# Get the ctsm tools
```
ekluzek: We talked about this a bit, and it's not too annoying, but I think it might be uneccessary.

This should be handled in both the testing framework and the top level skeleton scripts with path_utils type things.

Looking at it: plumber2_surf_wrapper does this same thing. And run_ctsm_py_tests does it as well.

So we only need this if we want to support running this standalone outside of those contexts. Is that something you are doing? And that you find useful? 

I think I side on not doing this in individual python modules and just assume it'll be handled by either the top level driver script, or the run_ctsm_py_tests environment. Maintaining this for all python module files seems to cumbersome to me. But, I'm happy to hear other thoughts on it as well.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
```diff
@@ -45,137 +51,154 @@ def get_parser():
...
+            list of args for command that we want to run.
     Raises:
-        Error with the return code from shell.
+        Whatever error subset_data gives, if any.
     """
     print("\n", " >>  ", *command, "\n")
 
-    try:
-        subprocess.check_call(command, stdout=open(os.devnull, "w"), stderr=subprocess.STDOUT)
+    sys.argv = command
```
ekluzek, line 86: Yes! This is a nice victory to call it directly from python rather than through subprocess.

Good job on doing this!

python/ctsm/site_and_regional/plumber2_usermods.py
```diff
@@ -11,9 +11,15 @@
 from __future__ import print_function
 
 import os
+import sys
 import tqdm
 
-import pandas as pd
+# Get the ctsm tools
```
ekluzek: Same comment here as  on plumber2_surf_wrapper.py.

python/ctsm/test/test_sys_plumber2_surf_wrapper.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python3
+
```
ekluzek, line 2: Awesome, new testing!

python/ctsm/test/test_unit_plumber2_surf_wrapper.py
```diff
@@ -26,12 +26,60 @@ class TestPlumber2SurfWrapper(unittest.TestCase):
     Basic class for testing plumber2_surf_wrapper.py.
     """
 
-    def test_parser(self):
+    def setUp(self):
```
ekluzek, line 29: The previous test was really just performative, but this is doing real testing. So this is so good to see.

ekluzek (approved):
@samsrabin and I went over this. And this is great work and I'm glad to see it come in. It fixes problems, but also makes the code more robust, with better error handling, AND it adds good testing for all this. So this will save the project time downstream.

I added a few comments/questions here that @samsrabin will decide on. I'm happy to approve it and have it come in shortly.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
samsrabin: Good point; deleting.

python/ctsm/site_and_regional/plumber2_usermods.py
samsrabin: Resolved with 56a36f1ad.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
samsrabin: Done with 56a36f1ad.

python/ctsm/site_and_regional/plumber2_shared.py
samsrabin: Resolved with cf86bfc05.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
```diff
@@ -45,137 +46,154 @@ def get_parser():
         help="Verbose mode will print more information. ",
         action="store_true",
         dest="verbose",
-        default=False,
     )
 
     parser.add_argument(
-        "--16pft",
-        help="Create and/or modify 16-PFT surface datasets (e.g. for a FATES run) ",
+        "--78pft",
```
samsrabin: Change to `--crop` to match standard elsewhere.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
```diff
@@ -45,137 +46,154 @@ def get_parser():
...
-    except subprocess.CalledProcessError as err:
-        # raise RuntimeError("command '{}' return with error
-        # (code {}): {}".format(e.cmd, e.returncode, e.output))
-        # print (e.ouput)
-        print(err)
+def is_valid_pft(pft_num):
+    """
+    Given a number, check whether it represents a valid PFT
+    """
+    return pft_num >= 1
```
samsrabin: Bare ground should be valid, so change from 1 to 0. Also check that it doesn't exceed max PFT, and that max PFT varies based on whether it's a 16-pft or 78-pft run.

samsrabin: Done with b7f7af386.

python/ctsm/site_and_regional/plumber2_surf_wrapper.py
samsrabin: Done with ddca30ff.

Issue #3116, opened 2025-05-08
Answer changing tag to modify snow thermal conductivity defaults


wwieder:
Need an answer changing tag that sets namelist defaults for CLM6 physics to turn on Jordan over glacier land units (and use Sturm everywhere else).

```
 snow_thermal_cond_glc_method = 'Jordan1991'
 snow_thermal_cond_lake_method = 'Sturm1997'
 snow_thermal_cond_method = 'Sturm1997'
```

This needs to be confirmed with CAM team and CESM project meeting.

wwieder:
This is required for our alpha07 tag

Issue #3005, opened 2025-03-11
Bring in ctsm5.4.0 IC files for clm6_0


ekluzek:
Add the new ctsm5.4.0 IC files for clm6_0 to the XML so they are used by default.

wwieder:
This will be on the 5.4 branch tag that goes into the alpha07 tag @slevis-lmwg will make

PR #3268, opened 2025-06-20
ctsm5.3.060: Preliminary update of ctsm54 defaults (answer changing)

slevis-lmwg:
### Description of changes
Brings to master the work done in #3206, which I merged to the ctsm5.4 alpha branch recently as tag
`alpha-ctsm5.4.CMIP7.02.ctsm5.3.055`.

Allows Cecile to run coupled without having to adjust clm things manually. In particular, updates namelist defaults, including new IC files, that have been limited to the ctsm5.4 branch so far.

[Link to the upcoming branch tag](https://github.com/orgs/ESCOMP/projects/46/views/1?pane=issue&itemId=112950697) for #3206 which I merged to the ctsm5.4 alpha branch recently.

### Specific notes

@ekluzek @wwieder 

Resolves https://github.com/ESCOMP/CTSM/issues/3116
Resolves https://github.com/ESCOMP/CTSM/issues/3005

Are answers expected to change (and if so in what way)? Yes, greater than roundoff, same climate:
- Reverted this in https://github.com/ESCOMP/CTSM/pull/3268/commits/4a5dfb5087c35bacd3aae120dec87f1a9eaf7041: Prior to #3206 the ctsm5.4 branch had already changed the raw datasets in mksurfdata_esmf, causing the MKSURFDATAESMF test to differ from the baseline.
- https://github.com/ESCOMP/CTSM/commit/a3ce6a7a703401c75af6c42b91961500080a6f10 changes the default snow thermal conductivity schemes over lakes and glaciers in clm6 cases
- https://github.com/ESCOMP/CTSM/commit/a75e4888baf30fba9c25f0b6bc0ba606e86100ac introduces a new paramfile to clm6 cases
- Reverted this here since we're still working in ctsm5.3: https://github.com/ESCOMP/CTSM/commit/b1890ac5ebbf3fd72c807b54abf7495c5f657591 introduces new f09 & ne30 fsurdat/landuse files
- https://github.com/ESCOMP/CTSM/commit/3a8c43262c425882f2ee10686684158ff2349477 introduces new f09 & ne30 finidat files for 1850 and 2000 clm6 cases

Namelist defaults changes

Probably. Not at this time.

Testing performed, if any:
#3206 testing appears [in this post](https://github.com/ESCOMP/CTSM/pull/3206#issuecomment-2946233295) and will be repeated before merging this PR to master.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,13 +1,13 @@
 ===============================================================
 Tag name: ctsm5.3.060
 Originator(s): slevis (Samuel Levis,UCAR/TSS,303-665-1310)
-Date: Mon 23 Jun 2025 04:01:13 PM MDT
+Date: Tue 24 Jun 2025 02:13:05 PM MDT
 One-line Summary: Preliminary update of ctsm54 defaults (answer changing)
 
 Purpose and description of changes
 ----------------------------------
 
- Brings to master the work done in #3206, which I merged to the ctsm5.4 alpha branch recently as tag alpha-ctsm5.4.CMIP7.02.ctsm5.3.055.
+ Brings to master some of the work done in #3206, which I merged to the ctsm5.4 alpha branch recently as tag alpha-ctsm5.4.CMIP7.02.ctsm5.3.055.
 
  Allows Cecile to run coupled without having to adjust clm things manually: updates namelist defaults and IC files that have been limited to the ctsm5.4 branch so far.
 
@@ -41,9 +41,9 @@ Changes made to namelist defaults (e.g., changed parameter values):
 
 Changes to the datasets (e.g., parameter, surface or initial files):
  a75e488 introduces new paramfile to clm6 cases
- b1890ac introduces new f09 and ne30 fsurdat/landuse files
  3a8c432 introduces new f09 and ne30 finidat files for 1850 and 2000 clm6 cases
- Reverted changes to the default raw datasets that came in with #3206 (preexisting #3206 in the ctsm54 branch)
+ Reverted preexisting changes to the default raw datasets that came in with #3206 (from the ctsm54 branch)
+ Reverted changes to the f09 and ne30 fsurdat/landuse files that came in with b1890ac
 
 Changes to documentation:
  None, yet
@@ -51,11 +51,9 @@ Changes to documentation:
 Notes of particular relevance for developers:
 ---------------------------------------------
 Changes to tests or testing:
- The next 4 tests are now labeled EXPECTED FAILURE in the RUN phase, to be addressed in issues #3250 and #3252:
+ The next 2 tests are now labeled EXPECTED FAILURE in the RUN phase, to be addressed in issue #3252:
  LII2FINIDATAREAS_D_P256x2_Ld1.f09_g17.I1850Clm50BgcCrop.derecho_intel.clm-default--clm-matrixcnOn_ignore_warnings
  LII2FINIDATAREAS_D_P256x2_Ld1.f09_g17.I1850Clm50BgcCrop.derecho_intel.clm-default
- SMS_D_Ld5.f09_g17.ISSP126Clm50BgcCrop.derecho_intel.clm-default
- SMS_D_Ld5.f09_g17.ISSP245Clm60BgcCropCrujra.derecho_intel.clm-default
 
 Testing summary:
 ----------------
@@ -84,7 +82,7 @@ Changes answers relative to baseline: Yes
     - what platforms/compilers: all
     - nature of change: larger than roundoff/same climate
 
- See above in changes to namelist defaults and datasets for various sources of change.
+ See above in changes to namelist defaults and datasets for the sources of change.
 
 Other details
 -------------
--- a/doc/ChangeSum
+++ b/doc/ChangeSum
@@ -1,6 +1,6 @@
 Tag                   Who      Date  Summary
 ============================================================================================================================
-       ctsm5.3.060   slevis 06/23/2025 Preliminary update of ctsm54 defaults (answer changing)
+       ctsm5.3.060   slevis 06/24/2025 Preliminary update of ctsm54 defaults (answer changing)
        ctsm5.3.059     erik 06/23/2025 Bring in various cleanup efforts found in previous testing after the chill changes came in
        ctsm5.3.058 samrabin 06/16/2025 Fix clm6 compset aliases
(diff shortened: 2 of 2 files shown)


slevis-lmwg:
Repeat #3206 testing as explained [here](https://github.com/ESCOMP/CTSM/pull/3206#issuecomment-2946233295) (after merging ctsm5.3.059 to this branch)
- [x] ./build-namelist_test.pl
- [x] derecho aux_clm and fates bc this is an answer-changing tag
- [x] izumi aux_clm and fates

tools/mksurfdata_esmf/gen_mksurfdata_namelist.xml
slevis-lmwg, line 1: Changes to this file preceded #3206 in the ctsm54 alpha branch and explain why the MKSURFDATAESMF test differs from the baseline. I will revert this file back to the one found in escomp/master and will bring these updates to master when we finalize ctsm54.

slevis-lmwg:
As in #3206, I get `use_c1* = .false.` for Clm60Bgc Hillslope tests, which I do not consider a problem for the purposes of this PR.

slevis-lmwg:
@ekluzek I am requesting this review from you because you reviewed #3206 and the code updates are the same.

cime_config/testdefs/ExpectedTestFails.xml
```diff
@@ -62,6 +62,34 @@
...
+    </phase>
+  </test>
+  <test name="LII2FINIDATAREAS_D_P256x2_Ld1.f09_t232.I1850Clm60BgcCrop.derecho_intel.clm-default--clm-matrixcnOn_ignore_warnings">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3252</issue>
+      <comment>Works with finidat = 'ctsm53041_54surfdata_snowTherm_100_pSASU.clm2.r.0161-01-01-00000.nc' and fails with finidat = 'ctsm53041_54surfdata_snowTherm_100_pSASU.clm2.r.0161-01-01-00000_64bitoffset.nc'.</comment>
+    </phase>
+  </test>
+  <test name="SMS_D_Ld5.f09_g17.ISSP126Clm50BgcCrop.derecho_intel.clm-default">
```
ekluzek: These should be able to go away with going back to the ctsm5.3 surface datasets.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -1708,7 +1717,7 @@ lnd/clm2/surfdata_esmf/ctsm5.3.0/surfdata_4x5_hist_1850_16pfts_c241007.nc</fsurd
 <fsurdat hgrid="360x720cru"  sim_year="1850">
```
ekluzek: I was thinking we wouldn't bring the ctsm5.4 surface datasets to master yet. Since it's labeled as a ctsm5.3 tag, you expect it to have ctsm5.3 datasets. Although the advantage of this is that we wouldn't have to do branch tags for cesm alpha tags. But, I think it's clearer if we don't and alpha tags are infrequent enough that it's OK.

So I'd revert the changes from here to the end of the file.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -1289,14 +1298,14 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
          ic_ymd="18500101" sim_year="1850" do_transient_pfts=".false." use_excess_ice=".true."
          ic_tod="0" glc_nec="10" use_crop=".true." irrigate=".false."
```
ekluzek, line 1299: Below I suggest that we don't bring in the ctsm5.4 surface datasets. However, I think we *should* bring in the finidat files to get us closer to that work. So these changes would stay.

But, they each should have

`use_init_interp=".true."`

Which it looks like they do, except this first one.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -150,7 +150,7 @@
       <machine name="izumi" compiler="nag" category="aux_clm"/>
     </machines>
     <options>
-      <option name="wallclock">00:30:00</option>
```
ekluzek, line 153: Hmmm. You shouldn't have had to make this change, because the run time is only 535 seconds. It is up from 441 seconds for the ctsm5.3.059. So I'm curious what the fail looked like for you. It's not necessarily bad to bump it up slightly either though...

ekluzek (changes requested):
This is identical to the work on the branch, so in general not a lot to say. But, I think there needs to be one important difference on master. 

That is I think we keep the surface datasets all ctsm5.3, rather than updating any to ctsm5.4. This does mean that we'll need to continue to keep the alpha branch up to date and used as the final tag for each CESM alpha tag. But, I think that's OK. If we mix ctsm5.4 datasets into ctsm5.3 tags, it'll be forever confusing as to what datasets are in place for the tags.

So I think we remove those changes, but still include everything else. Including the ctsm5.4 finidat files. We should be able to use them, but each entry needs a use_init_interp=".true." in the XML.

I also was somewhat surprised you had to increase some of the wallclock times on izumi. But, it's not necessarily bad to do so. But, I wanted to ask you about that as to what behavior you saw.

cime_config/testdefs/ExpectedTestFails.xml
slevis-lmwg: Ok, removing ISSP tests from expected failures.

bld/namelist_files/namelist_defaults_ctsm.xml
slevis-lmwg: Ok, makes sense. Done.

bld/namelist_files/namelist_defaults_ctsm.xml
slevis-lmwg: Ok, added it.

cime_config/testdefs/testlist_clm.xml
slevis-lmwg: This and several others for which I increased the wallclock failed with exceeding wallclock. It may have been a temporary izumi glitch. If 30 minutes will not seriously delay the test's start, then I would leave it.

slevis-lmwg: Resolving in prep for merge.

ekluzek (approved):

PR #3299, opened 2025-07-02
Add a testmod that allows transient ERI tests to work

billsacks:
### Description of changes

Add a testmod that allows transient ERI tests to work.

With this testmod, the following test passes: `ERI_Ld9.ne30pg3_t232.BHISTC_LTso.derecho_intel.allactive-defaultio--clm-transientERI`.

I'm not sure yet if we want to take this approach or a different approach.

I haven't added any tests using this new testmod. CTSM should probably add a transient ERI test, since it sounds like CAM and possibly CESM will be adding transient ERI tests and it would probably be good for CTSM to catch any problems before they make it to those levels. (That said, I'm not sure if there is actually much additional ERI-specific code covered in a transient ERI test compared with non-transient ERI tests. I haven't given this much thought.)

See https://github.com/ESMCI/cime/issues/4811 for some context

### Specific notes

Testing performed, if any:
ERI_Ld9.ne30pg3_t232.BHISTC_LTso.derecho_intel.allactive-defaultio--clm-transientERI

### DIFF
--- a/cime_config/testdefs/testmods_dirs/clm/transientERI/user_nl_clm
+++ b/cime_config/testdefs/testmods_dirs/clm/transientERI/user_nl_clm
@@ -1,0 +1,5 @@
+
+! The following is needed in a transient ERI test so that the hybrid case works,
+! given the 2-year jump from the initial case to the hybrid case:
+! https://github.com/ESMCI/cime/blob/f0b9d218e97eb92ea84decaf6ae1bfc263b8ddc9/CIME/SystemTests/eri.py#L93
+check_finidat_year_consistency = .false.
--- a/cime_config/testdefs/testmods_dirs/clm/transientERI/README
+++ b/cime_config/testdefs/testmods_dirs/clm/transientERI/README
@@ -1,0 +1,6 @@
+This testmod has changes that are needed in a transient (e.g., HIST) ERI test.
+
+This testmod is meant to be a mix-in with other testmods (e.g.,
+'clm-something--clm-transientERI'), so it does not have an include of the default testmod
+or any other testmods (it is assumed that any needed includes come from the primary
+testmod of the test, and that this is just added on top of that).


cime_config/testdefs/testmods_dirs/clm/transientERI/user_nl_clm
```diff
@@ -0,0 +1,5 @@
+
+! The following is needed in a transient ERI test so that the hybrid case works,
+! given the 2-year jump from the initial case to the hybrid case:
+! https://github.com/ESMCI/cime/blob/f0b9d218e97eb92ea84decaf6ae1bfc263b8ddc9/CIME/SystemTests/eri.py#L93
```
ekluzek: I find the referencing of the specific hash in cime a bit confusing here. I'm not sure that the code reference is helpful, just the description is probably fine.

It does make me wonder if this should be the default for a hybrid case though? I'm siding for it not to be, but I see reasons to make it the default for hybrid.

ekluzek (dismissed):
This is great. Thanks for figuring this out and proposing a PR for it @billsacks.

I wonder about the test-mod name, if it should be about what it does, rather than how it's used. And I think you should remove the line referencing the code.

Actually for that matter we should add a transient test to aux_clm so that we can show this works and continues to work. We should probably just change an existing ERI test to use a transient compset.

We'll talk about this in a minute though. So this is great.

billsacks:
Thanks for your good feedback, @ekluzek . I think I have addressed your thoughts.

billsacks:
As @ekluzek pointed out, we should have a transient ERI test. For this to be most useful, it should be a decStart test where both the hybrid (ref2) and branch (main) cases cross the year boundary. (The continue run crossing the year boundary would be a bonus, but less important, since that functionality is presumably covered via an ERS test.)

ekluzek:
I'm planning on converting:

``` xml
<test name="ERI_D_Ld9" grid="f10_f10_mg37" compset="I1850Clm60Bgc" testmods="clm/default">
```

to something like:

``` xml
<test name="ERI_D_Ld9" grid="f10_f10_mt232" compset="IHistClm60BgcCrop" testmods="clm/decStart--clm/allowTransientYearJump">
```

ekluzek:
OK, I think this might be an artifact of doing the decStart, but doing the above test I see this sort of error:

```
 set_curr_delta ERROR: found unexpected non-zero delta mid-year
 Dribbler name: dwt
 i, delta =           16  2.081914839658153E-006
 Start of time step date (yr, mon, day, tod) =         2003          12
          30           0
 This indicates that some non-zero flux was generated at a time step
 other than the first time step of the year, which this dribbler was told not to
  expect.
 If this non-zero mid-year delta is expected, then you can suppress this error
 by setting allows_non_annual_delta to .true. when constructing this dribbler.
iam = 0: local  patch    index = 16
.
.
.
iam = 0: landunit type         = 1
 ENDRUN:
 set_curr_delta: found unexpected non-zero delta mid-year: ERROR in AnnualFluxDr
 ibbler.F90 at line 277
```

billsacks:
Hmmm, yes, this is probably due to a combination of decStart, transient and the 2-year jump in the hybrid run. But actually, my rationale for why this could be expected makes it seem to me like it should also be expected in a non-decStart transient ERI test (because I think that the hybrid case in that test also started mid-year), yet I wasn't seeing it in my testing, so now I don't understand why you're seeing different behavior here than in my test....

In any case, I think you can bypass this guardrail by adding this in user_nl_clm:

`for_testing_allow_non_annual_changes = .true.`.

I'm hopeful that that will let this test pass, but let me know if that just leads to another error.

ekluzek:
That does get the test to run. However, one of the comparisons is recorded as being different, even though the cprnc.out files don't seem to show this:

PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart CREATE_NEWCASE
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart XML
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart SETUP
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart SHAREDLIB_BUILD time=94
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart MODEL_BUILD time=38
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart SUBMIT
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart RUN time=404
FAIL ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart COMPARE_base_hybrid
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart COMPARE_base_rest
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart MEMLEAK insufficient data for memleak test
PASS ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart SHORT_TERM_ARCHIVER

The TestStatus.log shows one file comparison as different, but I don't see this when I look at the cprnc.out files. The casedir is here:

/glade/work/erik/ctsm_worktrees/b4b-dev/cime/scripts/ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart.20250704_192412_4dp8gf

The file is a mosart file:

from the TestStatus.log:
```
tail -n20 /glade/derecho/scratch/erik/ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart.20250704_192412_4dp8gf/run/ERI_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart.20250704_192412_4dp8gf.mosart.h0.2003-12.nc.base.cprnc.out

              259200  (   101,   289,     1) (   182,   239,     1)
          avg abs field values:    5.282888198086358E+00    rms diff: 1.8E+01   avg rel diff(npos):  1.5E-01
                                   5.297771514872078E+00                        avg decimal digits(ndif):  0.7 worst: -0.3
 RMS TOTAL_DISCHARGE_TO_OCEAN_LIQ     1.8096E+01            NORMALIZED  3.4207E+00

************************************************************************************************************************************
  
SUMMARY of cprnc:
 A total number of     27 fields were compared
          of which      8 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
               and      0 had different data types
 A total number of      2 fields could not be analyzed
 A total number of      0 time-varying fields on file 1 were not found on file 2.
 A total number of      0 time-constant fields on file 1 were not found on file 2.
 A total number of      0 time-varying fields on file 2 were not found on file 1.
 A total number of      0 time-constant fields on file 2 were not found on file 1.
  diff_test: the two files seem to be DIFFERENT 

```

I think I'll rerun the test from scratch, to see if I still see this.

ekluzek:
Two glitches I noticed:

REST_OPTION is fixed at "ndays" in env_test.xml. I assume this is an issue with the eri.py code in cime.
BFBFLAG wasn't set and probably should be

I'll add both of these to the testmod for now.

ekluzek:
And @billsacks one last concern I want to ask about. From what you said here..

> Hmmm, yes, this is probably due to a combination of decStart, transient and the 2-year jump in the hybrid run. But actually, my rationale for why this could be expected makes it seem to me like it should also be expected in a non-decStart transient ERI test (because I think that the hybrid case in that test also started mid-year), yet I wasn't seeing it in my testing, so now I don't understand why you're seeing different behavior here than in my test....

I'm concerned that setting:

`for_testing_allow_non_annual_changes = .true.`

means it will allow bad behavior in other parts of the test? Is this actually OK? I suppose the history file checking is sufficient for any kind of problems that could happen here. Anyway, I wanted to hear your thoughts on that. Thanks

billsacks:
Regarding the cprnc.out file not showing it: I noticed this issue when looking at some ERI test stuff last week: it looks like, due to the way cprnc output files are named, the outputs from the second set of comparisons will clobber the outputs from the first. So the diffs still show up in the TestStatus.log and TestStatus files, but the files actually showing these diffs have disappeared. You *might* be able to fix this by changing one or both of these lines at the bottom of the eri test:

```
        self._component_compare_test("base", "hybrid")
        self._component_compare_test("base", "rest")
```

to swap the order (putting "base" second).

A failure in this test comparison does seem like a problem....

I'm not sure that BFBFLAG actually does anything anymore.

I share your concern that setting `for_testing_allow_non_annual_changes = .true.` may allow bad behavior rather than catching it. I gave that some thought before suggesting it and my feeling is that setting that is still probably the best / right thing to do. That said, I'd be more comfortable with this if I understood why the decStart ERI test requires it when the non-decStart test doesn't... (as I mentioned above, now that I've convinced myself that it's needed in the decStart test, I would have thought that it would also be needed in the non-decStart test with a 2-year jump...).

billsacks:
Ah, I think I see why the decStart case requires this setting while the non-decStart ERI case does not: I had thought that the hybrid (ref2) case was offset by 2 years from the refdate, but in fact it is offset by 2 years from the ref1 start date. So, in a non-decStart test, I think the hybrid/ref2 case starts at the beginning of the year, whereas in a decStart ERI test, I think the hybrid/ref2 case starts at the same date as the initial/ref1 case.

This is a little weird, in that it implies a further mismatch between the start date and refdate, but I think this explains the behavior, and I'm now more comfortable with what's going on here.

ekluzek:
@billsacks OK, applying the above and doing this change:

``` diff
diff --git a/CIME/SystemTests/eri.py b/CIME/SystemTests/eri.py
index 6d80a8fd8..d981ad633 100644
--- a/CIME/SystemTests/eri.py
+++ b/CIME/SystemTests/eri.py
@@ -299,4 +299,4 @@ def run_phase(self):
         self.run_indv(suffix="rest")
 
         self._component_compare_test("base", "hybrid")
-        self._component_compare_test("base", "rest")
+        self._component_compare_test("rest", "base")
```

Get's me to see the differences in the file, and they are large.

```
SUMMARY of cprnc:
 A total number of    650 fields were compared
          of which    574 had non-zero differences
               and     13 had differences in fill patterns
```

It's comparing these two files:

 file 1=
 /glade/derecho/scratch/erik/ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel
 .clm-decStart--clm-allowTransientYearJump.20250708_014011_g8or3s/run/ERI_D_Lh72
 .f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJ
 ump.20250708_014011_g8or3s.clm2.h0.2004-01-01-00000.nc.base
 file 2=
 /glade/derecho/scratch/erik/ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel
 .clm-decStart--clm-allowTransientYearJump.20250708_014011_g8or3s/run/ERI_D_Lh72
 .f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJ
 ump.20250708_014011_g8or3s.clm2.h0.2004-01-01-00000.nc.hybrid


The hybrid is the "ref2" case, and base is the plain original case. Base is branching from ref2, and ref2 is a hybrid from "ref1". Here's the difference in the env_run.xml file for the two cases:

``` diff
diff env_run.xml  ../ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump.20250708_014011_g8or3s/
179c179
<     <entry id="RUN_TYPE" value="hybrid">
---
>     <entry id="RUN_TYPE" value="branch">
241c241
<     <entry id="RUN_REFCASE" value="ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump.20250708_014011_g8or3s.ref1">
---
>     <entry id="RUN_REFCASE" value="ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump.20250708_014011_g8or3s.ref2">
247c247
<     <entry id="RUN_REFDATE" value="2001-12-30">
---
>     <entry id="RUN_REFDATE" value="2003-12-31">
253c253
<     <entry id="RUN_REFTOD" value="43200">
---
>     <entry id="RUN_REFTOD" value="32400">
267c267
<     <entry id="RUN_STARTDATE" value="2003-12-30">
---
>     <entry id="RUN_STARTDATE" value="2001-12-30">
286c286
<     <entry id="STOP_N" value="60">
---
>     <entry id="STOP_N" value="12">
299c299
<     <entry id="REST_OPTION" value="nhours">
---
>     <entry id="REST_OPTION" value="never">
306c306
<     <entry id="REST_N" value="33">
---
>     <entry id="REST_N" value="15">
319c319
<     <entry id="DRV_RESTART_POINTER" value="rpointer.cpl.2003-12-31-32400">
---
>     <entry id="DRV_RESTART_POINTER" value="rpointer.cpl.2004-01-01-00000">
370c370
<     <entry id="CONTINUE_RUN" value="FALSE">
---
>     <entry id="CONTINUE_RUN" value="TRUE">
```

Since, the start date for base is a two year jump from hybrid, it seems like this should be expected? But, why doesn't this show up in non Dec-12-30 tests? And maybe if I do a longer test where the restart is after Jan/1 -- it'll work then?

billsacks:
Can you try getting things working with an ERI test that has a length of days rather than hours? I'd bet that ERI has *never* been run with a mid-day restart. Ideally it would work for that, but let's try to get things working with restarts on the day boundary first.

billsacks:
> Can you try getting things working with an ERI test that has a length of days rather than hours? I'd bet that ERI has _never_ been run with a mid-day restart. Ideally it would work for that, but let's try to get things working with restarts on the day boundary first.

I recognize that accomplishing what we discussed - i.e., crossing the year boundary in the branch case - may then require pushing the start date a few days earlier than what's in the standard decStart test.

ekluzek:
@billsacks so I've done more tests. I did also look at ERI tests run in cesm3_0_alpha07a and CAM runs some 18 step tests. So midday is possible. The nonstandard lengths I see in CESM tests are: 18 steps, 9 days, 20 days,  60 days, 24 months, 9 years, 15 years, and 44 years.

72 days or hours fail with a December start, as well as 9 days

FAIL ERI_D_Ld72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-allowTransientYearJump COMPARE_base_hybrid
FAIL ERI_D_Ld9.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump COMPARE_base_hybrid
FAIL ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump COMPARE_base_hybrid

What PASSes is 9 days without a December start:

PASS ERI_D_Ld9.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-allowTransientYearJump COMPARE_base_hybrid

Without crossing the year boundary in some fashion we aren't really testing transient cases.

This is probably another time to brainstorm together again. But, I can also send a couple tests if you see others to try...

billsacks:
@ekluzek - with the changes you're making to this testmod I'm wondering if we should go back to something like the old name of transientERI, since some of these modifications seem specific to that purpose and possibly confusing in the context of the new, more general name. I'm happy to let you make the call on that, but just thought I'd raise the question.

billsacks:
@ekluzek and I looked into this:

> FAIL ERI_D_Ld9.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump COMPARE_base_hybrid

Fails due to a MOSART h0 file being present in one case but not in the other:

```
    File 'ERI_D_Ld9.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump.20250709_083239_b9jw9p.mosart.h0.2003-12.nc.hybrid' had no original counterpart in '/glade/derecho/scratch/erik/cases/ERI_D_Ld9.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump.20250709_083239_b9jw9p/run' with suffix 'base'
```

Possible fix: daily MOSART output

> FAIL ERI_D_Ld72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-allowTransientYearJump COMPARE_base_hybrid

Fails due to the time and time_bounds values differing on the MOSART h0 files between the hybrid and branch cases (time of 44.9375 vs. 45).

Possible fix: again, daily MOSART output

> FAIL ERI_D_Lh72.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-decStart--clm-allowTransientYearJump COMPARE_base_hybrid

This one is possibly concerning: most of the fields on the h0 file differ. It's good news that the final cpl hist file is identical, so the simulation itself seems fine, but the CLM output file seems messed up in the branch case. What might be happening here is that the history file being compared may be spanning the branch time, so that it's averaging values from before and after the branch. And this leads us to wonder if this situation is actually being tested - where the values on the history-restart (rh) file are being used....

Some things to try:
- ERS_Lh24 or ERS_Lh48 (transient or non-transient)
- ERI_Lh72, non-transient

ekluzek:
OK, I tried those options and they mostly worked:

These worked:

ERI_D_Ld20.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-default--clm-midDecStart--clm-allowTransientYearJump
ERS_Lh24.f10_f10_mg37.I1850Clm60Bgc.derecho_intel.clm-default--clm-decStart
ERS_Lh24.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-default--clm-decStart

This failed:

ERI_D_Lh72.f10_f10_mg37.I1850Clm60Bgc.derecho_intel.clm-default--clm-midDecStart--clm-allowTransientYearJump    (NLCOMP COMPARE_base_hybrid BASELINE)

kdraeder:
@billsacks @ekluzek 
github says that all checks have passed, but the comments indicate that not everything is working yet.
What's the status and plan?
Thanks for implementing and testing this!

ekluzek:
@kdraeder I added the test that I showed worked. And I'm testing on Izumi now. I'll merge this into b4b-dev and it'll come in on Thursday to master, depending on when Derecho comes back up.

kdraeder:
> @kdraeder I added the test that I showed worked. And I'm testing on Izumi now. I'll merge this into b4b-dev and it'll come in on Thursday to master, depending on when Derecho comes back up.

Thanks for pushing this forward!

ekluzek:
Rerunning the proposed test, I get a fail in base to hybrid comparison for the clm.h0a files

FAIL ERI_D_Ld20.f10_f10_mt232.IHistClm60Bgc.derecho_gnu.clm-midDecStart--clm-allowTransientYearJump COMPARE_base_hybrid

So this needs more work...

ekluzek:
I'll look at this again and talk with @billsacks again. And @slevis-lmwg will join as well.

ekluzek:
OK, this test passes:

ERI_D_Ld20.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-default--clm-midDecStart--clm-allowTransientYearJump

Getting the clm-default, in the list of testmods is critical in order to have the history output for the branch align correctly.

kdraeder:
I'm wondering about 2 things, and I don't know which order they should have.
Is this upgrade general enough that it could be ported back to cime, as discussed in [cime #4811](https://github.com/ESMCI/cime/issues/4811)?
What does the schedule look like for completing this issue?
Thanks for any insights.
Kevin

billsacks:
@ekluzek I'm just looking back at this, prodded by @kdraeder 's comment in https://github.com/ESMCI/cime/issues/4811. I've lost track of where things stand with this PR. Do you have a sense, or should we get together and look this over together?

One specific thing I'm wondering about is your addition of the REST_OPTION setting in the testmod. I see your [comment](https://github.com/ESCOMP/CTSM/pull/3299#issuecomment-3046363166) where you say

> REST_OPTION is fixed at "ndays" in env_test.xml. I assume this is an issue with the eri.py code in cime.

But I'm wondering if this actually caused you problems in practice: from my quick look at the eri test, it looks like the initial value is probably irrelevant, because it gets adjusted by eri.py. That said, I do agree this should probably be fixed in config_tests.xml, but for now I'd suggest backing it out here unless it's really needed, to avoid any confusion (at least I was confused looking at this, wondering why it's needed and what it accomplishes) and to avoid setting this in other test types where it could do more harm than good (e.g., test types that expect REST_OPTION=never/none).

billsacks:
Over in https://github.com/ESMCI/cime/issues/4811, @kdraeder said:

> It sounds like you're recommending that if I want to run this test for CAM in a data assimilation context,
> then I should import code from CTSM for that test, because changes to code outside of CTSM
> are not justified at this time. That may work fine for me, but my end goal is that these tests
> will be run automatically before a tag andor release of CESM, to ensure that the capabilities DART needs
> are still working. Would that work?
> And is that actually defining a new test, which should have a new name? That's fine too.

I'm moving that conversation here....

@kdraeder - CAM or CESM will already contain CTSM, so there's an option of leveraging this testmod directly. That said, in an email exchange that included some CAM folks in July (Brian Eaton & Cheryl Craig; I just saw that you weren't included in that exchange), Brian Eaton expressed a preference for just putting the necessary CTSM namelist change in the CAM testmods directly. I'll reply to that thread and include you.

kdraeder:
@billsacks Thanks for separating these issues.
I'd like to see the new test_mod before I proceed with more comments or coding, but I'm not finding it.
I see that it was developed in CTSM branch eri_transient_testmod, but that's not available in the list of branches in ESCOMP/CTSM.  Am I missing a link or a clue?

billsacks:
@kdraeder - you can see the changes in this PR here: https://github.com/ESCOMP/CTSM/pull/3299/files. The main relevant changes are just in this file: https://github.com/ESCOMP/CTSM/pull/3299/files#diff-7a2fabd152ce7a9427f0186a25424511eb80cd9c46288d5bfab1c95b1bdc08e1

slevis-lmwg:
From meeting with @wwieder and @ekluzek:
@slevis-lmwg to
- [x] update to latest b4b-dev and
- [x] started testing on izumi and derecho
```
OK ./run_sys_tests -s aux_clm -c ctsm5.3.085 --skip-generate
```
...so I can merge this to b4b-dev next.

slevis-lmwg:
@ekluzek may have discovered a problem:
https://github.com/ESCOMP/CTSM/pull/3640#issuecomment-3608956093

Issue #3295, opened 2025-06-30
Some changes to the self_test initialzation testing structure


ekluzek:
In #3276 I'm working on initialization self_tests for a specific purpose. This is something that can be useful for other types of testing at initialization. There are also self_tests that require to be used within time-looping and can be distinguished from the initialization ones. The initialization ones can also be run in a bare-bones framework and exit early as well, to make it as efficient as possible.

Therefore, I'm planning on doing the following:

- [ ] Change the name of the testmod run_self_tests to run_init_self_tests
- [ ] Change usage of self_test, self test or self-test to self_tests consistently everywhere
- [x] Add namelist variable: for_testing_bypass_init_after_tests
  - [x] Have it skip as much of the code as possible patterned after PTS_MODE where the point is a non-land point
  - [ ] Enable running this mode with stub ATM
  - [ ] Have this mode run without history or restart output
  - [x] Have it exit as early as possible in initialization

ekluzek:
@billsacks since you are the one behind the self_tests I'd love to hear any thoughts that you have here. If not that's fine as well.

billsacks:
Thanks for looping me in on this. I don't have many thoughts, just a general thought that, while I support the idea of making these self-tests efficient, I wouldn't go too far with that: I could imagine a quest for efficiency here that takes considerable time and effort and could make it hard to add future self-tests if you bypass too much code.

ekluzek:
This is due by end of September because of funding. But, not strictly needed for the ctsm5.4 work.

Issue #2995, opened 2025-03-09
Scalability issue in decompInit_lnd


briandobbins:
In the decompInit_lnd routine (in src/main/decompInitMod.F90), there's a performance issue in the section below:

https://github.com/ESCOMP/CTSM/blob/7ff60613ddb8235396e0098b67bf0dfa79195e28/src/main/decompInitMod.F90#L226-L236

  This happens because the main loop, outside of that, is over every cell, and this internal loop is over the number of clumps, which is at least equal to the number of PEs.  For km-scale runs, this ends up being quite large - eg, on the 3.75km test case, the outer loop is ~42M, and the inner loop is >~40K, since that's the minimum we're able to run the case on.  The conditional does restrict the inner loop to only happening on ~12.4M of the 42M cells, but it's still a problem to have things scale by the number of cores.

  In terms of data, the loop above took between 754 - 1044 seconds, averaging 804 across all PEs.  That's ~86% of the InitializeRealize call. 

  I've got a few ideas on things to try, including simply changing the complex conditionals as a temporary work-around, as well as saving/reading a decomposition, but would welcome insights from folks who know the land model.  

  Perhaps the other issue here, which may be more important in the end, is that it seems like 'clumps' is allocated on every rank _for_ every rank? 

https://github.com/ESCOMP/CTSM/blob/7ff60613ddb8235396e0098b67bf0dfa79195e28/src/main/decompInitMod.F90#L118

Again, I welcome input here - this is a challenge for memory scalability, at least, and unless this is needed elsewhere, we should move to a local-to-the-rank structure.

  Anyway, just getting the issue in - I think I can create work-arounds for the near-term needs, but would be happy to chat with any land folks on this and see if we can get a SIF or some other way to focus on addressing this soon, too.

  Thanks!

GOAL: 

- Memory usage and initialization time should be nearly constant with number of tasks

Grids for low resolution testing on:

- ne3pg{2,3} (do we need both of these?)
- ne30pg3

higher resolutions: 

- ne120
- mpas15

Ultra high resolution grids for final testing:

- mpas7p5
- mpas3p75

Definition of done:
- [x] Communicate with John about what we need to do
- [x] Erik learns about MPI_Scan from the article
- [x] Possibly meet with John if we need to figure things out
- [x] Figure out grids that we should test
- [x] Clarify goal
- [x] Start adding self-tests and self-test namelist control to test decomposition
- [x] Add a unit test for decompInit?
[...]

PR #3301, opened 2025-07-02
[WIP] Work with self_tests so that there's an initialization only option

### LINKED ISSUE #3371: Add self_tests for the decomposition
Add some self_test options to check that the decomposition is created correctly.

ekluzek:
### Description of changes

Have a self_tests mode that essentially only does initialization. Change the self_tests testmod to run in this mode.

### Specific notes

   This enables work in #2995
   Some work here #3295

   Some new namelist items for testing

Testing performed, if any: So far just running specific test cases

Definition of done:
- [ ] Merge #3425 into here, and reconcile it
- [ ] Figure out what needs to be added to the bypass logical functions to make them more robust
- [ ] Figure out how to combine these bypass functions with the bypass already put in place for PTS_MODE without a land point
- [ ] Add functions/handling needed for the PTS_MODE configuration
- [ ] Make sure up to date with the b4b-dev branch
- [ ] Run aux_clm and decomp_init test lists, bring into b4b-dev as normal


ekluzek:
In sprint 20 I spent a few days on this work.

This is stalled right now, because I had trouble getting the setup with SATM working. To do that, CTSM HAS to have a run section as it becomes the clock to start and stop the simulation.

PR #3347, opened 2025-07-23
Urban mesh was pointing to a 3x3min instead of a 0.25x0.25 file

slevis-lmwg:
Correcting in this PR.


slevis-lmwg:
I will merge this and then tag is as
`alpha-ctsm5.4.CMIP7.05.ctsm5.3.063`

PR #3348, opened 2025-07-24
Pass nutrient supplementation status to FATES

rgknox:
### Description of changes

This set of changes passes nitrogen supplementation status to FATES. This status is needed when deciding whether or not to dynamically adjust leaf to fine root ratios based on nutrient storage. This status prevents plants from reducing root mass while soils have artificially high nitrogen during supplementation.

### Specific notes

This is coupled with FATES PR: [1443](https://github.com/NGEET/fates/pull/1443)

Contributors other than yourself, if any: @sharma-bharat @glemieux @walkeranthonyp

Are answers expected to change (and if so in what way)? No, this should have not impact on answers on any configuration. This will only be relevant when we fully couple FATES-CLM-CN.

No

Not yet needed, no N limitations with FATES

Testing performed, if any:

### DIFF
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -1151,10 +1151,11 @@ module CLMFatesInterfaceMod
       logical  :: after_start_of_harvest_ts
       integer  :: iharv
       logical  :: nitr_suppl                             ! Is CLM currently supplementing N
-      logical, parameter :: phos_dummy_suppl = .false.   ! This argument is needed for FATES
+      logical, parameter :: phos_dummy_suppl = .true .   ! This argument is needed for FATES
                                                          ! to specify if phosphorus is being
                                                          ! supplemented, this is not cycled in CLM
-                                                         ! so we set it to false
+                                                         ! so we set it to TRUE (which essentially
+                                                         ! means it is NOT limiting)
       !-----------------------------------------------------------------------
 
       ! ---------------------------------------------------------------------------------
--- a/src/utils/clmfates_interfaceMod.F90
+++ b/src/utils/clmfates_interfaceMod.F90
@@ -79,6 +79,7 @@ module CLMFatesInterfaceMod
    use clm_varctl        , only : use_lch4
    use clm_varctl        , only : fates_history_dimlevel
    use clm_varctl        , only : nsrest, nsrBranch
+   use clm_varctl        , only : carbon_only
    use clm_varcon        , only : tfrz
    use clm_varcon        , only : spval
    use clm_varcon        , only : denice
@@ -1149,6 +1150,11 @@ module CLMFatesInterfaceMod
       real(r8) :: s_node, smp_node         ! local for relative water content and potential
       logical  :: after_start_of_harvest_ts
       integer  :: iharv
+      logical  :: nitr_suppl                             ! Is CLM currently supplementing N
+      logical, parameter :: phos_dummy_suppl = .false.   ! This argument is needed for FATES
+                                                         ! to specify if phosphorus is being
+                                                         ! supplemented, this is not cycled in CLM
+                                                         ! so we set it to false
       !-----------------------------------------------------------------------
 
       ! ---------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rgknox:
Test ERS_D_Ld60.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-Fates all PASS, didn't compare with base yet

ekluzek:
I started aux_clm testing on Izumi. I'll start fates testing later after we see other results.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -80,6 +80,7 @@ module CLMFatesInterfaceMod
    use clm_varctl        , only : use_lch4
    use clm_varctl        , only : fates_history_dimlevel
    use clm_varctl        , only : nsrest, nsrBranch
+   use clm_varctl        , only : CNAllocate_Carbon_only
```
ekluzek: Since this method is in clm_varctl I think I like the use of it for FATES. But, I think the CN prefix makes it confused with CN BGC methods, which is explicitly off for FATES. So I think if we also use this method for FATES -- we should change the name to drop the "CN" prefix.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -1161,6 +1145,12 @@ subroutine dynamics_driv(this, nc, bounds_clump,      &
       real(r8) :: s_node, smp_node         ! local for relative water content and potential
       logical  :: after_start_of_harvest_ts
       integer  :: iharv
+      logical  :: nitr_suppl                             ! Is CLM currently supplementing N
+      logical, parameter :: phos_dummy_suppl = .true.    ! This argument is needed for FATES
+                                                         ! to specify if phosphorus is being
+                                                         ! supplemented, this is not cycled in CLM
+                                                         ! so we set it to TRUE (which essentially
+                                                         ! means it is NOT limiting)
```
ekluzek:
```suggestion
      logical  :: nitr_suppl                     ! true -> CLM is supplementing Nitrogen
      logical, parameter :: phos_dummy_suppl = .true.    ! true -> Phosphorus is NOT limited (i.e. supplemented)
                                                         ! This argument is needed for FATES
                                                         ! to specify if phosphorus is being
                                                         ! supplemented, Phosphorous is not limited in CLM
                                                         ! so we set it to TRUE
```

src/utils/clmfates_interfaceMod.F90
```diff
@@ -474,23 +475,6 @@ subroutine CLMFatesGlobals2()
         end if
         call set_fates_ctrlparms('use_tree_damage',ival=pass_tree_damage)
         
-        ! These may be in a non-limiting status (ie when supplements)
```
ekluzek, line 477: These seem to be related but slightly different than the other things going on here. And just making sure this is intentionally removed. It looks to me like you'd also need to make changes in FATES for this.

I'm out of my depth here, but it seems to me the difference in use_nitrif_denitrif still needs to be captured somewhere right?

ekluzek (commented):
The CTSM changes are small and pretty straightforward. My main request is I think we should change CNAllocate_Carbon_only to Allocate_Carbon_only in order to use it both for CN/BGC and FATES. Otherwise it makes the code confusing as we explicitly have to choose between CN/BGC or FATES.

I have a suggestion for changes to comments that made it clearer to me, feel free to adjust as you see fit.

We should probably have a quick discussion on this if we can make that work.

src/utils/clmfates_interfaceMod.F90
rgknox: Discussed with @ekluzek in a call.  Key difference here is that 1) FATES doesn't care if the N came from ammonium or nitrate, once its in the plant it does not distinguish the source, and the acquisition from the different sources happens in CLM and 2) limitation status is no longer time-invariant, so setting it as an initialized parameter is no-longer appropriate and is passed in as a boundary condition.

src/utils/clmfates_interfaceMod.F90
rgknox: this has been resolved.

.gitmodules
```diff
@@ -27,8 +27,8 @@
 #
 [submodule "fates"]
 path = src/fates
-url = https://github.com/NGEET/fates
-fxtag = sci.1.87.2_api.41.0.0
+url = https://github.com/rgknox/fates
+fxbranch = cnp-l2fr-conly
```
ekluzek:
```suggestion
url = https://github.com/rgknox/fates
#fxtag = cnp-l2fr-conly     # This is the branch being pointed to, the hash used on the branch is below
fxtag = 8c5a24ce880ac117dffa7f5b87427a57250af8ed
```

@rgknox it looks like fxbranch isn't something that git-fleximod understands. It might have had it in the past, but doesn't now. When you run baselines I also recommend to point to either a tag or a hash so that you know exactly what you tested with. Naming a branch means the code at the tip of the branch may change after the tests are run.

So personally I usually have a comment that names what the branch is, and then use fxtag to point to the hash along the branch.

ekluzek: I'm going to go ahead and apply the above suggestion here, so that I can run Izumi tests that way.

ekluzek: When you get back to this it would be good to merge your PR in fates and make the tag. So we can point to the FATES tag at this point. That is one thing we need anyway before making the final tag.

ekluzek:
aux_clm passed as expected on Izumi. Everything there was b4b.

We're now pointing to a FATES tag. @rgknox and I will create a baseline with ctsm5.3.082 and then compare to it for these changes.

ekluzek:
OK, the baselines are now all done and looking as expected.  Ryan will do the ChangeLog files and we'll tag after that's done.

PR #3372, opened 2025-07-29
ctsm5.3.070: FATES API41: default paramfile update and new namelist option

glemieux:
### Description of changes

Updates the default FATES parameter file and update FATES to the associated tag.  This also adds a new namelist option `use_fates_managed_fire`.

### Specific notes

To be coordinated with ngeet/fates#1444

Testing performed, if any:


glemieux:
Regression testing on `derecho` and `izumi` are underway.

glemieux:
Regression testing on `derecho` is complete.  Nearly all expected tests are B4B with the expected exception of the fates testmods.  However there is one non-fates test failing RUN: `SUBSETDATAREGION_Ld5_D_Mmpi-serial.CLM_USRDAT.I2000Clm60BgcCropCrujra.derecho_intel.clm-default`.   From the logs it appears that there is a file missing: 

```
14 Abort with message No such file or directory in file /glade/derecho/scratch/jedwards/tmp/spack-stage/spack-stage-parallelio-2.6.6-6ijcyssbl2fmclu4hwqpgh2aqpjdafiv/spack-src/src/clib/pio    c_support.c at line 2857
```

Results: `/glade/u/home/glemieux/ctsm/tests_pr3372-aux_clm`

 A few tests on izumi seemed to have failed to submit, so I am submitting them manually now.

glemieux:
Regression testing on `izumi` is complete and showing all expected tests with b4b results. 

Results: `/home/glemieux/ctsm/tests_0815-165846iz`

glemieux:
> Regression testing on `derecho` is complete. Nearly all expected tests are B4B with the expected exception of the fates testmods. However there is one non-fates test failing RUN: `SUBSETDATAREGION_Ld5_D_Mmpi-serial.CLM_USRDAT.I2000Clm60BgcCropCrujra.derecho_intel.clm-default`. From the logs it appears that there is a file missing:
> 
> ```
> 14 Abort with message No such file or directory in file /glade/derecho/scratch/jedwards/tmp/spack-stage/spack-stage-parallelio-2.6.6-6ijcyssbl2fmclu4hwqpgh2aqpjdafiv/spack-src/src/clib/pio    c_support.c at line 2857
> ```
> 

@samsrabin and I were not able to replicate this with a fresh call to run the test.  We're racking this up as a glitch.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -750,6 +750,12 @@ ignitions.
 (Only relevant if FATES is on)
 </entry>
 
+<entry id="use_fates_managed_fire" type="logical" category="physics"
+        group="clm_inparm" valid_values="" value=".false.">
+Enable FATES managed fire mode.  Requires that fates_spitfire_mode is on (in any mode).
```
ekluzek, line 755: Add a little bit more description of what "managed fire mode" means. What kinds of managed fires are included here? This isn't crop prescribed burns right? And I think this is just about fires that are created by land managers to manage the fire loads right? It's not about suppression for example correct?

ekluzek (approved):
This looks good. I would like a little bit more added to the comment about what managed_fire means in the namelist definition.

However, you shouldn't rerun all your testing for just that. So we could either make it a small issue that comes in later. Or I think it's sufficient to do a limited amount of testing to make sure that change doesn't mess anything up. Running the namelist testing, and maybe one case for example.

The namelist changes are good. You moved the setting of fates_spitfire_mode from the list to a specific call, because it needs it know about managed fire. That all makes sense and is the right thing to do.

I did notice something outside the scope. It looks like we don't do checking to make sure fates settings are NOT done when FATES is off. That would be something good to add at some point for usability. But, that's outside of scope, so maybe we create an issue for it...

bld/namelist_files/namelist_definition_ctsm.xml
glemieux: Thanks for pointing this out Erik.  Updates made per a579be7.

glemieux:
> However, you shouldn't rerun all your testing for just that. So we could either make it a small issue that comes in later. Or I think it's sufficient to do a limited amount of testing to make sure that change doesn't mess anything up. Running the namelist testing, and maybe one case for example.

Running minimal set of testing on `derecho`

glemieux:
Minimal test case is done and results are as expected.

PR #3384, opened 2025-08-04
Change megan_use_gamma_sm to default false

### LINKED ISSUE #3016: Low MEGAN emissions in CESM3
Using the latest beta cases, we are seeing a significant underestimation of biogenic emissions, when running a F-case with -bgc sp:

beta02 was still OK, for example, isoprene emissions were around 460 Tg/yr, which is what we think is reasonable.
However, going to beta04, the biogenic emissions dropped to about 250 Tg/yr. 
We think that this is a case when MEGAN changes had been included.

We also run a recent beta05 version with @wwieder changes to the land and we saw some improvement with isoprene emissions going back to around 380 Tg/yr.

@shawnh @rrbuchholz @lkemmons

slevis-lmwg:
### Description of changes
One line change in namelist_defaults_ctsm.xml

### Specific notes

Resolves #3016 

Yes when megan is active

Changed megan_use_gamma_sm to default false

I did not

Testing performed, if any:
I will run aux_clm before merging

### DIFF
--- a/bld/namelist_files/namelist_defaults_ctsm.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm.xml
@@ -153,7 +153,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <megan configuration="nwp"          >0</megan>
 
 <!-- Setting to TRUE replicates previous ctsm5.3.063 answers, setting to false turns the soil moisture effect off -->
-<megan_use_gamma_sm>.true.</megan_use_gamma_sm>
+<megan_use_gamma_sm>.false.</megan_use_gamma_sm>
 <megan_min_gamma_sm>0.0d00</megan_min_gamma_sm>
 
 <!-- Supplmental Nitrogen mode -->


slevis-lmwg:
- [x] Merge ctsm5.3.067
- [x] Run aux_clm
  - [x] derecho
  - [x] izumi

doc/ChangeLog
```diff
@@ -1,4 +1,72 @@
...
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+
+Does this tag change answers significantly for any of the following physics configurations?
+(Details of any changes will be given in the "Answer changes" section below.)
+
+    [Put an [X] in the box for any configuration with significant answer changes.]
+
+[X] clm6_0
```
ekluzek: If we consider this significant, it would apply to all physics versions. But, since MEGAN is diagnostic for I compsets, I'd say it's NOT significant. So I'd leave these unchecked.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -153,7 +153,8 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <megan configuration="nwp"          >0</megan>
 
 <!-- Setting to TRUE replicates previous ctsm5.3.063 answers, setting to false turns the soil moisture effect off -->
```
ekluzek: I suggest changing the notes here and in line 156 slightly.

The soil moisture affect was added in ctsm5.2.020 with MEGAN2.1, and hardcoded on. So ctsm5.2.020 to ctsm5.3.067 have it turned on. In ctsm5.3.068 here, we are turning it default off. So it might be helpful to say it was on from ctsm5.2.020 to ctsm5.3.067 and then in ctsm5.3.067 it was turned default off, which is closer to the previous version of MEGAN, and hence we have it off for all CLM physics versions.

doc/ChangeLog
```diff
@@ -1,4 +1,72 @@
...
+
+    derecho -----
+    izumi -------
+
+Answer changes
+--------------
+
+Changes answers relative to baseline: Yes
+
+  Summarize any changes to answers, i.e.,
```
ekluzek, line 58: Should only change answers for tests with MEGAN and should just change answers for MEGAN fields. So you can verify that the answer changes are limited.

slevis-lmwg:
Note to self:
After merging to master and making the ctsm5.3.068 tag, merge ctsm5.3.068 to alpha-ctsm5.4.CMIP7 as per @wwieder's comment [here](https://docs.google.com/document/d/1F1lMRa1oDE3WGp6PEpGjXRHiTMAcq3jo9nMtMFi14h0/edit?disco=AAABo4-nso0)

PR #3389, opened 2025-08-06
[IGNORE] Add cirrus-testing workflow

samsrabin:
### Description of changes
This PR is just to test GitHub Workflows on the CIRRUS cloud runners.


samsrabin:
The ubuntu runner is currently failing because the `CESMDATAROOT` variable is undefined. Try giving it some garbage value; it might not actually be needed.

Issue #3388, opened 2025-08-06
Avoid in-code checks of PFT type


samsrabin:
We have a bunch of integer variables in `pftconMod` that are unique IDs for each PFT ("vegetation type constants"):
https://github.com/ESCOMP/CTSM/blob/d6f950acd665d59d6342237fda95f5a2d3001fc6/src/main/pftconMod.F90#L19-L101

These are used throughout the code in checks like `veg_type == ndllf_evr_tmp_tree`. We should move to using the parameter file for this instead, to facilitate changing PFT characteristics and adding new ones. For example, the ISIMIP runs may require some unused PFTs to be replaced with different ones, which will necessitate not just changing the parameter file but also looking through the code for these manual checks. In addition, it would make the affected code easier to read.

This especially makes things hard for non-developers trying to develop new crops ([e.g.](https://bb.cgd.ucar.edu/cesm/threads/rainfed-corn-yield-underestimation-after-2000-and-activation-of-potato-crop-in-clm.11812/post-63140)). They have to scour the code for places where PFT integers might need to be used.

Although I think this would be an easy job, it wouldn't have to happen all at once. There's some very low-hanging fruit out there. For instance, this could be replaced with a `gdd20_type` parameter:
https://github.com/ESCOMP/CTSM/blob/d6f950acd665d59d6342237fda95f5a2d3001fc6/src/biogeochem/CNPhenologyMod.F90#L2878-L2896

Note that fully resolving this would also necessarily involve solving Issue #21.

samsrabin:
From `src/`, this command will work to find all instances of any vegetation type constant, including `noveg`, `npcropmin`, and `npcropmax`:
```shell
[...]

PR #3390, opened 2025-08-06
Make npcropmin/max private to pftconMod

samsrabin:
### Description of changes
Making `npcropmin` and `npcropmax` private to `pftconMod` helps set up for adding an `is_prognostic_crop` parameter to the parameter file, contributing to #3388. Making these private required adding an `is_prognostic_crop()` function, the contents of which will eventually be replaced by a read of that new parameter.

### Specific notes

**CTSM Issues Fixed:**
- Contributes to #3388

- [x] `aux_clm` passes

### DIFF
--- a/src/biogeochem/CNFUNMod.F90
+++ b/src/biogeochem/CNFUNMod.F90
@@ -1271,7 +1271,7 @@ fix_loop:   do FIX =plants_are_fixing, plants_not_fixing !loop around percentage
                !           Calculate appropriate degree of retranslocation
                !-------------------------------------------------------------------------------
       
-               if(leafc(p).gt.0.0_r8.and.litterfall_n_step(p,istp)* fixerfrac>0.0_r8.and. (.not. is_prognostic_crop(ivt(p))))then
+               if (leafc(p) > 0.0_r8 .and. (litterfall_n_step(p,istp) * fixerfrac) > 0.0_r8 .and. (.not. is_prognostic_crop(ivt(p)))) then
                   call fun_retranslocation(p,dt,npp_to_spend,&
                                 litterfall_c_step(p,istp)* fixerfrac,&
                                 litterfall_n_step(p,istp)* fixerfrac,&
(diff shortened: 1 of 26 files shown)


src/biogeochem/CNFUNMod.F90
```diff
@@ -1271,7 +1271,7 @@ subroutine CNFUN(bounds,num_soilc, filter_soilc,num_soilp&
                !           Calculate appropriate degree of retranslocation
                !-------------------------------------------------------------------------------
       
-               if(leafc(p).gt.0.0_r8.and.litterfall_n_step(p,istp)* fixerfrac>0.0_r8.and.ivt(p) <npcropmin)then
+               if(leafc(p).gt.0.0_r8.and.litterfall_n_step(p,istp)* fixerfrac>0.0_r8.and. (.not. is_prognostic_crop(ivt(p))))then
```
slevis-lmwg:
```suggestion
               if (leafc(p) > 0.0_r8 .and. (litterfall_n_step(p,istp) * fixerfrac) > 0.0_r8 .and. (.not. is_prognostic_crop(ivt(p)))) then
```
While you're updating this line, I would recommend making it a bit more readable :-)

src/main/pftconMod.F90
```diff
@@ -1607,5 +1640,52 @@ subroutine Clean(this)
...
+  elemental logical function is_prognostic_crop(veg_type)
+    !
+    ! !DESCRIPTION:
+    ! Given a vegetation type (pft, integer), return whether it's a prognostic crop. Does not
+    ! include generic crops (those and natural PFTs will return .false.).
+    !
+    ! NOTE(wjs, 2017-02-02) This isn't a completely robust way to check if this is a
+    ! prognostic crop patch (at the very least it should also check if <= npcropmax;
+    ! ideally it should use a prognostic_crop flag that doesn't seem to exist
+    ! currently).
```
slevis-lmwg: You moved Bill's comment here from where it used to be (if I'm not mistaken). I would rather you replaced Bill's comment with something like...
```suggestion
    ! This isn't a completely robust way to check if this is a
    ! prognostic crop patch. At the very least it should also check if <= npcropmax
    ! and ideally the prognostic_crop flag will be set in clm's paramfile.
```
...if putting it in the paramfile is the plan (I may have misunderstood).

samsrabin: Improved with bd0ba2039.

PR #3391, opened 2025-08-07
Move nml parameters

swensosc:
### Description of changes
move namelist parameters to parameter file.  parameters moved are: baseflow_scalar, precip_repartition_nonglc_all_rain_t, precip_repartition_nonglc_all_snow_t, precip_repartition_glc_all_rain_t, precip_repartition_glc_all_snow_t

### Specific notes

Contributors other than yourself, if any:

CTSM Issues Fixed (include github issue #):

Are answers expected to change (and if so in what way)?

Any User Interface Changes (namelist or namelist defaults changes)?

Does this create a need to change or add documentation? Did you do so?

Testing performed, if any:
(List what testing you did to show your changes worked as expected)
(This can be manual testing or running of the different test suites)
(Documentation on system testing is here: https://github.com/ESCOMP/ctsm/wiki/System-Testing-Guide)
(aux_clm on derecho for intel/gnu and izumi for intel/gnu/nag/nvhpc is the standard for tags on master)

**NOTE: Be sure to check your coding style against the standard
(https://github.com/ESCOMP/ctsm/wiki/CTSM-coding-guidelines) and review
the list of common problems to watch out for
(https://github.com/ESCOMP/CTSM/wiki/List-of-common-problems).**

PR #3405, opened 2025-08-11
Distributed parameters

swensosc:
### Description of changes

Make some parameters spatially distributed, with data ingested via streams file.

### Specific notes

   See this discussion:
      https://github.com/ESCOMP/CTSM/discussions/2395

  New namelist parameters: 
     - use_distributed_parameters
     - stream_fldfilename_distparam
     - stream_meshfile_distparam

Does this create a need to change or add documentation? Did you do so? Yes No

Testing performed, if any: @ekluzek will do regular


ekluzek:
We should be able to show this as b4b-dev, so will likely rebase to b4b-dev, once we can show that.

slevis-lmwg:
@swensosc raised the question whether this will be in clm6. I will add a milestone to the PR.

ekluzek:
This has been blocked by #3391 

This was also a desired feature highlighted in the LMWG meeting, so we need to start working on this so it can be used by more people.

ekluzek:
I'll update this to the latest, so that we can create a branch tag with this for some coupled model testing that is desired.

wwieder:
@swensosc the following were previously shown to have strong influence on arctic snow melt (based on attributes in our latest parameter file).
```
xdrdt=5, 
scvng_fct_mlt_sf=0.5, 
snw_rds_refrz=1500, 
fresh_snw_rds_max=400
```
From these, the most influential is likely xdrdt.

@olyson summarized the following based on our dead veg work a few years ago:

_Regarding increasing albedo, our best bets are probably xdrdt and upplim_destruct_metamorph.  Increasing xdrdt from 1 to 5 decreased albedo by maybe 2-4% over Antarctica so reversing that would hopefully result in an increase by that amount.  Decreasing upplim_destruct_metamorph from 250 to 100 decreased albedo by 1-2%, so maybe we could go to 400 to increase albedo (we are at 250 now).  fresh_snw_rds_max is also a possibility, it is now 400 and an increase from about 204 to 400 decreased albedo by 1-2% also, so we could try going back down to 200 or so.  Not clear that the combined effects of these would be a simple sum, there must be some kind of upper limit on fresh snow albedo.  Maybe first try a couple of these in land-only mode to make sure we get the magnitude and direction we are expecting in Antarctica._

ekluzek:
@swensosc do you have a distributed parameter file/mesh that I can test with?

ekluzek:
OK, I updated to the latest on b4b-dev, which is up to ctsm5.4.021. A good number of tests passed, but a lot failed, many with a build error. Hopefully, the build error is simple. Here's the summary....

```
Test summary
297 Total tests
198 Tests passed
1 Tests compare different to baseline
0 Tests are new where there is no baseline
0 Tests pending
99 Tests failed
```

I'll push the update, and then see if the build error is easy. I also think we should test on Izumi with nag, as that can often find legit problems.

src/cpl/share_esmf/DistParamStreamMod.F90
```diff
@@ -0,0 +1,368 @@
...
+         call getfil (control%stream_fldFileName_distparams, locfn, 0)
+         call ncd_pio_openfile (ncid, trim(locfn), 0)
+
+         call ncd_inqnvars(ncid,nvariables=nvariables)
+
+         allocate(stream_varnames_in(nvariables))
+         nparams = 0
+         do n=1,nvariables
+            call ncd_inqvname(ncid,n,varname,var_desc)
+            if (.not. any(varname == (/'time','lat','lon'/))) then
```
ekluzek: This actually results in a build problem with at least the gnu compiler. The '(/ /)' syntax for strings requires that the strings be of the same length. So a space could be added to the end of lat and lon for example. But, then the string match isn't going to work. A "trim" can certainly be added to "varname" and probably to the array? So maybe this will work...


```suggestion
            if (.not. any(trim(varname) == trim( (/'time','lat','lon'/) ) ) ) then 
```

ekluzek: Nope, trim only takes scalar arguments. I'm trying to think of a way to use index to check for substrings, but not seeing a way to get that to work. So it probably needs an additional loop to go through the dimension names so trim can be used on both sides. Which means some overhead to save the array and such. A little annoying, but not hard to do...

swensosc:
I didn't know about the string length constraint.
Would it work to do this?

any((/trim(varname) == 'time',trim(varname) == 'lat',trim(varname) ==
'lon'/)) )

ekluzek:
> I didn't know about the string length constraint. Would it work to do this? any((/trim(varname) == 'time',trim(varname) == 'lat',trim(varname) == 'lon'/)) )

It does! So I'll go with that for now. the above isn't as clean as the original, but better than the loop suggestion. So longer term, we'll probably do something else -- but this is a decent starting point.

ekluzek:
Things that I see needed to get a branch tag done today:

- [x] Get gnu working
- [x] Add the script to make distparams to tools/contrib
- [x] Make a test distparams file and testmod to use it
- [ ] Update to ctsm5.4.022 (rebase this to master)
- [ ] Make a branch tag: branch_tags/distparams.n02.ctsm5.4.022
- [ ] Make sure most tests work (the same tests should fail from #3391)
- [ ] Run and add a distparam test
- [ ] Test on Izumi and show most tests work
- [ ] Make a branch tag: branch_tags/distparams.n02.ctsm5.4.022

swensosc:
you could do something like this (maybe in a function), but I don't know
that it's worth the effort:

character(len=4)               :: ex_varnames(3)     ! variable names to
ignore
...
ex_varnames = (/'time','lat','lon'/)
...
any((/(trim(varname) == trim(ex_varnames(ig)),ig=1,3)/))

ekluzek:
gnu tests are now working. And the same tests fail as for #3391 with only one new additional one: 

SETPARAMFILE_Ld5.f10_f10_mg37.I1850Clm60BgcCrujra.derecho_gnu.clm-default

but we won't worry about that for now.

slevis-lmwg:
Simulation https://github.com/NCAR/LMWG_dev/issues/148 uses this distributed paramfile provided by Sean:
`paramdata_ctsm60_params_c260204_ne30np4.pg3_hist_1850_78pfts_c251022.nc`
So far I have copied this to /inputdata but I have not ./rimport-ed, yet.

swensosc:
For cases when parameter streams file is of a different spatial resolution than the surface data, the map_algo default should perhaps be 'nn' instead of 'bilinear'

Issue #3407, opened 2025-08-11
Fixes needed to use h2i files in GDD-generating Python scripts


samsrabin:
I encountered various errors during testing for #3320. They are now fixed and will come in on PRs soon.

samsrabin:
This was resolved by... one or other of my recent PRs.

PR #3336, opened 2025-07-14
[WIP] Update crop calendar inputs for CRU-JRA

samsrabin:
# WORK IN PROGRESS! DO NOT REVIEW!
# SUBJECT TO FORCE PUSHES!

### Description of changes

This PR will update the default `stream_fldFileName_cultivar_gdds` and `stream_fldFileName_gdd20_baseline` to point to files created from CRU-JRA runs. It will also include any code, testing, or documentation changes spurred by generating those files. (These can be separated into other PRs if needed.)

### Specific notes

**Contributors other than yourself, if any:** None

**CTSM Issues Fixed:**
- Eventually: #3321

**Are answers expected to change (and if so in what way)?** Yes; new crop maturity requirements.

**Any User Interface Changes (namelist or namelist defaults changes)?** See description above.

**Does this create a need to change or add documentation? Did you do so?** Maybe; not yet

**Testing performed, if any:** None so far

Issue #3319, opened 2025-07-09
Rethinking (variables in) h2 files for GDD-generating workflow


samsrabin:
The GDD-Generating phase of the RXCROPMATURITY test requests two variables, `GDDACCUM` and `GDDHARV`, on its h2 tape(s). From ctsm5.3.062:
https://github.com/ESCOMP/CTSM/blob/400bfa03d3b97da2422235c468c2c8ac30afde23/cime_config/SystemTests/rxcropmaturity.py#L134-L149

Those variables are used to generate crop growing degree-day requirement files to be used in the second part of the test. That happens via a Python script that, before the split-up of accumulated vs. instantaneous files in ctsm5.3.062, were looking for a file with `.h2.` in it. From the tag before that:
https://github.com/ESCOMP/CTSM/blob/4c379f215c93180f374445b3da00ab26843c66e7/python/ctsm/crop_calendars/generate_gdds_functions.py#L567-L569

Both of those variables are time-averaged by default:
https://github.com/ESCOMP/CTSM/blob/400bfa03d3b97da2422235c468c2c8ac30afde23/src/biogeochem/CropType.F90#L288-L291
https://github.com/ESCOMP/CTSM/blob/400bfa03d3b97da2422235c468c2c8ac30afde23/src/biogeochem/CNVegStateType.F90#L329-L331

I thus expected that, for ctsm5.3.062, the Python script would need to be changed to look for `.h2a.` in the filenames. However, during that work (PR #2445), @slevis-lmwg noticed that didn't solve the problem—only `.h2i.` worked.

This was pretty confusing until I had a look at the `user_nl_clm` file. I noticed that, far above the `hist_fincl3 = 'GDDACCUM', 'GDDHARV'` line requested by RXCROPMATURITY, there was `hist_avgflag_pertape(3) = 'I'`. This is because the RXCROPMATURITY tests in `aux_clm` use the `cropMonthOutput` testmod, which inherits from `crop`, which has this in its `user_nl_clm`:
https://github.com/ESCOMP/CTSM/blob/400bfa03d3b97da2422235c468c2c8ac30afde23/cime_config/testdefs/testmods_dirs/clm/crop/user_nl_clm#L12-L20

Initially, I thought I would just add another line to the RXCROPMATURITY GDD-Generating phase `user_nl_clm` changing it back:
```
hist_avgflag_pertape(3) = 'A'
```

[...]

PR #3408, opened 2025-08-11
Logging improvements for GDD-generation workflow

samsrabin:
### Description of changes

Improves logging during GDD-generating runs and postprocessing. A few commits cherry-picked from branch at #3336, making a more focused PR for easier review.

### Notes to reviewer

I strongly recommend going through the commits one-by-one rather than just looking at the final diffs. I did a lot of work to consolidate the commits to make the review easier.

In CTSM:
- Suppresses a CTSM warning about cold temperatures during GDD-generating runs. (These get printed a LOT when you put all crops everywhere, as is standard for that workflow.)

In crop calendar Python tools:
- Moves `log()` and `error()` functions from `generate_gdds_functions.py` to `cropcal_utils.py` so they can be used in more places. This is responsible for most of the changed lines.
- Adds datetime string to those functions.
- Adds some `log()` calls w/r/t files being opened.

### Specific notes

**CTSM Issues Fixed:**
- Contributes to #3407

**Testing performed, if any:** Worked in a run using the branch at #3336, but I haven't tested this branch specifically.
- [x] Test with `rxcropmaturity` suite
- [x] Test with `aux_clm` suite


samsrabin:
Blocking further work on #3319: specifically, bringing in the actually-important changes from the #3336 branch.

python/ctsm/crop_calendars/cropcal_utils.py
```diff
@@ -9,6 +9,25 @@
 from ctsm.utils import is_instantaneous
 
 
+def log(logger_in, string):
```
ekluzek: This doesn't need to be done now, but I wonder if these shouldn't be general use methods that are added to the ctsm_logging.py module?

ekluzek: Actually, similarly with the leading_datetime_string function above -- should it be moved to one of the general ctsm utils?

python/ctsm/crop_calendars/cropcal_module.py
```diff
@@ -49,11 +49,12 @@ def check_and_trim_years(year_1, year_n, ds_in):
     return ds_in
 
 
-def open_lu_ds(filename, year_1, year_n, existing_ds, ungrid=True):
+def open_lu_ds(filename, year_1, year_n, existing_ds, *, logger, ungrid=True):
     """
     Open land-use dataset
     """
     # Open and trim to years of interest
+    utils.log(logger, f"open_lu_ds(): Opening this_ds_gridded: {filename}")
```
ekluzek: It took me a minute to figure this out but the "open_lu_ds(): " in the logger prefix is the name of the function. This seems useful to have. With the downside that the name needs to be maintained in case the subroutine name changes.

I'm not sure it's worth it, but in case the subroutine names change, you can use things like inspect.current_frame() or sys._getframe() to get the subroutine name. But, an idea to think about...

ekluzek (approved):
These changes make sense and are good to bring in. I was able to both go through each commit as well as see the total changes in the PR and figure it out.

As usual I have a few suggestions and/or questions to think about, but approving so you don't have to wait on me.

python/ctsm/crop_calendars/cropcal_module.py
```diff
@@ -49,11 +49,12 @@ def check_and_trim_years(year_1, year_n, ds_in):
     return ds_in
 
 
-def open_lu_ds(filename, year_1, year_n, existing_ds, ungrid=True):
+def open_lu_ds(filename, year_1, year_n, existing_ds, *, logger, ungrid=True):
```
ekluzek, line 53: I had to look up what the * argument means, and it's so that everything after HAS to have keyword arguments. That does seem like a useful thing to do in our python coding.

So just a thumbs up about using it.

samsrabin: Yup, it's especially useful because `pylint` doesn't like it when a function has too many positional arguments.

python/ctsm/crop_calendars/cropcal_utils.py
samsrabin: Done with ed3223fca.

python/ctsm/crop_calendars/cropcal_module.py
samsrabin: Good idea; done.

PR #3418, opened 2025-08-15
Endrun work

### LINKED ISSUE #3417: Use shr_abort_mod for endrun, and add optional arguments for file and line
Start using the shr_abort_abort subroutine to abort in abortutils::endrun making use of it's improvements on shr_sys_abort. Also allow optional arguments for file and line so that errMsg is used to report on where the call was made.

Also make sure shr_sys_flush is called in endrun, so log files are output.

This came from this discussion in a CTSM SE meeting:

https://github.com/ESCOMP/CTSM/issues/1452#issuecomment-3177063129

### LINKED ISSUE #3420: Don't have endrun abort on bad subgrid_level
### Brief summary of bug

Don't have endrun short circuit the abort if a bad subgrid_level is given into it. We need endrun to give as much information as the error as possible, without 

### General bug information

**CTSM version you are using:** ctsm5.3.068

**Configurations affected:** endrun that use the pt context option
### Details of bug

By code inspection endrun will abort early if a bad subgrid_level is given to the pt context version of endrun. Since, we need the error messaging to be reported about the actual error, we don't want it give the error about the call to endrun, but to finish it's writing about the actual problem.

ekluzek:
### Description of changes

Add a unit test to endrun, so I can do a little work on it, and use those improvements in the decompInit work.

### Specific notes

  Fixes #3417
  Fixes #3420

Testing performed, if any: So far running unit tests will run aux_clm


ekluzek:
@samsrabin this has everything I need in it. unit tests are fine on Derecho. I'm running aux_clm testing now.

ekluzek:
aux_clm on Derecho and Izumi are as expected.

src/main/test/endrun_test/test_endrun.pf
```diff
@@ -0,0 +1,200 @@
+module test_endrun
```
samsrabin: Could this module (and file) this be renamed to something like `test_abortutils` since `endrun` subroutines aren't the only things in there? I.e., there's also `write_point_context` that we might eventually want unit testing for.

src/main/test/endrun_test/test_endrun.pf
```diff
@@ -0,0 +1,200 @@
...
+    @assertExceptionRaised(endrun_msg(msg))
+
+  end subroutine endrun_addmsg_vanilla_aborts
+
+  @Test
+  subroutine endrun_addmsg_pt_context_aborts(this)
+    use decompMod, only : subgrid_level_lndgrid, subgrid_level_gridcell
+    use decompMod, only : subgrid_level_landunit, subgrid_level_column, subgrid_level_patch
+    use decompMod, only : subgrid_level_cohort
+    ! Test pt_context operation of endrun with an additional message sent in
```
samsrabin: In this test and subsequent ones, could you move the explanatory comment up to immediately below the `subroutine` line?

src/main/test/endrun_test/test_endrun.pf
ekluzek: Yep, make sense.

src/main/test/endrun_test/test_endrun.pf
ekluzek: Oh, sure. That is how we do it elsewhere, so a nice consistency change to make.

ekluzek: a139a42cc142f686d010f1e2dd7522ec3f2d7bd1

src/main/test/endrun_test/test_endrun.pf
ekluzek: 09400a301c154ab2120fd3d87d4949cfb5b4477a and 6e818f60833ff2467ca9a231d5ee32f30ec94c03

Issue #3276, opened 2025-06-23
Create self-test namelist control for decompInit


ekluzek:
With work in decomp and decompInit it will be helpful to have some self-tests controlled via namelist to add checks to ensure the decomp is correct. It can also be setup to terminate after the decomposition is finished for example.

ekluzek:
The self-tests are structured to have namelist control names like:

`for_testing_<description of test to do>`

Package self-tests are run from self_test_driver called in initialize2. The one example is "for_testing_run_ncdiopio_tests" which calls test_ncdio_pio from the TestNcdioPio module which is a test module around ncdio_pio. It's written in Fortran rather than PFunit as it's compiled with the rest of the model.

There are other for_test_* control variables that do other testing that's in given modules, or allow specific behavior that's only desired for test environments. So modeling that here we'd have something like:

`for_testing_run_decomp_init_tests`

with 

`src/self_test/TestDecompInit.F90`

being the tester around decompInit.

The current self_test is: SMS_D_Ln1.f10_f10_mg37.I2000Clm50BgcCropQianRs.derecho_intel.clm-run_self_tests

It currently just sets:

`for_testing_run_ncdiopio_tests = .true.`

ekluzek:
Some namelist control to add:
- [ ] decomp_init_dbug: option (dbug)
- [ ] decomp_init_log: logging of processor names

These should be default off.

PR #3425, opened 2025-08-21
[WIP] For testing add a framework to bypass parts of the code

ekluzek:
### Description of changes

This adds a framework for adding some structure to tell the model that you want to bypass some parts of the model for some different types of testing that you may want to do. It's controlled by a namelist read as part of the SelfTestDriver code. Then there are some methods to control the bypass functionality mainly in clm_initialize and clm_driver.

There is some more work I think that should happen before this comes in. But, this gives the rough structure. I need this to get to this point so that I can use this branch in my other ones.

This does part of #3301 but in a cleaner way.

### Specific notes

Contributors other than yourself, if any: @samsrabin and others for ideas

  This enables work in https://github.com/ESCOMP/CTSM/issues/2995
  Starts https://github.com/ESCOMP/CTSM/issues/3276
  Some work here https://github.com/ESCOMP/CTSM/issues/3295

   New namelist items for self tests

Does this create a need to change or add documentation? Did you do so? Maybe? But, no to second

Testing performed, if any: So far running a self test case. Will run the decomp_init testlist as well as aux_clm

Definition of done:
- [x] Get feedback on the general structure
- [x] Apply the suggested changes
- [ ] Figure out what the error checking should look like
- [ ] Decide on what things above should come in now, vs opened in an issue and done later in separate PR's
- [ ] Have CTSM SE group review it
- [ ] Add error checking in build-namelist
- [ ] Add error checking in the code
- [ ] Changes to the logical functions for robustness
- [ ] Add some documentation around this


bld/CLMBuildNamelist.pm
```diff
@@ -5290,6 +5290,7 @@ sub write_output_files {
   push @groups, "clm_canopy_inparm";
   push @groups, "prigentroughness";
   push @groups, "zendersoilerod";
+  push @groups, "for_testing_options";
```
ekluzek, line 5318: I think I should add more error checking to the build-namelist for these bypass options. I also think that these testing options should maybe do things like ensure that history and restart files are off the like. There also might be some error checking that some of the advanced options are NOT turned on with the bypass options, and that sort of thing. And it will be important to make sure it's clear that these bypass and testing options are turned on -- and they don't get turned on accidentally. It will take some thinking to figure that out.

cime_config/testdefs/testmods_dirs/clm/run_self_tests/user_nl_clm
```diff
@@ -1,3 +1,5 @@
+for_testing_bypass_init = .true.
+for_testing_bypass_run = .true.
```
ekluzek: I should leave the bypass init part here, but move the run part to the for_testing* testmod.

src/main/clm_driver.F90
```diff
@@ -165,6 +166,7 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
     ! CalcIrrigationNeeded. Simply declaring this variable makes the ICE go away.
     real(r8), allocatable :: dummy1_to_make_pgi_happy(:)
     !-----------------------------------------------------------------------
+    if ( for_testing_bypass_run_except_clock_advance() ) return
```
ekluzek, line 169: Because this short circuits the code -- it's worth thinking if this should be a simple return statement or a more explicit if structure.

src/main/clm_initializeMod.F90
```diff
@@ -467,6 +471,7 @@ subroutine initialize2(ni,nj, currtime)
        call bgc_vegetation_inst%Init2(bounds_proc, NLFilename)
     end if
 
+    if ( .not. for_testing_bypass_init_after_self_tests() )then
```
ekluzek: I need to indent the lines below. I left it unindents until I saw if it could run, and that this is the spot where it should be.

src/main/clm_initializeMod.F90
```diff
@@ -510,6 +515,7 @@ subroutine initialize2(ni,nj, currtime)
     if (nsrest == nsrContinue ) then
        call htapes_fieldlist()
     end if
+    end if
```
ekluzek: Indent lines above.

src/main/clm_initializeMod.F90
```diff
@@ -684,6 +690,7 @@ subroutine initialize2(ni,nj, currtime)
        call hist_htapes_build()
     end if
 
+    if ( .not. for_testing_bypass_init_after_self_tests() )then
```
ekluzek: Same thing about the indents.

It's also not completelly clear how many of these type of things there should be.

One reason that I need to do this here -- rather than having return statements, is that I need the timers to work. So if a timer has been started in the code above, I can't return in the middle without that timer being messed up. So it needs to exucute the part of the code where the timer is stopped.

src/main/clm_instMod.F90
```diff
@@ -269,7 +270,9 @@ subroutine clm_instInit(bounds)
     call humanindex_inst%Init(bounds)
 
     ! Initialize urban time varying data
-    call urbantv_inst%Init(bounds, NLFilename)
+    if ( .not. for_testing_bypass_init_after_self_tests() )then
```
ekluzek: In some of my decompInit testing work, I found where the urbantv calll specifically was problematic, so I'm explicitly just avoiding this one call here.

src/self_tests/SelfTestDriver.F90
```diff
@@ -46,4 +54,85 @@ subroutine self_test_driver(bounds)
...
+    !
+    ! !LOCAL VARIABLES:
+    integer :: ierr ! error code
+    integer :: unitn ! unit for namelist file
+
+    ! Namelist name: this has to be matched with the name in the read stqatement
+    character(len=*), parameter :: nmlname = 'for_testing_options'
+    !-----------------------------------------------------------------------
+
+    namelist /for_testing_options/ for_testing_bypass_init, for_testing_bypass_run
```
ekluzek: Some of the other for_testing namelist options should move to here as well.

src/self_tests/SelfTestDriver.F90
```diff
@@ -46,4 +54,85 @@ subroutine self_test_driver(bounds)
...
+       else
+          call endrun(msg="ERROR finding "//nmlname//"namelist", file=sourcefile, line=__LINE__)
+       end if
+       close(unitn)
+    end if
+
+    call shr_mpi_bcast (for_testing_bypass_init, mpicom)
+    call shr_mpi_bcast (for_testing_bypass_run, mpicom)
+
+    if (masterproc) then
```
ekluzek, line 119: There also should be some error checking done here in the Fortran code.

src/self_tests/SelfTestDriver.F90
```diff
@@ -46,4 +54,85 @@ subroutine self_test_driver(bounds)
...
+       write(iulog,nml=for_testing_options)
+       write(iulog,*) ' '
+    end if
+
+  end subroutine self_test_readnml
+
+  !-----------------------------------------------------------------------
+
+  logical function for_testing_bypass_init_after_self_tests()
+    ! Determine if should exit initialization early after having run the self tests
```
ekluzek, line 131: Both of these should get more sosphisticated and do things like NOT return true until after the self-tests are run. And ensure that the self_test namelist was read in, or else abort. The run phase bypass should maybe NOT return true until after the init phase has passed and things like that.

samsrabin:
@ekluzek This is the PR I was asking about earlier. Is it ready for review or still under development?

ekluzek:
> @ekluzek This is the PR I was asking about earlier. Is it ready for review or still under development?

Ahh, actually both. I want to get some feedback on it now. But, I think there are several things I need to do, before a final review. I want to get feedback on if this looks like even a good direction to pursue, as if not I'll brainstorm on a different direction. I also want to get some feedback now, if it does look promising on the list of things to do before merging it into b4b-dev.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -1242,6 +1242,22 @@ Whether to use subgrid fluxes for snow
...
 </entry>
 
+<!-- ========================================================================================  -->
+<!-- Namelist items for doing specific things for testing                                      -->
+<!-- for_testing section:                                                                      -->
+<!-- ========================================================================================  -->
+
+<entry id="for_testing_bypass_init" type="logical" category="default_settings"
+       group="for_testing_options" >
+For testing whether to bypass the rest of the initiatlization after the self test driver is run
```
samsrabin:
```suggestion
For testing whether to bypass the rest of the initialization after the self test driver is run
```

samsrabin (commented):
This is exactly how I was envisioning you implementing the self tests, so feel free to continue!

cime_config/testdefs/testmods_dirs/clm/run_self_tests/user_nl_clm
ekluzek: I did this.

src/main/clm_driver.F90
ekluzek: @samsrabin this is one thing I'd like to hear feedback on. Let me know what you think about this.

src/main/clm_instMod.F90
ekluzek: Actually, this isn't where the problem was.

ekluzek: So I should probably remove this one.

ekluzek (commented):
Some notes I'm making on things to do.

src/main/clm_initializeMod.F90
ekluzek: @samsrabin this and the one above about the return statements are things I'd like to hear from you on. In here they need to be if statements as I say above, but in the above section they could be either.

Although, maybe because this is a weird pathway in the code -- the return statements should be preferred so it doesn't disrupt the code flow as much as if statements. It's easy to miss returns in the code, flow but that's only important when it's something that might happen commonly enough. There was a code standard in my past that recommended to not have return statements in the midst of subroutines, because they are easy to miss. And I do see that point...

samsrabin: I prefer early returns as much as possible, especially since as you said these are rare cases. Avoiding returns by using `if` statements creates its own readability issues: How far indented am I right now? Under what conditions?

The bigger readability problem with this module is how enormous the subroutines are. I'd prefer to see things like `initialize2` getting refactored before we start worrying about ifs vs. returns. Maybe that could be something you do before this PR? I.e., refactor that subroutine to create at least two new ones:
1. Everything inside the first new `if`
2. Everything inside the second new `if`

Maybe also (3) everything between the new `if`s.

I wouldn't want to see that done in this PR, though, in the interest of keeping things easy to review and test.

samsrabin: Refactoring could also avoid return statements too: Break subroutines into two smaller subroutines, with the call of the first being wrapped in an `if` statement.

src/main/clm_driver.F90
samsrabin: Well this one isn't just an early return; it actually short-circuits the entire subroutine. So why not just wrap its call in an `if` statement?

samsrabin: I'm seeing now that the other `return`s are the same way, so same question there.

ekluzek: I really like your idea @samsrabin. For performance it's of course better to do at the higher level. For readability it might depend on where you want these things to be seen.

But, in this case it would remove this strange testing option from code that's a mix of science/software infrastructure to the higher level in the NUOPC cap that is pretty much just for SE's. So it helps with readability as well as separation of concerns.

src/main/clm_initializeMod.F90
ekluzek: All good points, thanks for the comments. I'll be thinking about all this.  I can't do a lot of work in refactoring clm_initialize and won't do it here. But maybe it would help to do some simple things in a different small PR. Hmmm...

Issue #3346, opened 2025-07-23
C Isotope data for CMIP7 for clm6_0 physics


wwieder:
Raw data for CO₂ Isotope data for CMIP7 are available have been dowloaded:
`/glade/campaign/cesm/cesmdata/input4MIPs_raw/input4MIPs/CMIP7/C4MIP/ImperialCollege/ImperialCollege-3-0`

CMIP7 data start in 1700 and include 4 sectors for 14C data, whereas CMIP6 data started in 1850 and had only 3 sectors for 14C

Sample cmip6 data are here `/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/isotopes/atm_delta_C13_CMIP6_1850-2015_yearly_v2.0_c190528.nc` (global values) and `/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/isotopes/atm_delta_C14_CMIP6_3x1_global_1850-2015_yearly_v2.0_c190528.nc` (which had 3 sector for Northern Hemisphere, Tropics, & Southern Hemisphere)

Is additional work needed to process the raw data?  
I'm imagining changes will be needed to define the new sectors for 14C data?  
Who has institutional knowledge about these data?

Definition of done:
- [x] Scope out what needs to be done and the options for moving forward
- [x] @slevis-lmwg Get a copy of the dataset into inputdata on Derecho
- [x] @slevis-lmwg  @ekluzek Check if there is any added metadata we need (also make sure the metadata is there for use with streams in the future)
- [x] @slevis do the [list below](https://github.com/ESCOMP/CTSM/issues/3346#issuecomment-3237669273)
- [x] @ekluzek do the [list below](https://github.com/ESCOMP/CTSM/issues/3346#issuecomment-3237669273)
- [x] Do some checking to make sure it's correct
- [x] Make an issue of problems
- [x] Add RC13_CANAIR/RC14_CANAIR to b4b-dev so cmip6 can be examined
- [ ] Do some checking to see if the problems occur in the cmip6 version
- [x] Make a tag on the ciso branch
- [x] Bring into the ctsm5.4 alpha branch and make a tag with it there
- [x] Make a script to convert to streams format @slevis-lmwg 
- [x] Create the new data files @slevis-lmwg 
- [x] Change the code to use streams @ekluzek 
- [x] Test with streams, compare to older version
- [x] Decide what the settings for spatial and time interpolation should be
- [ ] Bring into b4b-dev branch
[...]

PR #3465, opened 2025-08-30
Update Carbon isotope data to CMIP7

### LINKED ISSUE #3528: Preliminary CMIP7 C14 Carbon isotope dataset is upside down!
### Brief summary of bug

It appears that the simulation run with the preliminary CMIP7 C14 dataset has the latitude bands upside down.

### General bug information

**CTSM version you are using:** branch_tags%2FCMIP7_ciso.n01.alpha-ctsm5.4.CMIP7.11.ctsm5.3.075

**Does this bug cause significantly incorrect results in the model's science?** Yes -- just for C14 by latitude

**Configurations affected:** When use_c14 and use_c14_bombspike are on

### Details of bug

In preindustrial periods C14 on the input dataset goes from low in the SH to higher values in the NH. The pattern changes once the values skyrocket due to nuclear testing. But, in looking with @wwieder for one specific year it looks like it is clearly upside down.

This can be seen by looking at RC14_CANAIR or looking at variables influenced by C14

ekluzek:
### Description of changes

Start bringing in updates for the CMIP7 Carbon isoptope data.

### Specific notes

Contributors other than yourself, if any: @slevis-lmwg 

  Work on #3346
  Fixes #3528

Are answers expected to change (and if so in what way)? When it's turned on

Any User Interface Changes (namelist or namelist defaults changes)? Default for C13/C14 data will be updated

Definition of Done:
- [x] Change units of time expected on file from years to days
- [x] Handle the opposite direction that the latitude sectors are for CMIP6 and CMIP7
- [x] Do enough testing of a hardcoded version to CMIP7 that science experiements can run
- [x] Make a branch tag of it for those purposes
- [x] Fix the latitude problem #3528

### DIFF
--- a/src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
+++ b/src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
@@ -228,7 +228,6 @@ contains
     c13_previous_time_idx = ind_below
 
     ! interpolate between nearest two points in atm c13 timeseries
-    ! cdknotes. for now and for simplicity, just use the northern hemisphere values (sector 1)
     if (ind_below .eq. 0 ) then 
        delc13o2_atm = atm_delta_c13(1)
     elseif (ind_below .eq. ntim_atm_ts ) then


ekluzek:
Note, NEON/PLUMBER2 cases use a SSP scenario, which means they will point to CMIP6 isotope data. We'll have to do some reworking for them to point to the CMIP7 datasets.

ekluzek:
Not fully tested, but seems to have the minimum required to get this to work now.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
```diff
@@ -158,7 +158,10 @@ subroutine C14_init_BombSpike()
     call ncd_pio_closefile(ncid)
 
     ! check to make sure that time dimension is well behaved
+    atm_c14file_time(1) = atm_c14file_time(1) / get_average_days_per_year()
     do t = 2, ntim
+       ! Convert time in days to years
+       atm_c14file_time(t) = atm_c14file_time(t) / get_average_days_per_year()
```
slevis-lmwg: @ekluzek here and elsewhere, does get_average_days_per_year() always return 365? The c13 and c14 files use a 365-day calendar:
`time:calendar = "365_day"`
...so for an accurate conversion from days to years the denominator needs to equal 365 every time.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
```diff
@@ -132,7 +132,7 @@ subroutine C14_init_BombSpike()
     call ncd_pio_openfile (ncid, trim(locfn), 0)
 
     call ncd_inqdlen(ncid,dimid,ntim,'time')
-    call ncd_inqdlen(ncid,dimid,nsec,'sector')
+    call ncd_inqdlen(ncid,dimid,nsec,'lat')
     if ( nsec /= nsectors_c14 )then
        call endrun(msg="ERROR: number of sectors on file not what's expected"//errMsg(sourcefile, __LINE__))
     end if
```
slevis-lmwg, line 150: A few concerns about "sectors", with my third concern being most important:
1) Change nsec and other similar variables to n_lat_sec or similar to indicate latitude sectors?
2) The comment in new line number 210 (originally 205) is wrong. It may have been wrong before, too. The c13 file has global values and does not include sectors of any kind.
3) Most importantly, the CMIP7 sectors are in reverse order relative to the CMIP6 sectors:
CMIP6 `sector:lat_bnds = "1: 30.0, 90.0; 2: -30.0, 30.0; 3: -90.0, -30.0"`
CMIP7 `lat:lat_bnds = "1: -90.0, -30.0; 2: -30.0, 0.0; 3: 0.0, 30.0; 4: 30.0, 90.0"`
To address this in the code, you may just need to reverse the sector loop. Unless...
- you handled this some other way, and I missed it
- or unless my suggestion is wrong and there's a different solution
- or unless the order ultimately does not matter because we convert the c14 data to global in our simulations

slevis-lmwg (changes requested):
Thank you @ekluzek 
1) I see [your point](https://github.com/ESCOMP/CTSM/issues/3346#issuecomment-3244006436), we may not need to preprocess the c13 and c14 files, which is good I think. But we may want to add a comment stating that the new code is not backwards compatible with CMIP6.
2) I looked at your code changes. In this review I expressed a couple of concerns based on my inspection of the files. Let me know if you want to follow up.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: Yes, this is one of the things I was going to point out, that we'll need to fix in a final version. And we might need to do something slightly different depending on what calendar the model is running. Right now assuming the default calendar of no_leap -- it works correctly. But, if the model is running "gregorian" it will be a slightly off, as you point out.

So thanks for bringing this up.

ekluzek: I also think we should convert the units the code uses from "years" to "days", so we should do that as well.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: 1. Sector name: 

Naming of sectors is a good question. I'm not sure what would be best. If we change it -- I'd go for longer something like "num_latitude_sectors".

4. Comment for C13:

Ahh, good point it was probably wrong before.

7. Sector order:

Yep, I did not pick that up -- so I'm glad you did. This is something the code should really check for. In the end we want the CMIP6 and CMIP7 versions to work so it'll need to be a general solution that can work for both. I think we'll want to include the expected latitude bnds for both CMIP6 and CMIP7 so that it can do either one.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
slevis-lmwg: > Yes, this is one of the things I was going to point out, that we'll need to fix in a final version. And we might need to do something slightly different depending on what calendar the model is running. Right now assuming the default calendar of no_leap -- it works correctly. But, if the model is running "gregorian" it will be a slightly off, as you point out.

Sounds good. So for now a clear comment along those lines would suffice.

ekluzek:
I realized there's another part where it's hardcoded to the old dataset. Here in PhotosynthesisMod

```
               ! determine latitute sector for radiocarbon bomb spike inputs
               if ( grc%latdeg(g) .ge. 30._r8 ) then
                  sector_c14 = 1
               else if ( grc%latdeg(g) .ge. -30._r8 ) then            
                  sector_c14 = 2
               else
                  sector_c14 = 3
               endif
```

So that needs to be handled.

ekluzek:
The history variables for this are default off, but are: RC14_CANAIR, RC13_CANAIR

ekluzek:
OK, and it looks like RC14_CANAIR needs to be added as a history output variable.

ekluzek:
Here's a conversion to Delta using ncap2, I'm trying out:

``` shell
 ncap2 -s 'c14ratio = 1.e-12' -s 'PDB = 0.0112372' -s 'preind_atm_del13c = -6.' \
-s 'preind_atm_ratio = PDB + (preind_atm_del13c * PDB)/1000.0' -s 'c13ratio = preind_atm_ratio/(1.0+preind_atm_ratio)' \
-s 'Deltac13 = (RC13_CANAIR/PDB - 1.0)/1.0e-3' -s 'Deltac14 = (RC14_CANAIR/c14ratio - 1.0)/1.0e-3'  infile.nc outfile.nc
```

ekluzek:
Results are looking good so far. But, I'm running a test for a few years to make sure time is correct.

ekluzek:
Ok. It looks like there's something wrong with the time interpolation. So I'm looking into this more.

ekluzek:
I added some logging for time, and saw there were problems with both c13 and c14 time. So I've modified it so that the time is in units of CE year, and can then be directly compared to dateyear for the current time. This should resolve the issues with time. Running a test for this now.

wwieder:
Thanks for working on this and providing updates, Erik.

ekluzek:
Tag to make:
branch_tags/CMIP7_ciso.n01.alpha-ctsm5.4.CMIP7.11.ctsm5.3.075

wwieder:
I'm spinning this up in an [ne30 bgc-crop test](https://github.com/NCAR/LMWG_dev/issues/123). If we're OK with how these results look, I suggest we **don't** worry about maintaining the flexibility of using CMIP6 vs. CMIP7 datasets for carbon isotopes. 

Other input data are easier to maintain with streams files (e.g., #3471 to switch between CMIP6 and CMIP7 data), but given the brittle nature of how we handle C isotopes I suggest we only support CMIP7 13C and 14C datasets moving forward. It's not not worth the overhead costs of maintaining code to ingest both files at this point.  Happy to have further discussion on this if it's helpful.

ekluzek:
@slevis-lmwg and I went over this, and I'll bring this into the ctsm5.4 alpha branch largely as is at this point, even though it has problems. It's already been used to run simulations so we want to preserve that version. And the work to do this with streams should be an independent PR.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: By converting to streams this will be handled more robustly because streams can take different calendars. So I'm resolving for the purposes of this PR.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: I've removed the incorrect comment about C13 sectors.

ekluzek: The issues about naming of sectors goes away with the move to streams, so resolving this comment for now.

ekluzek: I resolved the sector order issue incorrectly at first, but with the fix to #3528in b7786b580a0911b4e790789b5843991a74adfc8e  I did get it working correctly.

src/biogeophys/PhotosynthesisMod.F90
slevis-lmwg, line 1: I see that this is the needed correction.

ekluzek:
A new problem I realized was that we have CMIP6 SSP datasets, which don't work in this setup. The way forward we decided on is to turn Carbon isotopes off for SSP scenarios even for clm6_0. And we might as well turn carbon isotopes off for some compsets such as single point cases.

I'm going to make a tag of this branch, and then close without merging once we have the streams way forward up and working.

ekluzek:
I tagged the fixed version as  branch_tags/CMIP7_ciso.n02.alpha-ctsm5.4.CMIP7.11.ctsm5.3.075, so the fix is preserved in a tag in case we need it. This will also allow us to compare the new answer with streams to this version to validate the streams method.

ekluzek:
Replaced by work on b4b-dev with the latest bit being in #3627 so closing.

PR #3470, opened 2025-09-04
Add fortls to ctsm_pylib, add a .vscode directory and a couple settings

ekluzek:
### Description of changes

Add fortls to ctsm_pylib so it can be used in VS Code FORTRAN-IntelliSense extension.

Add a .vscode directory and a couple settings:
- Ignore the same files in VSCode that git does
- Handle .pf files as Fortran

### Specific notes

Are answers expected to change (and if so in what way)? No No

Any User Interface Changes (namelist or namelist defaults changes)? No No

Testing performed, if any: 

Use py_env_create to create ctsm_pylib on Derecho and Izumi
Ran python unit/sys tests on Derecho


.vscode/settings.json
samsrabin, line 1: I'm not sure it's a good idea to add this file, because devs and users will often have their own version and/or want to customize this one. Perhaps instead we could have a `settings_template.json` or something similar that VS Code doesn't actually use.

samsrabin (commented):
Good idea! One thing to think about, though.

.vscode/settings.json
adrifoster: I agree with @samsrabin. In fact, I think we should add anything in the .vscode folder to our .gitignore. As a pretty heavy users of that folder, I really don't want my own personal stuff getting fiddled with. And I'd like to have this folder for every branch of CTSM (rather than only on my specific branches, etc.).

I like the idea of having a sample template to set some basic things up for folks. We can add things here as we find things that are helpful. We could also have some others in the tools/contrib folder.

ekluzek: Interesting. Thanks for the feedback.

One thing about this though is that there already are different levels of .vscode settings directories for the User level, and remote level. So I'm not sure you HAVE to keep this separate from your personal settings. You can still set the user and remote level (provided your remote points to a directory that includes the git sandbox and not just the git sandbox).

So I would envision this would be a barebones level settings that we could could agree would be useful to all. And then you add user and remote specifics on top of it. Provided that doesn't mess anyone up, I'd prefer that as it just automagically happens for someone who's new to VS Code. I'd like to make using VS Code easier for everyone.

But, I'm also fine with making it a template. I do like the idea of making sure there is at least the directory in the repository. The other advantage of a template is that we could fill out and example of the fortls path, as that's a part that you have to customize for your account and machine.

@adrifoster I also suspect that there are things in your .vscode settings that would be useful to others. So it would be nice to have them available. So getting this started will allow you and others to start adding things here.

@adrifoster let's meet when you have a chance. Thanks so much!

samsrabin: @ekluzek There are different levels of VS Code settings, yes, but the "Workspace" one is not any more or less important than the others. It sounds like you are open to removing this file, so let's do that. Thanks!

samsrabin: Oh, but also: I would say go ahead and delete the file then re-request review, rather than waiting to meet to figure out about a perfect .vscode template and how to distribute it. That can be a separate PR.

ekluzek: OK, I moved it to a file with the template ending. So I think it's good to go now.

samsrabin (approved):

PR #3475, opened 2025-09-08
ctsm5.3.075: Change default glcmec_downscale_longwave from true to false for clm6

### LINKED ISSUE #3467: ctsm5.3.075: turn off LW downscaling over glacier elevation classes by default
In tuning up the SMB over Greenland, @whlipscomb, @gunterl and I found that turning off the LW downscaling improves the melt and runoff biases, and would like it be the default in the CESM3 release. Specifically,

```
glcmec_downscale_longwave = .false.
```

as it currently defaults to true.

slevis-lmwg:
### Description of changes
See title.

### Specific notes

@adamrher @whlipscomb @gunterl

Resolves #3467 

Yes, clm6 only.

See title.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,74 @@
 ===============================================================
+Tag name: ctsm5.3.075
+Originator(s): slevis (Samuel Levis,UCAR/TSS,303-665-1310)
+Date: Tue 09 Sep 2025 03:12:17 PM MDT
+One-line Summary: Change default glcmec_downscale_longwave from true to false for clm6
+
+Purpose and description of changes
+----------------------------------
+
+ Why: Adam Herrington, Bill Lipscomb, Gunter Leguy found that turning off the LW downscaling improves the melt and runoff biases.
+
+ How: Change glcmec_downscale_longwave to false for clm6 in namelist_defaults_ctsm.xml.
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+
+Does this tag change answers significantly for any of the following physics configurations?
+(Details of any changes will be given in the "Answer changes" section below.)
+
+    [Put an [X] in the box for any configuration with significant answer changes.]
+
+[ ] clm6_0
+
+[ ] clm5_0
+
+[ ] ctsm5_0-nwp
+
+[ ] clm4_5
+
+
+Bugs fixed
+----------
+List of CTSM issues fixed (include CTSM Issue # and description) [one per line]:
+ Resolves #3467 Turn off LW downscaling over glacier elevation classes by default
+
+Notes of particular relevance for users
+---------------------------------------
+Changes made to namelist defaults (e.g., changed parameter values):
+ As explained above.
+
+Testing summary:
+----------------
+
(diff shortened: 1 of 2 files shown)


slevis-lmwg:
`./run_sys_tests -s aux_clm -c ctsm5.3.073 --skip-generate`
- izumi OK

`./run_sys_tests -s aux_clm -c ctsm5.3.074 -g ctsm5.3.075`
- [x] izumi OK, repeating with https://github.com/ESCOMP/CTSM/pull/3475/commits/4b521805f21584d2a8d29db28ef4384f62eb4dd7
- [x] derecho OK, repeating with https://github.com/ESCOMP/CTSM/pull/3475/commits/4b521805f21584d2a8d29db28ef4384f62eb4dd7

build-namelist_test OK

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -227,7 +227,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <repartition_rain_snow               >.true.</repartition_rain_snow>
 <repartition_rain_snow phys="clm4_5" >.false.</repartition_rain_snow>
 
-<glcmec_downscale_longwave>.true.</glcmec_downscale_longwave>
+<glcmec_downscale_longwave>.false.</glcmec_downscale_longwave>
```
slevis-lmwg, line 237: @ekluzek 
aux_clm seems promising so far, so I requested your review.

ekluzek: The question is if we change it for all physics version, or just clm6_0. I think we should just do this for clm6_0, which would allow you to show that answers don't change for the other physics modes. This also keeps our older physics as close as possible to the settings they had when they were set.

It's also good that you are currently testing it for all physics, because that will ensure it doesn't cause problems for any of our tests in case we decide to do it for all tests.

wwieder: I'd be inclined to just do this for CLM6 physics to avoid changing answers in CLM5.

slevis-lmwg: Thanks @ekluzek, makes sense.

ekluzek: I agree with @wwieder especially after seeing it was set this way for both clm4_5 and clm5_0.

ekluzek: I also checked and didn't think there is documentation that needs to be updated.

But, I think you should add a comment above this that talks about what Adam and Gunter found. This is more or less an advanced option, typically we turn on newer options with newer physics, but in this case we are doing the opposite of the normal flow. Here we are turning a more complex option off for the newer physics. So it needs a comment to explain why we are doing that.

ekluzek (changes requested):
We just need to keep it the same for the older physics versions. And also add a comment  about why we are doing this in the newer physics.

slevis-lmwg:
Thank you both for the reviews and helpful comments.

PR #3364, opened 2025-07-26
For ctsm5.4 replace WhatsNewInCTSM5.3.md with WhatsNewInCTSM5.4.md

slevis-lmwg:
### Description of changes
I used WhatsNewInCTSM5.3.md as the starting point for WhatsNewInCTSM5.4.md and will start modifying in subsequent commits.

### Specific notes

Contributors other than yourself, if any:
@ekluzek 

CTSM Issues Fixed (include github issue #):
Resolves #3003 

Are answers expected to change (and if so in what way)?
No

Any User Interface Changes (namelist or namelist defaults changes)?
No

Does this create a need to change or add documentation? Did you do so?
This PR updates WhatsNewInCTSM5.3.md to WhatsNewInCTSM5.4.md

Issue #3250, opened 2025-06-12
ISSP tests need to point to CMIP6 ctsm5.3 fsurdat/landuse files to PASS


slevis-lmwg:
### Brief summary of bug

Fails in the 5.4 branch because we have not generated ctsm54 SSP landuse files, yet. The test should work with the clm54 landuse file that we generated out to 2023, but this would mask the problem, and we would likely forget to fix it correctly later. I also considered changing to a grid that still uses the ctsm53 data, but such alternate grid is not available for SSP126. As a result I'm labeling the test as EXPECTED FAILURE and filing this issue.

### General bug information

**CTSM version you are using:** output of `git describe` is ctsm5.3.055-24-g84dcbe0ac PR #3206 soon to be merged to the 5.4 branch

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** SSP in the 5.4 branch

slevis-lmwg:
#3254 intends to change all SSP tests from 5.0 to 6.0. This issue remains the same regardless.

slevis-lmwg:
Marked "blocked" due to the dependency on CMIP7 SSP raw datasets that are unavailable as of yet.

slevis-lmwg:
Same problem when running
`./run_sys_tests -s aux_clm -g alpha-ctsm5.4.CMIP7.09.ctsm5.3.068 --skip-compare`
but the test was `SMS_D_Ld5.f09_g17.ISSP245Clm60BgcCropCrujra.derecho_intel.clm-default`
so I will make the issue title more general.

ekluzek:
What we will do here is to point to the CMIP6 ctsm5.3 fsurdat/landuse files for SSP compsets.

ekluzek:
This will be resolved by #3471 and when it's completed the expected fail file can be updated to remove this fail.

slevis-lmwg:
For Sprint planning purposes this is done.
I will keep the issue open until I merge #3501.

slevis-lmwg:
addressed in #3501

PR #3501, opened 2025-09-24
Add support for CLM_CMIP_ERA (= "cmip6" or "cmip7")

### LINKED ISSUE #3471: Add new CLM_CMIP_ERA XML variable in order to support CMIP6 SSP scenarios
In order to support future scenarios for cesm3.0 we will be supporting the ctsm5.3 SSP surface datasets. In order to do this, we will add a new XML variable to control it. By default it will be set to cmip6 for the SSP's and cmip7 for everything else. The user will be able to change it to cmip6 for a cmip7 default compset, but there may not be resolutions supported for what they are trying to do -- and it'll come back with saying we don't have a dataset for this resolution/compset. We will limit the support to only 1850 and/or SSP and for fewer resolutions.

There's a few other datasets where we have a single dataset to define it, and we can easily keep both the cmip6 version and the cmip7 version. So we might as well do this. And with the variable being explicit and required for the landuse.timeseries files -- we might as well have other datasets be pointing to the cmip6 version at the same time for consistency.

When we have future scenarios for CMIP7 -- we plan to remove the CMIP6 support. This is expected to be a ctsm6.1 activity for cesm3.1.

We started discussing this in CTSM SE meeting today and notes are in this document:

https://docs.google.com/document/d/1Np69hB9Rx0L19OqOlhtI6bH48D4oYxPwgmrGofrdgVM

Definition of done:
- [x] Add XML variable to config_component.xml in cime_config: CLM_CMIP_ERA
- [x] In config_component.xml have valid_options be cmip6/cmip7 and same in namelist_definition_ctsm.xml
- [x] Read it in the perl build-namelist so it can be used, similar to other XML variables read in
- [x] Add a function to handle it being set in build-namelist in $nl_flags
- [x] In namelist_defaults_ctsm.xml add an attribute: cmip_era="cmip6" or cmip_era="cmip7" to the places it's needed (1850 fsurdat and SSP's)
- [x] Error check: What happens when you ask for an SSP with cmip7? Fail in a graceful manner that tells you what to do
- [x] Add a test for above to the namelist tester 
- [x] Run namelist tester to make sure it still works
- [x] Run aux_clm and ctsm_sci (which runs some of these SSP's)
- [x] Address in https://github.com/ESCOMP/CTSM/issues/3255: Run all the SSPs in ctsm5.4 branch and make sure they all work and use the CMIP6 datasets (f09 for all, SSP2-4.5 for a bunch) (for ones not in aux_clm or ctsm_sci)

slevis-lmwg:
### Description of changes
Definition of DONE in the issue #3471 

### Specific notes

@ekluzek 

Resolves #3471 

No.

The CLM_CMIP_ERA flag became necessary because we will not have SSP and NOANTHRO pft raw datasets for CTSM5.4. This means that SSP and NOANTHRO cases default to cmip6 fsurdat/landuse files, while other cases default to cmip6 fsurdat/landuse files.

May need to update #3364 

Testing performed, if any:
So far...
```
./create_test SMS_Ld5_Mmpi-serial.1x1_brazil.IHistClm60Bgc.derecho_gnu.clm-mimics
./create_test SMS_D_Ld5.f10_f10_mg37.ISSP245Clm60BgcCropCrujra.derecho_intel.clm-default
./create_test SMS_D_Ld5.f09_t232.ISSP245Clm60BgcCropCrujra.derecho_intel.clm-default
```
- The first two PASS. They correctly set cmip_era = "cmip7" and "cmip6" respectively, but they do not use that information because these grids do not have cmip7 1850 fsurdat files, yet.
- The third incorrectly picks up the cmip7 fsurdat file because I need corrections in CLMBuildNamelist.pm.


slevis-lmwg:
@ekluzek this is how far I got today. The next two days will be very busy, and I may take Friday off. I would benefit from your corrections and/or comments on where to focus next, as this perl/xml interface is not my forte :-)

bld/namelist_files/namelist_defaults_cmip_era.xml
slevis-lmwg, line 1: From meeting with @ekluzek 
Better in namelist_defaults_ctsm.xml.
The separate ones are separate because their namelists are in driver.

bld/CLMBuildNamelist.pm
```diff
@@ -735,6 +735,11 @@ sub setup_cmdl_resolution {
     $log->verbose_message( "This is NOT a NEON site" );
   }
 
+  #
+  # To determine CMIP era
+  #
+  $nl_flags->{'cmip_era'} = $envxml_ref->{'CLM_CMIP_ERA'};
```
slevis-lmwg, line 744: From meeting with @ekluzek 
Placement seems fine in this subroutine.

bld/namelist_files/namelist_defaults_cmip_era.xml
```diff
@@ -0,0 +1,23 @@
...
+are: RESOLUTION, defaults, mask, ic_ymd, ic_tod, sim_year and all configuration
+attributes from the config_cache.xml file (with keys converted to upper-case).
+-->
+
+<!-- Defaults for cmip_era_inparm namelist  -->
+<cmip_era>cmip7</cmip_era>
+<cmip_era sim_year="PtVg" >cmip6</cmip_era>
+<cmip_era sim_year_range="1850-2100" >cmip6</cmip_era>
+<cmip_era sim_year_range="2000-2100" >cmip6</cmip_era>
+<cmip_era sim_year_range="2000-2025" >cmip6</cmip_era>
```
ekluzek: It looks to me like 2000-2025 could be removed. And looking at other sim_year_range values it looks like many of them can be removed

So remove the line regarding it. And I'll open an issue about removing more sim_year_range options.

slevis-lmwg: The last of these sim_year_ranges is obsolete and never gets invoked, so remove as unnecessary.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -1825,8 +1829,10 @@ lnd/clm2/surfdata_esmf/ctsm5.3.0/surfdata_mpasa120_hist_1850_78pfts_c240908.nc</
 lnd/clm2/surfdata_esmf/ctsm5.3.0/surfdata_ne30np4_hist_1850_78pfts_c240908.nc</fsurdat>
 <fsurdat hgrid="ne30np4.pg2" sim_year="1850">
 lnd/clm2/surfdata_esmf/ctsm5.3.0/surfdata_ne30np4.pg2_hist_1850_78pfts_c240908.nc</fsurdat>
-<fsurdat hgrid="ne30np4.pg3" sim_year="1850">
+<fsurdat hgrid="ne30np4.pg3" sim_year="1850" cmip_era="cmip7">
 lnd/clm2/surfdata_esmf/ctsm5.4.0/surfdata_ne30np4.pg3_hist_1850_78pfts_c250723.nc</fsurdat>
+<fsurdat hgrid="ne30np4.pg3" sim_year="1850" cmip_era="cmip6">
+lnd/clm2/surfdata_esmf/ctsm5.3.0/surfdata_ne30np4.pg3_hist_1850_78pfts_c240908.nc</fsurdat>
```
slevis-lmwg, line 1837: From meeting with @ekluzek 
- [x] Adding cmip7 flag to the same three grids for historical landuse files but not the cmip6 flag.
- [x] Check how asking for cmip6 fails for such a case (used one of the three tests above):
`SMS_Ld5_Mmpi-serial.1x1_brazil.IHistClm60Bgc.derecho_gnu.clm-mimics.20250926_165339_s5tc79`
This one PASSed because finidat = ' ' so try...
`SMS_D_Ld5.f09_t232.IHistClm60BgcCrop.derecho_intel.clm-default.20250930_162652_qtf9jl`
Currently returns the correct error with the correct question:
```
No default value found for flanduse_timeseries.
Are defaults provided for this resolution and land mask?
```
Add "CLM_CMIP_ERA (in env_run.xml)" at the end of the question above.
- [x] Also check how asking for cmip7 fails for an SSP case (used one of the three tests above):
`SMS_D_Ld5.f09_t232.ISSP245Clm60BgcCropCrujra.derecho_intel.clm-default.20250926_165802_qsb9xj`
Currently returns an error message with the correct instructions:
```
Confirm that the year of your surface dataset corresponds to the first year of your landuse_timeseries file (e.g., for a landuse_timeseries file starting at year 1850, which is typical, you should be using an 1850 surface dataset), and that your landuse_timeseries file is compatible with the surface dataset.

If you are confident that you are using the correct landuse_timeseries file and the correct surface dataset, then you can bypass this check by setting:
   check_dynpft_consistency = .false.
in user_nl_clm
```
From meeting with @ekluzek: Add cmip_era="cmip6" to all SSPs in namelist_defaults.
- [x] Ensure graceful failures
- [x] Add test(s) to the namelist tester: 
  - cmip7 w ISSP needs to FAIL
  - cmip6 w I2000 needs to FAIL (not true as of [this later post](https://github.com/ESCOMP/CTSM/pull/3501#issuecomment-3362408446))
  - cmip6 w I1850 needs to PASS
  - cmip6 w IHist needs to PASS

bld/namelist_files/namelist_defaults_cmip_era.xml
slevis-lmwg: Turns out I can remove this new code. The flag will be set by config_component.xml and the user may override with xmlchange to env_run.xml.

bld/CLMBuildNamelist.pm
slevis-lmwg: TODO Add the cmip_era flag to all the add_default calls for fsurdat and flanduse.

ekluzek: @slevis-lmwg after our meeting I did more looking at the current design and realized that each separate command line option has a subroutine that goes with it. The reason that NEONSITE is handled in this one is that NEONSITE is essentially a resolution. It says it's going to be a single point site for a NEON location.

ekluzek: So I think what you should do is

- [x] Add a subroutine with a name something like prcoess_envxml_flags
- [x] Call it from process_namelist_user_input

For now it'll just have the bit about setting cmip_era from the XML CLM_CMIP_ERA, but then it allows other such XML variables to be set there as well.

bld/namelist_files/namelist_defaults_cmip_era.xml
slevis-lmwg: Removing this file.

slevis-lmwg:
Submitting the same three tests as above.

slevis-lmwg:
> Submitting the same three tests as above.

Same behavior in the first two, as I expected and hoped.
Behavior corrected now in the third, also as expected and hoped.
- [x] aux_clm on izumi and derecho
- [x] ctsm_sci retest on derecho (see checkbox below):
```
./run_sys_tests -s ctsm_sci --skip-compare -g ctsm_sci-alpha-ctsm5.4.CMIP7.12.ctsm5.3.075
./run_sys_tests -s aux_clm -c alpha-ctsm5.4.CMIP7.11.ctsm5.3.075 -g alpha-ctsm5.4.CMIP7.12.ctsm5.3.075
```
- [x] aux_clm FAIL RUNs to address:
```
ERP_D_Ld3_PS.f09_g17.I2000Clm50Sp.derecho_intel.clm-prescribed
```
floating divide by zero
899  CanopyFluxesMod.F90
Did it pass before the fsurdat changed? If I were to switch back the fsurdat, I think I would need to re-test the ERI case below.

ctsm_sci FAIL RUNs to address:
```
ERI_Ld9.f09_t232.I2000Clm60BgcCropCrujra.derecho_intel.clm-default
SMS.f09_g17.I1850Clm50BgcCropG.derecho_intel.clm-default
```
The ERI test points to 5.4 finidat and 5.3 fsurdat; update the fsurdat?
From meeting with @ekluzek: Yes, make a temporary fsurdat and update namelist_defaults.
- [x] This did NOT fix the need for use_init_interp. Looking at `git blame` in namelist_defaults_ctsm.xml current line 1515 suggests that this line has included `use_init_interp` historically. I removed it last month with 71f2db2 and may have not run ctsm_sci on it. So I'm adding back use_init_interp to this line and retesting.

The SMS test points to 5.3 finidat and 5.4 fsurdat; update the finidat?
From meeting with @ekluzek: No, update namelist_defaults to use_init_interp = TRUE.
- [x] Retest is in the queue
- [x] Confirm that tests failing in #3250 passed and remove from expected fails
- [x] After resolving the last failure, I plan on submitting aux_clm and ctsm_sci again, while I work on build-namelist_test.pl

slevis-lmwg:
Regarding most recent commit 14f647d:
For the record, this MODEL_BUILD passes also in my b4b-dev testing (ctsm5.3.077-30-gefe185a89).

On a different note, as per today's software meeting (2025/10/2):
Add cmip6 to IHist namelist_defaults.

slevis-lmwg:
@ekluzek now that I've added back the cmip6 defaults for IHist cases, it seems silly to omit them for I2000 cases, so I will add them for consistency.

slevis-lmwg:
Troubleshooting this aux_clm test:
/glade/derecho/scratch/slevis/tests_1001-161326de/ERP_D_Ld3_PS.f09_g17.I2000Clm50Sp.derecho_intel.clm-prescribed.GC.1001-161326de_int
See second lnd.log file which ends with `z0mv = 0` before entering a log function...

git blame suggests that the code used to have an `else` section for non-woody plants. Did that become obsolete because we assigned sai > 0 everywhere? Check that. The error appears now because we have sai = 0 (and displa = 0) somewhere. From what I see, the lai stream file contains ONLY lai (no sai) and ONLY for 16 pfts.

Tested new if statements forcing displa and z0mv > 0: Now I get "floating invalid" so do I need to set sai>0 too?

MIGHT WE RATHER MOVE THIS TEST TO CLM6, WHERE WE WOULD RUN Meier2022 INSTEAD OF ZengWang2007? Or at least make a clm6 version of the test? Also we could force pre-clm6 versions to use cmip_era = "cmip6" to bypass this problem...

Trying a clm6 version of the test.

slevis-lmwg:
The clm6 test gave the same "floating invalid" error.
Comparing the 5.4 (failing) and 5.3 (works) fsurdats, I see that pfts 15 and 16 have htop, hbot, lai, sai changed to zero in the 5.4:
1) Is this due to changes in the raw data or in the code?
2) Try putting non-zero values in the 5.4 file to see if it fixes the problem?

ekluzek:
> The clm6 test gave the same "floating invalid" error. Comparing the 5.4 (failing) and 5.3 (works) fsurdats, I see that pfts 15 and 16 have htop, hbot, lai, sai changed to zero in the 5.4:
> 
> 1. Is this due to changes in the raw data or in the code?
> 2. Try putting non-zero values in the 5.4 file to see if it fixes the problem?

Hmmm. The thing is that for this case it should be using the values from the LAI dataset, and not the fsurdat. But, also it seems like these values should never be identically zero for any PFT at any gridcell. And for both the LAI dataset and for the fsurdat file. What matters though is that PCT_CFT for 15 and 16 is non-zero AND that these values are zeroed.

Possibly we should trap for zero values in mksurfdata_esmf and CTSM.

slevis-lmwg:
> The clm6 test gave the same "floating invalid" error. Comparing the 5.4 (failing) and 5.3 (works) fsurdats, I see that pfts 15 and 16 have htop, hbot, lai, sai changed to zero in the 5.4:
> 
> 1. Is this due to changes in the raw data or in the code?

@ekluzek @lawrencepj1 I have identified a problem in the raw data. I compared these two files:
```
lnd/clm2/rawdata/CTSM53RawData/globalctsm53histMKSRFDeg025_240709/mksrf_landuse_ctsm53_pftlai_CLIM.c240709.nc
lnd/clm2/rawdata/CTSM54RawData/globalctsm53histMKSRFDeg025_250417/mksrf_landuse_ctsm53_pftlai_CLIM.c250417.nc
```
...and found that the 5.3 file has htop, hbot, mlai, msai to pft 15, while the 5.4 file stops at pft 14. This may explain why the tests
```
ERP_D_Ld3_PS.f09_g17.I2000Clm50Sp.derecho_intel.clm-prescribed
ERP_D_Ld3_PS.f09_g17.I2000Clm60Sp.derecho_intel.clm-prescribed
```
are segfaulting.

slevis-lmwg:
@ekluzek I didn't fully answer your question. The reason the "prescribed" test was the ONLY one to fail is twofold:
- The LAI stream only contains LAIs, so htop, hbot, and sai are 0 for pft 15 (this part I told you)
- Other tests find lai = 0 for pft 15 in the fsurdat file, so they skip CanopyFluxes and instead run through BareGroundFluxes and thus avoid crashing

bld/namelist_files/namelist_defaults_ctsm.xml
slevis-lmwg: This test `SMS_Ld1.f10_f10_mg37.ISSP245Clm60BgcCropCrujra.casper_intel.clm-default` fails as it should when I change it to cmip7. So far, I have not gotten the namelist test to fail. See test.out8 for latest attempt.

slevis-lmwg:
UPDATE FROM MEETING WITH @ekluzek 

Skip this list below. Instead, consider the "lai streams" test an expected failure and merge this PR. Resolve the expected failure in the later PR where I add Peter's new raw lai dataset and new raw noanthro dataset.

- Expect new raw lai dataset from Peter L.
- Pointing now to that, repeat `make all` to ensure it still works
- Point to the new fsurdat/landuse files in namelist_defaults
- Repeat this PR's complete testing:
  - ./build-namelist_test.pl
  - ctsm_sci
  - aux_clm
  - python

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -2364,6 +2364,15 @@ A sim_year_range of 1000-1002 or 1000-1004 corresponds to data used for testing
 A sim_year_range that goes beyond 2005 corresponds to historical data until 2005 and then scenario data beyond that point.
 </entry>
 
+<entry id="cmip_era" type="char*5" category="default_settings"
```
ekluzek: Just remove this and leave it as the previous version.

bld/unit_testers/build-namelist_test.pl
```diff
@@ -569,12 +569,32 @@ sub cat_and_create_namelistinfile {
 &make_config_cache($phys);
 foreach my $options (
```
ekluzek, line 571: Probably add a comment above the loop?

bld/unit_testers/build-namelist_test.pl
```diff
@@ -1832,15 +1857,21 @@ sub cat_and_create_namelistinfile {
...
       if ( ($usecase eq "1850_control") || ($usecase eq "20thC_transient") ) {
          $startymd = 18500101;
+         &make_env_run();
       } elsif ( $usecase eq "2000_control") {
          $startymd = 20000101;
+         &make_env_run();
       } elsif ( $usecase eq "2010_control") {
          $startymd = 20100101;
+         &make_env_run();
+      } elsif ( $usecase eq "1850-2100_SSP2-4.5_transient") {
```
slevis-lmwg: possibly change to any SSP usecase, not this exact one.

slevis-lmwg:
@ekluzek this PR is ready for me to merge and needs your approval first.

bld/namelist_files/namelist_definition_ctsm.xml
slevis-lmwg, line 1: I neglected to remove these changes (from Erik's last review).

slevis-lmwg: Ok, now done.

cime_config/config_component.xml
```diff
@@ -156,6 +156,23 @@
...
+      <value compset="NOANTHRO" >cmip6</value>
+      <value compset="SSP" >cmip6</value>
+    </values>
+    <group>run_component_ctsm</group>
+    <file>env_run.xml</file>
+    <desc>Whether to use CMIP6 or CMIP7 fsurdat/landuse files.
+    As of ctsm5.4 we do not have CMIP7 pft raw datasets for SSPs and NOANTHRO, so default to cmip6 for these. For other compsets default to cmip7. User may modify to a non-default value in env_run.xml. Caveats:
+1) We supply only cmip7 C13/C14 isotope datasets, so these get used regardless.
+2) We supply only cmip7 population density with ctsm6 in non-SSP cases, because the fire model is calibrated to that; conversely, for pre-ctsm6 and for SSP we supply only cmip6 population density.
+3) We supply only cmip6 nitrogen deposition (ndep), so this gets used regardless.</desc>
```
ekluzek: I think we should also add a note about at least aerosols from DATM. Right now CO2 is also cmip6. 

So when we get CMIP7 CO2 in place, do we trigger which one based off of CLM_CMIP_ERA as well? Or should that be independent of this? The compset will set the appropriate thing, so it could be done either way.

ekluzek: In the documentation lines I suggest always using CMIP as uppercase. Just in the code and xml and namelist options we use it as lowercase. Although I think you are using lowercase in some cases because that's the setting you use. I think it would probably be best to use uppercase, unless it's explicit in saying that it refers to the option.

ekluzek:
```suggestion
As of ctsm5.4 we do not have CMIP7 PFT raw datasets for SSPs and NOANTHRO, so default to CMIP6 for these. For other compsets default to CMIP7. User may modify to a non-default value in env_run.xml. 
Caveats:
1) We supply only CMIP7 C13/C14 isotope datasets, so these get used regardless.
2) We supply only CMIP7 population density with ctsm6 in non-SSP cases, because the fire model is calibrated to that; conversely, for pre-ctsm6 and for SSP we supply only CMIP6 population density.
3) We supply only CMIP6 nitrogen deposition (ndep), so this gets used regardless.
4) For DATM we supply only CMIP6 aerosols.
5) For DATM we supply only CMIP6 CO2</desc>
```

bld/CLMBuildNamelist.pm
slevis-lmwg: @ekluzek I hadn't seen the changes in this post until now...

Since
- as written this is a single line of new code
- and testing and new baselines are done

I'm adding a TODO comment with your suggestion, and I will open issue to keep it on our radar (#3547)

bld/unit_testers/build-namelist_test.pl
slevis-lmwg: Yes, adding.

cime_config/config_component.xml
slevis-lmwg: From meeting with @ekluzek 
we decided to leave the CO2 up to DATM to determine and to not add a communication requirement between DATM and CLM where we have to pass CLM_CMIP_ERA.

bld/CLMBuildNamelist.pm
slevis-lmwg: I checked off the boxes above not as "done" but as having moved to issue #3547.

slevis-lmwg:
My testing is up-to-date:
- aux_clm in tests_1001-161326de and tests_1001-163228iz (baselines: alpha-ctsm5.4.CMIP7.12.ctsm5.3.075)
- ctsm_sci in tests_0930-102805de (baseline: ctsm_sci-alpha-ctsm5.4.CMIP7.12.ctsm5.3.075)
- ./build-namelist_test.pl in `/glade/derecho/scratch/slevis/temp_work/ctsm5.4/mk_ctsm54_datasets/bld/unit_testers/test.out`

PR #3506, opened 2025-09-24
Internal subroutines for decompInit_lnd and some decompInit timer updates

ekluzek:
### Description of changes

Spin off some internal subroutines in decompInit to organize/modularize the code better, as well as to make it easier to test.

### Specific notes

Testing performed, if any: will do aux_clm


src/main/decompInitMod.F90
```diff
@@ -73,30 +73,18 @@ subroutine decompInit_lnd(lni, lnj, amask)
...
-          write(iulog,*) 'decompInit_lnd(): Number of gridcell clumps= ',nclumps, &
-               ' is less than the number of processes = ', npes
-          call endrun(msg=errMsg(sourcefile, __LINE__))
-       end if
-    else
-       write(iulog,*)'clump_pproc= ',clump_pproc,'  must be greater than 0'
-       call endrun(msg=errMsg(sourcefile, __LINE__))
-    end if
+    ! Do some error checking and also set nclumps and numg
+    call decompInit_lnd_check_errors( ier )
```
ekluzek, line 80: I should either change the name of this to include the fact that it also sets nclumps and numg. Or have a separate routine for doing that. I think the later, but want some feedback on this.

src/main/decompInitMod.F90
```diff
@@ -306,6 +256,164 @@ subroutine decompInit_lnd(lni, lnj, amask)
     end if
     call shr_sys_flush(iulog)
 
+  !------------------------------------------------------------------------------
+  ! Internal subroutines for this subroutine
```
ekluzek, line 261: The purpose of using internal subroutines is so that all of the declarations at the top of decompInit_lnd are assumed in the internal subroutines below. This makes it an easier transition to having these calls here. It also makes them private and only usable within decompInit_lnd.

ekluzek:
This is ready to review. I'm running standard testing to validate, but don't expect further changes. This is at a point it would be good to go through with someone else to approve.

src/main/clm_initializeMod.F90
```diff
@@ -369,20 +367,20 @@ subroutine initialize2(ni,nj, currtime)
...
     call InitDaylength(bounds_proc, declin=declin, declinm1=declinm1, obliquity=obliqr)
+    call t_stopf('clm_init2_part2')
+    call t_startf('clm_init2_part3')
 
     ! Initialize Balance checking (after time-manager)
     call BalanceCheckInit()
 
     ! History file variables
-    if (use_cn) then
+    if (use_cn .and. .not. use_noio ) then
```
ekluzek, line 383: There's a couple things here around use_noio which is a small addition to the other things going on here. It just moves more things regarding output under it, so that they aren't done when you are testing with use_noio.

I think it's small enough to include this little bit here.

ekluzek: Meeting with @slevis-lmwg we thought this was ok, so resolving.

src/main/decompInitMod.F90
ekluzek: Meeting with @slevis-lmwg we agreed that I should move the set part into it's own subroutine called before the check errors. This will make the code more clear.

- [x] Move set of nclumps/numg to own subroutine

ekluzek: 614dc1438ccaeba67f89ecd4ae4b42df0c7cf919 handles this.

ekluzek:
OK, I made the update that was part of the review, and testing on aux_clm on both Izumi and Derecho are as expected. So merging in.

ekluzek:
Crap, so this was accidentally put on master rather than b4b-dev. So I'll fix that.

Issue #3494, opened 2025-09-19
ERI_Ld90 tests fail with mosart.h0a having differences


mvdebolskiy:
### Brief summary of bug

Test fails in COMPARE_base_hybrid

### General bug information

**CTSM version you are using:** ctms5.3.075

**Does this bug cause significantly incorrect results in the model's science?** [Yes / No]

No

### Details of bug

[Fill in details here.]

### Important details of your setup / configuration so we can reproduce the bug

run f.e. ERI_Ld90.ne16pg3_ne16pg3_mg17.I2000Clm60Fates.derecho_gnu.clm-FatesCold

### Important output or errors that show the problem

TestStatus gives:

```
FAIL ERI_Ld90.ne16pg3_ne16pg3_mg17.I2000Clm60Fates.derecho_gnu.clm-FatesCold COMPARE_base_hybrid
```

TestStatus.log gives:
```
ERI_Ld90.ne16pg3_ne16pg3_mg17.I2000Clm60Fates.derecho_gnu.clm-FatesCold.ne16icomp.mosart.h0a.2002-02.nc.base did NOT match ERI_Ld90.ne16pg3_ne16pg3_mg17.I2000Clm60Fates.derecho_gnu
.clm-FatesCold.ne16icomp.mosart.h0a.2002-02.nc.hybrid
```
and:
```
tail -n20 /glade/derecho/scratch/mvdebolskiy/ERI_Ld90.ne16pg3_ne16pg3_mg17.I2000Clm60Fates.derecho_gnu.clm-FatesCold.ne16icomp/run/ERI_Ld90.ne16pg3_ne16pg3_mg17.I2000Clm60Fates.derecho_gnu.clm-FatesCold.ne16icomp.mosart.h0a.2002-02.nc.base.cprnc.out

              259200  (   258,   179,     1) (   421,   243,     1)
          avg abs field values:    3.207239101982755E+00    rms diff: 3.3E+01   avg rel diff(npos):  3.9E-02
                                   3.104853528468416E+00                        avg decimal digits(ndif):  1.2 worst: -0.3
 RMS TOTAL_DISCHARGE_TO_OCEAN_LIQ     3.3383E+01            NORMALIZED  1.0578E+01

************************************************************************************************************************************
  
SUMMARY of cprnc:
 A total number of     27 fields were compared
          of which      8 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
               and      0 had different data types
[...]

PR #3516, opened 2025-10-02
Add a failing test as a reminder for mosart history issue

mvdebolskiy:
### Description of changes

Add a longer ERI test to show that any ERI test with MOSART will fail if their length is not years or longer than a month

### Specific notes

No

Testing performed, if any:

All tests in #3494  plus:
`ERI_Ld41.f10_f10_mg37.I2000Clm60BgcCrop.derecho_intel.clm-default `
Will result in `COMPARE_base_hybrid (EXPECTED FAILURE)`

### DIFF
--- a/cime_config/testdefs/ExpectedTestFails.xml
+++ b/cime_config/testdefs/ExpectedTestFails.xml
@@ -403,5 +403,10 @@
       <issue>#3383</issue>
     </phase>
   </test>
-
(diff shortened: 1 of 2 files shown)


samsrabin (changes requested):
Thanks @mvdebolskiy! One request: Could you have this test use the gnu compiler instead? The intel test list is a bit overloaded.

Issue #3509, opened 2025-09-26
Internal subroutines for decompInit_lnd


ekluzek:
Add some internal subroutines for decompInit_lnd to make it more modular, easier to follow, and add testing possibilities. It also helps to make it easier to inspect allocates and deallocates to see that they match.

Base this off the design notes in #3456

The list of subroutines is:
- decompInit_lnd_set_nclumps_numg_lns
- decompInit_lnd_check_error
- decompInit_lnd_allocate
- decompInit_lnd_gindex_global_allocate
- decompInit_lnd_cleans

There is an order dependence and they do  need to be called in that order.

Issue #3508, opened 2025-09-26
Add decompMod_clean method


ekluzek:
Add a method to clean memory and set pointers to null in decompMod. This helps facilitate testing and ensure memory is handled.

Issue #3387, opened 2025-08-05
Bring in more timers for initialization


ekluzek:
In order to assess scalability issues in initialization we need more timers especially for clm_init2 and in the lnd_set_domain code. Put timers for whole  subroutines around the calls, and have internal timers for sections of a subroutine be inclusive of the entire subroutine.

Also remove some of the timers that are around inconsequential parts of the code (time is so low that it can be ignored).

ekluzek:
I've started doing this in PR #3335 on the decomp init study branch. I'll cherry pick the commits and send them to b4b-dev.

One thing I learned from John Dennis and from doing this work is that the timers should be constructed so the sub timers inside of a subroutine should add up to the timer around the entire call. You sometimes think you can guess which parts are expensive and only have timers around those sections -- but in practice you'll guess wrong or things change and you'll need more timers that show all parts of the code beneath the higher level timer.

Issue #3448, opened 2025-08-25
Improve design and error checking in decompMod and decompInitMod


ekluzek:
It turns out there are some design issues and lack of error checking decompMod, and decompInitMod that lead to problems like in #3447. This is where a segmentation fault happens in unrelated places in the code that aren't obvious to track down, because of this. This is an example of brittle code that's hard to work with, that also doesn't have sufficient testing around it, so is harder to work with and takes more time to do so.

Design issues I'm seeing now:
- [x] decompInitMod allocates and sets module data from decompMod, rather than having decompMod do this.
- [x] decompMod doesn't have clean allocate/deallocate methods
- [ ] decompMod has many functions that don't do error checking that the structures are even setup
- [ ] decompMod should initialize values before they are setup such that errors will be triggered if used before set
- [ ] decompMod doesn't do error checking for valid values from the decomp setup
- [ ] The subroutines in decompInit are too long, without clear purposes
- [x] get_proc_bounds is initially used to only return begg/endg, and later for everything, this usage isn't clear from the code...
- [x] The code in decompInit is ordered, without it being easy/ovious to change the order things are done
- [ ] Code is duplicated between get_proc_bounds and get_proc_clumps (maybe OK now, but with additional error checking, would be good to have the duplicated code in a private subroutine shared by both)
- [x] Pointers aren't set to null() initially so that you can check they are valid before using them

Error checking:
- [x] Error checking of namelist items should be done as close to the top as possible
- [x] Error checking that nsegspc is valid should be done as it can be set by the user
- [ ] Should check that the decomp has been setup, before you can use the functions in decompMod
- [ ] Appropriate range checking of inputs, and the internal arrays and dimension sizes should be done in the decompMod functions

Testing:
- [x] Adding self tests would be useful to ensure changes are correct and stay correct with development
- [x] Adding a serial unit tester for the most basic things and error checking testing would be helpful
- [x] Should add a standard case test to run to the decomp_init test, to make sure a normal test without the special testing framework still works

PR #3517, opened 2025-10-02
Merge b4bdev 20251002

slevis-lmwg:
### Description of changes
New work in b4b-dev:
PR #3516 @mvdebolskiy 
PR #3511 @ekluzek 

### Specific notes

Makes visible #3494 
Fixes https://github.com/ESCOMP/CTSM/issues/3387
Some work in https://github.com/ESCOMP/CTSM/issues/3448
Fixes https://github.com/ESCOMP/CTSM/issues/3508
Fixes https://github.com/ESCOMP/CTSM/issues/3509

No

New test added; added same test to expected failures

No

Testing performed, if any:
- Starting aux_clm
- OK ./build-namelist_test.pl

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.3.078
 Originator(s): slevis (Samuel Levis,UCAR/TSS,303-665-1310)
-Date: Thu 02 Oct 2025 05:19:57 PM MDT
+Date: Fri 03 Oct 2025 09:31:06 AM MDT
 One-line Summary: Merge b4b-dev to master
 
 Purpose and description of changes
@@ -30,7 +30,7 @@ Does this tag change answers significantly for any of the following physics conf
 Bugs fixed
 ----------
 List of CTSM issues fixed (include CTSM Issue # and description) [one per line]:
-Fixes #3494 ERI_Ld90 tests fail with mosart.h0a having diffs
+Makes visible #3494 ERI_Ld90 tests fail with mosart.h0a having diffs
 Fixes #3387 Bring in more timers for initialization
 Some work in #3448 Improve design and error checking in decompMod and decomInitMod
 Fixes #3508 Add decompMod_clean method
--- a/doc/ChangeSum
+++ b/doc/ChangeSum
@@ -1,6 +1,6 @@
 Tag                   Who      Date  Summary
 ============================================================================================================================
-       ctsm5.3.078   slevis 10/02/2025 Merge b4b-dev to master
+       ctsm5.3.078   slevis 10/03/2025 Merge b4b-dev to master
        ctsm5.3.077   rgknox 10/02/2025 Adjustment of timestep albedo filtering with FATES
        ctsm5.3.076     erik 09/24/2025 Merge b4b-dev to master
        ctsm5.3.075   slevis 09/10/2025 Change default glcmec_downscale_longwave from true to false for clm6


slevis-lmwg:
aux_clm
- [x] izumi
- [x] derecho

slevis-lmwg:
The test-suites are in progress, so I asked @mvdebolskiy to review/approve, with the idea that he may do so before I return to this on Friday. But it's also fine if anyone else wishes to review/approve.

src/main/decompInitMod.F90
```diff
@@ -306,6 +257,174 @@ subroutine decompInit_lnd(lni, lnj, amask)
...
+
+         !--- set and verify nclumps ---
+         if (clump_pproc > 0) then
+            if (nclumps < npes) then
+               ier = 1
+               write(iulog,*) 'Number of gridcell clumps= ',nclumps, &
+                     ' is less than the number of processes = ', npes
+               call endrun(msg="Number of clumps exceeds number of processes", &
+                           file=sourcefile, line=__LINE__)
+               return
```
mvdebolskiy, line 400: @ekluzek I am confused about these two messages. The latter one is says nclumps exceeds npes, but the condition says otherwise. Also, can it be PEs instead of processes? Does the rest of CESM use processes to describe threads*tasks?

doc/ChangeLog
```diff
@@ -1,4 +1,71 @@
...
+
+[ ] ctsm5_0-nwp
+
+[ ] clm4_5
+
+
+Bugs fixed
+----------
+List of CTSM issues fixed (include CTSM Issue # and description) [one per line]:
+Fixes #3494 ERI_Ld90 tests fail with mosart.h0a having diffs
```
mvdebolskiy:
```suggestion
```
That PR does not fix the issue, It just makes it visible during testing.

mvdebolskiy (changes requested):

doc/ChangeLog
slevis-lmwg: Right, good point!

slevis-lmwg: Thank you @mvdebolskiy!

src/main/decompInitMod.F90
mvdebolskiy: I see, that processes are used throughout the error messages, so just the error message should be only the one that is in the write statement above the endrun.

slevis-lmwg: To prevent me from having to repeat testing, @ekluzek will address this concern in #3412

slevis-lmwg: Also note that these lines of code were preexisting, even if they appear changed here, and even though @mvdebolskiy's point is valid.

slevis-lmwg:
@mvdebolskiy are you able to approve this now?

mvdebolskiy (approved):

slevis-lmwg:
Thank you @mvdebolskiy :-)

src/main/decompInitMod.F90
ekluzek: @mvdebolskiy thanks so much for pointing this out! Yes, the description of the error reported is wrong (and has been wrong for years before this). I'll point you to the other PR where I'll fix this.

src/main/decompInitMod.F90
```diff
@@ -73,30 +73,19 @@ subroutine decompInit_lnd(lni, lnj, amask)
...
-    lns = lni * lnj
+    ! Do some error checking
+    call decompInit_lnd_check_errors( ier )
+    if (ier /= 0) return
 
-    !--- set and verify nclumps ---
-    if (clump_pproc > 0) then
-       nclumps = clump_pproc * npes
-       if (nclumps < npes) then
-          write(iulog,*) 'decompInit_lnd(): Number of gridcell clumps= ',nclumps, &
```
ekluzek, line 83: @mvdebolskiy this is the original error message, which looks like it was correct and I messed it up below. This is actually something where I couldn't add a unit test for it in #3412 because I couldn't think of a way to test for it. Which on second thought means that this is an error condition to remove.

I'll create an issue to talk about this.

slevis-lmwg: @ekluzek I happened upon this post and wondered whether the issue should still be created or the post could be resolved.

ekluzek: @slevis-lmwg OK, I looked into this, in the latest code. The error message is now in 

decompInit_lnd_check_errors

as follows...

``` fortran
            if (nclumps < npes) then
               ier = 1
               write(iulog,*) 'Number of gridcell clumps= ',nclumps, &
                     ' is less than the number of processes = ', npes
               call endrun(msg="Number of clumps exceeds number of processes", &
                           file=sourcefile, line=__LINE__)
               return
            end if
```

and the endrun message contradicts the message in the write. So it should be fixed or removed.

slevis-lmwg: Thanks @ekluzek I see the issue that you opened, so I'm resolving this post.

Issue #3542, opened 2025-10-10
generate_gdd20_baseline needs SystemTest


samsrabin:
It turns out `generate_gdd20_baseline` has been broken for a while, and I didn't notice because it wasn't being tested. Ideally it would have complete Python unit and system testing, but at the very least we should have a SystemTest in the `ctsm_sci` and `crop_calendars` suites like `RXCROPMATURITY`:
1. Do a multi-year run to generate outputs.
2. Call `generate_gdd20_baseline` on the outputs, generating new `stream_fldFileName_gdd20_baseline`.
3. Run at least a few days with the new `stream_fldFileName_gdd20_baseline`.

wwieder:
Is there an issue with how gdd20 is currently being handled in the code (and in #3543), or is this just a testing issue to prevent future issues?

samsrabin:
The latter: gdd20 in the FORTRAN code is fine as far as I know; #3543 is just to fix the Python script.

wwieder:
requires 2 years of simulation, could be low resolution (f10)

PR #3543, opened 2025-10-10
Initial fixes to generate_gdd20_baseline

samsrabin:
### Description of changes

Some Python changes needed to get the `generate_gdd20_baseline` script working, along with some tests to prevent those exact problems from happening again. We should also add a SystemTest, but I've filed that as a separate issue: #3542.

Note that more changes are coming. This PR is the first of several that will bring this script and related scripts up to date with the versions I used during the crop reparameterization, which have a complete set of fixes.

**Note to reviewer**: It'd probably be best to walk through the commits one-by-one.

### Specific notes

- [x] Python unit and system tests (including new unit tests)

I've listed expectations for these suites, but if any fail, I will add them to the expected failure list. This is acceptable because I have more PRs to come in this series that will eventually fix the tests, if needed.
- [x] `clm_pymods` suite completes successfully (diffs okay in tests exercising affected scripts **but none occurred**). **No unexpected failures.**
- [x] `rxcropmaturity` suite completes successfully (no baseline for comparison). **No unexpected failures.**


slevis-lmwg (approved):
@samsrabin thank you for going over this work with me.

PR #3548, opened 2025-10-14
Finishing adding RC14_CANAIR history output

### LINKED ISSUE #3488: Add RC14_CANAIR as optional to history output so that C14 time series data can be verified
To validate the new C13/C14 ratio timeseries data is correct, we need to output RC14_CANAIR similar to RC13_CANAIR  so we can ensure the data being used by the model is correct. When we did this before, we must have just assumed the code was correct by inspection. On examining more closely it looks like there are problems with both C13 and C14 data.

### LINKED ISSUE #3490: Change history averaging of RC13_CANAIR and RC14_CANAIR so that zero's aren't averaged in
When you plot RC13_CANAIR (and the same will be true when RC14_CANAIR is added) each point varies because points that are set to zero (special landunits as well as non-exposed vegetation) end up being averaged into the gridcell average. Since, the primary purpose of these variables is to make sure the input file is read correctly, that means it's impossible to use the variable to ensure it's right.

ekluzek:
### Description of changes

Bring in addition of RC14_CANAIR on history so that Carbon isotope data reading can be validated.

### Specific notes

  Fixes #3490 
  Fixes #3488

Any User Interface Changes (namelist or namelist defaults changes)? Add RC14_CANAIR as an output history option

Testing performed, if any: So far ran SMS_Ly3.f10_f10_mt232.IHistClm60BgcCropCrujra.derecho_intel.clm-monthly--clm-clm60cam7LndTuningMode_2013Start
   Will run complete test suite

### DIFF
--- a/src/biogeophys/PhotosynthesisMod.F90
+++ b/src/biogeophys/PhotosynthesisMod.F90
@@ -583,7 +583,7 @@ contains
        this%rc13_canair_patch(begp:endp) = spval
        call hist_addfld1d (fname='RC13_CANAIR', units='proportion', &
             avgflag='A', long_name='C13/C(12+13) for canopy air', &
-            ptr_patch=this%rc13_canair_patch, default='inactive')
+            ptr_patch=this%rc13_canair_patch, set_spec=spval, default='inactive')
 
        this%rc13_psnsun_patch(begp:endp) = spval
        call hist_addfld1d (fname='RC13_PSNSUN', units='proportion', &
@@ -600,7 +600,7 @@ contains
        this%rc14_canair_patch(begp:endp) = spval
        call hist_addfld1d (fname='RC14_CANAIR', units='proportion', &
             avgflag='A', long_name='C14/C(12+13) for canopy air', &
-            ptr_patch=this%rc14_canair_patch, default='inactive')
+            ptr_patch=this%rc14_canair_patch, set_spec=spval, default='inactive')
     end if
 
     ! Canopy physiology
@@ -1189,12 +1189,12 @@ contains
             .or. lun%itype(l) == istice &
             .or. lun%itype(l) == istwet) then
           if (use_c13) then
-             this%rc13_canair_patch(p) = 0._r8
+             this%rc13_canair_patch(p) = spval
              this%rc13_psnsun_patch(p) = 0._r8
              this%rc13_psnsha_patch(p) = 0._r8
           end if
           if (use_c14) then
-             this%rc14_canair_patch(p) = 0._r8
+             this%rc14_canair_patch(p) = spval
           end if
        end if
     end do
@@ -1216,13 +1216,13 @@ contains
     if ( use_c13 ) then
        this%alphapsnsun_patch(p) = 0._r8
        this%alphapsnsha_patch(p) = 0._r8
-       this%rc13_canair_patch(p) = 0._r8
+       this%rc13_canair_patch(p) = spval
        this%rc13_psnsun_patch(p) = 0._r8
        this%rc13_psnsha_patch(p) = 0._r8
     endif
 
     if ( use_c14 ) then
-       this%rc14_canair_patch(p) = 0._r8
+       this%rc14_canair_patch(p) = spval
     end if
 
     this%psnsun_patch(p) = 0._r8


src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -2142,6 +2166,8 @@ subroutine PhotosynthesisTotal (fn, filterp, &
                   sector_c14 = 3
                endif
 
+               rc14_canair(p) = rc14_atm(sector_c14)
+
```
slevis-lmwg, line 2170: Do we expect to still need the sectors once this comes in as a stream?

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -1197,11 +1216,15 @@ subroutine NewPatchInit (this, p)
...
        this%alphapsnsun_patch(p) = 0._r8
        this%alphapsnsha_patch(p) = 0._r8
-       this%rc13_canair_patch(p) = 0._r8
+       this%rc13_canair_patch(p) = spval
        this%rc13_psnsun_patch(p) = 0._r8
        this%rc13_psnsha_patch(p) = 0._r8
     endif
 
+    if ( use_c14 ) then
+       this%rc14_canair_patch(p) = spval
```
slevis-lmwg, line 1225: Here and in subroutine TimeStepInit:
Do you know if we need rc14 equivalents of the _psnsun _psnsha variables that I see in the c13 section above?

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -579,7 +583,7 @@ subroutine InitHistory(this, bounds)
        this%rc13_canair_patch(begp:endp) = spval
        call hist_addfld1d (fname='RC13_CANAIR', units='proportion', &
             avgflag='A', long_name='C13/C(12+13) for canopy air', &
-            ptr_patch=this%rc13_canair_patch, default='inactive')
+            ptr_patch=this%rc13_canair_patch, set_spec=spval, default='inactive')
```
slevis-lmwg, line 586: Nit: Organize all c13 calls under one if-statement for computational efficiency?

doc/source/tech_note/External_Nitrogen_Cycle/CLM50_Tech_Note_External_Nitrogen_Cycle.rst
```diff
@@ -133,6 +133,45 @@ where :math:`{WS}_{tot\_soil}` (kgH\ :sub:`2`\ O m\ :sup:`-2`) is the total mass
 
    NF_{leached} =\min \left(NF_{leached} ,\frac{NS_{sminn} sf}{\Delta t} \right).
 
+Alternative way of evaluating the Leaching Losses of Nitrogen
```
slevis-lmwg, line 136: The .rst changes seem tangential; however, it's great that they are coming in!

src/biogeophys/PhotosynthesisMod.F90
ekluzek: Yes, all of the sector stuff goes away with streams.

doc/source/tech_note/External_Nitrogen_Cycle/CLM50_Tech_Note_External_Nitrogen_Cycle.rst
ekluzek: Hmmm. That's strange. This shouldn't be part of this one. I'll need to figure that out.

src/biogeophys/PhotosynthesisMod.F90
ekluzek: It probably matters more for readability. But, each if does add the check again. So this is a good point.

ekluzek: Since, I ran testing I think I'll keep this in mind and do later since there's more work coming. But, this is a good thing to do.

src/biogeophys/PhotosynthesisMod.F90
ekluzek: Thats a good question. From looking at the code these are diagnostic history variables that are just there to help show what's going on. There could be c14 versions of these, but it seems like no one thought it important enough to add. There is the "c14_psnsun_patch" version (without the starting "r") that actually does the science.

So I think the answer is no, unless we do have a scientist working in this space that tells us they need those for c14 as well.

doc/source/tech_note/External_Nitrogen_Cycle/CLM50_Tech_Note_External_Nitrogen_Cycle.rst
ekluzek: Ahh, the problem was that this was setup to go to master rather than b4b-dev. So I corrected that and now that change goes away.

ekluzek:
@slevis-lmwg it looks like you didn't hit the approval part. I could override, but I'll wait for you to hit approval.

src/biogeophys/PhotosynthesisMod.F90
slevis-lmwg: Ah, but that will happen in a follow-up PR, not in this one.

Issue #3546, opened 2025-10-13
Design notes around Carbon Isotopes read in as streams


ekluzek:
A few design notes on how to bring Carbon Isotopes in the code as streams.

One change is to store the C13/C14 on the model grid (this is an increase in memory, but it scales with processors).

A common pattern for stream handling is to have: Init, Clean, Advance, and Interp methods, so we use that here. 

New namelist data:
- stream_fldfilename_atm_c14
- stream_year_first_atm_c14
- stream_year_last_atm_c14

- stream_fldfilename_atm_c13
- stream_year_first_atm_c13
- stream_year_last_atm_c13

Hardcoded options:
- atm_c14_mapalgo (hardcode this to nn for now)
- atm_c14_tintalgo (hardcode this to linear for now)
- atm_c14_taxmode (hardcode this to extend for now)
- atm_c13_mapalgo  (hardcode this to nn for now)
- atm_c13_tintalgo (hardcode this to linear for now)
- atm_c13_taxmode (hardcode this to extend for now)
- stream_offset --- hardcode to 0 for now

New files:

- src/cpl/share_esmf/CTSMForcStreamBase2DType.F90
- Description: Base type for streams that can be extended for specific uses
- Data:
    - Stream data
- Methods:
    - InittBase
    - Init (deferred)
    - Clean (deferred)
    - CleanBase
    - Advance
    - Interp (deferred)
    - CheckUnits -- Move the check_units subroutine from CNCIsoAtmTimeSeriesReadMod to here (check both units and info, remove time check)

- unit_test_stubs/share_esmf/CTSMForcStreamBase2DType.F90
- Description: Stub version for unit tests

- src/biogeochem/AtmCarbonIsotopeStreamType.F90 (extends CTSMForcStreamBase2DType.F90 for atm_delta_c13_stream_type/atm_delta_c14_stream_type)
- Description: Carbon isotope stream data for C13 and C14
- Data:
   - with atm_delta_c13_stream_type, atm_delta_c14_stream_type extending the base type
    - Have the data in here have the same units as the dataset, in the deltac13co2_in_air and Delta14cos_in_air units
- Methods: (both delta_c13_stream_type and delta_c14_stream_type
    - Interp
    - Init (extended)
    - InitAllocate (private local)
    - C1?Clean

Changes to these files:

- src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
- Data:
   - Remove: *_time, time_axis_offset, nsectors_c14 data
[...]

PR #3561, opened 2025-10-21
Start bringing in carbon isotopes in streams

### LINKED ISSUE #3502: Time interpolation for Carbon isotope datasets is done incorrectly
### Brief summary of bug

The time interpolation of the C13 and C14 isotope datasets is supposed to be linear between each year. It looks like in practice it mostly uses the yearly values, with some smoothing between years.

### General bug information

**CTSM version you are using:** branch_tags/CMIP7_ciso.n01.alpha-ctsm5.4.CMIP7.11.ctsm5.3.075

**Configurations affected:** when use_c13 or use_c14 is on with transient datasets

Start with this test case:

`SMS_Ly3.f10_f10_mt232.IHistClm60BgcCropCrujra.derecho_intel.clm-monthly--clm-clm60cam7LndTuningMode_2013Start`

Edit the user_nl_clm file and add this line:

``` fortran
 hist_fincl1   += 'FCO2', 'RC13_CANAIR', 'RC14_CANAIR'
```

And make sure output is going to be monthly

Run the case then do the following to look at the output:

``` shell
# Concatenate the monthly files together in one file
ncrcat -O -v RC13_CANAIR,RC14_CANAIR $RUNDIR/SMS*.clm2.h0a*.nc allmonths.h0a.nc
# Use ncap2 to convert RC13_CANAIR and RC14_CANAIR to Delta format
ncap2 -s 'c14ratio = 1.e-12' -s 'PDB = 0.0112372' -s 'preind_atm_del13c = -6.' -s 'preind_atm_ratio = PDB + ( preind_atm_del13c * PDB ) /1000.0' -s 'c13ratio = preind_atm_ratio/ ( 1.0+preind_atm_ratio ) ' -s 'Deltac13 = ( RC13_CANAIR/PDB - 1.0 ) /1.0e-3' -s 'Deltac14 = ( RC14_CANAIR/c14ratio - 1.0 ) /1.0e-3' -O allmonths.h0a.nc test.nc
```

The output Deltac13 and Deltac14 will be close to a step function with some smoothing between years.

ekluzek:
### Description of changes

Bring in the capability to read in Carbon isotopes in streams. There will be other changes to fully flesh this out. And a version that allows it to be turned on instead of the current file method. The final thing will be a tag on master to change the default to only using streams. And then on b4b-dev we can remove the old version.

### Specific notes

  Design notes in #3546
  Fixes #3502
  Work on #3346

   New namelist options for stream handling will come in:

- stream_fldfilename_atm_c14
- stream_meshfile_atm_c14
- stream_year_first_atm_c14
- stream_year_last_atm_c14
- stream_fldfilename_atm_c13
- stream_year_first_atm_c13
- stream_year_last_atm_c13

Testing performed, if any: Single case testing so far, will run aux_clm
  Currently just running an SMS test: SMS_D_Ld3.f10_f10_mt232.IHistClm60BgcCropCrujra.derecho_intel.clm-clm60cam7LndTuningMode
  Will transition to the 3-year test starting in Dec/2013 used previously


ekluzek:
The read is working now, so it needs more testing to ensure it's correct as well as review and some further updates we'll figure out in the reveiw.

src/cpl/share_esmf/CTSMForce2DStreamBaseType.F90
```diff
@@ -0,0 +1,178 @@
+module CTSMForce2DStreamBaseType
+
```
ekluzek, line 2: @slevis-lmwg and I went over the basic name and it makes sense to both of us.

I will need to add comments and descriptions, now that the code is working and we think the names are OK.

src/cpl/share_esmf/CTSMForce2DStreamBaseType.F90
```diff
@@ -0,0 +1,178 @@
...
+  use decompMod , only : bounds_type
+  use clm_varctl, only : FL => fname_len
+
+  implicit none
+  private
+
+  type, abstract, public :: ctsm_force_2DStream_base_type
+     private
+     type(shr_strdata_type), public :: sdat  ! Stream data type
+     character(len=FL) :: stream_filename ! The stream data filename (also in sdat)
```
ekluzek, line 35: Do I really need these here? Can I use the data in sdat?

src/biogeochem/AtmCarbonIsotopeStreamType.F90
```diff
@@ -0,0 +1,193 @@
...
+    subroutine C13Interp( this, bounds )
+         class(atm_delta_c13_stream_type), intent(inout) :: this 
+         type(bounds_type), intent(in) :: bounds
+
+         ! Local Variables
+         integer :: g
+         real(r8), pointer :: dataptr1d(:)
+         integer :: rc ! error return code
+
+         ! Get pointer for stream data that is time and spatially interpolated to model time and grid
```
ekluzek, line 144: This bit about getting the pointer to data could be put in the base type.

src/biogeochem/AtmCarbonIsotopeStreamType.F90
```diff
@@ -0,0 +1,193 @@
+module AtmCarbonIsotopeStreamType
+  use shr_kind_mod , only : r8 => shr_kind_r8
```
ekluzek, line 8: Need to add notes and descriptions and general comments, now that the general structure is approved.

src/biogeochem/AtmCarbonIsotopeStreamType.F90
```diff
@@ -0,0 +1,193 @@
+module AtmCarbonIsotopeStreamType
```
ekluzek, line 1: This puts both C13 and C14 in the same file -- it could be separated into two files.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
```diff
@@ -33,13 +35,33 @@ module CIsoAtmTimeseriesMod
...
+  type(atm_delta_c14_stream_type), private :: atm_c14_stream  ! Atmospheric C14 stream object
+
   ! !PRIVATE TYPES:
   real(r8), allocatable, private :: atm_c14file_time(:)       ! time for C14 data
   real(r8), allocatable, private :: atm_delta_c14(:,:)        ! Delta C14 data
   real(r8), allocatable, private :: atm_c13file_time(:)       ! time for C13 data
   real(r8), allocatable, private :: atm_delta_c13(:)          ! Delta C13 data
   real(r8), parameter :: time_axis_offset = 1850.0_r8         ! Offset in years of time on file
 
+  ! Private data for the control namelist:
```
ekluzek, line 64: This needs to become an actual namelist and added to namelist defaults all the things.

ekluzek: At least the filenames need to be done this way.

src/biogeochem/AtmCarbonIsotopeStreamType.F90
ekluzek: We decided one file makes sense here.

src/biogeochem/AtmCarbonIsotopeStreamType.F90
ekluzek: I figured out what to do here in #3566 the data can be directly accessed from here.

bld/CLMBuildNamelist.pm
```diff
@@ -3753,48 +3757,92 @@ sub setup_logic_c_isotope {
...
+}
+
+#-------------------------------------------------------------------------------
+
+sub setup_logic_c14_streams {
+  my ($opts, $nl_flags, $definition, $defaults, $nl) = @_;
+  #
+  # C14 stream file settings
+  #
+}
```
slevis-lmwg, line 3844: Currently these don't do much. Do you plan to move code from `sub setup_logic_c_isotope` to here later?

src/biogeochem/AtmCarbonIsotopeStreamType.F90
ekluzek: I put this in the base type which turned out to be needed to get unit testing working.

bld/CLMBuildNamelist.pm
ekluzek: This will grow as there becomes handling for all the streams namelist items. So it's just a starting point right now.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: This is done.

src/cpl/share_esmf/CTSMForce2DStreamBaseType.F90
ekluzek: OK, I've added comments and descriptions.

src/cpl/share_esmf/CTSMForce2DStreamBaseType.F90
ekluzek: In looking at this we'd need to go several type layers down

sdat%stream%file(1)%name

And it looks like stream_name isn't really kept after initialization.

src/biogeochem/AtmCarbonIsotopeStreamType.F90
ekluzek: I did this now.

ekluzek:
Reviewing this for post Sprint 25. It looks like time for this is divided up as follows:

- Few days for putting the basic structure in place
- Few days for figuring out the options and flow and test that it's correct
- Few days for working through miscellaneous issues
- Few days for adding the new namelist items and a new test with them

PR #3562, opened 2025-10-23
alpha-ctsm5.4.CMIP7.15.ctsm5.3.079: Updates to default_data_[1850,2000].cfg & modify_smallville.sh

slevis-lmwg:
### Description of changes
Having made CTSM5.4 fsurdat/landuse files with `make all` (all-subset excluded), we then update a few .cfg and .sh files before running `make all-subset`. Then we can also proceed with making new NEON/PLUMBER datasets.

### Specific notes

Addresses a checkbox in #2851.

Using the new fsurdat/landuse files will lead to new answers.

The .cfg files listed above are changing.

Testing performed, if any:
- `make all-subset` worked with the changes in this PR
- `neon_surf_wrapper` and `plumber2_surf_wrapper` also worked

### DIFF
--- a/tools/modify_input_files/modify_smallville.sh
+++ b/tools/modify_input_files/modify_smallville.sh
@@ -7,14 +7,14 @@ module load nco
 # This script runs from the mksurfdata_esmf/Makefile.
 # When running standalone, it may need "subset_data_single_point/" in front
 # of each landuse.timeseries file name.
- file_to_2100="landuse.timeseries_1x1_smallvilleIA_SSP2-4.5_1850-2100_78pfts_c$(date +%y%m%d).nc"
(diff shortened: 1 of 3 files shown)


ekluzek (approved):
Looks good. Just updating the datasets so that the subset data part will work.

The only change to comment on is the change of name from "file_to_2100" to "file_whole", which could be maybe something like "file_to_histendyear", but really doesn't matter enough. In the future we'll want it to be from an SSP again, so I wondered about adding a comment. But, I don't think that matters that much.

slevis-lmwg:
@ekluzek thank you for the review. I decided not to add a comment in the script. It seemed pretty obvious what I had to do, so I'm hoping it will seem equally obvious next time. Also, I decided that the name "file_whole" would also apply to the SSP file later, so we would not need to keep changing this variable name.

PR #3566, opened 2025-10-24
Have C13/C14 data stored on the model grid

### LINKED ISSUE #3565: Have C13/C14 data arrays on a begg:endg gridcell array rather than in a form that requires knowledge of the grid
The C13 timeseries data has been stored as a single global value. However, this use doesn't allow the data grid to change. The C14 for CMIIP6 was with three latitude bands, but the CMIP7 data has four., and the code is hardcoded for this handling, and can't easily adapt between the two, or a future more complex grid. It also exposes the data grid to the points where the data is used, rather than have that information isolated to the file handling part.  As a first step to moving the data to streams, this needs to change.

The upside is flexibility, readability and code maintenance. The only downside is an increase in memory, but because the memory is scalable this is acceptable.

ekluzek:
### Description of changes

Store the C13/C14 data on begg:endg gridcell arrays rather than a single value or hardcoded latitude band array. This makes the code more: flexible, readable, maintainable, similar to other CTSM code, isolates the grid from the rest of the code, and enables the switch to streams.

This is a small increase in scalable memory. Since, it's scalable this is perfectly fine.

### Specific notes

   Fixes #3565

Testing performed, if any: single case tests so far will run aux_clm
  SMS_D_Ld10.f09_f09_mt232.IHistClm60BgcCrop.derecho_intel.clm-f09_FillMissingW_Urban is identical to the ctsm5.3.080 baseline


src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
```diff
@@ -142,6 +167,12 @@ subroutine C14_init_BombSpike()
     allocate(atm_delta_c14(nsectors_c14,ntim))
     atm_delta_c14(:,:) = 0.0_r8
 
+    ! Allocate the gridcell arrays
+    allocate(atm_delta_c14_grc(bounds%begg:bounds%endg))
```
ekluzek, line 212: With allocates, there really should be a clean subroutine to do the deallocation. But, since we don't have a place it would be called, this doesn't seem helpful right now.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
```diff
@@ -256,6 +300,12 @@ subroutine C13_init_TimeSeries()
     allocate(atm_c13file_time(ntim))
     allocate(atm_delta_c13(ntim))
 
+    ! Allocate the gridcell arrays
+    allocate(atm_delta_c13_grc(bounds%begg:bounds%endg) )
```
ekluzek, line 382: Same as above on Clean method.

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -2157,19 +2157,10 @@ subroutine PhotosynthesisTotal (fn, filterp, &
             endif
             if ( use_c14 ) then
 
-               ! determine latitute sector for radiocarbon bomb spike inputs
```
ekluzek, line 2160: This is moved to go inside of the C14_Bombspike subroutine which isolates knowledge of the grid to the place where it's read.

src/main/clm_initializeMod.F90
```diff
@@ -476,10 +476,10 @@ subroutine initialize2(ni,nj, currtime)
           call SatellitePhenologyInit(bounds_proc)
        end if
        if ( use_c14 .and. use_c14_bombspike ) then
```
ekluzek: I think it would be better for the code here, to just trigger over use_c14 and have the if regarding use_c14_bombspike/use_c13_timeseries to be contained within there rather than at this level.

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -2116,15 +2116,15 @@ subroutine PhotosynthesisTotal (fn, filterp, &
 
       if ( use_c14 ) then
          if (use_c14_bombspike) then
```
ekluzek: Similar to below for initialization, it would be better for the code here, to just trigger over use_c14 and have the if regarding use_c14_bombspike/use_c13_timeseries to be contained within there rather than at this level.

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -2116,15 +2116,15 @@ subroutine PhotosynthesisTotal (fn, filterp, &
...
-            call C14BombSpike(rc14_atm)
+            call C14BombSpike(bounds)
          else
-            rc14_atm(:) = c14ratio
+            rc14_atm_grc(:) = c14ratio
```
ekluzek: When held constant this part could be within the C14Bombspike subroutine

ekluzek: Possibly this shouldn't really be in the guts of Photosynthesis and done at a high level like in VegetationFacade or clm_driver.

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -2142,7 +2142,7 @@ subroutine PhotosynthesisTotal (fn, filterp, &
          if (use_cn) then
             if ( use_c13 ) then
                if (use_c13_timeseries) then
-                  rc13_canair(p) = rc13_atm
+                  rc13_canair(p) = rc13_atm_grc(g)
                else
                   rc13_canair(p) = forc_pc13o2(g)/(forc_pco2(g) - forc_pc13o2(g))
```
ekluzek: Same here, this would be better to be setting rc13_atm_grc and have it inside the C13_TimeSeries for even the constant version.

src/biogeophys/PhotosynthesisMod.F90
ekluzek: So the code would look like this:

``` fortran
if (use_c14 ) call C14BombSpike(bounds)
.
.
.
if (use_c13) call C13TimeSeries(bounds)
```

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: Documented under #3568 so resolving.

ekluzek:
OK, I need to add a test for use_c13/c_14 without timeseries files to make sure the preind constants work. Then I just need to run aux_clm and show it's b4b to do the merge.

ekluzek:
aux_clm testing showed some problems with threaded cases, that I fixed.  So this is pretty ready to be merged now.

PR #3581, opened 2025-11-04
Generalize some paths so unit testing works in a CESM checkout

### LINKED ISSUE #2837: FUNIT tests don't work on CESM/CAM checkouts
### Brief summary of bug

The pfunit, unit test framework, currently doesn't work in CESM or CAM checkouts.

I had noticed this a while back and @billsacks noticed it recently as well.

### General bug information

**CTSM version you are using:** ctsm5.3.0 but before

**Configurations affected:**
Just FUNIT sys test and running by hand in he src directory

billsacks:
### Description of changes

Previously these paths assumed a standalone CTSM checkout

### Specific notes

  Fixes #2837

Testing performed, if any:

Ran unit testing on my Mac in:
- Standalone CTSM checkout (`../cime/scripts/fortran_unit_testing/run_tests.py --build-dir unit_tests.temp`)
- CESM checkout (`../../../cime/scripts/fortran_unit_testing/run_tests.py --build-dir unit_tests.temp`)

### DIFF
--- a/cime_config/testdefs/testlist_clm.xml
+++ b/cime_config/testdefs/testlist_clm.xml
@@ -4399,6 +4399,7 @@
   <test name="FUNITCTSM_P1x1" grid="f10_f10_mg37" compset="I2000Clm50Sp">
     <machines>
       <machine name="derecho" compiler="intel" category="aux_clm"/>
+      <machine name="derecho" compiler="intel" category="prealpha"/>
       <machine name="izumi" compiler="intel" category="aux_clm"/>
     </machines>
     <options>


ekluzek (approved):
Excellent, thanks for this update @billsacks! The code changes look correct to me. And I don't have any suggestions on changes there.

There is one thing that I'll plan to do (unless you want to do it) -- which is to add the FUNITCTSM unit test to the prealpha testlist.

- [ ] Add FUNITCTSM unit test on Derecho to the prealpha testlist @ekluzek

With Derecho down, we can't bring this in immediately, but I should be able to bring it in once it is. I'll test both in a CTSM and CESM checkout there, and make sure the FUNITCTSM system test works as well.

So the things I'll do to double check:
- [ ] Test under CTSM checkout on Derecho
- [ ] Test under CESM checkout on Derecho
- [ ] Run aux_clm just to see that the FUNITCTSM test works there (CTSM checkout)
- [ ] Run prealpha in a CESM checkout just to see that the FUNITCTSM test works

billsacks:
Adding FUNITCTSM to prealpha seems like a good idea - thanks!

billsacks:
> Adding FUNITCTSM to prealpha seems like a good idea - thanks!

@ekluzek - inspired by this idea, I'm about to add a new FUNITSHARE test to the prealpha test list (running the CESM_share unit tests). Since the CTSM unit tests are run for each CTSM tag, and the share unit tests are run via a GitHub action for each share PR, I see the point of this prealpha test being less about actually testing the code and more about making sure the unit testing infrastructure is continuing to work on derecho. Therefore, it seems like having the share unit tests in prealpha may be sufficient, but I also don't object to your adding FUNITCTSM as well if you want to, since these tests are quick to run.

ekluzek:
Working so merging in.

PR #3582, opened 2025-11-04
Use a CMake function provided by CIME

billsacks:
### Description of changes

Use a new CMake function provided by CIME in place of some messy CMake code.

Requires cime6.1.137

### Specific notes

Testing performed, if any:
- Ran unit tests on my Mac


src/CMakeLists.txt
```diff
@@ -63,29 +63,16 @@ add_subdirectory(${CLM_ROOT}/src/self_tests clm_self_tests)
...
-  # on PIO. This removal is needed on some systems but not on others: the unit test build
-  # works without this removal on a Mac with a pre-built PIO library, but failed (with
-  # error message, "Cannot open module file 'pio.mod'") on a Mac without a pre-built PIO
-  # (where ESMF was built with its internal PIO).
-  string(REGEX MATCH "shr_pio_mod.F90" match_found ${sourcefile})
-  if(match_found)
-    list(REMOVE_ITEM share_sources ${sourcefile})
-  endif()
-endforeach()
+# Remove shr_mpi_mod from share_sources. This is needed because we want to use the mock
```
ekluzek, line 81: I like using the new CMake function to make the code better.

I'm wondering how much we need this anymore. From the comments in the file in order to work on various systems (i.e. Mac's) it looks like we need to eliminate shr_pio_mod.F90, so that probably needs to stick around. But, you have a cime update that means we could use the standard shr_mpi_mod.F90 right (cime6.1.137)?

billsacks: Good question. There's probably a way to get this working with an mpi library, but I think it would take some work in terms of getting the build set up right to include the mpi dependencies.

Even though I partially enabled use of a real mpi library in cime6.1.137, I don't necessarily recommend using that with the unit tests: it was a solution to a problem I encountered in the CESM_share GitHub Actions, but I think it's still simpler to use mpiserial – and for some reason, I can't get one of the share unit tests working with a real mpi library on my Mac (though it works on derecho and the GitHub runner).

But actually, the changes in cime6.1.137 aren't necessarily a pre-requisite for changing this behavior: you could do what's done in the share code and typically use an mpi-serial library (which would be assumed to be pre-built on the system). Again, though, I don't necessarily recommend this because it complicates the build. I'd only go there if for some reason it isn't working to use this stub mpi library. And note that I have *not* tried to actually run any parallel tests: even with the changes in cime6.1.137, all of the testing I did was still single-processor tests that just linked to a real mpi library.

billsacks:
> I'll test it on Derecho just to show it works for me and then merge it in.

It seems like you saw this, but I just wanted to call out: Note that this requires cime6.1.137 or later. That's the main reason I kept this separate from #3581 -- #3581 can come in any time, but this one will need to be coordinated with a cime update.

src/CMakeLists.txt
ekluzek: OK, this is all good to know. For the work under #2995 it would've been nice to have the possibility of using MPI. So I was hoping this would've been more helpful for that work. But, since I'm not working on that right now, I don't have a need for this. So letting things lie as they are seems like the best. But, good to know the limitations of what was done. It is really nice that some progress was made regarding MPI though.

ekluzek:
This is blocked by #3583 for the submodule update part of this.

src/CMakeLists.txt
billsacks: Just to give you a little more information on this mpi bit: The unit test framework was originally set up to work with parallel tests in principle, but I'm not sure whether that capability was ever tested. When I updated this framework to switch from pfunit3 to pfunit4 (https://github.com/ESMCI/cime/pull/4397), I saw that the mechanism for parallel testing with pfunit had changed, and since we weren't actually doing any parallel testing at the time, I took a YAGNI approach and didn't bother trying to figure out what we'd need to do to support this. To keep things clean and clear, I removed all aspects of support for a real mpi library at that time.

The recent changes I made to cime bring back just a small piece of this support for using a real mpi library, and actually do so in a kind of weird way, in that they involve using a software stack with a real mpi library even though you're running serial tests.

billsacks:
It looks like this is no longer blocked so can come in whenever is convenient.

wwieder:
@ekluzek do you want to bring this to b4b_dev?  If not, maybe assign someone else?

PR #3614, opened 2025-11-20
Carbon isotopes in streams

ekluzek:
### Description of changes

Brings in some fixes needed to Isotope streams to get the correct results. Adds back the C14 meshfile as well as the tintalgo and taxmode for C13/C14 streams to allow easier checking of results.

### Specific notes

Any User Interface Changes (namelist or namelist defaults changes)? New namelist items

Testing performed, if any: So far just running some test cases that start in 2013

### DIFF
--- a/src/biogeochem/AtmCarbonIsotopeStreamType.F90
+++ b/src/biogeochem/AtmCarbonIsotopeStreamType.F90
@@ -177,12 +177,22 @@ module AtmCarbonIsotopeStreamType
          integer, intent(in) :: year_last ! last  year to use
          integer, intent(in) :: model_year_align ! align yearFirst with this model year
 
+         character(len=len(mapalgo)) :: str_mapalgo ! Temporary copy of mapalgo so can change input if meshfile is none
+
          ! Since C14 data has latitude bands mapalgo and meshfile can neither be set to none
          call shr_assert( trim(mapalgo) /= "none", "mapalgo MUST NOT be none for C14 streams"//errMsg( file=sourcefile, line=__LINE__) )
          ! TODO: Uncomment this error check when we are ready for the test for this to change answers
          !call shr_assert( trim(meshfile) /= "none", "meshfile MUST NOT be none for C14 streams"//errMsg( file=sourcefile, line=__LINE__) )
+         ! TOD: Remove this tempoary bit at the same time
+         if ( trim(meshfile) == "none" )then
+            str_mapalgo = "none"
+         else
+            str_mapalgo = mapalgo
+         end if
+         ! TODO: End of temporary bit
+
          call this%InitBase( bounds, varnames = (/ varname_c14 /), fldfilename=fldfilename, meshfile=meshfile, &
-                             mapalgo=mapalgo, tintalgo=tintalgo, taxmode=taxmode, name=varname_c14, &
+                             mapalgo=str_mapalgo, tintalgo=tintalgo, taxmode=taxmode, name=varname_c14, &
                              year_first=year_first, year_last=year_last, model_year_align=model_year_align )
          call this%C14InitAllocate( bounds )
          call this%Advance( )


ekluzek:
This gives correct results when using these two files:

``` fortran
 stream_fldfilename_atm_c14 = '$DIN_LOC_ROOT/lnd/clm2/isotopes/rawdata_copies/workspace/ctsmforc.Graven.atm_delta_C14_CMIP7_360x720_1700-2023_yearly_v3.0_c251117.nc'
 stream_meshfile_atm_c14 = '$DIN_LOC_ROOT/share/meshes/360x720_120830_ESMFmesh_c20210507.nc'
```

slevis-lmwg:
@ekluzek let me know if we could look this over together. I will not be available after 1p MT.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -149,6 +149,58 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <atm_c14_filename use_c14=".true." use_c14_bombspike =".true." ssp_rcp="SSP5-3.4" >lnd/clm2/isotopes/atm_delta_C14_CMIP6_SSP534os_3x1_global_1850-2100_yearly_c181209.nc</atm_c14_filename>
 <atm_c14_filename use_c14=".true." use_c14_bombspike =".true." ssp_rcp="SSP5-8.5" >lnd/clm2/isotopes/atm_delta_C14_CMIP6_SSP5B_3x1_global_1850-2100_yearly_c181209.nc</atm_c14_filename>
 
+<!-- Settings for Carbon Isotope streams -->
+<stream_fldfilename_atm_c14 use_c14=".true." use_c14_bombspike =".true." ssp_rcp="hist" >lnd/clm2/isotopes/rawdata_copies/workspace/ctsmforc.Graven.atm_delta_C14_CMIP7_360x720_1700-2023_yearly_v3.0_c251117.nc</stream_fldfilename_atm_c14>
+<stream_meshfile_atm_c14    use_c14=".true." use_c14_bombspike =".true." >share/meshes/360x720_120830_ESMFmesh_c20210507.nc</stream_meshfile_atm_c14>
+
+<!-- The following are only set if both use_c14 and use_c14_bonbspike are TRUE -->
```
slevis-lmwg:
```suggestion
<!-- The following are only set if both use_c14 and use_c14_bombspike are TRUE -->
```

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -149,6 +149,58 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <atm_c14_filename use_c14=".true." use_c14_bombspike =".true." ssp_rcp="SSP5-3.4" >lnd/clm2/isotopes/atm_delta_C14_CMIP6_SSP534os_3x1_global_1850-2100_yearly_c181209.nc</atm_c14_filename>
 <atm_c14_filename use_c14=".true." use_c14_bombspike =".true." ssp_rcp="SSP5-8.5" >lnd/clm2/isotopes/atm_delta_C14_CMIP6_SSP5B_3x1_global_1850-2100_yearly_c181209.nc</atm_c14_filename>
 
+<!-- Settings for Carbon Isotope streams -->
+<stream_fldfilename_atm_c14 use_c14=".true." use_c14_bombspike =".true." ssp_rcp="hist" >lnd/clm2/isotopes/rawdata_copies/workspace/ctsmforc.Graven.atm_delta_C14_CMIP7_360x720_1700-2023_yearly_v3.0_c251117.nc</stream_fldfilename_atm_c14>
```
slevis-lmwg: Now that you've tested it, I think this will move a couple of directories up from here.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -149,6 +149,58 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
...
+<stream_year_last_atm_c14  sim_year="1850" sim_year_range="constant">1850</stream_year_last_atm_c14>
+
+<stream_year_first_atm_c14 sim_year="PtVg" sim_year_range="constant">1700</stream_year_first_atm_c14>
+<stream_year_last_atm_c14  sim_year="PtVg" sim_year_range="constant">1700</stream_year_last_atm_c14>
+
+<stream_year_first_atm_c14 sim_year_range="1850-2000" >1700</stream_year_first_atm_c14>
+<stream_year_last_atm_c14  sim_year_range="1850-2000" >2023</stream_year_last_atm_c14>
+<stream_model_year_align_atm_c14  sim_year_range="1850-2000" >1700</stream_model_year_align_atm_c14>
+
+<stream_fldfilename_atm_c13 use_c13=".true." use_c13_timeseries=".true." ssp_rcp="hist" >lnd/clm2/isotopes/ctsmforc.Graven.atm_delta_C13_CMIP7_global_1700-2023_yearly_v3.0_c251013.nc</stream_fldfilename_atm_c13>
```
slevis-lmwg, line 179: New c13 hasn't been tested, yet, if I'm interpreting correctly.

bld/CLMBuildNamelist.pm
```diff
@@ -3792,6 +3797,8 @@ sub setup_logic_c_isotope {
       $use_c13_timeseries = $nl->get_value('use_c13_timeseries');
       if ( &value_is_true($use_c13_timeseries) ) {
          if ( defined($stream_fldfilename_atm_c13) ) {
+            &add_logical_to_nl_flags( $nl_flags, $nl, "use_c13_timeseries" );
+            &add_logical_to_nl_flags( $nl_flags, $nl, "use_c13_timeseries" );
```
slevis-lmwg:
```suggestion
            &add_logical_to_nl_flags( $nl_flags, $nl, "use_c13_timeseries" );
```

slevis-lmwg: OR should the second be
`  &add_logical_to_nl_flags( $nl_flags, $nl, "use_c13_bombspike" );`

AND I'm not looking into it, so I will just ask:
Are lines 3833-3834 needed or are they redundant?

bld/CLMBuildNamelist.pm
```diff
@@ -3832,6 +3841,23 @@ sub setup_logic_c13_streams {
...
+
+  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_fldfilename_atm_c13',
+              'use_c13'=>$nl_flags->{'use_c13'}, 'use_c13_timeseries'=>$nl_flags->{'use_c13_timeseries'},
+              'ssp_rcp'=>$nl_flags->{'ssp_rcp'});
+
+  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_year_first_atm_c13',
+              'sim_year'=>$nl_flags->{'sim_year'}, 'sim_year_range'=>$nl_flags->{'sim_year_range'});
+  add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_year_last_atm_c13',
+              'sim_year'=>$nl_flags->{'sim_year'}, 'sim_year_range'=>$nl_flags->{'sim_year_range'});
+  # Set align year, if first and last years are different
```
slevis-lmwg, line 3855: Is "align year" ignored when first = last, or is it hardwired to the same as first and last?

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
```diff
@@ -127,13 +131,18 @@ subroutine CIsoAtmReadNML( NLFilename )
    end if
    ! Broadcast namelist values to all tasks
    call shr_mpi_bcast( stream_fldfilename_atm_c14, mpicom )
+   call shr_mpi_bcast( stream_meshfile_atm_c14, mpicom )
```
slevis-lmwg, line 134: Similar lines referring to the mesh will be added for c13 I assume.

bld/CLMBuildNamelist.pm
ekluzek: This is the same logic as used in other streams. It's essentially ignored when first == last. There must have been a reason we needed to NOT set it for these type of cases though. If only to not add clutter to the namelist file.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: Oh, yes we might as well do that. I can do that, unless you get to it first.

We also should update the script to add the matlab script that did this to the CMIP7_inputdata place. I suppose we are done with the data processing at this point right?

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: There is a testmod and test in aux_clm that tests this and the previous c14 file.

So you are partially right. The cmip7 c13 is being tested, but the new c14 isn't right now. Although you probably meant c14 rather than c13.

I've tested the namelist building on it's own. The change to the testmod will be answer changing so I'll do that as part of another tag on master.

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
ekluzek: Actually c13 is hardcoded to just use the meshfile == "none" option right now, since it's a single global value. If a new dataset would require a mesh -- we'd need to add it in later. But, it would be easy to do.

But, that does remind me that I should add some checking for these things in the code.

bld/CLMBuildNamelist.pm
ekluzek: 3833-3834 are added to ensure that these are set in all cases. It's also done here as it's going to be used in the next call. So has to be done in that case.

I could add a comment to say this for clarity.

bld/namelist_files/namelist_defaults_ctsm.xml
slevis-lmwg: Ok, I copied the files to the /isotopes directory.

I think we are done with the data processing. I will open a TODO card on my Sprint board.

bld/namelist_files/namelist_defaults_ctsm.xml
slevis-lmwg: I was confused (probably still am). Seemed as though you were pointing to the new c14 file (above) but the old c13 file (here) . I interpreted that to mean that you were now reading the new c14 but not the new c13 yet. If I have misunderstood, it may help me to see a compete list of your stepwise approach, so that I can understand which step you are addressing with this PR. Sorry...

src/biogeochem/CNCIsoAtmTimeSeriesReadMod.F90
slevis-lmwg: This may be part of my confusion: 
I thought we were replacing both the c14 and the c13 with the 360x720 files that I generated.

ekluzek: Ahh, I didn't even notice you created a c13 version. Nope we don't need that one. It's only when there are latitude bands > 1 that we do.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: No problem. I'll try to make the steps more clear in one of the issues. And in the definition of this PR. The purpose evolved a bit as I got into it.

But, also see below that we don't need c13 on the half degree grid -- just c14. And it only because we need to regrid and need to have more elements on the file/mesh than LND_TASKS.

ekluzek: So actually you are right, I just didn't pick up what you were saying.

slevis-lmwg: Ok, it makes sense now :-)

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: OK, moved it up.

ekluzek:
Testing is passing as expected at this point, so ready to merge. I just need @slevis-lmwg to finalize approval, and I'll merge it in.

slevis-lmwg (approved):
Thank you @ekluzek.

PR #3631, opened 2025-11-25
Use CMIP7 Carbon isotopes by default

ekluzek:
### Description of changes

This turns on the CMIP7 Carbon isotopes by default. It's answer changing for BGC cases that use Carbon Isotopes.

### Specific notes

Are answers expected to change (and if so in what way)? Yes -- only with use_c13/use_c14 on

  Carbon isotope streams will resond to the CLM_CMIP_ERA setting

Testing performed, if any: None, so far will do regular testing
  Right now this is the same as the branch with b4b ciso changes in it
  Will be adding the changes in a bit


ekluzek:
Replaced by #3634

PR #3633, opened 2025-11-26
Add history outputs: Crop biomass and LAI per harvest

### LINKED ISSUE #3696: CropPhenology: Rename leafout to gddtsoi
Because this is confusing:

https://github.com/ESCOMP/CTSM/blob/19d9b8a35cb433486c5e8d914a11e1c45f4e92c3/src/biogeochem/CNPhenologyMod.F90#L2095

Seemingly this is the only place it's used.

samsrabin:
### Description of changes

Adds a bunch of history outputs with the `mxharvests` dimension (max number of crop harvests per year; 2).

`MAX_TLAI_PERHARV` gives the maximum total LAI seen across the growing season.

20 other outputs give the biomass for five C pools (leaf, grain, stem, fine root, coarse root) at three crop phase transitions (emergence, anthesis, maturity) and harvest (which may not happen at maturity).

Also adds an option to suppress a warning about low `gddmaturity`, which I used during the CRU-JRA crop calendar work.

### Specific notes

- Resolves #3696

**Does this create a need to change or add documentation? Did you do so?** Will need to update non-FATES output list; no.

- [x] Update non-FATES output list

### ✅ New unit tests
Pass.

### ✅ aux_clm
~~`aux_clm` reveals what seem to be threading problems with `MAX_TLAI_PERHARV`, which fails `COMPARE_base_rest` in `ER[PS]_Ly3_P64x2` tests but not an `ERS_Ly3` test.~~ Resolved with a7e66aa.
- [x] Redo

### ✅ Sense checks
- [x]  Using [this script](https://gist.github.com/samsrabin/65062431fdf12e96a0a5af15c4ac0a86)

## To do

- [x] Fix threading issue.
- ~~Add HARVORGANC equivalent but JUST for grain. Will match what I have there for now, because grain is all CLM simulates by default. But it'll differ with the options underneath. (Do not name GRAIN unless you fix LPREPRSTRUCT test to handle this.)~~ People should just use existing `GRAINC_TO_(FOOD|SEED)_PERHARV` outputs.
- ~~Add support for `for_testing_use_second_grain_pool` and/or `for_testing_use_repr_structure_pool` true (or throw error in that case).~~ See above.


samsrabin:
Adding "next" label until we get a reviewer.

cime_config/testdefs/testmods_dirs/clm/crop/user_nl_clm
```diff
@@ -13,7 +13,12 @@ hist_fincl2  += 'DYN_COL_SOIL_ADJUSTMENTS_C'
 ! Note that, under normal circumstances, these should only be saved annually.
 ! That's needed for the mxsowings and mxharvests axes to make sense.
 ! However, for testing purposes, it makes sense to save more frequently.
-hist_fincl3 = 'SDATES', 'SDATES_PERHARV', 'SYEARS_PERHARV', 'HDATES', 'GRAINC_TO_FOOD_PERHARV', 'GRAINC_TO_FOOD_ANN', 'GRAINN_TO_FOOD_PERHARV', 'GRAINN_TO_FOOD_ANN', 'GRAINC_TO_SEED_PERHARV', 'GRAINC_TO_SEED_ANN', 'GRAINN_TO_SEED_PERHARV', 'GRAINN_TO_SEED_ANN', 'HDATES', 'GDDHARV_PERHARV', 'GDDACCUM_PERHARV', 'HUI_PERHARV', 'SOWING_REASON_PERHARV', 'HARVEST_REASON_PERHARV', 'SWINDOW_STARTS', 'SWINDOW_ENDS', 'GDD20_BASELINE', 'GDD20_SEASON_START', 'GDD20_SEASON_END'
+hist_fincl3 = 'SDATES', 'SDATES_PERHARV', 'SYEARS_PERHARV', 'HDATES', 'GRAINC_TO_FOOD_PERHARV', 'GRAINC_TO_FOOD_ANN', 'GRAINN_TO_FOOD_PERHARV', 'GRAINN_TO_FOOD_ANN', 'GRAINC_TO_SEED_PERHARV', 'GRAINC_TO_SEED_ANN', 'GRAINN_TO_SEED_PERHARV', 'GRAINN_TO_SEED_ANN', 'HDATES', 'GDDHARV_PERHARV', 'GDDACCUM_PERHARV', 'HUI_PERHARV', 'SOWING_REASON_PERHARV', 'HARVEST_REASON_PERHARV', 'SWINDOW_STARTS', 'SWINDOW_ENDS', 'GDD20_BASELINE', 'GDD20_SEASON_START', 'GDD20_SEASON_END', 'MAX_TLAI_PERHARV'
```
wwieder: can the requested output just be included in the list with everything else?

wwieder:
@slevis-lmwg can you take this one one, I think you have the most familiarity with the crop model. Is it helpful to go through the PR with @samsrabin ?

src/biogeochem/CropType.F90
```diff
@@ -940,6 +1221,54 @@ subroutine CropIncrementYear (this, num_pcropp, filter_pcropp)
...
+       this%leafc_emergence_patch(p) = leafc
+       this%reprc_emergence_patch(p) = reprc
+    else if (this%cphase_patch(p) == cphase_grainfill) then
+       this%frootc_anthesis_patch(p) = frootc
+       this%livecrootc_anthesis_patch(p) = livecrootc
+       this%livestemc_anthesis_patch(p) = livestemc
+       this%leafc_anthesis_patch(p) = leafc
+       this%reprc_anthesis_patch(p) = reprc
+    else if (this%cphase_patch(p) == cphase_harvest) then
+       this%frootc_maturity_patch(p) = frootc
```
samsrabin: Only save to maturity variable if it's actually mature. Always save to a new `harvest` variable.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -2532,6 +2557,9 @@ subroutine CropPhenology(num_pcropp, filter_pcropp                     , &
                ! changes to the offset subroutine below
 
             else if (do_harvest) then
+               cphase(p) = cphase_harvest
+               call crop_inst%CropPhaseTransitionBiomass(p, cnveg_carbonstate_inst)
```
slevis-lmwg: In other phases you include
`if (cphase(p) /= cphase_orig) then` before calling CropPhaseTransitionBiomass.
Here do you not because harvest occurs in one tstep always?

src/biogeochem/CropType.F90
samsrabin: Done with a7e66aa.

src/biogeochem/CNPhenologyMod.F90
samsrabin: More like: By definition, if we're harvesting, then it wasn't in `cphase_harvest` at the beginning of the subroutine.

samsrabin: Another way of saying this: `do_harvest` means "`cphase(p)` isn't `cphase_harvest` but now it should be."

slevis-lmwg: I recommend a simple comment in the code to explain that.

src/biogeochem/CropType.F90
```diff
@@ -1016,6 +1076,31 @@ subroutine Restart(this, bounds, ncid, cnveg_state_inst, flag)
...
+           call restartvar(ncid=ncid, flag=flag, varname='leafc_maturity_thisyr_patch', xtype=ncd_double,  &
+                dim1name='pft', dim2name='mxharvests', switchdim=.true., &
+                long_name='Leaf C at maturity for harvests in this patch this year', units='unitless', &
+                scale_by_thickness=.false., &
+                interpinic_flag='interp', readvar=readvar, data=this%leafc_maturity_thisyr_patch)
+           call restartvar(ncid=ncid, flag=flag, varname='reprc_maturity_thisyr_patch', xtype=ncd_double,  &
+                dim1name='pft', dim2name='mxharvests', switchdim=.true., &
+                long_name='Reproductive C at maturity for harvests in this patch this year', units='unitless', &
+                scale_by_thickness=.false., &
+                interpinic_flag='interp', readvar=readvar, data=this%reprc_maturity_thisyr_patch)
```
slevis-lmwg: should these have been "harvest" rather than maturity?

slevis-lmwg: For now I will wait for successful aux_clm testing before I hit "approve" if you don't mind @samsrabin

src/biogeochem/CNPhenologyMod.F90
samsrabin: Done with 7e480b9.

src/biogeochem/CropType.F90
samsrabin: Great catch! Resolved with 71bc137.

samsrabin:
@slevis-lmwg This is ready for re-review. Note that I realized it's possible to skip a crop phase (i.e., increasing crop phase by 2 or more during a single timestep), so I had to rework `CropPhaseTransitionBiomass()` a bit. I also added tests to make sure it behaves correctly in such situations (as well as typical cases).

src/biogeochem/CropType.F90
```diff
@@ -1366,19 +1366,24 @@ subroutine CropPhaseTransitionBiomass (this, p, cnveg_carbonstate_inst, is_matur
...
        this%frootc_anthesis_patch(p) = frootc
        this%livecrootc_anthesis_patch(p) = livecrootc
        this%livestemc_anthesis_patch(p) = livestemc
        this%leafc_anthesis_patch(p) = leafc
        this%reprc_anthesis_patch(p) = reprc
-    else if (this%cphase_patch(p) == cphase_harvest) then
+    end if
+
+    ! Don't need >= etc. here because harvest is the maximum (final) phase
+    if (this%cphase_patch(p) == cphase_harvest) then
```
slevis-lmwg: Might it be of use, even just so that users may interpret the code more easily, to add an error check (or assert statement of some kind) confirming that cphase_patch falls within the correct range?

src/biogeochem/CropType.F90
```diff
@@ -1366,19 +1366,24 @@ subroutine CropPhaseTransitionBiomass (this, p, cnveg_carbonstate_inst, is_matur
...
+    if (this%cphase_patch(p) >= cphase_leafemerge .and. this%frootc_emergence_patch(p) < 0._r8) then
        this%frootc_emergence_patch(p) = frootc
        this%livecrootc_emergence_patch(p) = livecrootc
        this%livestemc_emergence_patch(p) = livestemc
        this%leafc_emergence_patch(p) = leafc
        this%reprc_emergence_patch(p) = reprc
-    else if (this%cphase_patch(p) == cphase_grainfill) then
+    end if
+
+    if (this%cphase_patch(p) >= cphase_grainfill .and. this%frootc_anthesis_patch(p) < 0._r8) then
```
slevis-lmwg: I may have noticed this before and didn't address at the time:
Seems inconsistent to use two terms in one of the phases, namely anthesis and grainfill, while using a single term to identify other phases. Is there a logical reason for it?

I think of anthesis as similar to emergence and harvest, i.e. as events that occur in one time step, while grainfill goes on for a little while. But we also treat emergence as a non-instantaneous phase, don't we? Just trying to clarify.

samsrabin: I agree that it's inconsistent—I would prefer for `cphase_leafemerge` to be `cphase_postemergence` instead. I initially would have said `cphase_vegetative`, but technically I think the pre-emergence phase is included in the usual usage of "vegetative phase."

src/biogeochem/CropType.F90
samsrabin: Done with 08b93e7.

src/biogeochem/CropType.F90
slevis-lmwg: If you feel ready to finish this PR, I would be open to inserting a comment pointing out the inconsistency, as something that could be addressed eventually. I don't know if it's worth opening an issue...

samsrabin: I've just gone ahead and opened an issue (#3690), because that makes more sense to me than an arbitrarily-located comment in the code.

samsrabin:
It's also possible for harvest to occur without having reached anthesis, or even emergence.

- [x] Handle this by calling `CropPhase` from inside `CropPhaseTransitionBiomass` to check what phase the crop is actually in.

This will require moving `CropPhase` (or at least a procedure doing what happens inside its `for` loop) from `CNPhenologyMod` to `CropType`... Or moving `CropPhaseTransitionBiomass` to `CNPhenologyMod`. That'd probably be better, to keep all the phenology stuff there.

samsrabin:
Blocked by #3713.

samsrabin:
#3713 is merged, so this is unblocked.

PR #3640, opened 2025-12-01
Restructure Carbon Isotope streams for OpenMP

ekluzek:
### Description of changes

This moves Carbon isotope streams calls to clm_driver outside of OMP loops. And also adds a missing Advance call for C13Streams. This should be b4b-dev except maybe the ciso_cmip7 test, and if so this will need to be a tag on master. Or just the answer changing part moved to the answer changing tag (which I think is just the addition of the Advance call).

### Specific notes

Are answers expected to change (and if so in what way)? No (but maybe the ciso cmip7 test?)

Testing performed, if any: will run standard testing

### DIFF
--- a/src/biogeophys/PhotosynthesisMod.F90
+++ b/src/biogeophys/PhotosynthesisMod.F90
@@ -2067,8 +2067,7 @@ contains
     !
     ! Determine total photosynthesis
     !
-    !use CIsoAtmTimeseriesMod, only : C14BombSpike, C13TimeSeries
-    use CIsoAtmTimeseriesMod, only : rc13_atm_grc, rc14_atm_grc
+    use CIsoAtmTimeseriesMod, only : rc13_atm_grc, rc14_atm_grc  ! Atmospheric Carbon isotope timeseries data
     ! !ARGUMENTS:
     type(bounds_type)      , intent(in)    :: bounds
     integer                , intent(in)    :: fn                             ! size of pft filter


src/main/clm_driver.F90
```diff
@@ -408,6 +409,10 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
        call PrescribedSoilMoistureAdvance( bounds_proc )
        call t_stopf('prescribed_sm')
     endif
+      ! Get the current C13/C14 ratio in the atmosphere from timeseries data or the fixed values
```
ekluzek: This isn't properly indented.  And I wonder if it should be within a use_cn if statement?

ekluzek: I added the use_cn if statement, and it's indented correctly now.

ekluzek: Based on what I'm gathering from #3641 I think being in a use_cn statement makes sense. And it probably should be moved to inside of CNVegetationFacade. And there likely should be more checking around this to make sure this is all consistent. But, this is a good start for now.

src/biogeophys/PhotosynthesisMod.F90
```diff
@@ -2067,7 +2067,7 @@ subroutine PhotosynthesisTotal (bounds, fn, filterp, &
     !
     ! Determine total photosynthesis
     !
-    use CIsoAtmTimeseriesMod, only : C14BombSpike, C13TimeSeries
+    !use CIsoAtmTimeseriesMod, only : C14BombSpike, C13TimeSeries
```
ekluzek: PRoperly remove that line.

slevis-lmwg (approved):
@ekluzek and I looked over this, and I'm approving.

Erik has a few refactor ideas in clm_driver, but we agreed that they were out of scope for now.

src/biogeophys/PhotosynthesisMod.F90
ekluzek: Got it

ekluzek:
Testing is now done on Izumi and passes as expected other than this test:

ERI_D_Ld20.f10_f10_mg37.IHistClm60Bgc.izumi_nag.clm-default--clm-midDecStart--clm-allowTransientYearJump

which also doesn't pass on derecho_intel either it turns out. But, it also doesn't pass on the raw b4b-dev branch, so I'm merging this in.

@slevis-lmwg we'll need to figure this out on b4b-dev, or revert #3299. There has been work on ERI tests in cime, and perhaps that might be needed. We could also bring it in and mark them as expected fails. It looks like it's failing in mosart -- which is really odd. It might mean we should just turn off MOSART or something like that?

slevis-lmwg:
> Testing is now done on Izumi and passes as expected other than this test:
> 
> ERI_D_Ld20.f10_f10_mg37.IHistClm60Bgc.izumi_nag.clm-default--clm-midDecStart--clm-allowTransientYearJump
> 
> which also doesn't pass on derecho_intel either it turns out. But, it also doesn't pass on the raw b4b-dev branch, so I'm merging this in.
> 
> @slevis-lmwg we'll need to figure this out on b4b-dev, or revert #3299. There has been work on ERI tests in cime, and perhaps that might be needed. We could also bring it in and mark them as expected fails. It looks like it's failing in mosart -- which is really odd. It might mean we should just turn off MOSART or something like that?

In case this helps with troubleshooting, see the same test in #3299:
`/glade/derecho/scratch/slevis/tests_1125-171338de/ERI_D_Ld20.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-default--clm-midDecStart--clm-allowTransientYearJump.C.1125-171338de_int/TestStatus`
Mine fails in COMPARE_base_hybrid. Is this what you see? This is the new test introduced in #3299, and I see the following relevant posts:
https://github.com/ESCOMP/CTSM/pull/3299#issuecomment-3112000970 test without `clm-default` fails in the same phase
https://github.com/ESCOMP/CTSM/pull/3299#issuecomment-3169132636 the test with `clm-default` passed

Issue #3589, opened 2025-11-06
355+ K land temperature triggers RRTMGP error


olyson:
### Brief summary of bug

From @adamrher : I've gotten out of bounds surface temperature errors from RRTMGP in 3 of 5 coupled runs I've performed this week. I tracked down 1 of those errors to 355+ K surface temperatures at a point in S.America (RRTMGP triggers an error when temperatures fall outside of 160 K - 355 K). 

### General bug information

**CTSM version you are using:** cesm3_0_alpha07c

**Does this bug cause significantly incorrect results in the model's science?** Model crashes

**Configurations affected:** Compset longname is 1850C_CAM70%LT_CLM60%BGC-CROP_CICE_MOM6_MOSART_DGLC%NOEVOLVE_WW3_SESP

### Details of bug

<img width="1728" height="1117" alt="Image" src="https://github.com/user-attachments/assets/6707030b-c81f-4078-bee2-838025a5414e" />

### Important details of your setup / configuration so we can reproduce the bug

/glade/derecho/scratch/aherring/b.e30_alpha07c_cesm.B1850C_LTso.ne30_t232_wgx3.239 (and several other CESM3_dev simulations)

olyson:
Replicated crash and diagnostic output with:

/glade/derecho/scratch/oleson/b.e30_alpha07c_cesm.B1850C_LTso.ne30_t232_wgx3.239.test

olyson:
The t_sfc in RRTMGP is derived by inverting the upward longwave from CTSM.  The gridcell-average upward longwave for the problematic gridcell must thus be in excess of 900 W/m2.  
So I've trapped for patch (pft) - level upward longwave above 900 W/m2. A particular patch in that grid cell has very high vegetation temperatures, > 360K, and upward longwave, up to 980 W/m2, associated with a patch of broadleaf deciduous tropical tree.  The characteristics of that patch are:

1. Very low leaf area, ~0.01
2. High stem area, 5-6
4. Very high incoming solar radiation, 1000-1100 W/m2
4. Very high absorbed solar radiation, 920-1020 W/m2, implying an albedo of ~0.08

[...]

PR #3643, opened 2025-12-02
post-5.4.n02.ctsm5.4.000: Fix for RRTMGP errors caused by high surface temperatures

olyson:
### Description of changes

               ! Add esai to sa_leaf if elai is less than threshold.
               ! Intended to avoid small sa_leaf which leads to small leaf conductance
               ! and high leaf temperature. This in turn can lead to unrealistically
               ! high surface temperatures passed to the atmospheric model (The RRTMGP
               ! component in particular, which returns an error and stops the model
               ! if the surface temperature is greater than 355K).
               ! See https://github.com/ESCOMP/CTSM/issues/3589 for more info.
               ! The 0.1_r8 value is fairly arbitrary but has been effective in
               ! avoiding RRTMGP errors in CESM3 development simulations.

### Specific notes

Contributors other than yourself, if any: @swensosc , @slevis-lmwg 

CTSM Issues Fixed (include github issue #): #3589

Are answers expected to change (and if so in what way)? Yes (similar climate though)

Does this create a need to change or add documentation? Did you do so? Yes. No.

Testing performed, if any:
See #3589 for testing description.  I have not run any CTSM system testing.


slevis-lmwg (approved):
Approving. I will start aux_clm to confirm nothing's broken. Also I will try rebasing to the post-5.4 branch to get this in more efficiently.

slevis-lmwg:
Rebasing didn't work, as it ended up with 44 files changed, so I reverted to master. I thought rebasing would apply the same changes but starting from post-5.4 instead of starting from master. Clearly there's something about rebasing that I don't get, because sometimes it does what I expect and sometimes not.

Maybe if I update the post_5.4 branch to ctsm5.3.085? I will think about it. I will hold off on aux_clm for now.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -754,6 +754,19 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+               ! Add esai to sa_leaf if elai is less than threshold.
+               ! Intended to avoid small sa_leaf which leads to small leaf conductance
+               ! and high leaf temperature. This in turn can lead to unrealistically
+               ! high surface temperatures passed to the atmospheric model (The RRTMGP
+               ! component in particular, which returns an error and stops the model
+               ! if the surface temperature is greater than 355K).
+               ! See https://github.com/ESCOMP/CTSM/issues/3589 for more info.
+               ! The 0.1_r8 value is fairly arbitrary but has been effective in 
+               ! avoiding RRTMGP errors in CESM3 development simulations.
+               if(elai(p) .lt. 0.1_r8) then
```
ekluzek:
```suggestion
               if(elai(p) < 0.1_r8) then
```

Just suggesting to use this syntax for human readability.

ekluzek: I also suggest making the 0.1_r8 a parameter so that it's meaning is more clear.

``` fortran
     real(r8), parameter :: min_lai = 0.1_r8 ! Minimum LAI to add SAI to sa_leaf
                                                                 ! Value is arbitrary but has been effective in avoiding RRTMGP errors in CESM3 development simulations
.
.
.
     if(elai(p) < min_lai) then
```

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -754,6 +754,19 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
+                  sa_leaf(p) = sa_leaf(p) + esai(p)
```
ekluzek, line 768: For consistency with the above section shouldn't sa_stem also be set to zero?

``` fortran
               frac_rad_abs_by_stem(p) = 0.0_r8
               sa_stem(p) = 0.0_r8
```

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -754,6 +754,19 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
                frac_rad_abs_by_stem(p) = 0.0_r8
```
ekluzek, line 756: This is not something you did. But, the comment starting in line 749 looked wrong to me, at first. And then I realized what it's saying and it is correct, just doesn't obviously correspond to the code logic.

The code logic says:

 if it is NOT a tree or shrub OR it is (tree or shrub) and dbh is less than some minimum value combine stem and leaf together, rather than calculate them separately)

ekluzek: So I wonder if the comment should be changed?

src/biogeophys/CanopyFluxesMod.F90
ekluzek: And actually maybe what would be better would be to make the original if even more complex so that if it's triggered it does the same thing. If you do that I'd suggest you make a little logical function to put it in though.

So something like:

``` fortran
         if ( combine_leaf_and_stem(p) )then
               frac_rad_abs_by_stem(p) = 0.0_r8
               sa_stem(p) = 0.0_r8
         end if
.
.
.
     logical function combine_leaf_and_stem( p )
         if ( is_tree(patch%itype(p)) .or. is_shrub(patch%itype(p) )then
              ! combine leaf and stem if dbh is less than a minimum diameter
              if ( dbh(p) < min_stem_diameter )then
                 combine_leaf_and_stem = .true.
              ! Also combine leaf and stem if LAI is smaller than a minimum value
              else if ( elai(p) < min_lai) then
                 combine_leaf_and_stem = .true.
              else
                  combine_leaf_and_stem = .false.
              end if
         else
             ! Always combine leaf and stem if it isn't a tree or shrub
             combine_leaf_and_stem = .true.
         end if
```

I think this makes the code much more readable.

ekluzek: This would also allow min_stem_diameter to be moved to just inside the function. And min_lai could be set in just the function as well.

ekluzek: If it were to be needed this could allow unit testing to be done on the function. I think making it into a function allows it to be broken up enough though that doing that isn't needed. It's easier to understand the logic for the function and see that it's correct. The complex if statement takes quite a bit of time to comprehend.

src/biogeophys/CanopyFluxesMod.F90
olyson: Done

src/biogeophys/CanopyFluxesMod.F90
olyson: From conversations with @swensosc I don't think sa_stem should be set to zero, but pinging @swensosc here to double-check.  We still have stems, we are just adding stems to very small leaf area to increase leaf conductance and sensible heat.
I don't have a problem with the idea of creating a new function to handle the existing code, allow for unit testing, etc.  But this is out of scope for this PR.

swensosc:
I think keith has it correct.  In some cases he found, stem area was
sufficiently large, but leaf area was too small.  The boundary layer
conductances are calculated separately, so the model doesn't know that the
tiny leaf is part of a larger structure that would presumably alter the
conductance in this case.  So the fix modifies the sa_leaf term for the
purpose of increasing the conductance.

src/biogeophys/CanopyFluxesMod.F90
olyson: Seems like this would clarify it:

            ! only calculate separate leaf/stem heat capacity for trees
            ! and shrubs if dbh is greater than some minimum value.
            ! otherwise, set surface area for stem and fraction absorbed by stem to zero, and add esai to sa_leaf.

"Otherwise" would cover cases where it is not tree or shrub, or if it is tree or shrub and dbh is not greater than the minimum value.

ekluzek: Yes, that is so much clearer!

src/biogeophys/CanopyFluxesMod.F90
ekluzek: Yep, I can make an issue for the suggested refactor. The important thing is to hear back from @swensosc on if the other things should be zeroed as in the top part. I read it as it should do the same thing for both cases -- but let's have @swensosc confirm.

olyson: To clarify my out of scope comment, my impression is that we are emphasizing separating non-bfb changes with bfb changes.  It seems like the new function would be a candidate for the bfb-dev branch, after the non-bfb is merged to master.

ekluzek: Yes, that's a really good point. We should make this just the minimal part that shows the change in answers. Other refactoring can be made on b4b-dev and shown to NOT change answers. That is the best mode of operating.

src/biogeophys/CanopyFluxesMod.F90
olyson: Ok, thanks, change committed.

src/biogeophys/CanopyFluxesMod.F90
ekluzek: Here's the comment from Sean that clarifies this:

https://github.com/ESCOMP/CTSM/pull/3643#issuecomment-3608823887

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -745,15 +747,28 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
...
                sa_leaf(p) = sa_leaf(p) + esai(p)
+            else
+               ! Add esai to sa_leaf if elai is less than threshold.
+               ! Intended to avoid small sa_leaf which leads to small leaf conductance
+               ! and high leaf temperature. This in turn can lead to unrealistically
+               ! high surface temperatures passed to the atmospheric model (The RRTMGP
+               ! component in particular, which returns an error and stops the model
+               ! if the surface temperature is greater than 355K).
+               ! See https://github.com/ESCOMP/CTSM/issues/3589 for more info.
+               ! The 0.1_r8 value is arbitrary but has been effective in
```
ekluzek: Since this part of the comment is above, you could remove these two lines.

olyson: Good point, thanks.  Done.

src/biogeophys/CanopyFluxesMod.F90
wwieder: Good discussion here and I agree that we should keep this as simple fix for a somewhat odd model behavior in edge cases.  Let's avoid a larger refactor or introduction of unit testing at this point.

slevis-lmwg:
I will try merging the latest from post-5.4 (tag post-5.4.n01.alpha-ctsm5.4.CMIP7.21.ctsm5.3.085) and rebasing to post-5.4. If all that works, then I can start the test-suites.

slevis-lmwg:
`./run_sys_tests -s aux_clm -c post-5.4.n01.alpha-ctsm5.4.CMIP7.21.ctsm5.3.085 -g post-5.4.n02.ctsm5.4.000`
derecho and izumi OK: BASELINE post-5.4.n01.alpha-ctsm5.4.CMIP7.21.ctsm5.3.085: DIFF

I looked at the cprnc file of the shortest tests (Ln9, Ld2) e.g.:
`SMS_Ln9.ne3pg3_ne3pg3_mt232.I2000Clm60Sp.derecho_gnu.clm-clm60cam7LndTuningMode--clm-nofireemis.GC.1208-175929de_gnu/SMS_Ln9.ne3pg3_ne3pg3_mt232.I2000Clm60Sp.derecho_gnu.clm-clm60cam7LndTuningMode--clm-nofireemis.GC.1208-175929de_gnu.cpl.hi.2000-01-01-16200.nc.cprnc.out`
All fields with diffs (greater than roundoff but small) have prefixes lnd[Imp,Exp] and rof[Imp,Exp].

src/biogeophys/CanopyFluxesMod.F90
slevis-lmwg: @olyson I'm resolving since you wrote "Done."

slevis-lmwg:
Keith and I looked at diffs in a 2-yr test (fields FSH, FIRE, TV, TG), and Keith approved the changes that we saw.

PR #3645, opened 2025-12-03
Fix eri transient jump tests by having mosart history daily

ekluzek:
### Description of changes

This fixes the problem brought in with #3299 for the transient ERI tests

### Specific notes

Any User Interface Changes (namelist or namelist defaults changes)? No No

Testing performed, if any: ran the test on Derecho and saw it works, will run on Izumi


ekluzek:
Passes on Izumi, so merging in.

Issue #3476, opened 2025-09-08
Add information on the gridcell 2D index on ouput


ekluzek:
With #2995 I'll have the ability to output the 2D gridcell index. This could be reported on a fail with the error checking, or could be stored as part of the decomposition information at startup.

Definition of done:
- [x] Add the i, j index to the write_point_context subroutine in abortutils
- [ ] Add it to the BalanceCheck and CNBalanceCheckMod fails
- [ ] Add it to error index reporting in lnd_import_export_utils.F90
- [ ] Add it to index reporting fails in subgridAveMod
- [ ] Add it to index reporting fails in initInterpMindist.F90
- [ ] Add it to the following miscellansous places:
    - [ ] DustEmis*.f90
    - [ ] UrbanFluxesMod
    - [ ] ch4Mod
    - [ ] HillslopeHydrologyMod
 - [ ] Check for  other occurances with something like: 
     -- `git grep write | grep "index"`
     -- git grep 'write(iulog' | grep "grc%lon"

ekluzek:
What is currently reported when the code aborts with a problem with a particular point is that it reports on the decomposition index, and the latitude/longitude of the point. So typically people have to figure out what global 2D gridcell that latitude/longitude corresponds to. Giving the i/j index when reporting on a bad point where the model is shutting down would make it easier to see for example what point on the surface dataset corresponds to the point with the problem.

samsrabin:
This seems like a great idea for usability! I do wonder, though, if it's something we should be prioritizing at the moment.

ekluzek:
> This seems like a great idea for usability! I do wonder, though, if it's something we should be prioritizing at the moment.

Always, a good question.

It came up as a natural part of my work in #2995, which is the only reason I bring it up. And as part of that work I can do some of this. I also want to make sure we agree on how this should be done. And some of the steps can be done over time.

wwieder:
What are we looking at for time commitment here, Erik?  

If it's beyond the scope of work for the SIF, maybe it's better to create an issue with a later milestone and some notes for your future self?

ekluzek:
> What are we looking at for time commitment here, Erik?
> 
> If it's beyond the scope of work for the SIF, maybe it's better to create an issue with a later milestone and some notes for your future self?

[...]

PR #3469, opened 2025-09-03
decomp: Testing for MPI_SCAN for the decomposition

ekluzek:
### Description of changes

Experimental branch for working with using MPI_SCAN for the decomposition to get the gridcell offsets.

This is starting off as very experimental with calls to asserts to make sure it gives the same results as before.

### Specific notes

Contributors other than yourself, if any: @johnmauf

CTSM Issues Fixed (include github issue #):
  Work for #2995
  Fixes #3370

Are answers expected to change (and if so in what way)? no

Any User Interface Changes (namelist or namelist defaults changes)? No 

Does this create a need to change or add documentation? Did you do so? No No

Testing performed, if any: Running the decomp_init testlist

Issue #3466, opened 2025-08-30
Dimension clumps object by clumps_proc rather than global nclumps


ekluzek:
The clumps object in decompMod is an array over all the clumps for the entire grid, and is on each task. When initializing the decomposition it's used to setup the decomposition, but after initialization only the clumps over the local processor are needed.

Hence, we can save memory at least by saving a copy of the local clumps for the given task, and then deallocating clumps and reallocating only the part for the local task.

So something like:

 ``` fortran
type(clump_type), allocatable  :: clumps_local(:)
.
.
.
nclumps = bounds_proc%nclumps
allocate( clumps_local(nclumps) )
clumps_local = clumps(nclumps)
deallocate( clumps )
allocate( clumps(nclumps) )
clumps(:) = clumps_local(:)
deallocate( clumps_local )
```

Definition of done:
- [ ] Dimension clumps as clump_pproc rather than nclumps
- [ ] Change decompMod: get_proc_global to do a mpi_reduce to get the totals over all processors to log about
- [ ] deallocate lcid sooner in decompInit_lnd
- [ ] Change code in decompInit_clumps as follows:
   - [ ] Change loops over nclumps to over clump_pproc
   - [ ] Remove mpi operations
   - [ ] Change or remove the logging at the end

ekluzek:
I'm running testing where I reset the clumps that aren't used by the local task to -1 values so if they are used it'll trigger an error. So far this is working which shows that we only need the local processor.

It also looks to me that there isn't a need to figure this out for every other processor. So we should be able to have each local processor figure out the subgrid levels for their own gricells. Which also means there's no need to gather and scatter non-local clumps to every other task. Some of the processing for this done over masterproc and then sent to the other processors. But, we can just have each processor figure it out for just their own gridcells and save memory from the get go.

ekluzek:
clump_type has 16 4 byte integers and there are two other arrays to remove. So for the mpasa3p75 case on 38,400 tasks that's 18*4*38400 = 2.7 MB so 307 MBytes per node. So not a lot of savings.

However, in talking with @johnmauff yesterday, he pointed out that there will be overhead around this in the MPI operations for getting this on each processor.

In looking at the code. There are two arrays of size npes for each task in decompInit_clumps. These should be changed to dimension clump_pproc, and the MPI operations can be removed. Loops over nclumps change to over clump_pproc. The output of clumps should either be removed or just done over the local

``` fortran
[...]

PR #3666, opened 2025-12-16
b4b: Move the changes to initialize the task decomposition from mpi_scan to main development

### LINKED ISSUE #3368: Deallocate memory in lnd_set_decomp_and_domain
It turns out the ESMF objects that are created in lnd_set_decomp_domain that could be cleaned after they are used as they are only needed in initialization. The objects are in local memory, but the memory that ESMF allocates for them isn't cleared without doing explicit Destroy or Release calls.

### LINKED ISSUE #3370: Convert decompInit_lnd to use MPI_SCAN
Convert decompInit_lnd to use MPI_SCAN rather than the current methodology.

### LINKED ISSUE #3672: Add add_simple_patch argument to unittest_add_column
Add an optional argument to unittest_add_column in unittestSubGridMod.F90, that adds a simple single bare-soil patch beneath the column created.

ekluzek:
### Description of changes

This moves the core code changes to initialize the processor decomposition from the mpi_scan branch in #3469 to b4b-dev. This removes some of the changes for memory checking and additional self testing as well as some of the additional timers that don't look useful now.

I created two previous branches before I created the process in https://github.com/ESCOMP/CTSM/discussions/3665 where I worked out the details to NOT make this have too many commits and be hard to do. I also figured out how to remove merge commits as they need special handling, and usually aren't wanted in a case like this. Another way to do this would be to do this outside of git, which might have been similar length as the final version, but could've missed some important changes.

The actual code changes is that the task/thread decomposition determined at initialization was updated to be based on mpi_scan, rather than using a lot of complex custom code that was a much higher order based on the number of tasks. Methods to calculate the 2D global indicies for a point was added so that it could be displayed on endrun calls. processor_type and clump_type were moved towards an Object-Oriented (OO) programming style, more error checking added, including checking that the decomposion is correct, deallocates some initialization memory that isn't needed.

- decompMod -- New subroutines:  decompmod_allocate_clumps, decompmod_allocate_gindex,, 
                             calc_ijindices_from_full_global_index
                             New methods for processor_type:    InitAllocate , calc_global_index_fromij, 
                             calc_globalxy_indices
                             New method for clump_type: Init
                             Add new data: nglob_x, nglob_y (global 2D indices)
                             Add more error checking
                             Add only_gridelll optional argument to get_proc_bounds (needed for calls used before the subgrid levels are set from calling decompInit_glcp)
- decompModInit -- Remove data no longer used, initialize procinfo data in procinfo%InitAllocate, 
                                   clumps in clumps%Init, use mpi_scan rather than higher order method,
                                   add error checking, and asserts that decomposition is correct,
                                   reset clumps data to -1 for non-local processors to show that we don't need it (shows #3466 can be done)
                                    add some timers
- lnd_set_decomp_and_domain -- add methods to deallocate memory when needed, change timers, 
- Some broad changes: some endrun/shr_sys_abort calls changed to use the file, line format, initialize more pointers to null, 

Changes of note for users:
   - On endrun the 2D indices of the gridpoint are written out for endrun calls that send the point
   - There is some checking that the PE decomposition is created correctly will now abort if it's wrong

Changes of note for developers:
  - set_decomp_info in unittestSubGridMod uses more decompMod procedures
  - Limitations of the decompMod testing are given in test_decompMod.pf

### Specific notes

Contributors other than yourself, if any: John Dennis

   Fixes #3370
   Fixes #3368
   Fixes #3672
   Some work on #3448
   Some work on #3476

Are answers expected to change (and if so in what way)? No (the determination of the PE decomposition is identical as well)

Changes to testing:
  - endrun unit tests -- more tests for write_pt_context with the new write of the global 2D indices
  - decompMod tests -- add more tests especially for the new calculation of 2D grid indices, 
                                         use the set_decomp_info to allocate the PE decompostion,
                                         change test size so ni/nj isn't symmetrical to test that size checks are correct
  - unittestSubGridMod -- add optional add_simple_patch argument to unittest_add_column,
                                            set more decompMod data in set_decomp_info, 
                                            returns after endrun added for unittests
  - many PF unit tests -- call the unittest_subgrid_teardown for tests that setup a PE decomposition

Testing performed, if any: ran standard testing
   The mpi_scan testing branch has had all the test lists run for it: aux_clm, ctsm_sci, decomp_init, decomp_init_uhr, and fates


src/cpl/share_esmf/lnd_set_decomp_and_domain.F90
```diff
@@ -163,7 +164,7 @@ subroutine lnd_set_decomp_and_domain_from_readmesh(driver, vm, meshfile_lnd, mes
     ! Get JUST gridcell processor bounds
     ! Remaining bounds (landunits, columns, patches) will be set after calling decompInit_glcp
     ! so get_proc_bounds is called twice and the gridcell information is just filled in twice
-    call get_proc_bounds(bounds)
+    call get_proc_bounds(bounds, allow_errors=.true.)
```
ekluzek: The new argument should be named only_gridcell, as at this point we can't check anything else because it isn't set.


```suggestion
    call get_proc_bounds(bounds, only_gridcell=.true.)
```

src/cpl/share_esmf/lnd_set_decomp_and_domain.F90
```diff
@@ -331,7 +373,7 @@ subroutine lnd_set_decomp_and_domain_for_single_column(scol_lon, scol_lat, scol_
     call t_stopf ('decompInit_lnd')
 
     ! Initialize processor bounds
-    call get_proc_bounds(bounds)
+    call get_proc_bounds(bounds, allow_errors=.true.) ! allow errors since decomp not fully initialized
```
ekluzek:
```suggestion
    call get_proc_bounds(bounds, only_gridcell=.true.) ! only_gridcell since decomp not fully initialized
```

src/main/clm_initializeMod.F90
```diff
@@ -222,8 +222,8 @@ subroutine initialize2(ni,nj, currtime)
     !-----------------------------------------------------------------------
 
     call t_startf('clm_init2_part1')
-    ! Get processor bounds for gridcells
-    call get_proc_bounds(bounds_proc)
+    ! Get processor bounds for gridcells, just for gridcells
+    call get_proc_bounds(bounds_proc, allow_errors=.true.)  ! Just get proc bounds for gridcells, other variables won't be set until adter decompInit_clumps
```
ekluzek:
```suggestion
    call get_proc_bounds(bounds_proc, only_gridcell=.true.)  ! Just get proc bounds for gridcells, other variables won't be set until after decompInit_clumps
```

src/main/decompInitMod.F90
```diff
@@ -171,77 +178,111 @@ subroutine decompInit_lnd(lni, lnj, amask)
...
-    end do
+       clumps(cid)%endg = clumps(cid)%begg + clumps(cid)%ncells - 1
+       cid_previous = cid
     end do
 
     ! Initialize global gindex (non-compressed, includes ocean points)
     ! Note that gindex_global goes from (1:endg)
-    call get_proc_bounds(bounds)    ! This has to be done after procinfo is finalized
-    call decompInit_lnd_gindex_global_allocate( bounds, ier ) ! This HAS to be done after prcoinfo is finalized
+    call get_proc_bounds(bounds, allow_errors=.true.)    ! This has to be done after procinfo is finalized
```
ekluzek:
```suggestion
    call get_proc_bounds(bounds, only_gridcell=.true.)    ! This has to be done after procinfo is finalized
```

src/main/decompInitMod.F90
```diff
@@ -342,15 +379,33 @@ subroutine decompInit_lnd_gindex_global_allocate( bounds, ier )
             call endrun(msg="allocation error for gindex_global", file=sourcefile, line=__LINE__)
             return
          end if
+         ! TODO: Remove the data, and only use the subroutine to calculate when needed
```
ekluzek: Remove this section and remove storing these bits of data, and only calculate them in endrun.

ekluzek: This is in accordance with this comment

https://github.com/ESCOMP/CTSM/issues/3476#issuecomment-3286811501

src/main/decompInitMod.F90
```diff
@@ -468,8 +523,9 @@ subroutine decompInit_clumps(lni,lnj,glc_behavior)
     character(len=32), parameter :: subname = 'decompInit_clumps'
     !------------------------------------------------------------------------------
 
+    call t_startf('decompInit_clumps')
     !--- assign gridcells to clumps (and thus pes) ---
-    call get_proc_bounds(bounds)
+    call get_proc_bounds(bounds, allow_errors=.true.)
```
ekluzek:
```suggestion
    call get_proc_bounds(bounds, only_gridcell=.true.)
```

src/main/decompInitMod.F90
```diff
@@ -574,25 +630,52 @@ subroutine decompInit_clumps(lni,lnj,glc_behavior)
...
+          call endrun(msg="allvecg error cohorts", file=sourcefile, line=__LINE__)
+          return
+       endif
        endif
     enddo
 
     deallocate(allvecg,allvecl)
     deallocate(lcid)
 
+    ! ------ Reset the clump type array for all non-local cid's to -1 to show it can be made smaller
```
ekluzek, line 578:
```suggestion
    ! ------ Reset the clump type array for all non-local cid's to -1 to show it can be made smaller
    ! TODO: Remove this when https://github.com/ESCOMP/CTSM/issues/3466 is done
```

src/main/decompInitMod.F90
```diff
@@ -171,77 +178,111 @@ subroutine decompInit_lnd(lni, lnj, amask)
...
+             procinfo%ggidx(g) = ln
+             g = g + 1
+          end if
+          end if
+          end if
+       end if
+    end do
+    end do
+
+    ! ---- Get the global index for each gridcell and save the i,j incices for ach gridcell on this processor
```
ekluzek: This section should be removed, and the subroutine only called in write_point_context.

src/main/decompInitMod.F90
```diff
@@ -171,77 +178,111 @@ subroutine decompInit_lnd(lni, lnj, amask)
...
+       if (amask(ln) == 1) then
+          cid = lcid(ln)
+          if ( cid > 0 )then
+          if (clumps(cid)%owner == iam) then
+          if ( procinfo%cid(lc) == cid ) then
+             if ( (g < procinfo%begg) .or. (g > procinfo%endg) )then
+                write(iulog,*) ' iam, g = ', iam, g
+                call endrun(msg='g out of bounds for MPI_SCAN', file=sourcefile, line=__LINE__)
+             end if
+             procinfo%ggidx(g) = ln
```
ekluzek: Remove the setting of ggidx

src/main/decompMod.F90
```diff
@@ -67,7 +68,10 @@ module decompMod
   !---global information on each pe
   type processor_type
      integer :: nclumps              ! number of clumps for processor_type iam
-     integer,pointer :: cid(:)       ! clump indices
+     integer,pointer :: cid(:) => null()  ! clump indices
+     integer,pointer :: ggidx(:) => null()   ! global vector index on the full 2D grid
```
ekluzek: Remove ggidx, gi, and gj.

src/main/decompMod.F90
```diff
@@ -121,6 +128,106 @@ module decompMod
...
+    ! Get the global i/j indices from the global vector grid index
+    ! !ARGUMENTS:
+    class(processor_type), intent(in) :: this
+    integer, intent(in) :: g ! gridcell index on this processor
+    integer, intent(in) :: ni, nj ! Global 2D size of full grid
+    integer, intent(out) :: i, j ! 2D indices in x and y on the full global 2D grid (j will be 1 for an unstructured grid)
+
+    integer :: global_index
+
+    if ( .not. associated(this%ggidx) )then
```
ekluzek: This will need to be handled differently

src/main/decompMod.F90
```diff
@@ -121,6 +128,106 @@ module decompMod
 
 contains
 
+  !-----------------------------------------------------------------------
+  function calc_global_index_fromij( this, g, ni, nj ) result(global_index)
+    ! Returns the full grid global vector index from the gridcell on this processor
```
ekluzek, line 175: This will need to be handled differently to not rely on the global i/j data.

src/main/decompInitMod.F90
ekluzek: This is complex enough that it should be done, in a different PR.

src/main/decompInitMod.F90
ekluzek: Although see below.

src/main/decompMod.F90
```diff
@@ -239,19 +356,44 @@ subroutine get_clump_bounds (n, bounds)
...
+    end if
+    if ( bounds%endl <= 0 )then
+       call shr_sys_abort( 'bounds%endl is not valid', file=sourcefile, line=__LINE__)
+       return
+    end if
+    if ( bounds%endg <= 0 )then
+       call shr_sys_abort( 'bounds%endg is not valid', file=sourcefile, line=__LINE__)
+       return
+    end if
+    ! End Cohort isn't necessarily valid, so don't do this error check
```
ekluzek: This should check if use_fates is TRUE

src/main/decompMod.F90
```diff
@@ -239,19 +356,44 @@ subroutine get_clump_bounds (n, bounds)
...
+    !end if
+
     bounds%level = bounds_level_clump
     bounds%clump_index = n
 
   end subroutine get_clump_bounds
 
   !------------------------------------------------------------------------------
-  subroutine get_proc_bounds (bounds, allow_call_from_threaded_region)
+  subroutine get_proc_bounds (bounds, allow_call_from_threaded_region, allow_errors)
```
ekluzek:
```suggestion
  subroutine get_proc_bounds (bounds, allow_call_from_threaded_region, only_gridcell)
```

src/main/decompMod.F90
```diff
@@ -239,19 +356,44 @@ subroutine get_clump_bounds (n, bounds)
...
     !
     ! !DESCRIPTION:
     ! Retrieve processor bounds
     !
     ! !ARGUMENTS:
     type(bounds_type), intent(out) :: bounds ! processor bounds bounds
+    logical, intent(in), optional :: allow_errors ! Don't do the normal error checking
```
ekluzek:
```suggestion
    logical, intent(in), optional :: only_gridcell ! Only return the gridcell bounds, other subgrid info assumed to not be set yet
```

src/main/decompMod.F90
```diff
@@ -292,6 +435,35 @@ subroutine get_proc_bounds (bounds, allow_call_from_threaded_region)
...
+    ! Always check that gridcells are set
+    if ( bounds%endg <= 0 )then
+       call shr_sys_abort( 'bounds%endg is not valid', file=sourcefile, line=__LINE__)
+       return
+    end if
+
+    ! Exit before checking if errors should be allowed
+    if ( present(allow_errors) ) then
+      if ( allow_errors ) return
+    end if
```
ekluzek:
```suggestion
    if ( present(only_gridcell) ) then
      if ( only_gridcell ) return
    end if
```

src/main/decompMod.F90
```diff
@@ -583,6 +756,18 @@ subroutine decompmod_clean()
     if ( allocated(clumps) )then
       deallocate(clumps)
     end if
+    if ( associated(procinfo%ggidx) )then
```
ekluzek: This is another group to remove and only do the calls to figure this out in endrun.

src/main/decompMod.F90
ekluzek: I don't think this subroutine is actually used.

src/main/decompMod.F90
```diff
@@ -121,6 +128,106 @@ module decompMod
...
+    end if
+    if ( (j < 1) .or. (j > nj) ) then
+       call shr_sys_abort( 'Computed global j value out of range', file=sourcefile, line=__LINE__)
+       return
+    end if
+  end subroutine calc_ijindices_from_full_global_index
+
+  !-----------------------------------------------------------------------
+  subroutine calc_globalxy_indices( this, g, ni, nj, i, j )
+    ! Get the global i/j indices from the global vector grid index
```
ekluzek, line 241: This really should be added to unit-testing.

ekluzek (commented):
There's a couple of things that should be done here before merging.

- [x] Change allow_errors argument to only_gridcell
- [x] Put the check for cohorts inside a use_fates if and uncomment
- [x] Add showing the grid indices ggidx, gi, and gj to endrun_point_context
- [x] Add the changes in mpi_scan for Assertions and buildnml in here (will be needed to build)
- [ ] Create an issue about removing the procinfo%ggidx, %gi and %gj

Things that could be added into the last one about the issue are:
- [ ] Add the %gi and %gj to unit testing (new unit testing would need to come in)
- [ ] Check %gi and %gj in the abortutils unit testing
- [ ] Add error check testing for decompInit into a new tester
- [ ] Remove global memory for %gi and %gj and modify the subroutines that figure it out
- [ ] Make local call to figure out i/j indices only from endrun

src/main/decompInitMod.F90
```diff
@@ -52,6 +53,7 @@ subroutine decompInit_lnd(lni, lnj, amask)
     use clm_varctl , only : nsegspc
     use decompMod  , only : gindex_global, nclumps, clumps
     use decompMod  , only : bounds_type, get_proc_bounds, procinfo
+    use Assertions, only : assert_equal
```
ekluzek: Rather than doing this with Assertions for self_test, which requires updates to Assertions and buildnml. I could use shr_assert_mod code instead. To always do it, you just need to use the `call shr_assert` form.

ekluzek: I went ahead with that method and the code is working.

ekluzek:
There's two single point cases that fail when I ran aux_clm on Izumi:

SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60FatesCrujraRs.izumi_nag.clm-FatesCold        (RUN)
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60FatesRs.izumi_nag.clm-FatesCold      (RUN)

It has to do with the deallocation of memory for the readmesh. Which probably has a difference in handling for single point verses global grids.

ekluzek:
The unit-tests are failing because I added error checking to get_proc_bounds. Here's how the code goes

unittestSubGridMod.F90

set_decomp_info subroutine does the following:

``` fortran
   procinfo%begg = begg
    procinfo%endg = gi
    procinfo%begl = begl
    procinfo%endl = li
    procinfo%begc = begc
    procinfo%endc = ci
    procinfo%begp = begp
    procinfo%endp = pi
```

which at this point pi for example is set to 0. So get_proc_bounds fails. Like this...

```
Failure
 in: 
test_subgridWeights_suite.test_is_gcell_all_ltypeX_true
  Location: 
[<unknown location>]
ABORTED: bounds%endp is not valid
```

I think at this point I could also create decomp from running decompInit_lnd, as an alternative, but I need to understand what the unittestSubGridMod does with the decomp.

This works for the unit testing cases, because it doesn't use the end* indices from bounds -- it uses the end* indices saved in the module. I tried setting them to the module values -- but that also failed for me see below.

```
 Start: <test_set_mindist_suite.multipleTypes_findsSameType>
. clm_ptrs_compdown ERROR: pcolumn            2       -9999           1
          50
iam = 0: local  patch    index = 2
iam = 0: global patch    index = 0
iam = 0: global column   index = -1
iam = 0: global landunit index = -1
iam = 0: global gridcell index = -1
forrtl: severe (408): fort: (3): Subscript #1 of the array LONDEG has value -9999 which is less than the lower bound of 1

initInterpMindist  000000000051A161  abortutils_mp_wri         234  abortutils.F90
initInterpMindist  0000000000516024  abortutils_mp_end          92  abortutils.F90
initInterpMindist  0000000000530A49  initsubgridmod_mp          95  initSubgridMod.F90
initInterpMindist  000000000048E76C  unittestsubgridmo         165  unittestSubgridMod.F90
initInterpMindist  000000000048CEBB  unittestsimplesub         137  unittestSimpleSubgridSetupsMod.F90
```

ekluzek:
> The unit-tests are failing because I added error checking to get_proc_bounds. Here's how the code goes
> 
> unittestSubGridMod.F90
> 
> set_decomp_info subroutine does the following:
> 
.
.
.
> ```

I talked this over with @billsacks. We realized that the setting of the decomp is OK like it is. It's just that unit tests don't always create the subgrid hierarchy beneath the level it needs. So one way to fix it would be to make sure that happens for the all the cases where it might need it. Another option is to just add the optional argument of "only_grid" to bypass the subgrid level checking in the unittestSubGridMod get_proc_bounds call.

I tried doing the latter and there's still one unit test that failed because there is a call to get_proc_bounds in the setup of the accumulators, which is in the main model code rather than just the unit test code. Note, as an aside this really should be getting the bounds from the higher level where you know what type of loop you are in.

So for this one case, I'll make sure the lower level hierarchy is setup by doing this sort of thing:

``` fortran
diff --git a/src/dyn_subgrid/test/dynInitColumns_test/test_init_columns.pf b/src/dyn_subgrid/test/dynInitColumns_test/test_init_columns.pf
index ade2e6d95..79ec506f9 100644
--- a/src/dyn_subgrid/test/dynInitColumns_test/test_init_columns.pf
+++ b/src/dyn_subgrid/test/dynInitColumns_test/test_init_columns.pf
@@ -66,24 +66,24 @@ contains
 
     ! The first landunit is neither natural veg nor crop
     call unittest_add_landunit(my_gi=gi, ltype=istwet, wtgcell=0.25_r8)
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.5_r8)
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.5_r8)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.5_r8, add_simple_patch=.true.)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.5_r8, add_simple_patch=.true.)
 
     call unittest_add_landunit(my_gi=gi, ltype=1, wtgcell=0.5_r8)
     this%l1 = li
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8)
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8, add_simple_patch=.true.)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8, add_simple_patch=.true.)
     ! This column (the second column on the landunit with ltype=1) will be the target for
     ! some tests of initialization of a new column
     this%c_new = ci
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8)
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8, add_simple_patch=.true.)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8, add_simple_patch=.true.)
 
     call unittest_add_landunit(my_gi=gi, ltype=2, wtgcell=0.25_r8)
     this%l2 = li
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8)
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8)
-    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.5_r8)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8, add_simple_patch=.true.)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.25_r8, add_simple_patch=.true.)
+    call unittest_add_column(my_li=li, ctype=1, wtlunit=0.5_r8, add_simple_patch=.true.)
 
     call unittest_subgrid_setup_end()
 
```

The new optional "add_simple_patch" argument means add a single simple patch below the column created. This could be done in other unittests to make sure the patch level is set when columns are added. The same sort of thing could be done for landunit and grid level as well.

samsrabin:
@ekluzek I see that this is still a draft and has several comments from you in it, including several suggested changes. I'll unsubscribe for now; please @ me when it's fully ready (or now, if it actually is).

ekluzek:
@samsrabin ok that makes sense. The one thing that I wanted input on is if all of the conversations I have need to be handled before we meet to go over it.

But I really think the answer is yes on that, so I'll try to complete them and see if we could meet next week to go over it. If something ends up taking too long I'll make sure we meet so we can discuss.

Thanks for the feedback.

src/main/decompMod.F90
ekluzek: This was fixed.

ekluzek:
I remember where I got stuck with this. I was making changes to comply with what we decided in #3476, and got stuck in some of the unit testing for that.

src/main/decompMod.F90
```diff
@@ -114,13 +123,126 @@ module decompMod
...
+    if ( (nglob_x < 1) .or. (nglob_y < 1) ) then
+       !write(iulog,*) 'WARNING: Global gridsize nglob_x/nglob_y is not set'
+       return
+    end if
+    if ( (this%gi(g) < 1) .or. (this%gi(g) > nglob_x) ) then
+       !write(iulog,*) 'this%gi(g) = ', this%gi(g)
+       !write(iulog,*) 'WARNING: Global gi index is out of bounds'
+       return
+    end if
+    if ( (this%gj(g) < 1) .or. (this%gj(g) > nglob_x) ) then
```
ekluzek: @briandobbins noticed a correction here. The above should be nglob_y

src/main/decompInitMod.F90
```diff
@@ -289,24 +335,21 @@ subroutine decompInit_lnd_allocate( ier )
             call endrun(msg="nclumps is NOT set before allocation", file=sourcefile, line=__LINE__)
             return
          end if
+         ! TODO: This will be moved to the other allocate and for a smaller size ----
```
ekluzek: Add the issue number.

samsrabin: Issue ESCOMP/CTSM#3466

samsrabin:
Notes from meeting today with @ekluzek and @briandobbins:
- Brian would like to get [this](https://github.com/ESCOMP/CTSM/issues/3466#issuecomment-4085126491) done soon. Erik will evaluate the required work and post there; Brian wants to help.
- As far as unit testing for this PR: Erik needs to fix the unit test that's currently failing, then add one covering [this bug](https://github.com/ESCOMP/CTSM/pull/3666/changes#r2955683826) (and any others that Brian posts). It should fail, then Erik needs to fix the bug and have it pass. Then any other unit testing will be up to Erik's judgment, balancing that effort with other work on his plate.

src/main/decompInitMod.F90
ekluzek: In our meeting with @samsrabin and @briandobbins we thought doing this would be worthwhile. But, we also want to get something in, by end of April so can be used for upcoming ultra high resolution simulations.

src/main/decompMod.F90
ekluzek: Add a unit-test that would detect this problem

src/main/decompMod.F90
ekluzek: Go through unit tests and make sure coverage is reasonable.

ekluzek:
> Notes from meeting today with @ekluzek and @briandobbins:
> 
> * Brian would like to get [this](https://github.com/ESCOMP/CTSM/issues/3466#issuecomment-4085126491) done soon. Erik will evaluate the required work and post there; Brian wants to help.

We later realized that this isn't critical to do. Since, the number of clumps is basically the number of processors, and even though that's on every task -- the memory for it isn't as big as for example the number of land grid cells.

So we can let #3466 happen when there's time for it.

ekluzek:
#3931 had an impact here and is the reason behind the unit tests not working. Taking this into account should allow me to make progress again.

src/main/decompMod.F90
ekluzek: It's not currently used. But, now I added a unit test for it, as it might need to be used with the changes needed.

src/main/decompMod.F90
ekluzek: OK I added a unit test that detects the problem and fails without the fix.

src/main/decompMod.F90
ekluzek: I do have unit tests for both of these now.

src/main/test/abortutils_test/test_abortutils.pf
```diff
@@ -1,6 +1,18 @@
...
 
   ! Tests of abortutils
+  !
+  ! IMPORTANT NOTES: Limitations of these tests
+  !                  These tests cover the different options to endrun including failure modes.
+  !                  It does NOT verify the log output of write_point_context however.
+  !                  The test_decompMod.pf unit tests test the 2D calculation of global indices however.
+  !                  Other unit tests, exercise endrun calls in various ways so test actual endrun calls.
+  !
+  !                  WORK THAT COULD BE DONE:
```
ekluzek, line 11: @briandobbins and @samsrabin this is ready to review now. This is the one thing that I wonder about doing before it's merged in. I think the test covered is pretty reasonable between the PF unit tests and the system tests. Co-pilot also agrees with that sentiment. I have unit tests that cover what the AI found when @briandobbins asked for a review. And there was another similar error in a test that Co-Pilot found for me.

I could add some testing to verify the point_context log output. The catch there is that I'd need to figure out how to capture log output in our unit testing. There must be ways of doing this, but I'm not familiar with it, so would need to investigate. I could also just do a little bit more in this area and limit the time I spend on doing it. Spending a few hours might be reasonable, but not spending many days. But, I figured I'd ask both your opinion on that before just doing it.

Outside of that, this is good to go. I'm running system tests and barring problems there I'd like to see it merged to b4b-dev.

Let me know if you spot anything else I should pay attention to. Thanks in advance.

ekluzek:
Running aux_clm on Derecho and Izumi. Will also run decomp_init and uhr_decomp_init before merging in.

ekluzek:
A few tests fail for odd grids/PE layouts that I'll need to look into:

ERS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm50FatesRs.derecho_gnu.clm-FatesCold    (RUN)
FSURDATMODIFYCTSM_D_Mmpi-serial_Ld1.5x5_amazon.I2000Clm50SpRs.derecho_gnu       (RUN)
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60Bgc.derecho_gnu.clm-HillslopeC       (RUN)
SMS_Ld1_PS.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldas.derecho_gnu.clm-default--clm-nofireemis (RUN)
SMS_Ld1_PS.nldas2_rnldas2_mnldas2.I2000Ctsm50NwpSpNldasRs.derecho_gnu.clm-default--clm-nofireemis       (RUN)
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60FatesCrujraRs.izumi_nag.clm-FatesCold        (NLCOMP RUN)
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60FatesRs.izumi_nag.clm-FatesCold      (NLCOMP RUN)
The 25 point amazon grid for mpi-serial, and the NLDAS 104k point grid on one node

ekluzek:
0.8 for Sprint 31

samsrabin:
@ekluzek Please ping me again to review once you have those tests working.

src/cpl/share_esmf/lnd_set_decomp_and_domain.F90
```diff
@@ -255,13 +258,52 @@ subroutine lnd_set_decomp_and_domain_from_readmesh(driver, vm, meshfile_lnd, mes
...
+       logical :: no_esmf_garbage = .true. ! If .true. release all ESMF data (which can be problematic if referenced again)
+
+       rc = ESMF_SUCCESS
+
+       if ( associated(lndfrac_loc_input) ) deallocate(lndfrac_loc_input)
+       deallocate(gindex_lnd)
+       deallocate(gindex_ocn)
+       deallocate(gindex_ctsm)
+       ! Destroy or release all of the ESMF objects
+       call ESMF_FieldRedistRelease( rhandle_lnd2ctsm, noGarbage=no_esmf_garbage, rc=rc)
```
ekluzek, line 285: It looks like the problem with the failing tests is that some things are being deallocated for cases where they aren't necessarily created first. So this deallocate method either needs to only be called for cases that need it, or it has to check that things were allocated first.

This won't be hard to do. But, will take a little fiddling around...

ekluzek: OK I seperated the dealloc for ESMF objects from the rest and now it's working for the cases that failed before.

ekluzek:
OK, all aux_clm tests are working now. The decomp_init testlists were already working, but I might send other testlists just to make sure there isn't some other strange configuration that has problems.

@samsrabin

ekluzek:
fates testlist ran as expected on both Izumi and Derecho. The ctsm_sci test list on Derecho has only three tests left to finish, but outside of those everything ran as expected.

src/main/test/accumul_test/test_accumul.pf
```diff
@@ -967,7 +991,8 @@ contains
     ! Verify
     reset_output_int = merge(1, 0, reset_output)
     @assertEqual(expected, reset_output_int)
-end subroutine markreset_allpoints_1level
+    call unittest_subgrid_teardown()
+  end subroutine markreset_allpoints_1level
```
samsrabin: Remove leading whitespace:
```suggestion
end subroutine markreset_allpoints_1level
```

samsrabin:
@ekluzek I think I'm going to need you to walk me through this. Your PR title and description mainly just talk about moving work from some other git branch to here. What does this PR actually _do_?

ekluzek:
> @ekluzek I think I'm going to need you to walk me through this. Your PR title and description mainly just talk about moving work from some other git branch to here. What does this PR actually _do_?

OK, let me update the PR description to go over what it actually does. And I'll try to lay it out in a way that doesn't require a lot of insider technical knowledge. I'd kind of like to have this in before b4b-dev on next Thursday, so maybe I'll schedule a quick code review for Wednesday. I think I'm up for that, but I also might have to change last minute with what's going on in the family. And if it has to be the next cycle that wouldn't be the worst thing to happen either.

src/main/decompInitMod.F90
```diff
@@ -1,9 +1,11 @@
-module decompInitMod
+ module decompInitMod
```
ekluzek: There's an extra space before module that got in here.

ekluzek:
OK @briandobbins and @samsrabin I've updated the description at the top so it tells more about what the code changes mean. Let me know if anything is unclear or missing and I'll fix it. We'll meet next week to finalize anything left to do, and hopefully it can come in pretty quickly after that.

samsrabin (approved):
@ekluzek walked @briandobbins and me through the changes, and it all sounds great! We didn't go through the actual CODE changes, but I don't think it would be helpful for me to do so anyway. Approving.

briandobbins (approved):
Thanks for the walkthrough; given the passing tests, and the higher degree of understanding, I'm approving without walking through all the code changes.  Glad to have this in, and thanks for the hard work on it!

ekluzek:
I met yesterday with @samsrabin and @briandobbins and they approved this coming in. I'll update and rerun the testing again and then bring it in. We did think it's important to run all the testing I did previously, especially the ctsm_sci testlist as that will run the CAM VR grids.

ekluzek:
I met yesterday with @samsrabin and @briandobbins and they approved this coming in. I'll update and rerun the testing again and then bring it in. We did think it's important to run all the testing I did previously, especially the ctsm_sci testlist as that will run the CAM VR grids.

ekluzek:
The aux_clm testing ran as expected. So I'm running these testlist now: ctsm_sci, decomp_init, uhr_decomp_init. If that all goes as expected I'll do the merge.

ekluzek:
The ctsm_sci testlist ran as expected (with one test still in the queue). The decomp_init test all passed as expected. The uhr_decomp_init test had one test run as expected, one that's in the queue and this one failed:

SMS_Ln1_PL.mpasa3p75_mpasa3p75_mt13.I2000Clm45Sp.derecho_intel.clm-for_testing_fastsetup_bypassrun--clm-mpasa3p75       ( RUN)

It ran out of the hour wallclock time it was given. This is concerning because based on previous timings an hour should have been overly generous. From looking at log files it looks like it failed in datm and likely hung so it didn't die until it reached the wallclock limit.

ekluzek:
OK, I documented more on that final failing test and marked it as an expected fail. I also added the default PE layout mpas3p75 grid test to the ctsm_sci testlist so that we test at least the basic setup for km scale runs more often. It uses a lot of nodes, but only 15 minutes of wallclock so it isn't that onorous.

PR #3670, opened 2025-12-19
Update the What's New ctsm5.4 file

ekluzek:
### Description of changes

Some corrections to the WhatsNewinCTSM5.4 document regarding Carbon Isotopes

I also updated a few README files and converted them to markdown format to make them easier to read.

### Specific notes

Contributors other than yourself, if any: @slevis-lmwg and @billsacks 

CTSM Issues Fixed (include github issue #):
  Work on #2761

Are answers expected to change (and if so in what way)? No

Any User Interface Changes (namelist or namelist defaults changes)? No

Does this create a need to change or add documentation? Did you do so? No No

Testing performed, if any: No, will do namelist testing because of the documentation changes in namelist_definition

PR #3622, opened 2025-11-23
compare_paramfiles tool

samsrabin:
### Description of changes

Adds a utility to compare parameter files. Compares both "raw" values and values after masking and scaling (applying `_FillValue` to convert missing values to NaN, multiplying by `scale_factor`, and adding `add_offset`).

Example usage; results truncated:

```shell
file0=/glade/campaign/cesm/cesmdata/cseg/inputdata/lnd/clm2/paramdata/clm_params78pftModDates.c130821.nc
file1=/glade/campaign/cesm/cesmdata/cseg/inputdata/lnd/clm2/paramdata/ctsm60_params.c251124.nc

tools/param_utils/compare_paramfiles $file0 $file1
```

```
File 0: /glade/campaign/cesm/cesmdata/cseg/inputdata/lnd/clm2/paramdata/clm_params78pftModDates.c130821.nc
File 1: /glade/campaign/cesm/cesmdata/cseg/inputdata/lnd/clm2/paramdata/ctsm60_params.c251124.nc

Variable(s) present in File 0 but not File 1:
   cn_s1_bgc
   cn_s2_bgc
   ...

Variable(s) present in File 1 but not File 0:
   C2_liq_Brun89
   FUN_fracfixers
   ...

...

allconss:
   Values differ:
      [pft 71 (miscanthus)] 0.0 → 2.0 (raw)
                            nan → 2.0 (masked/scaled)
      [pft 72 (irrigated_miscanthus)] 0.0 → 2.0 (raw)
                                      nan → 2.0 (masked/scaled)
      [pft 73 (switchgrass)] 0.0 → 2.0 (raw)
                             nan → 2.0 (masked/scaled)
      [pft 74 (irrigated_switchgrass)] 0.0 → 2.0 (raw)
                                       nan → 2.0 (masked/scaled)

...

atmch4:
   Attribute(s) present in File 1 but not File 0:
       _FillValue: nan
   Dimension names differ: File 0: ['allpfts']
                           File 1: []

baset:
   Attribute(s) present in File 1 but not File 0:
       _FillValue: nan
   Values differ:
      [pft 71 (miscanthus)] 0.0 → 8.0
      [pft 72 (irrigated_miscanthus)] 0.0 → 8.0
      [pft 73 (switchgrass)] 0.0 → 8.0
      [pft 74 (irrigated_switchgrass)] 0.0 → 8.0

...
```

### Specific notes

**Contributors other than yourself, if any:** None

**CTSM Issues Fixed (include github issue #):** None

[...]

PR #3113, opened 2025-05-07
Minor fixes to names and units of canopy turbulence variables

mpaiao:
### Description of changes

I was adding a few diagnostic variables to ELM based on the CLM code (E3SM-Project/E3SM#7328), and in the process I found a few typos and a potential bug in one of the diagnostic variables.

1. The description and the units of variable `ustar` were incorrect (file `src/biogeophys/FrictionVelocityMod.F90`).
2. The units listed for variable `'VPD'` in the call to sub-routine `hist_addfld1d` (file `src/biogeophys/FrictionVelocityMod.F90`) is Pa, however the calculation of VPD in file `src/biogeophys/CanopyFluxesMod.F90` had a 0.001_r8 multiplication factor, which I think implies conversion to kPa.
3. Replaced the "Monin-Obukhov length" occurrences in comments and output name with "Obukhov length scale", which is consistent with Bonan (2015) and the [American Meteorological Society](https://glossary.ametsoc.org/wiki/Obukhov_length).

### Specific notes

Contributors other than yourself, if any:
@olyson

CTSM Issues Fixed (include github issue #): 
Fixes #3112 that has the proposed fixes

Are answers expected to change (and if so in what way)? 
This should be bit-for-bit, only metadata and code comments updated.

Any User Interface Changes (namelist or namelist defaults changes)?
No

Does this create a need to change or add documentation? Did you do so?
No

Testing performed, if any: will run regular testing

Issue #2761, opened 2024-09-16
Transition most text README files to markdown or restructured text format so will be visualized in github/VS code


ekluzek:
Many of our README files are in raw text format. Since github will visualize markdown or restructured text, we should update them to one of those formats. Which will make them easier to read on either github or VS code, but still readable in raw text format. This will make URL's proper links and issue or PR's links as well.

wwieder:
let's do this in markdown

slevis-lmwg:
For Sprint accounting only: I marked the issue done and will continue to track in the PR.

slevis-lmwg:
Reopening until this makes it to master...

PR #3673, opened 2025-12-20
Merge b4bdev 20251219

### LINKED ISSUE #29: add namelist option to turn on all history fields, add a CLM test that uses this

### LINKED ISSUE #3112: Proposed minor updates to canopy biogeophysics diagnostic variables
I was going through the code to incorporate a few additional output variables in ELM based on the CLM code (E3SM-Project/E3SM#7328), and in the process I found the following: 

1. The units and description of variable `USTAR` are incorrect. I guess the description should be friction velocity, and the units should be m/s.

https://github.com/ESCOMP/CTSM/blob/e0104b91fef35b06aa1458b3023a7d83f568ad03/src/biogeophys/FrictionVelocityMod.F90#L286-L288

2. A few lines later, variable `VPD` seems to be reported in Pa:

https://github.com/ESCOMP/CTSM/blob/e0104b91fef35b06aa1458b3023a7d83f568ad03/src/biogeophys/FrictionVelocityMod.F90#L314-L316

But the actual calculation has a factor of `0.001_r8`, which seems to be a conversion factor to kPa.

https://github.com/ESCOMP/CTSM/blob/e0104b91fef35b06aa1458b3023a7d83f568ad03/src/biogeophys/CanopyFluxesMod.F90#L1103-L1111

3. This one is just a very minor detail, but the term "Monin-Obukhov length" is incorrect, it should be "Obukhov length" according to the [American Meteorological Society](https://glossary.ametsoc.org/wiki/Obukhov_length).

I have a commit ready for pull request that addresses all these points, but I am first opening an issue to see if there are any concerns with my suggested changes.

### LINKED ISSUE #3284: ERI tests fail for transient cases
### Brief summary of bug

@jedwards found this in testing a FHIST compset with an ERI test.

### General bug information

**CTSM version you are using: ?

**Configurations affected:**

ERI tests with HIST compsets.

### Details of bug

There's a discussion in CIME that this is maybe due to how ERI tests work, and that transient cases should be disallowed for this.

```
ERROR: Current model year does not match year on initial conditions file (finid
 at)
 Current year:         2014
 Year on initial conditions file:         2012
  
 This match is a requirement when both:
 (a) The current run is a transient run, and
 (b) The initial conditions file was generated from a transient run
  
 Possible solutions to this problem:
 (1) Make sure RUN_STARTDATE is set correctly
 (2) Make sure you are using the correct initial conditions file (finidat)
 (3) If you are confident that you are using the correct start date and initial 
 conditions file,
     yet are still experiencing this error, then you can bypass this check by se
 tting:
       check_finidat_year_consistency = .false.
     in user_nl_clm
  
 ENDRUN:
 ERROR in restFileMod.F90 at line 1039
```

ekluzek:
### Description of changes

Bring in b4b-dev to master

Some documentation updates: Update the ctsm5.4 What's New file with text around the Carbon isotopes
Fixes to names and units of canopy turbulence variables. 

New CTSM Tool: compare_parmfiles to nicely compare CTSM NetCDF parameter files in a clear way

New namelist option: hist_all_fields == TRUE flags to output ALL available history fields by default

Test-mod updates: Add a test-mod so that transient ERI tests work

#3670 Update ctm5.4 What's new and some README files
#3622 compare_parmfiles tool
#3113 Fixes to names and units of canopy turbulence vars
#2061 Add hist_all_fields to output ALL history fields by default
#3645 Have mosart history output daily so that ERI transient tests work
#3299 Testmod that allows transient ERI tests to work

### Specific notes

  @slevis-lmwg @samsrabin @mpaiao @johnpaulalex @billsacks 

  Work on #2761 change README files to markdown
   Fixes #3112 Updates to names and units of canopy turbulence vars
   Fixes #29 Namelist option to run on all history fields, testmod for it
   Fixes #3284 ERI tests fail for transient cases

   New namelist item hist_all_fields

   New compare_paramfiles tool to compare two parameter files

Does this create a need to change or add documentation? Did you do so? No But, documentation updates did happen

Testing performed, if any: regular

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.4.006
 Originator(s): erik (Erik Kluzek)
-Date: Wed Jan  7 04:09:34 AM MST 2026
+Date: Wed Jan  7 02:49:18 PM MST 2026
 One-line Summary: Merge b4b-dev to master
 
 Purpose and description of changes


ekluzek:
Testing is almost done, and looks as expected so far. Finished on Derecho, a couple tests running on Izumi.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -2125,6 +2125,15 @@
       <option name="comment">Among other extra outputs, ensure that writing the list of all history fields to a separate file does not cause failure</option>
     </options>
   </test>
+  <test name="ERP_P36x2_D_Ld3" grid="f10_f10_mg37" compset="I1850Clm50BgcCrop" testmods="clm/default--clm/all_outputs">
```
ekluzek: This test is for 36x2 processors, so should be updated for Derecho processors. Also should be Clm60 so that excess-ice and Carbon isotopes are turned on. And also drydep and megan should be turned on. Also Hist as I think some history fields are only activated for transient cases.

To cover more fields we probably need one with Sp mode as well. So the testmod is run with: Fates, BgcCrop and Sp.

ekluzek: Also we should turn the namelist item on for the SaveHistFieldList tests to make sure all history fields are turned on for those tests.

ekluzek: Also add to the extra_outputs test mod.

ekluzek: So I did the above things that apply.

I did NOT add a Sp test though.
And I didn't add drydep to it

ekluzek: So @slevis-lmwg my inclination is to bring this in as is, and maybe make an issue about those last things that could be done. Just asking if you have thoughts on that idea?

slevis-lmwg: @ekluzek thank you for all this. Yes, I am comfortable with bringing in as is, because it already increases our test coverage. You could open an issue about additional tests if you like. I think I would have expected SP to output a subset of fields relative to BGC (I could be wrong).

doc/source/users_guide/setting-up-and-running-a-case/README_history_fields_files
```diff
@@ -4,8 +4,8 @@
 The files history_fields_nofates.rst and history_fields_fates.rst each contain a
 table of the history fields, active and inactive, available in the CTSM cases
 that get generated by these tests:
-SMS_Ld1.f10_f10_mg37.I1850Clm50BgcCrop.cheyenne_intel.clm-SaveHistFieldList
-SMS_Ld1.f10_f10_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-SaveHistFieldList
+SMS_Ld1.f10_f10_mt232.IHistClm60BgcCropCrujra<.derecho_intel.clm-defaults--clm-SaveHistFieldList
```
slevis-lmwg: Is the "<" in this line a typo most likely?

slevis-lmwg (approved):
Erik, thank you for the clean-up work that you ended up including here. I pointed out what seems like a typo in doc/source/users_guide/setting-up-and-running-a-case/README_history_fields_files

doc/source/users_guide/setting-up-and-running-a-case/README_history_fields_files
ekluzek: Yep! Good catch!

cime_config/testdefs/testlist_clm.xml
ekluzek: OK, undewr #3681

doc/source/users_guide/setting-up-and-running-a-case/README_history_fields_files
ekluzek: Fixed.

Issue #3679, opened 2026-01-07
placeholder issue soil water movement


wwieder:
In an email exchange @dlawrenncar asked about:
_"This [paper](https://scholar.google.com/scholar_url?url=https://iopscience.iop.org/article/10.1088/1748-9326/ae2f72/pdf&hl=en&sa=X&d=12047054256090494578&ei=4QRKabn4O56u6rQPotyMwAQ&scisig=ALhkC2QmG9zgdgzr8PTiZnQvk9jw&oi=scholaralrt&hist=0HohESgAAAAJ:17650382304653952560:ALhkC2SNSKgY4jSk8MZLjE8vFvr6&html=&pos=1&folt=rel&fols=) popped up in my feed.  It looks like there may be some issues with our calculation of the 10cm soil moisture and an issue with CMORization of soil bounds.  Maybe we can collectively take a look at this after the holidays.  The CMORization issue, if true, is something we should resolve.  I wonder if the remaining discrepancy at high latitudes has anything to do with ponded water?"_

@olyson responded:
_"Regarding the calculation of 10cm soil moisture being too early in the model computation, it looks like this was addressed in this issue, presumably too late for some (all?) of the CMIP6 simulations. see #932"_

@swensosc noted:
_"Under near-saturated conditions and esp. when high organic amounts are present, this oversaturation is common.  (I believe this is analogous to a CFL condition, where the velocities are too fast for the grid spacing, and the water travels deeper than it should).  When the water table is subsequently calculated, it is based on this distribution of water, and tends to be deeper than it should.  Later the soil moisture is corrected, but presumably it is then somewhat inconsistent with the water table value.  In my working branch, I put a correction in SoilWaterMovement to fix this issue immediately after it occurs, so that the water table is correctly diagnosed later. "_

To close this issue:

- [ ] @olyson will work with @jedwards4b to make sure we're CMORIZING history variables correctly for CMIP7 (suggested CESM3 release milestone)
- [ ] @swensosc will open a PR that changes the order of soil state variable calculations so that soil moisture and water table depth are consistent (suggested beta08 milestone?)

ekluzek:
Simulations for the second checkbox above are detailed here:

https://github.com/NCAR/LMWG_dev/issues/132

olyson:
[...]

PR #3683, opened 2026-01-08
ctsm5.4.008: Oversaturation

swensosc:
### Description of changes
Add a correction for oversaturated soil layers immediately after the soil water flux calculation in SoilWaterMovementMod.  

### Specific notes

 Fixes second checkbox in #3679 

Are answers expected to change (and if so in what way)? Yes climate should be similar, but > roundoff
  Simulations and diagnostics are reported on here:
   https://github.com/NCAR/LMWG_dev/issues/132

Does this create a need to change or add documentation? Did you do so? No?

Testing performed, if any: will run regular

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.4.008
 Originator(s): swensosc (Sean Swenson)
-Date: Mon Jan 12 05:27:29 PM MST 2026
+Date: Mon Jan 12 11:19:38 PM MST 2026
 One-line Summary: Add a correction for oversaturated soil layers in SoilWaterMovement
 
 Purpose and description of changes
@@ -45,21 +45,12 @@ Testing summary: regular
 ----------------
  [PASS means all tests PASS; OK means tests PASS other than expected fails.]
 
-  build-namelist tests (if CLMBuildNamelist.pm has changed):
-
-    derecho - 
-
-  python testing (if python code has changed; see instructions in python/README.md; document testing done):
-
-    derecho - 
-
   regular tests (aux_clm: https://github.com/ESCOMP/CTSM/wiki/System-Testing-Guide#pre-merge-system-testing):
 
-    derecho -----
-    izumi -------
+    derecho ----- OK
+    izumi ------- OK
 
 If the tag used for baseline comparisons was NOT the previous tag, note that here:
-
 
 Answer changes
 --------------


wwieder:
Thanks for opening this PR, Sean.  Have you done any scientific tests of the code or can you describe expected behavior / answer changes?    Would a short SP run be enough to quantify differences?  Are we safer doing a BGC run to see if there are C cycle changes?  

Fixes part two listed in #3679

samsrabin:
CTSM SE meeting 2026-01-08: @wwieder will do some scientific testing.

src/biogeophys/SoilWaterMovementMod.F90
```diff
@@ -1403,6 +1404,13 @@ subroutine soilwater_moisture_form(bounds, num_hydrologyc, &
          !  save number of adaptive substeps used during time step
          nsubsteps(c) = nsubstep
 
+         ! check for over-saturated layers and move excess upward
+         do j = nlayers,2,-1
+            over_saturation   = max(h2osoi_liq(c,j)-(eff_porosity(c,j)*m_to_mm*dz(c,j)),0._r8)
+            h2osoi_liq(c,j)   = min(eff_porosity(c,j)*m_to_mm*dz(c,j), h2osoi_liq(c,j))
+            h2osoi_liq(c,j-1) = h2osoi_liq(c,j-1) + over_saturation
+         end do
```
samsrabin, line 1409: @wwieder says this is already happening somewhere else. Let's convert to a function that can be reused.

samsrabin (changes requested):

wwieder:
I'm running a 30 year BCG-Crop case from my: 
- 5.4.004 sandbox; (~/CTSM/); on an 
- _Oversaturation_ branch with these changes to SoilWaterMovement; that is
- cloned from ctsm5.4.004_bgc_f09_131_HIST; and using 
- initial conditions from 
`finidat = '/glade/derecho/scratch/wwieder/archive/ctsm5.4.004_bgc_f09_131_HIST/rest/1994-01-01-00000/ctsm5.4.004_bgc_f09_131_HIST.clm2.r.1994-01-01-00000.nc'`. 

The case directory is:
`/glade/work/wwieder/ctsm5.4.0/ctsm5.4.004_bgc_f09_Oversaturation_131b_HIST`

src/biogeophys/SoilWaterMovementMod.F90
ekluzek: I couldn't find a place where it's doing "exactly" the same thing. But, I found places that likely are equivalent.

``` fortran
git grep -i h2osoi_liq | grep ' = ' | grep -i 'min(' | grep -i dz
biogeophys/SnowHydrologyMod.F90:             vol_liq(c,j)      = min(eff_porosity(c,j),h2osoi_liq(c,j)/(dz(c,j)*frac_sno_eff(c)*denh2o))
biogeophys/SoilHydrologyMod.F90:             h2osoi_liq(c,j)   = min(eff_porosity(c,j)*dzmm(c,j), h2osoi_liq(c,j))
biogeophys/SoilHydrologyMod.F90:          h2osoi_ice(c,1) = min(max(0._r8,pondmx+watsat(c,1)*dzmm(c,1)-h2osoi_liq(c,1)), h2osoi_ice(c,1))
biogeophys/SoilHydrologyMod.F90:             h2osoi_liq(c,j)   = min(eff_porosity(c,j)*dzmm(c,j), h2osoi_liq(c,j))
biogeophys/SoilHydrologyMod.F90:          h2osoi_ice(c,1) = min(max(0._r8,pondmx+watsat(c,1)*dzmm(c,1)-h2osoi_liq(c,1)), h2osoi_ice(c,1))
biogeophys/SoilMoistStressMod.F90:             vol_liq(c,j) = min(eff_porosity(c,j), h2osoi_liq(c,j)/(col%dz(c,j)*denh2o))
biogeophys/SoilWaterMovementMod.F90:            h2osoi_liq(c,j)   = min(eff_porosity(c,j)*m_to_mm*dz(c,j), h2osoi_liq(c,j))
biogeophys/SurfaceHumidityMod.F90:                     vol_liq = min(eff_porosity, h2osoi_liq(c,j)/(dz(c,j)*denh2o))
```

There's a function in SoilMoistStress "calc_volumetric_h2oliq" that does the restriction (but only for liquid and not ice) in units of volume. Something around it could possibly be used.

I'm always in favor of using little functions to make sure things are done consistently and in one place for when they need to change. But, I think the refactoring for it in this case is probably beyond the scope we need here.

ekluzek: @wwieder and or @swensosc do you know of another place that duplicate this that could be refactored? If we do some refactoring we'd also want that to be a part where we show b4b so we know it's correct. So that could be in a different tag and/or on b4b-dev. But, let me know your thoughts on this thanks.

src/biogeophys/SoilWaterMovementMod.F90
```diff
@@ -1403,6 +1404,13 @@ subroutine soilwater_moisture_form(bounds, num_hydrologyc, &
          !  save number of adaptive substeps used during time step
          nsubsteps(c) = nsubstep
 
+         ! check for over-saturated layers and move excess upward
+         do j = nlayers,2,-1
+            over_saturation   = max(h2osoi_liq(c,j)-(eff_porosity(c,j)*m_to_mm*dz(c,j)),0._r8)
```
ekluzek, line 1406: The one thing I think we should add is to define the units of over_saturation at the top with

``` fortran
    real(r8) :: over_saturation. ! Amount of water that over saturates this soil layer [kg/m2]
```

ekluzek (approved):
Glad you all were able to figure this out and have a fix for it.

I agree with @samsrabin about refactoring to have a function to do this in that would allow this to be checked and made consistent everywhere. But, I think that should be done in b4b-dev.

The only suggestion I have is to add a comment about the purpose and units of the new temporary variable added in it's type declaration. Still marking as approve, because I can just do that change myself.

ekluzek:
@swensosc question for you. eff_pososity*dz gives you how much available room is in the soil for both ice and liquid right? So if both ice and liquid exist the limitation would have to change to account for that right? Or is that handled elsewhere?

This is just an example of a kind of detail that's good to have in one place so it can be ensured it's correct there. And that promotes the idea of a function to handle these sorts of things...

Anyway, I wondered about that wanted to make sure we have it right since we are looking at it. Thanks in advance...

swensosc:
no, watsat is the variable for soil porosity (total available pore space).
eff_porosity is the variable for available space for liquid, i.e.
watsat-ice_volume.  In this case, eff_porosity is appropriate because only
liquid water is redistributed in soilwatermovement.  Ice stays the same, so
you want to make sure the liquid doesn't go above eff_porosity; otherwise
the sum of liquid and ice will be greater than porosity (watsat).

swensosc:
I disagree that this should be made a function.  As it is, it is readable
and clear and its purpose is obvious.  It is only a few lines of code, and
a function would be more code and less clear.

ekluzek:
@swensosc thanks for getting back to me. I figured that would've had to be handled, but wasn't sure how it's done. So thanks for the explaination.

I'm going to move forward with this as is other than the comment on the new temporary variable.

We can decide to talk about any refactoring later, and decide if we want to bring it up again or not...

@samsrabin for now could you approve as is?

src/biogeophys/SoilWaterMovementMod.F90
wwieder: Let's not worry about refactoring the code here.  I've tested this PR and it's simple enough that I suggest we leave it as is.

wwieder:
> I disagree that this should be made a function. As it is, it is readable and clear and its purpose is obvious. It is only a few lines of code, and a function would be more code and less clear.
 
I agree with Sean here.  Let's not over complicate our lives here, the PR is very readable and strait forward.

wwieder:
Sean also said that the simulations and diagnostics I posted on [LWMG_dev#132](https://github.com/NCAR/LMWG_dev/issues/132) look as expected, so I think we can merge this PR.

samsrabin:
I'm not going to hold up this PR for it, but I do want to explain my rationale for requesting a refactor to a reused subroutine.

1. It's either complicate our lives now (making a new subroutine) or later (when this behavior needs to be updated across multiple locations in the code that may not all look alike).
2. To me, it would be more readable and clear if it had more obviously-named variables and better commenting. It would take up less space to do this just once, in its own subroutine.

doc/.ChangeLog_template
```diff
@@ -146,13 +146,12 @@ Changes answers relative to baseline:
    only by roundoff level (because order of operation changes for example). Roundoff
    changes to state fields usually grow to greater than roundoff as the simulation progresses.
 
-   If this tag changes climate describe the run(s) done to evaluate the new
-   climate (put details of the simulations in the experiment database)
-       - casename:
-
-   URL for LMWG diagnostics output used to validate new climate:
-
+   If this tag changes climate list the run(s) done to evaluate the new
```
ekluzek, line 149: FYI: @wwieder and @slevis-lmwg I updated the ChangeLog template so that LMWG_dev issues are easily referenced. Which makes it easier to fill out and give the full details in LMWG_dev. Comment if you'd like changes in the wording.

slevis-lmwg: Thanks for doing that, Erik.

ekluzek:
Testing looks as expected. Most tests change answers from baseline, other than a few short 1 day tests or single point ones.

ekluzek:
I failed to put this in the right PR, so adding now:

I tested SMS_Lm1.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode to show it changes answers (and to show that my latest commit didn't change answers from the previous.

Lot's of fields are different as expected. So starting standard testing now.

wwieder:
Given some of the changes we saw in the LDF diagnostics I'm not surprised that there are lots of (small) differences.  Soil moisture changes should influence the rest of the system in ways that aren't going to produce b4b changes, but they should be functionally / statistically similar.

Issue #1099, opened 2020-08-07
QBOT is zero for 2014 for GSWP3v1 data


ekluzek:
QBOT is zero for all of 2014 for the GSWP3v1 data. In order to fix this we need to bring in a cime update that points to the new data. The cime issue that documents this problem is here: https://github.com/ESMCI/cime/issues/3653

@negin513 @billsacks

ekluzek:
We think the first stop gap measure we should put in place is to change the end year to 2013 rather than 2014. And try to add some notes about this problem.

billsacks:
The stop-gap measure was done in https://github.com/ESMCI/cime/pull/3679 . From discussion with @dlawrenncar we might stop using the GSWP data, so let's hold off on doing anything major until that's decided.

ekluzek:
We did the stop-gap measure, and that is all we plan to do at this point. So closing.

PR #3688, opened 2026-01-12
ctsm5.4.009 Implement dewpoint temperature check for bare ground to fix RRTMGP error

olyson:
### Description of changes

Implement dewpoint temperature check for bare ground to fix RRTMG error

### Specific notes

See the discussion in https://github.com/ESCOMP/CTSM/issues/3589 beginning Dec 30, 2025.
This bug fix eliminates dew flux (sets evaporation conductance to zero) x if t_grnd > forc_dewpoint, where forc_dewpoint is the dewpoint temperature at the forcing height.
This prevents large spurious negative latent heat fluxes that increase heat flux into the soil and signficantly increase surface soil temperature (t_grnd) such that soil temperature > 355K, triggering a RRTMGP error.

Contributors other than yourself, if any: @swensosc , @wwieder , @slevis-lmwg , @dlawrenncar 

  Fixes https://github.com/ESCOMP/CTSM/issues/3589

Are answers expected to change (and if so in what way)? Yes.  Likely same climate however.

Does this create a need to change or add documentation? Did you do so? Yes. No.

Testing performed, if any: regular aux_clm on Derecho and Izumi
 Ran two Clm60SP I2000 simulations to evaluate the impact of the bug fix:

/glade/work/oleson/ctsm_dew/cime/scripts/Clm60SP_ctsmdew_1deg_Control_2000
/glade/work/oleson/ctsm_dew/cime/scripts/Clm60SP_ctsmdew_1deg_Dew_2000

ILAMB results are [here](https://webext.cgd.ucar.edu/I2000/Clm60SP_ctsmdew_1deg_Dew_2000/lnd/BUGFIX/_build_DEW/)
Standard diagnostics results are [here](https://webext.cgd.ucar.edu/I2000/Clm60SP_ctsmdew_1deg_Dew_2000/lnd/BUGFIX/Clm60SP_ctsmdew_1deg_Dew_2000_2005_2019-Clm60SP_ctsmdew_1deg_Control_2000_2005_2019/setsIndex.html)
NOTE: 01/15/2026: Links to these diagnostics have been updated to new simulations run with a bug fix for the check for ice/liq coefficients.  The bug fix had minimal impact on the simulations, but note that the discussion of results as indicated below may not be entirely consistent with the newest simulations.

See https://github.com/ESCOMP/CTSM/issues/3589 for discussion of results

Also ran a branch beginning at 0037-08-01 of the b.e30_alpha07g.B1850C_LTso.ne30_t232_wgx3.276_4xco2 simulation, which passed the point at which the model crashed with the RRTMGP error.  This simulation (/glade/derecho/scratch/oleson/b.e30_alpha07g.B1850C_LTso.ne30_t232_wgx3.276_4xco2.testrh1) was continued through year 65.  Diagnostics are [here](https://webext.cgd.ucar.edu/BCO2x4/b.e30_alpha07g.B1850C_LTso.ne30_t232_wgx3.276_4xco2.testrh1/lnd/b.e30_alpha07g.B1850C_LTso.ne30_t232_wgx3.276_4xco2.testrh1_46_65-b.e30_alpha07g.B1850C_LTso.ne30_t232_wgx3.276_4xco2_46_65/setsIndex.html)

Test results:
- Almost ALL tests show differences (269 on Derecho, 39 Izumi)
- 8 tests on Izumi were identical
- 18 tests on Derecho were identical

### DIFF
--- a/doc/.ChangeLog_template
+++ b/doc/.ChangeLog_template
@@ -148,10 +148,8 @@ Changes answers relative to baseline:
 
    If this tag changes climate list the run(s) done to evaluate the new
    climate (from https://github.com/NCAR/LMWG_dev)
-       - issue number:
+       - issue number(s):
 
-   URL for LDF diagnostics output used to validate new climate:
-       - 
 Other details
 -------------
 [Remove any lines that don't apply. Remove entire section if nothing applies.]
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.4.009
 Originator(s): olyson (Keith Oleson)
-Date: Fri Jan 16 02:27:00 PM MST 2026
+Date: Mon Jan 19 12:53:18 PM MST 2026
 One-line Summary: Dewpoint Temperature check for bare ground
 
 Purpose and description of changes
@@ -87,9 +87,7 @@ Changes answers relative to baseline: Yes! Almost ALL cases if run long enough
     - nature of change: same climate
 
    climate (from https://github.com/NCAR/LMWG_dev)
-       - issue number:
-
-       casename (oleson): b.e30_alpha07g.B1850C_LTso.ne30_t232_wgx3.276_4xco2.test
+       - issue number(s): 135 and 136
 
    URL for LDF diagnostics output used to validate new climate:
 
--- a/doc/ChangeSum
+++ b/doc/ChangeSum
@@ -1,6 +1,6 @@
 Tag                   Who      Date  Summary
 ============================================================================================================================
-       ctsm5.4.009   olyson 01/16/2026 Dewpoint Temperature check for bare ground
+       ctsm5.4.009   olyson 01/19/2026 Dewpoint Temperature check for bare ground
        ctsm5.4.008 swensosc 01/12/2026 Add a correction for oversaturated soil layers in SoilWaterMovement (erik)
        ctsm5.4.007 samrabin 01/09/2026 Update (irrigated_)rice grnfill and lfemerg
        ctsm5.4.006     erik 01/07/2026 Merge b4b-dev to master


src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -144,6 +144,17 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
     real(r8) :: e_ref2m                          ! 2 m height surface saturated vapor pressure [Pa]
     real(r8) :: qsat_ref2m                       ! 2 m height surface saturated specific humidity [kg/kg]
     real(r8) :: www                              ! surface soil wetness [-]
+    real(r8) :: forc_dewpoint                    ! dewpoint temperature at forcing height [K]
+    real(r8) :: forc_e                           ! vapor pressure at forcing height [Pa]
+    ! Coefficients from Alduchov and Eskridge (1996) as used by Lawrence (2005) for Magnus dewpoint
+    ! equation over liquid
+    real(r8) :: A1_liq = 17.625_r8                   ! [-]
```
ekluzek: Does [-] mean unitless?

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -144,6 +144,17 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
     real(r8) :: e_ref2m                          ! 2 m height surface saturated vapor pressure [Pa]
     real(r8) :: qsat_ref2m                       ! 2 m height surface saturated specific humidity [kg/kg]
     real(r8) :: www                              ! surface soil wetness [-]
+    real(r8) :: forc_dewpoint                    ! dewpoint temperature at forcing height [K]
+    real(r8) :: forc_e                           ! vapor pressure at forcing height [Pa]
+    ! Coefficients from Alduchov and Eskridge (1996) as used by Lawrence (2005) for Magnus dewpoint
```
ekluzek, line 151: @olyson question on these new expressions and constants. I got the impression that QSatMod.F90 and HumanIndexMod.F90 might have essentially duplicates of at least of these constants (C1). If so it would be good to use those.

But, the references seem to be different, so they might be different enough that they can't be used interchangably.

I did look for signs that the same constants where someone else in CTSM and didn't find it, including in share, cmeps and cdeps. So I don't think there is really anything to do, but wanted to check on your thoughts on C1.

src/biogeophys/BareGroundFluxesMod.F90
olyson: Yes, as, for example, www defined in the original code just above.

src/biogeophys/BareGroundFluxesMod.F90
olyson: That is an interesting question, thanks.
I had originally searched the code for a dewpoint calculation and there is a dewpoint calculation in src/fates/fire/SFNesterovMod.F90, which points to the reference Lawrence (2005).  
That reference is what I used to implement a dewpoint calculation in CTSM.
However, FATES uses Equation (8) which is based on relative humidity while I am using Equation (7) which is based on vapor pressure.  The two equations provide the same answer as long as relative humidity in Equation (8) is calculated from vapor pressure.  Based on what I had available in BareGroundFluxesMod.F90, I chose Equation (7).
But the constants in Lawrence (2005) and the ones used by FATES are not exactly the same.  For instance, FATES defines dewpoint_a as 17.62_fates_r8 (in src/fates/main/FatesConstantsMod.F90) while the value that is documented in Lawrence (2005) for Equations (7) and (8) is A1 = 17.625_r8.  And FATES defines dewpoint_b = 243.12_fates_r8 while the value that is documented in Lawrence (2005) is B1 = 243.04_r8.   The Equation (8) used by FATES doesn't require C1.
I did not see the FATES dewpoint_a and dewpoint_b values documented in Lawrence (2005).
Also, Lawrence (2005) references  Alduchov and Eskridge (1996) for the source of these constants and I verified the values that I'm using appear in that paper.  Furthermore,  Alduchov and Eskridge (1996) provide equations for dewpoitnt temperature over ice, which doesn't appear to be addressed in FATES.  I've used those constants over ice.

ekluzek: Ok so should we try to use what's done in FATES here, or go as is? What do you think?

Since FATES is always included and will be the default veg model in the future it might be good to go with what FATES does right?

ekluzek: Ok so should we try to use what's done in FATES here, or go as is? What do you think?

Since FATES is always included and will be the default veg model in the future it might be good to go with what FATES does right?

olyson: Well, no, I don't think so.  As I mentioned, I can't confirm the constants used in FATES.  And relative humidity calculated from vapor pressure is not available in BareGroundFluxesMod.F90.  There is a 2-m relative humidity but it is calculated from specific humidity, not vapor pressure, and I don't want to introduce a forcing relative humidity that is calculated from vapor pressure.  That might cause confusion. Furthermore, FATES doesn't deal with the concept of dewpoint temperature over ice.

ekluzek: Ahh, OK. So we'll go forward as is. Somehow my comment got replicated, so I deleted the second one.

ekluzek:
I tested  SMS_Lm1.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode to show it changes answers (and to show that my latest commit didn't change answers from the previous.

Lot's of fields are different as expected.  So starting standard testing now.

ekluzek:
Whoops put a comment in the wrong PR, so moving.

ekluzek:
OK, submitted testing for this one.

ekluzek:
32 tests failed on Derecho and 2 on Izumi. There's a divide by zero that I think should be straightforward to fix.

olyson:
@ekluzek  and I looked at this.  The failures are due to forc_q (forcing height specific humidity) being zero.  GSWP3V1 apparently has some zero QBOT for some gridcells, e.g., in the 2000-02 forcing.  We had an issue filed for zero QBOT in 2014 (https://github.com/ESCOMP/CTSM/issues/1099), but apparently there are some random zero QBOT in other files.  In order for the GSWP3V1 tests to pass, we'd probably need to set a non-zero minimum forc_q in the new code.

olyson:
One thing we could do to avoid the error caused by zero forc_q (which causes zero forc_e) is to restrict forc_e to be no lower than the value corresponding to 1% relative humidity (a bit arbitrary but this is what FATES chooses in fire/SFNesterovMod.F90).
So,

```
real(r8) :: forc_qs                         ! saturated specific humidity at forcing height [kg/kg]
real(r8) :: forc_esat                       ! saturated vapor pressure at forcing height [Pa]

! Get saturated vapor pressure at forcing height
call QSat(forc_t(c), forc_pbot(c), qs = forc_qs, es = forc_esat)

! Calculate vapor pressure at forcing height and restrict to no less than value corresponding to 1% relative humidity
forc_e = max (forc_q(c)Xforc_pbot(c)) / (forc_q(c)+0.622_r8), 0.01_r8Xforc_esat)
```

olyson:
@ekluzek , I've implemented the changes I suggeested above and tested them and they look good.  I committed them and tried pushing them to my branch but I guess you have updated the branch to latest master.  Can you suggest the proper steps for me to take to push my new changes?  Then you can run the test suite on them.

ekluzek:
> Can you suggest the proper steps for me to take to push my new changes? Then you can run the test suite on them.

Yes. You need to pull the changes from the branch into your local, and then you'll be able to push your changes. If there's conflicts you'll have to resolve them at the pull step. So

``` shell
git pull olyson dew
# resolve conflicts if they exist
# if conflicts exist you'll need to
$EDIT <files_w_conflicts>
git add <files_w_conflicts>
git commit
# Then push your changes
git push olyson dew
```

olyson:
Ok, initially, after the pull I got:

hint: You have divergent branches and need to specify how to reconcile them.
hint: You can do so by running one of the following commands sometime before
hint: your next pull:
hint:
hint:   git config pull.rebase false  # merge
hint:   git config pull.rebase true   # rebase
hint:   git config pull.ff only       # fast-forward only
hint:
hint: You can replace "git config" with "git config --global" to set a default
hint: preference for all repositories. You can also pass --rebase, --no-rebase,
hint: or --ff-only on the command line to override the configured default per
hint: invocation.
fatal: Need to specify how to reconcile divergent branches.

I chose to do git config pull.rebase false.
And then pulled again.
Then pushed.  Which seemed to work.

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -144,6 +144,19 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
     real(r8) :: e_ref2m                          ! 2 m height surface saturated vapor pressure [Pa]
     real(r8) :: qsat_ref2m                       ! 2 m height surface saturated specific humidity [kg/kg]
     real(r8) :: www                              ! surface soil wetness [-]
+    real(r8) :: forc_dewpoint                    ! dewpoint temperature at forcing height [K]
+    real(r8) :: forc_e                           ! vapor pressure at forcing height [Pa]
+    real(r8) :: forc_qs                          ! saturated specific humidity at forcing height [kg/kg]
+    real(r8) :: forc_esat                        ! saturated vapor pressure at forcing height [Pa]
+    ! Coefficients from Alduchov and Eskridge (1996) as used by Lawrence (2005) for Magnus dewpoint
+    ! equation over liquid
+    real(r8) :: A1_liq = 17.625_r8                   ! [-]
```
ekluzek: I realized an important change here is that these constants should all be made into parameters,

src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -405,13 +418,37 @@ subroutine BareGroundFluxes(bounds, num_noexposedvegp, filter_noexposedvegp, &
...
+         forc_e = max( (forc_q(c)*forc_pbot(c)) / (forc_q(c)+0.622_r8), 0.01_r8*forc_esat)
+
+         ! Magnus expression for dew point
+         ! Equation (7) from Lawrence (2005)
+         if (t_grnd(c) < 0._r8) then ! ice coefficients
+            forc_dewpoint = B1_ice*log(forc_e/C1_ice) / (A1_ice - log(forc_e/C1_ice))
+         else                        ! liquid water coefficients
+            forc_dewpoint = B1_liq*log(forc_e/C1_liq) / (A1_liq - log(forc_e/C1_liq))
+         end if
+         forc_dewpoint = forc_dewpoint + tfrz
```
ekluzek, line 435: Also in thinking about this I think we should make dewpoint a little function that's part of QSat. That isolates it's complexity in one place, allows for the possibility of unit testing, or testing specific inputs that might fail, and it's a move to better modularization. Right now dewpoint isn't used anywhere else in the code, but it's a basic enough thing that it might be.

This is likely a refactoring that we should do later. But, I did want to bring it up.

ekluzek:
I started testing again. Let's see how it does this time!

src/biogeophys/BareGroundFluxesMod.F90
olyson: Like this example I guess:

main/FatesConstantsMod.F90:  real(fates_r8), parameter, public :: dewpoint_a = 17.62_fates_r8
main/FatesConstantsMod.F90:  real(fates_r8), parameter, public :: dewpoint_b = 243.12_fates_r8 ![degrees C]

olyson: Or more CTSM-like, e.g.,

SoilTemperatureMod.F90:  real(r8), private, parameter :: thin_sfclayer = 1.0e-6_r8

ekluzek:
OK, testing looks as expected! Almost all tests change answers (with the exception of 18 on Derecho and 2 on Izumi).

ekluzek:
@olyson we need to update the ChangeLog. I can do that based on the information at the top, or you can. Which do you prefer?

src/biogeophys/BareGroundFluxesMod.F90
ekluzek: Since, they are local right now, they don't need "private" since they already are. The parameter keyword is the important thing.

olyson:
If you can do it that would be great, it's been a long time since I've gone through that procedure.  Happy to reivew it if need be.
Fyi, I did change those constants to parameters and pushed that.

src/biogeophys/BareGroundFluxesMod.F90
wwieder: This is good idea, especially when / if we consider doing this over vegetation in canopy fluxes.  For now, let's leave this as is, since it's important to get this change into an alpha tag quickly.

olyson:
Actually, we are going to have to hold off on this.  There is an error in the check for liq/ice coefficients.

wwieder:
@olyson can you share with @cecilehannay where she can get these sourcemods for a new case she's about to start?

olyson:
> @olyson can you share with @cecilehannay where she can get these sourcemods for a new case she's about to start?

I'd like to see what the diagnostics look like for the bugfix I implemented today.  I don't expect any big differences but just want to make sure before I pass it on.  Should be this afternoon.
What tag is she going to run with?  Need to make sure there haven't been any changes to BareGroundFluxesMod.F90 since the tag my changes are based on (alpha07g).

olyson:
Ok, I think the bugfix is looking fine.  If anything, the differences from the control are even smaller, particularly over ice sheets.  The updated ILAMB diagnostics are [here](https://webext.cgd.ucar.edu/I2000/Clm60SP_ctsmdew_1deg_DewRH1_2000/lnd/BUGFIX/_build_DEWRH1/).
The standard diagnostics are [here](https://webext.cgd.ucar.edu/I2000/Clm60SP_ctsmdew_1deg_DewRH1_2000/lnd/BUGFIX/Clm60SP_ctsmdew_1deg_DewRH1_2000_2005_2019-Clm60SP_ctsmdew_1deg_Control_2000_2005_2019/setsIndex.html).

I've pushed the change and the branch should now be ready for re-testing @ekluzek

ekluzek:
OK, testing looks good! The only glitch was that NEON tests failed because of a glitch in downloading data. So I'm attending to them which is just a little bit of work. So I'll make the ChangeLog as well and make the tag today.

src/biogeophys/BareGroundFluxesMod.F90
ekluzek: Made into an issue in #3694

ekluzek:
Crap, I didn't get @olyson's latest update in my testing on Derecho (although I thought I had). So I'll rerun testing there before making the tag.

Issue #3396, opened 2025-08-07
Add CUPiD to the CTSM checkout and start hooking it up to enable using it in cases


ekluzek:
CUPiD has advanced far enough that we now need to start planning on bringing it into the CTSM checkout and enable it to be able to be used in I compset cases that people create with CTSM.

Definition of Done:
- [ ] Add to .gitmodule and tools/CUPiD directory
- [ ] Make sure it can be easily turned on from an I case setting RUN_POSTPROCESSING=TRUE
- [ ] Add a test to aux_clm that tests turning it on
- [ ] Add to the github actions for conda env to build the CUPiD envs

This started as a discussion that is referenced below. And there is a list of requirements from @TeaganKing before we should do this listed there.

I asked @TeaganKing an update on this and she sees just a few things that should happen before we do this. Here's the list of suggestions she had as requirements before this comes in:

_We should bring CUPiD into CTSM as a submodule once the following PRs are included in CUPiD:_

- [x] _[Include ILAMB in key_metrics and CESM workflow](https://github.com/NCAR/CUPiD/pull/210/files) (needs a bit more work but getting close)_
- [x] _[Enable LDF in CUPiD](https://github.com/NCAR/CUPiD/pull/226) (under final review)_
- [x] _[Allow users to specify what CUPiD install to use in CESM Workflow](https://github.com/NCAR/CUPiD/issues/257) (is on Mike's high priority list)_
- [x] _[Create a lnd-specific example](https://github.com/NCAR/CUPiD/issues/262) (This will be super quick & easy to do but hasn't been started)

_I'd guess this may take a month-ish before everything is ready..._

_Teagan_

_Originally posted by @ekluzek in https://github.com/ESCOMP/CTSM/discussions/2864#discussioncomment-14038391_

wwieder:
Thanks for renewing this topic, Erik.  I'm supper supportive of having CUPiD available with CTSM and suggest we do this for the CLM6 release

ekluzek:
The CUPiD things have been done in the list at the top, other than create a land example. It says that should be easy to do. Adding to next to discuss if we should have someone do that step and start having CUPiD available in CTSM checkouts.

TeaganKing:
[...]

PR #3689, opened 2026-01-13
Changes to enable easily turning on CUPiD diagnostics from your CTSM case

ekluzek:
### Description of changes

The changes that are needed to turn CUPiD diagnostics from a CTSM case.

### Specific notes

  @TeaganKing @samrabin

  Work on #3396 

Any User Interface Changes (namelist or namelist defaults changes)? Allows 

   RUN_POSTPROCESSING==TRUE to be used in CTSM cases

Does this create a need to change or add documentation? Did you do so? Yes
  I worked on the README files, but we'll need to add to the User's Guide

Testing performed, if any: 

I'm running a single test case to show the workflow. But, there are issues...

### DIFF
--- a/cime_config/usermods_dirs/clm/CUPiD/shell_commands
+++ b/cime_config/usermods_dirs/clm/CUPiD/shell_commands
@@ -33,6 +33,8 @@ fi
 #
 # General setup based on the start and length of the case being run
 #
+./xmlchange CUPID_CLIMO_END_YEAR=2004
+./xmlchange CUPID_CLIMO_N_YEAR='$STOP_N'
 # TODO: Have this set based on what CLM_BLDNML_OPTS has
 #./xmlchange CUPID_RUN_TYPE=SP
 ./xmlchange CUPID_STARTDATE='$RUN_STARTDATE'
@@ -45,6 +47,7 @@ fi
 #
 # Setting for specific CUPiD diagnostics to run
 #
+./xmlchange CUPID_GEN_TIMESERIES=FALSE
 # NOTE: Create webpages for Tutorial, for testing have this done seperately
 ./xmlchange CUPID_GEN_HTML=TRUE
 # TODO: Test with running more tasks


README.DIAGNOSTICS
```diff
@@ -0,0 +1,44 @@
+# QuickStart to Running Diagnostics of your Case with CUPiD (CESM Unified Postprocessing and Diagnostics)
```
TeaganKing, line 1: Can we point to the documentation for running CUPiD via the CESM workflow rather than duplicating and ending up with outdated information?
I'm thinking it would also be good for me to put the land-specific 'cheat sheet' as a docs page for component-specific example of running a STANDALONE version

cime_config/usermods_dirs/clm/CUPiD/shell_commands
```diff
@@ -0,0 +1,40 @@
...
+
+#
+# Set CUPiD parameters for what a CTSM case needs to do
+#
+
+# Set the standard CUPiD baseline case to compare to
+./xmlchange --force CUPID_BASELINE_CASE='$CASE'
+./xmlchange --force CUPID_BASE_STARTDATE='$RUN_STARTDATE'
+./xmlchange --force CUPID_BASE_STOP_N='$STOP_N'
+# TODO: Make sure STOP_OPTION is nyears
```
TeaganKing: Adding these in now in [this PR](https://github.com/NCAR/CUPiD/pull/357):

CUPID_CLIMO_N_YEARS
CUPID_CLIMO_START_YEAR
CUPID_BASE_CLIMO_START_YEAR
CUPID_BASE_CLIMO_N_YEARS

.gitmodules
```diff
@@ -121,6 +121,18 @@ fxrequired = ToplevelRequired
 # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
 fxDONOTUSEurl = https://github.com/ESMCI/mpi-serial
 
+[submodule "cupid"]
+# CUPiD does model post-processing
+path = tools/CUPiD
+#url = https://github.com/NCAR/CUPiD.git
+url = https://github.com/TeaganKing/CUPiD.git
+#fxtag = v0.4.1
```
TeaganKing: The tag will need to be updated.

cime_config/usermods_dirs/clm/CUPiD/shell_commands
samsrabin, line 1: Let's go through this and figure out which of these should be default settings for ALL CTSM standalone cases—i.e., not just in this usermod. (And also—which of these are already set correctly by default without us in CTSM needing to do anything.)

README.DIAGNOSTICS
```diff
@@ -0,0 +1,44 @@
...
+
+[!NOTE]
+Land Diagnostics Cheat Sheet is here:
+
+This document has more details on more options and such
+
+https://docs.google.com/document/d/1ziZWGgaj9FxgR6WRHyCAZnzIc6-4ezqfUDDpQv4MzUI/edit?tab=t.0
```
ekluzek: Point to this:

https://ncar.github.io/CUPiD/run_cesm.html

README.DIAGNOSTICS
```diff
@@ -0,0 +1,44 @@
...
+
+## Initial setup steps to do first time
+
+This step is something you need to do the first time you want to run CUPiD for a case.
+And you only need to do it again if the CUPiD environment changes.
+
+``` shell
+# Setup the conda environments
```
ekluzek: Here, point to the CUPiD documentation

cime_config/usermods_dirs/clm/CUPiD/shell_commands
samsrabin: Maybe in `cime_config/config_component.xml`. But need to check that these would only apply in a standalone CTSM checkout.

ekluzek:
We talked about this in the CTSM SE meeting this morning. I need to move onto other things, so I can't complete everything here. But, here are tasks that we do need:
- [x] Erik will create a project board and issues for CUPiD
- [ ] CUPiD is working on [v0.6.0](https://github.com/NCAR/CUPiD/issues/368) which will verify that CESM cases are working
- [ ] Update to that version here
- [x] Check if Meg's key_metrics test (of atm/lnd) is expected to work without CAM (it doesn't because of PSL)
- [x] Will W: Create config.yml file for LDF and ILAMB for tutorial land_only example (https://github.com/NCAR/CUPiD/issues/381)
- [x] Erik: Create a shorter one for CTSM testing, land_only_testing example
- [ ] The CUPiD test needs to create a notebook that can be checked by hand
- [x] Adding a proper system test that validates that things worked will be later (part of the larger CUPiD project) (will be done in #3710)

TeaganKing:
A few notes:
1. Meg's notebook only reads in land files, but I have no idea whether or not it's valid without CAM turned on.
2. Just added a link to the issue ticket describing the PRs that will go into new tag
3. I feel like "The CUPiD test needs to create a notebook that can be checked by hand" may be outside the scope of this PR and theoretically should be in CUPiD.

samsrabin:
@TeaganKing It should be in both places, I'd think. We want to make sure that whatever we have on the CTSM side does everything necessary.

cime_config/usermods_dirs/clm/CUPiD/shell_commands
ekluzek: I've been looking at other repositories for other things, but I see this sort of thing in config_compsets.xml for several components.

``` xml
  <entries>
    <entry id="RUN_STARTDATE"> 
      <values>
        <value compset="1850_">0001-01-01</value>
      </values>
    </entry>
```

So it looks like that's the sort of thing that can be done. The entry is defined elsewhere, but you can set the value according to the compset as above. I think in our case we want it to be in config_component.xml, so I'll check if it can be added there.

If not we could add %CUPID as part of the CLM compset options. I don't think I like that, but it's probably something to discuss. We do have compsets that set a particular user-mod directory as part of their definition. And this would be doing that same thing here.

cime_config/usermods_dirs/clm/CUPiD/shell_commands
ekluzek: This is merged in now, so resolving this. The CUPiD tag will still need to be updated.

cime_config/usermods_dirs/clm/CUPiD/shell_commands
```diff
@@ -0,0 +1,52 @@
...
+if [ "$TEST" == "TRUE" ]; then
+   ./xmlchange DOUT_S=TRUE --file env_test.xml
+fi
+
+#
+# Set CUPiD parameters for what a CTSM case needs to do
+#
+
+# Set the standard CUPiD baseline case to compare to
+./xmlchange --force CUPID_BASELINE_CASE='$CASE'
```
ekluzek: For the base case I need to point to @wwieder base case.

/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/ctsm5.4.002_clm6_BGCcrop_crujra_4x5_HIST/lnd/hist

make the clm6_0 physics in the ctsm5.,4.002 tag 4x5 case the base case

ekluzek: Use the start and extent for the base case.

ekluzek: Do the case as a I2000 starting 2000-2004

cime_config/usermods_dirs/clm/CUPiD/shell_commands
```diff
@@ -0,0 +1,52 @@
...
+./xmlchange --force CUPID_STOP_OPTION='$STOP_OPTION'
+# TODO: This should maybe be turned on as a smoke test, and to check for the appropriate files in testing
+./xmlchange CUPID_GEN_HTML=TRUE
+# TODO: Test with running more tasks
+./xmlchange CUPID_NTASKS=1
+./xmlchange CUPID_RUN_ADF=FALSE
+./xmlchange CUPID_RUN_ALL=FALSE
+# Use the land_only example
+./xmlchange CUPID_EXAMPLE=land_only
+# TODO: Get CUPiD working with LDF under CESM
```
ekluzek: LND without either LDF or ILAMB won't work except with a ne30 case

ekluzek:
We did work on this in order to use it for the CUPiD Tutorial at the Spring 2026 LMWG meeting. We made some good progress, but more needs to be done.

We also created Tutorial pages on the Wiki, and plan to move them to the UG.

 https://github.com/ESCOMP/CTSM/wiki/LMWG-Winter-Meeting-2026:-CTSM%E2%80%90CUPID-Tutorial

PR #3699, opened 2026-01-20
Decomp mod unittest

ekluzek:
### Description of changes

This adds a unit test for decompMod, that I need for #3666 

### Specific notes

Testing performed, if any: So far just unit tests 

(since this doesn't change model code there doesn't seem a need to run aux_clm)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -63,12 +63,16 @@ add_subdirectory(${CLM_ROOT}/src/init_interp clm_init_interp)
 add_subdirectory(${CLM_ROOT}/src/self_tests clm_self_tests)
 
 # Add FATES source directories
-add_subdirectory(${CLM_ROOT}/src/fates/main fates_main)
-add_subdirectory(${CLM_ROOT}/src/fates/biogeochem fates_biogeochem)
-add_subdirectory(${CLM_ROOT}/src/fates/biogeophys fates_biogeophys)
-add_subdirectory(${CLM_ROOT}/src/fates/parteh fates_parteh)
-add_subdirectory(${CLM_ROOT}/src/fates/fire fates_fire)
-add_subdirectory(${CLM_ROOT}/src/fates/radiation fates_radiation)
+# Adding FATES directories is commented out -- because of the problem in:
+# https://github.com/ESCOMP/CTSM/issues/3704
+# add_subdirectory(${CLM_ROOT}/src/fates/main fates_main)
+# add_subdirectory(${CLM_ROOT}/src/fates/biogeochem fates_biogeochem)
+# add_subdirectory(${CLM_ROOT}/src/fates/biogeophys fates_biogeophys)
+# add_subdirectory(${CLM_ROOT}/src/fates/parteh fates_parteh)
+# add_subdirectory(${CLM_ROOT}/src/fates/fire fates_fire)
+# add_subdirectory(${CLM_ROOT}/src/fates/radiation fates_radiation)
+# NOTE: Need to add the fates library and dependencies below
+# ### End add of FATES directories
 
 # Add general unit test directories (stubbed out files, etc.)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -13,18 +13,20 @@ project(clm_tests Fortran C)
...
 include(CIME_utils)
 
-
 # find needed external packages
 # NetCDF is required -- because PIO and NetCDF are required by the standard default ESMF libraries
 find_package(NetCDF 4.7.4 REQUIRED Fortran)
 
 # The following - for finding ESMF - is copied from the share CMakeLists.txt
-if (DEFINED ENV{ESMF_ROOT})
+if(DEFINED ENV{ESMF_ROOT})
```
ekluzek, line 21: There's some whitespace changes here, that I didn't really intend to have. And I actually thought I had removed, but I either didn't get that in, or pushed it to the wrong branch. So I'm willing to remove these if reviewers think they aren't helpful.

src/CMakeLists.txt
```diff
@@ -60,6 +62,14 @@ add_subdirectory(${CLM_ROOT}/src/main clm_main)
 add_subdirectory(${CLM_ROOT}/src/init_interp clm_init_interp)
 add_subdirectory(${CLM_ROOT}/src/self_tests clm_self_tests)
 
+# Add FATES source directories
```
ekluzek, line 65: Adding FATES source actually isn't needed here for this PR. But, I feel like in general we do want to bring in compiling FATES source code for unit tests, so I think it's a worthwhile thing to include. But, we could remove it if reviewers think so.

src/main/test/decomp_test/test_decompMod.pf
```diff
@@ -0,0 +1,106 @@
...
+    call decompmod_clean()
+
+  end subroutine tearDown
+
+  subroutine create_simpleSingleDecomp(this)
+    use spmdMod, only : iam
+    class(TestDecompMod), intent(inout) :: this
+
+    integer :: clump_pproc
+    ! TOTO: When decompMod has it's own allocate method that could be used here
```
slevis-lmwg:
```suggestion
    ! TODO: When decompMod has it's own allocate method that could be used here
```

src/main/test/decomp_test/test_decompMod.pf
```diff
@@ -0,0 +1,106 @@
...
+    nclumps = 1
+    clump_pproc = nclumps
+    allocate(procinfo%cid(clump_pproc))
+    allocate(clumps(nclumps))
+    ! Set the procinfo and clumps values
+    ! TOD: Use initialization method when available (currently in decompInitMod)
```
slevis-lmwg:
```suggestion
    ! TODO: Use initialization method when available (currently in decompInitMod)
```

src/CMakeLists.txt
slevis-lmwg: Not worth worrying about in my opinion.

src/CMakeLists.txt
slevis-lmwg: I'm ok with allowing the change as a placeholder for future development.

ekluzek: Oh, actually adding this brings in errors, that we don't want to have right now. I get a bunch of errors on endrun calls in FATES code like the following:

```
/glade/work/erik/ctsm_worktrees/newbranch/src/fates/radiation/TwoStreamMLPEMod.F90(1748): error #6498: The use-name for this local-name is not defined.   [ERRMSG]
       call endrun(msg=errMsg(sourcefile, 1748))
```

It works, if the build is run twice, but if you delete the unit test build directory, and run from scratch it doesn't work. So I'll remove this part.

ekluzek: OK, issue added in #3704 
And I commented out the addition in the file, and reference the issue.

PR #3700, opened 2026-01-21
Fix mksurfdata CMakeLists.txt: use SHARED for .so PIO libraries

cdevaneprugh:
The CMakeLists.txt declares PIO libraries as STATIC IMPORTED but then points to shared library files (.so). This is an internal inconsistency that can cause build failures when using externally-built shared PIO libraries.

Fix: Change STATIC IMPORTED to SHARED IMPORTED for pioc and piof targets since they reference .so files.

Tested on: HiPerGator (University of Florida HPC) with GCC 14.2.0, OpenMPI 5.0.7, shared PIO 2.6.6

### Description of changes
See above.

### Specific notes
Affects users building mksurfdata_esmf with externally-built shared PIO libraries.

Contributors other than yourself, if any:

CTSM Issues Fixed (include github issue #):

Are answers expected to change (and if so in what way)?
No - this is a build configuration fix only.

Any User Interface Changes (namelist or namelist defaults changes)?
No

Does this create a need to change or add documentation? Did you do so?
No

Testing performed, if any:
Built mksurfdata_esmf successfully with external shared PIO on HiPerGator.

PR #3701, opened 2026-01-21
Fix mksurfdata.F90 format specifiers for GCC 10+ compatibility

cdevaneprugh:
The `I` format descriptor without a width is a legacy Fortran extension that fails with strict GCC 10+ compilers. GNU Fortran requires `-fdec-format-defaults` flag to enable this extension.

Fix:
- For write statements: use `I0` (minimum-width format, standard Fortran 95+)
- For read statements: use `I6` (read requires positive width)

This matches the coding style used elsewhere in CTSM (e.g., `I0` in src/main/abortutils.F90 for output formatting).

Changes:
- Line 271: '(2(a,I))' -> '(2(a,I0))'
- Line 295: '(i)' -> '(I6)'
- Line 328: '(a, I, a, I)' -> '(a, I0, a, I0)'

Tested on: HiPerGator (UF HPC) with GCC 14.2.0

Reference: https://gcc.gnu.org/onlinedocs/gfortran/Default-widths-for-F_002fG_002fI-format-descriptors.html

### Description of changes
See above.

### Specific notes
Portability fix for GCC 10+ compilers which enforce stricter Fortran standards.

No - diagnostic output formatting only.

No

No

Testing performed, if any:
Built mksurfdata_esmf successfully on HiPerGator with GCC 14.2.0.


ekluzek:
We'll want this to go to b4b-dev like #3700, but we'll undoubtably have the same problem rebasing this one as well. So same solutions apply.

ekluzek:
I verified the build works on Derecho. So merging.

PR #3710, opened 2026-01-22
Improve radiation balance check

### LINKED ISSUE #3057: error in errsol computation in BalanceCheckMod reports the wrong albedo values
### Brief summary of bug
When an error occurs in BalanceCheckMod for errsol - the wrong albedo values are reported. 

### General bug information
The reason for this is the following. 
In in clm_driver.F90 the following call sequence occurs:

```
       call SurfaceRadiation()
       ! Determine albedos for NEXT TIME STEP
       if (doalb) then
          call SurfaceAlbedo()
       endif
       call BalanceCheck()
```

In BalanceCheck the albedos that are reported for the energy balance error are those obtained from SurfaceAlbedo  for the next time step - and NOT the ones used in SurfaceRadiation. This gives misleading information as to the cause of the balance error. I believe that the correct solution is to have a separate routine to check energy balance errors that is called directly at the end of SurfaceRadiation - either from SurfaceRadiation or from clm_driver.

**Does this bug cause significantly incorrect results in the model's science?** no

samsrabin:
### Description of changes

Aimed at resolving #3057. Most of the changed lines have to do with factoring a new EnergyBalanceCheck subroutine out of BalanceCheck. This new subroutine is now called in `clm_drv()` just _before_ the surface albedo stuff happens. It is also still called after the albedo stuff.

**Remaining questions:**
- [x] Should this check be deleted from the existing `BalanceCheck`? Or maybe modified? **Keep it.**

### Specific notes

**Contributors other than yourself, if any:** @mvertens 

**CTSM Issues Fixed:**
- Resolves #3057

**Testing performed, if any:** `clm_short` suite: All pass.

### DIFF
--- a/src/biogeophys/BalanceCheckMod.F90
+++ b/src/biogeophys/BalanceCheckMod.F90
@@ -561,8 +561,10 @@ contains
           qflx_glcice_dyn_water_flux_col => waterflux_inst%qflx_glcice_dyn_water_flux_col  & ! Input: [real(r8) (:)]  column level water flux needed for balance check due to glc_dyn_runoff_routing (mm H2O/s) (positive means addition of water to the system)
           )
 
-       ! Get time step
+       ! Get step size and time step
        dtime = get_step_size_real()
+       nstep = get_nstep()
+       DAnstep = get_nstep_since_startup_or_lastDA_restart_or_pause()
 
        ! Determine column level incoming snow and rain
        ! Assume no incident precipitation on urban wall columns (as in CanopyHydrologyMod.F90).


src/biogeophys/BalanceCheckMod.F90
samsrabin, line 990: This check is failing with "floating invalid" during the new call of this subroutine during `SurfaceRadiation`. I'm going to try moving the check to after both SurfaceRadiation and UrbanRadiation have run.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -895,6 +841,110 @@ subroutine BalanceCheck( bounds, &
...
+     use CanopyStateType   , only : canopystate_type
+     use SurfaceAlbedoType , only : surfalb_type
+     !
+     ! !ARGUMENTS:
+     type(bounds_type)     , intent(in)    :: bounds
+     type(atm2lnd_type)    , intent(in)    :: atm2lnd_inst
+     type(solarabs_type)   , intent(in)    :: solarabs_inst
+     type(surfalb_type)    , intent(in)    :: surfalb_inst
+     type(energyflux_type) , intent(inout) :: energyflux_inst
+     type(canopystate_type), intent(inout) :: canopystate_inst
```
samsrabin, line 887: These are `inout` as they were in the original, but I don't like it! It's confusing for a balance check to modify anything.

Here are the fields that get modified:
- `energyflux_inst%errsol`
- `energyflux_inst%errseb`
- `energyflux_inst%errlon`
- `energyflux_inst%netrad`

The last one already has a comment saying it should be moved. The first three are _okay_ here, but it would be better for them to be in a different subroutine.

`canopystate_inst` has nothing modified and so it can be easily changed to `intent(in)`.

ekluzek: Yes, really good point there. Maybe the balance check variables should be in their own type. So something like

energyflux_balance_inst

?

samsrabin: I was thinking more along the lines of making a new member subroutine for `energyflux_type` that does the updating of the error variables.

wwieder:
If this is ready to go, should it go to b4b-dev and on the alpha08d tag or is the beta09 milestone acceptable?

ekluzek:
I suggest this go to bfb-dev and with a beta09 milestone. It's a refactoring and not critical for beta08.

samsrabin:
> If this is ready to go, should it go to b4b-dev and on the alpha08d tag or is the beta09 milestone acceptable?

@wwieder It could go in, but I'd like to hear from @mvertens as to whether the check should be removed from `BalanceCheck`.

src/biogeophys/BalanceCheckMod.F90
samsrabin: Filed as #3730.

PR #3713, opened 2026-01-22
Factor out new subroutine CropPhase_OnePatch

### LINKED ISSUE #3712: Need to make a version of CropPhase that works on individual patches
Will be necessary for #3633, which this blocks, and thus the crop CUPiD notebook. It will also enable simpler testing, if we ever want to do that.

samsrabin:
### Description of changes

Will be necessary for #3633.

### Specific notes

**CTSM Issues Fixed:**
- Resolves #3712

**Testing performed, if any:** `clm_short` tests are b4b.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -76,7 +76,7 @@ fxDONOTUSEurl = https://github.com/ESMCI/ccs_config_cesm.git
 [submodule "cime"]
 path = cime
 url = https://github.com/ESMCI/cime
-fxtag = cime6.1.145
+fxtag = cime6.1.146
 fxrequired = ToplevelRequired
 # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
 fxDONOTUSEurl = https://github.com/ESMCI/cime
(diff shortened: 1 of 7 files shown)


samsrabin:
Blocks #3633.

src/biogeochem/CNPhenologyMod.F90
```diff
@@ -2652,37 +2646,60 @@ subroutine CropPhase(bounds, num_pcropp, filter_pcropp, &
...
+
+    if (croplive) then
+       ! Start with cphase_planted, but this might get changed in the later
+       ! conditional blocks.
+       crop_phase = cphase_planted
+       if (leafout >= huileaf .and. hui < huigrain) then
+          crop_phase = cphase_leafemerge
+       else if (hui >= huigrain) then
+          ! Since we know croplive is true, any hui greater than huigrain implies that
+          ! we're in the grainfill stage: if we were passt gddmaturity then croplive
```
slevis-lmwg:
```suggestion
          ! we're in the grainfill stage: if we were past gddmaturity then croplive
```
This is a preexisting typo, but might as well correct it.

PR #3728, opened 2026-01-30
ctsm5.4.030: Add a FATES namelist option for land use transition logic

JessicaNeedham:
### Description of changes
This change is coordinated with FATES PR https://github.com/NGEET/fates/pull/1489. 
Together they move a hard coded value in FATES, which controls vegetation mortality during transition between land use classes, to a namelist option. 

### Specific notes

This PR sets the default to option 1 which is different than the current hard coded value of 4. Setting the value to 4 would be equivalent to existing code. 

Yes - see above. 

On the FATES side. Not done yet. 

Testing performed, if any:
Compiles and runs with NorESM-CTSM-FATES. 

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.4.030
 Originator(s): glemieux (Gregory Lemieux, LBNL, glemieux@lbl.gov)
-Date: Fri Apr  3 11:54:39 PM MDT 2026
+Date: Fri Apr  6 04:01:00 PM MDT 2026
 One-line Summary: Add FATES namelist option for land use transition logic
 
 Purpose and description of changes
@@ -19,6 +19,10 @@ Changes to CTSM's user interface (e.g., new/renamed XML or namelist variables):
 
 Testing summary:
 ----------------
+
+  build-namelist tests (if CLMBuildNamelist.pm has changed):
+
+    derecho - PASS
 
   regular tests (aux_clm: https://github.com/ESCOMP/CTSM/wiki/System-Testing-Guide#pre-merge-system-testing):
 


bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -2683,6 +2684,15 @@ lnd/clm2/surfdata_esmf/NEON/ctsm5.4.0/surfdata_1x1_NEON_TOOL_hist_2000_78pfts_c2
...
 <use_fates_luh                  use_fates=".true."                                >.false.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="1"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="2"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="3"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="4"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="5"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="6"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="7"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="8"   >.true.</use_fates_luh>
+<use_fates_luh                  use_fates=".true." fates_lu_transition_logic="9"   >.true.</use_fates_luh>
```
glemieux: @JessicaNeedham  I think we need to remove setting the `use_fates_luh` to `.true.` for landuse transition logic setting, since this means by default, full fates will run in landuse mode, which I don't think we want.  This would mean that the onus is on the user to set `use_fates_luh` to true.

Looking at the fates-side of the code, it appears we get the clearing matrix regardless of whether or not `use_fates_luh` is true (or the ruleset).  Does the transition logic setting matter for non-landuse mode?  The best I can tell it doesn't since the transition rates are zero in that case.  If not, we could create and default to a `0` option for the transition logic that would not turn on landuse mode.  We'd just need to add that as a viable case option on the fates side.  

cc @ckoven

glemieux:
FYI, issue to track adding the new namelist option to the FATES user's guide is here: https://github.com/NGEET/fates-users-guide/issues/110

bld/namelist_files/namelist_defaults_ctsm.xml
glemieux: Thinking about this more, it may be simpler to flip this and make sure that there is a default setting when `use_fates_luh` is true since this is an option necessary for land use mode.  I'll create a PR to this branch @JessicaNeedham.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -945,6 +945,13 @@ which processes the raw land use data from the THEMIS tool data sets
 (https://doi.org/10.5065/29s7-7b41)
 </entry>
 
+<entry id="fates_lu_transition_logic" type="integer" category="physics"
+        group="clm_inparm" valid_values="1,2,3,4,5,6,7,8,9">
+Select the logic for land use class transitions.
+Allowed values are  1-9. See the FATES user guide for an explanation of the options.
+(Only relevant if FATES with land use is on)
+</entry>
```
ekluzek, line 948: It's not currently in the FATES UG. And I didn't see documentation changes in the FATES PR for this (or is the FATES UG located in a different repo?).

The FATES PR does point to Table 1 in this paper:

https://gmd.copernicus.org/articles/13/3203/2020/

which does explain it. So pointing to that table might be the best thing here.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -2674,6 +2674,7 @@ lnd/clm2/surfdata_esmf/NEON/ctsm5.4.0/surfdata_1x1_NEON_TOOL_hist_2000_78pfts_c2
 <fates_spitfire_mode            use_fates=".true.">0</fates_spitfire_mode>
 <fates_spitfire_mode            use_fates=".true." use_fates_managed_fire=".true." >1</fates_spitfire_mode>
 <fates_harvest_mode             use_fates=".true.">no_harvest</fates_harvest_mode>
+<fates_lu_transition_logic      use_fates=".true.">4</fates_lu_transition_logic>
```
ekluzek: Since, it's just an integer, it would be good to have a comment to explain that 4 means

"default value of ruleset 4 above means that plants are not cleared during land use change transitions to rangeland, whereas plants are cleared in transitions to pasturelands and croplands."

(taking language from the FATES PR)

Some of the other integer values I find more acceptable, because there's a smaller list, and for example fates_spitfire_mode has a natural progression. This one is more complex and needs an explaination.

bld/namelist_files/namelist_definition_ctsm.xml
ekluzek: I think you are going for a principle of having the documentation in one place, which is a good goal. It's just that right now that place doesn't exist, so we need a source that can be looked up. It's also easier for users if it is in place and has more visibility. So it's more usable, but less maintainable. But, I also think this is a part that is unlikely to change -- except maybe on a future CMIP cycle. So even though it means duplicating some of this across HLM's the goal of making it easier for users might be OK

Although another thing that could be considered is to tell the user what FATES Fortran code file to look at. Although that's again something that can be changed out from under the documentation here. So I'm not sure the best way of doing this.

src/main/clm_varctl.F90
```diff
@@ -332,6 +332,7 @@ module clm_varctl
                                                                         ! > 1 for external data (lightning and/or anthropogenic ignitions)
                                                                         ! see bld/namelist_files/namelist_definition_clm4_5.xml for details
   logical, public            :: use_fates_managed_fire = .false.        ! true => turn on managed fire
+  integer, public            :: fates_lu_transition_logic = -9          ! controls logic around transition between land use classes
```
ekluzek, line 335: Does the code check and make sure it was changed from the default value of -9? And that it's within the right range in the Fortran code? The namelist scripts ensure it's between 1 and 9, but the Fortran code should as well.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek: @glemieux it's sounds like you'll make a PR to this branch, and we need to wait on that to bring this in. Just confirming I have that right.

ekluzek (commented):
This is great thanks @JessicaNeedham. I have comments/questions on how best to document this for CTSM users so they have ways of looking up how this works.

There are some things that will need to happen before this comes in. This will come to master as it will include a FATES update, which normally has answer changes that come with it.
- [x] @glemieux makes a PR to this one with a change he sees needs to be done
- [ ] Add in the FATES tag to the .gitmodule file
- [x] Resolve conversations
- [x] Decide who will do the testing and finalize the tag
- [ ] Go through the normal testing/tagging process

ekluzek:
@adrifoster I think this is the PR you mean. This one needs to go to master since it changes answers.

adrifoster:
> @adrifoster I think this is the PR you mean. This one needs to go to master since it changes answers.

oh okay...I thought it was B4B but that makes sense!

bld/namelist_files/namelist_definition_ctsm.xml
glemieux: I think pointing the user to the FATES user's guide should be the most direct option.  I added an issue to the FATES user's guide to make sure we get documentation on this there: https://github.com/NGEET/fates-users-guide/issues/110

src/main/clm_varctl.F90
glemieux: I've added a check in the CLMBuildNamelist to check for valid values: https://github.com/JessicaNeedham/ctsm/pull/4/changes#diff-c5398a7bf444c79516bff3ba46c0fcc5789d003af843d18b96448d2624d27b18

Is that accomplish what you're thinking about here @ekluzek?

bld/namelist_files/namelist_defaults_ctsm.xml
JessicaNeedham: @glemieux - yes that does sound better. We want it so that if use_fates_luh is true, then this namelist option needs to be set to some default value. I will go ahead and merge your PR - thanks!

bld/namelist_files/namelist_defaults_ctsm.xml
JessicaNeedham: Thanks @ekluzek, good idea. I've added a comment explaining what the default value means.

bld/namelist_files/namelist_definition_ctsm.xml
JessicaNeedham: Thanks @glemieux , I will add changes to FATES documentation to your PR.

glemieux: @ekluzek we've updated the user's guide to describe the various options via https://github.com/NGEET/fates-users-guide/pull/111.  @JessicaNeedham added the table that corresponds to the valid values.

glemieux:
> FYI, issue to track adding the new namelist option to the FATES user's guide is here: [NGEET/fates-users-guide#110](https://github.com/NGEET/fates-users-guide/issues/110)

This is resolved via NGEET/fates-users-guide#111

glemieux:
@ekluzek, my PR to @JessicaNeedham branch has been merged.  Would you give it another review when you get a chance and see if my changes look good to you?  I think this may be ready for testing (which I'm willing to run).

samsrabin:
Per SE meeting today: Don't go with @rgknox's suggestion of moving things to the namelist JSON.

glemieux:
FYI @ekluzek the fates-side testing is looking good so far: https://github.com/NGEET/fates/pull/1489#issuecomment-4119840850.  Where in the queue do you think this is?

src/main/clm_varctl.F90
ekluzek: @glemieux no actually I meant how does the Fortran code handle it if a bad value is entered? The change above in the build-namelist code isn't needed, because the list of valid_values already ensures that it has to be in the acceptable list.

ekluzek: In some cases I think it's OK to just have the build-namelist check for valid values, but sometimes the code should also do it as well. Especially if there are cases where the code running with a bad value will cause the code to abort in weird ways, or silently do the wrong thing.

I wasn't sure what kind of case this one corresponds to.

bld/namelist_files/namelist_definition_ctsm.xml
ekluzek: Cool I see it in the UG here:

https://fates-users-guide.readthedocs.io/en/latest/user/landuse-harvest-modes.html

I recommend pointing to that page directly, or saying something like "See the Namelist section of the FATES user guide".

src/main/clm_varctl.F90
glemieux: > @glemieux no actually I meant how does the Fortran code handle it if a bad value is entered? The change above in the build-namelist code isn't needed, because the list of valid_values already ensures that it has to be in the acceptable list.

I see, sorry for the misunderstanding.  The namelist option value is ultimately passed to a fates variables called `ruleset` that is used with `case select`.  If the case isn't 1-9, it'll go to the `default` case and end the run with a custom error here: https://github.com/JessicaNeedham/fates/blob/fe966fbd95a10f1f5f5626aced091c8d67e303c7/biogeochem/FatesLandUseChangeMod.F90#L264-L270

So it checks that it's in the right range, but not that it's been changed (which is current default behavior for the fates namelist value handover).

bld/namelist_files/namelist_definition_ctsm.xml
glemieux: @ekluzek thanks for the refinement.  I've updated the comment to reflect your suggestion.

src/main/clm_varctl.F90
ekluzek: OK, perfect, so the Fortran code also doesn't allow bad values and aborts if so. Resolving.

bld/CLMBuildNamelist.pm
```diff
@@ -5002,6 +5003,15 @@ sub setup_logic_fates {
                     }
                  }
               }
+              $var = "fates_lu_transition_logic";
```
ekluzek: Actually @glemieux go ahead and move this bit. It's already handled as part of being in the valid_values in namelist_definition.

glemieux:
Final `fates` suite regression testing on `izumi` and `derecho` are underway.

glemieux:
Regression testing on `derecho` and `izumi` is underway.  CC @samsrabin

glemieux:
Regression testing on `derecho`  and `izumi` against `ctsm5.4.029` is complete.  All expected tests are B4B, with known diffs on FATES testmods (due to `sci.1.91.3_api.43.1.0`). FATES sp mode tests are B4B.

Results:
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_0403-151012de`
`/scratch/cluster/glemieux/ctsm-tests/tests_0403-151127iz`

glemieux:
Changelog update.  This is ready to integrate @samsrabin.

glemieux:
Removed the namelist build check per Erik's request.  Retested  `build-namelist_test.pl` and using `./run_sys_tests ... --extra-create-test-args " --no-run"`.  All tests build successfully.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_0406-160019de`

Issue #605, opened 2019-01-07
Move settings from use_cases to namelist_defaults


billsacks:
Let me preface this by saying that I've never fully grasped the motivation for having use_cases (`bld/namelist_files/use_cases`), so I may be missing something here. But that said:

I'd like to see use cases removed and replaced with their analogous settings in the namelist_defaults xml file. My rationale is: even though use cases make sense for what they do, it's confusing and error-prone to have this additional mechanism that is just used for a handful of namelist variables. Even if the organization of namelist_defaults is less ideal for these variables, I think the upside of consistency and only needing to look in one place outweighs this downside.

I think this is actually a bigger problem because of how few variables are set via use cases. I'd say that 99% (or more) of the time when I'm messing with namelist defaults, I don't need to think about these use cases... which means that the 1% that I *do* need to look there, I forget to do so.

Definition of done:
- [ ] Simplify the use-cases to remove unneeded variables
- [ ] Have use-cases only have sim_year, sim_year_range and rcp in the use cases, and move all of the other variables that are currently in use cases to namelist_defaults (with dependence on those three).
- [ ] Move use-cases as a setting in namelist definition, and move the settings into the namelist_default file using use_case as the attribute that distinguishes the different options
- [ ] List of things in discussion below...

ekluzek:
I'm certainly willing to talk about this. It's possible the main motivation is really all around the simulation year, which we enter anyway. So it's possible all these things could be handled by looking at the sim_year or looking at if it's transient or not. It's also something where we adopted something that CAM was (and is) doing. So you could argue about the need for it in CTSM.

The principle is that the use-cases put together a set of defaults that should be used for a particular standard case. And as such the list of namelist variables is somewhat random.  And the mechanism allows any list of namelist variables to be included in the set, which makes it a convenient way to package things together. The only way to do that kind of packaging without use-cases is to have an attribute that controls it.

This is something that we've had in place for a while, so I'd like to also hear from users about this. Almost all COMPSETS that you choose, are going to pick a use-case. So users are using this even if they don't realize it.

ekluzek:
I'm going to add another wrinkle to this issue that makes use_cases difficult. The only valid attributes that can be used in the use_case files are the ones that it sets up. I was noticing it wasn't paying attention to an attribute, but that is the reason why. So that's causes a bit of an unexpected behavior. This could possibly, be it's own issue, but I'm just going to put it here for now.

billsacks:
[...]

PR #3745, opened 2026-02-05
ctsm5.4.018: Quick fix so that we can use Nitrogen Deposition sent from the ATM model

### LINKED ISSUE #3737: When coupled to CAM, ndep is still read in from CTSM rather than using what CAM is sending
### Brief summary of bug

When coupled to CAM for BGC cases ndep is being read in by CTSM rather than using what CAM is sending in.

### General bug information

**CTSM version you are using:** ctsm5.4.015 (but has always been this way)
 The CAM version that broke the behavior was: cam6_4_056 (first used in: cesm3_0_alpha06b)

**Does this bug cause significantly incorrect results in the model's science?** Yes -- for coupled cases with advanced Chemistry
For simpler cases there will be small differences due to interpolation differences. But, the data files used would be identical to only slightly different.

**Configurations affected:** When coupled to CAM

### Details of bug

We determine if we should use ndep sent from CAM by checking the drv_flds_in namelist ndep_inparm with the field ndep_nflds, from CMEPS shr_ndep_mod.F90 in the call shr_ndep_readnl. This is in advertise_fields in lnd_import_export.F90:

``` fortran
    ! from atm - nitrogen deposition
    call shr_ndep_readnl("drv_flds_in", ndep_nflds)
    if (ndep_nflds > 0) then
       call fldlist_add(fldsToLnd_num, fldsToLnd, Faxa_ndep, ungridded_lbound=1, ungridded_ubound=ndep_nflds)
       ! This sets a variable in clm_varctl
       ndep_from_cpl = .true.
    end if
```

And example coupled case is here:

https://github.com/NCAR/cesm_dev/issues/216

It's on /glade/campaign under the case: b.e30_alpha07c_cesm.B1850C_MTt4s.ne30_t232_wgx3.251.001

See the plots in:

https://github.com/ESCOMP/CMEPS/issues/627#issue-3876549256

### LINKED ISSUE #3742: Add ndep_inparm namelist to drv_flds_in
CTSM currently relies on CAM setting ndep_list in the ndep_inparm namelist in the drv_flds_in file. We didn't see a reason to be able to set this in CTSM, so we didn't enable setting this in CTSM, as we figured CAM always would. However, in cam6_4_056 (which came in with cesm3_0_alpha06b) the behavior changed so CAM no longer sets it.

So we need to be able to set ndep_list when we are coupled to CAM, which means having it added to the namelist definition and build-namelist for the CTSM version of drv_flds_in.

### LINKED ISSUE #3743: Ndep from ATM is now hard coded to 2 fields
The ndep_inparam namelist in drv_flds_in allows for different species names of up to 2 fields. But the code actually assumes that it's 2 fields for noy and nhx and adds them together for CTSM purposes.

Since this is fixed now we shouldn't allow it to be anything else. So a bit of checking on this should be done.

### LINKED ISSUE #3744: Have testmods with cam LndTuningMode use ndep from ATM
In order to help prevent problems like #3737 we need to test using ndep from the atmosphere model. So if we turn this on for all of the testmods that have some flavor of cam then we will be exercising this for ndep coming from the DATM.

We should also make sure that one or more of these tests are added to prealpha testing in case a change in CMEPS breaks CTSM.

Another part of this is to add range checking for ndep from ATM. I think it can't be identically 0. And 0 everywhere likely means something's gone wrong in the coupling. So would be a useful check.

### LINKED ISSUE #3747: Add CLM_NDEP_FROM_CPL XML variable
In order to make where NDEP is coming visible to the user (and in order to make it so there isn't duplicated information in both ATM models and CTSM) we'll add a 

CLM_NDEP_FROM_CPL xml variable to config_component.xml that build-namelist will use to determine if the drv_flds_in ndep_inparm namelist is activated.

By adding this it will at least be better documented where Nitrogen Deposition is coming from. So this could be an acceptable form for the release. Since, longer term we do want to always receive NDEP from the ATM model -- it's something that will go away. But, it would be acceptable for it to go away in ctsm6.1 and not make it into the release.

### LINKED ISSUE #3748: Difference between ndep from CTSM streams and from DATM
### Brief summary of bug

There's a significant difference in nitrogen deposition for a case that brings in NDEP from DATM than from read in from CTSM

### General bug information

**CTSM version you are using:** ctsm5.4.016-12-ge810c3744 (branch in #3745)

**Configurations affected:** BGC cases
when using ndep from DATM rather than read by CTSM

### Details of bug

In comparing the two all of the streams settings are the same between DATM and CTSM as well as the file used and the mesh file used. The only difference I saw was that in CTSM NDEP_month is used, wheras DATM uses the sum of NDEP_NOy_month and NDEP_NHx_month. 

However, NDEP_month is the sum of those two, and in comparing it with a sum of it with ncap2 I saw the difference on the order of 1.e-14. I thought of that as roundoff different but now realize for this small field I need to be more careful around this. The data is in single precision, and there might be a rounding difference from the original data that was in double.

``` shell
./create_test SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode -r . --compare ctsm5.4.015 --baseline-root /glade/campaign/cgd/tss/ctsm_baselines --walltime 00:20:00
```

``` shell
grep NDEP_TO_SMIN /glade/derecho/scratch/erik/SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode.C.20260206_172403_h7vep5/run/SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode.C.20260206_172403_h7vep5.clm2.h0a.2000-01-06-00000.nc.cprnc.out
 NDEP_TO_SMINN   (lndgrid,time)  t_index =      1     1
 RMS NDEP_TO_SMINN                    2.1687E-12            NORMALIZED  1.8103E-04
ncdiff /glade/derecho/scratch/erik/SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode.C.20260206_172403_h7vep5/run/SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode.C.20260206_172403_h7vep5.clm2.h0a.2000-01-06-00000.nc /glade/campaign/cgd/tss/ctsm_baselines/ctsm5.4.015/SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode/clm2.h0a.2000-01-06-00000.nc ncdiff.nc
```

The normalized difference is large. The absolute difference is fairly small because the field is small.

ncview of NDEP_TO_SMINN shows differences on the order of -0.3 to 0.4 e-11 when the field varies between 1e-11 to 1e-7

Image

ekluzek:
### Description of changes

Add the ndep_inparm namelist to drv_flds_in, so that CTSM can turn on ndep from the ATM model. Also turn on using the ndep from the ATM model for the cam LndTuningMode testmods. And make sure ndep comes from CAM when we are coupled to CAM.

### Specific notes

  - [x] Fixes #3742
  - [x] Fixes #3737
  - [x]  Fixes #3743
  - [x] Fixes #3744
  - [x] Fixes #3747
  - [x] Fixes #3748
  - [x] Some work on #605

Are answers expected to change (and if so in what way)? Yes (just for the cam LndTuningMode BGC tests)
   Answers change to a small amount because of small differences in interpolation and 
   sending the data through the system as well as the difference between using NDEP_month
   rather than the sum of NDEP_NOy_month+NDEP_NHx_month witn single precision data

Does this create a need to change or add documentation? Did you do so? Yes Yes

Testing performed, if any: Single cases will run aux_clm
   Currently only running SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode
   with  ndep_list = 'noy', 'nhx' at the end of user_nl_clm

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.4.018
 Originator(s): erik (Erik Kluzek,UCAR/TSS,303-497-1326)
-Date: Wed Feb 11 11:54:03 AM MST 2026
+Date: Thu Feb 12 11:28:51 AM MST 2026
 One-line Summary: Quick fix so that we can use Nitrogen Deposition sent from the ATM model
 
 Purpose and description of changes
@@ -73,7 +73,7 @@ Caveats for developers (e.g., code that is duplicated that requires double maint
 
 Changes to tests or testing:
    clm*cam*LndTuningMode testmods changed to have ndep come from the ATM model (so DATM)
-   build-namelist tester updated to accomidate the addition of CLM_NDEP_FROM_CPL
+   build-namelist tester updated to accommodate the addition of CLM_NDEP_FROM_CPL
 
 
 Testing summary: regular
@@ -83,7 +83,7 @@ Testing summary: regular
 
   build-namelist tests (if CLMBuildNamelist.pm has changed):
 
-    derecho - OK (some changes relative to ctsm5.4.018)
+    derecho - OK (some changes relative to ctsm5.4.017)
 
   regular tests (aux_clm: https://github.com/ESCOMP/CTSM/wiki/System-Testing-Guide#pre-merge-system-testing):
 
--- a/doc/ChangeSum
+++ b/doc/ChangeSum
@@ -1,6 +1,6 @@
 Tag                   Who      Date  Summary
 ============================================================================================================================
-       ctsm5.4.018     erik 02/11/2026 Quick fix so that we can use Nitrogen Deposition sent from the ATM model
+       ctsm5.4.018     erik 02/12/2026 Quick fix so that we can use Nitrogen Deposition sent from the ATM model
        ctsm5.4.017   slevis 02/09/2026 Introduce CRUJRA2024b datm option (Greenand snow hole filled)
        ctsm5.4.016 samrabin 02/06/2026 Merge b4b-dev to master
        ctsm5.4.015     erik 02/03/2026 Update CISM to latest version with answer changes


ekluzek:
OK, started aux_clm on Derecho and Izumi. I'll work on the change files and bring this in after testing looks good.

doc/ChangeLog
```diff
@@ -1,4 +1,120 @@
...
+
+
+Testing summary: regular
+----------------
+
+ [PASS means all tests PASS; OK means tests PASS other than expected fails.]
+
+  build-namelist tests (if CLMBuildNamelist.pm has changed):
+
+    derecho - OK (some changes relative to ctsm5.4.018)
```
slevis-lmwg: Should this say instead:
```suggestion
    derecho - OK (some changes relative to ctsm5.4.017)
```

doc/source/tech_note/Ecosystem/CLM50_Tech_Note_Ecosystem.rst
```diff
@@ -292,7 +292,3409 @@ The current state of the atmosphere (:numref:`Table Atmospheric input to land mo
...
 
-:sup:`4`\ The nitrogen deposition rate is required by the biogeochemistry model when active and represents the total deposition of mineral nitrogen onto the land surface, combining deposition of NO\ :sub:`y` and NH\ :sub:`x`. The rate is supplied either as a time-invariant spatially-varying annual mean rate or time-varying for a transient simulation. Nitrogen deposition rates were calculated from the same CAM chemistry simulation that generated the aerosol deposition rates.
+:sup:`4`\ The nitrogen deposition rate is required by the biogeochemistry model when active and represents the total deposition of mineral nitrogen onto the land surface, combining deposition of NO\ :sub:`y` and NH\ :sub:`x`. The rate is supplied either as a time-varying spatially-varying monthly mean rate fixed for a particular year for control simulations or advancing yearly for transient simulation. This is for the case when CTSM reads it's own datasets, but when coupled to CAM it can also use the Nitrogen deposition calculated or read in by it. For the datasets that CTSM uses Nitrogen deposition rates were calculated from the same CAM chemistry simulation that generated the aerosol deposition rates.
```
slevis-lmwg: Small, potentially helpful clarifications:
```suggestion
:sup:`4`\ The nitrogen deposition rate is required by the biogeochemistry model when active and represents the total deposition of mineral nitrogen onto the land surface, combining deposition of NO\ :sub:`y` and NH\ :sub:`x`. The rate is supplied either as a time-varying spatially-varying monthly mean rate fixed for a particular year for control simulations or advancing yearly for transient simulation. This is for the case when CTSM reads it's own datasets, but when coupled to CAM it can also use the Nitrogen deposition calculated or read in by CAM. For the datasets that CTSM uses, Nitrogen deposition rates were calculated from the same CAM chemistry simulation that generated the aerosol deposition rates.
```

doc/source/tech_note/Ecosystem/CLM50_Tech_Note_Ecosystem.rst
```diff
@@ -292,7 +292,3409 @@ The current state of the atmosphere (:numref:`Table Atmospheric input to land mo
...
+   "Acceleration of gravity", :math:`g`, 9.80616,  m s\ :sup:`-2`
+   "Standard pressure", :math:`P_{std}`, 101325, "Pa"
+   "Stefan-Boltzmann constant", :math:`\sigma`, 5.67 :math:`\times 10^{-8}`, W m :sup:`-2` K :math:`{}^{-4}`
+   "Boltzmann constant", :math:`\kappa`, 1.38065 :math:`\times 10^{-23}`, J K :sup:`-1` molecule :sup:`-1`
+   "Avogadro's number", :math:`N_{A}`, 6.02214 :math:`\times 10^{26}`, molecule kmol\ :sup:`-1`
+   "Universal gas constant", :math:`R_{gas}`, :math:`N_{A} \kappa`, J K :sup:`-1` kmol :sup:`-1`
+   "Molecular weight of dry air", :math:`MW_{da}`, 28.966, kg kmol :sup:`-1`
+   "Dry air gas constant", :math:`R_{da}`, :math:`{R_{gas} \mathord{\left/ {\vphantom {R_{gas}  MW_{da} }} \right.} MW_{da} }`, J K :sup:`-1` kg :sup:`-1`
+   "Molecular weight of water vapor", :math:`MW_{wv}`, 18.016, kg kmol :sup:`-1`
+   "Water vapor
```
slevis-lmwg: I did not read through all these other additions, but something seems wrong... Why did all these new lines (from line 296 to 3697) appear here?

slevis-lmwg (changes requested):
@ekluzek thanks for this PR! I requested some minor changes. I also asked about the strange appearance of over 3000 new lines in the .rst file.

ekluzek:
I have one test with an unexpected fail:

LILACSMOKE_D_Ld2.f10_f10_mg37.I2000Ctsm50NwpSpAsRs.derecho_intel.clm-lilac

doc/source/tech_note/Ecosystem/CLM50_Tech_Note_Ecosystem.rst
ekluzek: Yeah, this is something that happened in bringing in the changes I wanted to bring in. Someone I added repeats of the file content. I've fixed it now so only the parts that change are in.

doc/ChangeLog
ekluzek: Yes

ekluzek:
Changes to answers are to the cam LndTuningMode test cases as expected:

```
ERP_D_Ld9.f10_f10_mg37.I1850Clm60BgcCrop.izumi_nag.clm-clm60cam7LndTuningModeLDust

ERI_D.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningModeLDust
ERP_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam6LndTuningMode
ERP_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningModeLDust
ERP_D_Ld9.ne30pg3_t232.IHistClm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningModeLDust
ERP_P64x2_Ld396.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-monthly--clm-matrixcnOn_ignore_warnings            EXPECTED POSSIBILITY
ERS_D_Ld9.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode
SMS.ne30pg3_t232.I2000Clm60BgcCrop.derecho_intel.clm-clm60cam7LndTuningMode
SMS_Lm1.ne30pg3_t232.I1850Clm60BgcCropCrujraG.derecho_intel.clm-clm60cam7LndTuningMode
SMS_Lm1.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningMode
SMS_Ln9.mpasa480_mpasa480.I1850Clm50BgcCrop.derecho_gnu.clm-clm50cam6LndTuningMode
SMS_Ln9.ne30pg2_ne30pg2_mg17.I2000Clm50BgcCrop.derecho_intel.clm-clm50cam6LndTuningMode
SMS_Ln9_P256x2.C96_C96_mt232.IHistClm50BgcCrop.derecho_intel.clm-clm50cam6LndTuningMode
```

PR #3757, opened 2026-02-12
Add FATES tree recruitment scheme tests 

### LINKED ISSUE #2760: Add regression test to `fates` test suite to exercise Tree Recruitment Scheme
This is the CTSM side copy of NGEET/fates#1251.  It looks like we may have overlooked an update that broken `fates_regeneration_model = 2`.  To alleviate this in the future we should add another `fates` test to generate the appropriate parameter file update on the fly.

glemieux:
### Description of changes

This simple pull request adds tests for the two recruitment namelist options for the [FATES tree recruitment scheme](https://fates-users-guide.readthedocs.io/projects/tech-doc/en/latest/fates_tech_note.html#environmentally-sensitive-tree-recruitment).  There are two different modes for turning on the scheme, which now have a test.  No FATES parameter file modifications are necessary to activate the test.  

Resolves #2760.

Note that there is an known FATES mass balance issue, https://github.com/NGEET/fates/issues/1246, that might be triggered by the `FatesColdTRS` testmod.  This will be added to the expected fail list if it comes up during testing.

### Specific notes

  Fixes #2760 

Testing performed, if any: ran fates test list

### DIFF
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -1893,11 +1893,6 @@ sub process_namelist_inline_logic {
   #####################################
   setup_logic_canopy($opts,  $nl_flags, $definition, $defaults, $nl);
 
-  ########################################
-  # namelist group: soilhydrology_inparm #
-  ########################################
-  setup_logic_hydrology_params($opts,  $nl_flags, $definition, $defaults, $nl);
-
   #####################################
   # namelist group: irrigation_inparm #
   #####################################
@@ -3349,28 +3344,6 @@ sub setup_logic_supplemental_nitrogen {
         $log->warning("There is no need to use a bgc_spinup mode when supplemental Nitrogen is on for all PFT's, as these modes spinup Nitrogen" );
       }
     }
-  }
-}
-
-#-------------------------------------------------------------------------------
-
-sub setup_logic_hydrology_params {
-  #
-  # Logic for hydrology parameters
-  #
-  my ($opts, $nl_flags, $definition, $defaults, $nl) = @_;
-
-  my $lower = $nl->get_value( 'lower_boundary_condition'  );
-  my $var   = "baseflow_scalar";
-  if ( $lower == 1 || $lower == 2 ) {
-     add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl,
-                 $var, 'lower_boundary_condition' => $lower );
-  }
-  my $val   = $nl->get_value( $var );
-  if ( defined($val) ) {
-     if ( $lower != 1 && $lower != 2 ) {
-        $log->fatal_error("baseflow_scalar is only used for lower_boundary_condition of flux or zero-flux");
-     }
   }
 }
 
@@ -4866,18 +4839,6 @@ sub setup_logic_atm_forcing {
       }
    }
 
-   foreach $var ("precip_repartition_glc_all_snow_t",
-                 "precip_repartition_glc_all_rain_t",
-                 "precip_repartition_nonglc_all_snow_t",
-                 "precip_repartition_nonglc_all_rain_t") {
-      if ( &value_is_true($nl->get_value("repartition_rain_snow")) ) {
-         add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, $var);
-      } else {
-         if (defined($nl->get_value($var))) {
-            $log->fatal_error("$var can only be set if repartition_rain_snow is true");
-         }
-      }
-   }
 }
 
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


glemieux:
Running the TRS test mods on `derecho` currently.

glemieux:
The `FatesColdTRSNoSeedDynamics` runs successfully, but the `FatesColdTRS` test mod is failing with an issue different from  https://github.com/NGEET/fates/issues/1246.  I'll create a new issue on the fates side and update the expected failure list.

Alternatively, we could attempt to fix the issue and turn this PR into an API update associated with whichever FATES tag fixes the TRS issue.

glemieux:
TO DO:

- [x] Review history output to add to test
- [x] Add expected failure

glemieux:
Running `aux_clm` regression tests against `ctsm5.4.021`.

glemieux:
Testing on `izumi` against `ctsm5.4.021` is B4B on all expected tests.

Testing on `derecho` against the same baseline is showing almost all expected tests are B4B with the following exceptions:

```
pr3757-aux_clm-ctsm54021_int: 208 tests
    PEND ERP_P256x2_Ld30.f45_f45_mg37.I2000Clm60FatesRs.derecho_intel.clm-mimicsFatesCold COMPARE_base_rest
    FAIL RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput RUN time=70
    FAIL SUBSETDATAPOINT_Ld5_D_Mmpi-serial.CLM_USRDAT.I2000Clm60BgcCropCrujra.derecho_intel.clm-default RUN time=24
    FAIL SUBSETDATAREGION_Ld5_D_Mmpi-serial.CLM_USRDAT.I2000Clm60BgcCropCrujra.derecho_intel.clm-default RUN time=92
```
@ekluzek @slevis-lmwg did you see these fail during the testing for the last `b4b-dev` tag?  Apologies in advance if I missed a conversation in the chat or elsewhere about these.

Results (`izumi`): `/home/glemieux/scratch/ctsm-tests/tests_0222-215030iz`
Results (`derecho`): `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr3757-aux_clm-ctsm54021`

slevis-lmwg:
> Testing on `derecho` against the same baseline is showing almost all expected tests are B4B with the following exceptions:
> 
> ```
> pr3757-aux_clm-ctsm54021_int: 208 tests
>     PEND ERP_P256x2_Ld30.f45_f45_mg37.I2000Clm60FatesRs.derecho_intel.clm-mimicsFatesCold COMPARE_base_rest
>     FAIL RXCROPMATURITYSKIPGEN_Ld1097.f10_f10_mg37.IHistClm60BgcCrop.derecho_intel.clm-cropMonthOutput RUN time=70
>     FAIL SUBSETDATAPOINT_Ld5_D_Mmpi-serial.CLM_USRDAT.I2000Clm60BgcCropCrujra.derecho_intel.clm-default RUN time=24
>     FAIL SUBSETDATAREGION_Ld5_D_Mmpi-serial.CLM_USRDAT.I2000Clm60BgcCropCrujra.derecho_intel.clm-default RUN time=92
> ```
> 
> @ekluzek @slevis-lmwg did you see these fail during the testing for the last `b4b-dev` tag? Apologies in advance if I missed a conversation in the chat or elsewhere about these.

I confirmed that the RX and SUBSETDATA tests passed in ctsm5.4.021 by looking at Teststatus here:
`/glade/campaign/cgd/tss/ctsm_baselines/ctsm5.4.021`

glemieux:
Looking at the cesm.log error message it appears to be an instance of #3590 as I'm seeing the same thing as mentioned in https://github.com/ESCOMP/CTSM/issues/2930#issuecomment-3576790263.  I'll try rerunning these without defining a custom test-id.

glemieux:
> Looking at the cesm.log error message it appears to be an instance of #3590 as I'm seeing the same thing as mentioned in [#2930 (comment)](https://github.com/ESCOMP/CTSM/issues/2930#issuecomment-3576790263). I'll try rerunning these without defining a custom test-id.

Looks like this is the cause of the issue I was seeing.  Re-running these tests individually with the default test-id resulted in all three previous run failures passing b4b.

cime_config/testdefs/testmods_dirs/clm/FatesColdTRS/user_nl_clm
```diff
@@ -0,0 +1,27 @@
+fates_regeneration_model = 'trs'
+hist_fincl1       = 'FATES_NCOHORTS', 'FATES_TRIMMING', 'FATES_AREA_PLANTS',
```
ekluzek, line 2: @glemieux just one question here. This is a really long list of variables to add. It seems like maybe one of three things should happen here:

1.) Most of these should be switched to default="active" when fates_regenerate_model=='trs'
2.) We should maybe turn on all output -- and just exclude a few (if the list would be shorter)
3.) Maybe just run with all output on? It doesn't hurt to have extra output...

glemieux: These are basically a duplicate of the `Fates` user_nl_clm list of variables, plus an addition of seed specific variables that aren't included in that `Fates` list which the TRS testmods include.  Some of these are default active and some are inactive.  We could potentially add the seed variables to the `Fates` testmod instead and reduce the code duplication.  I don't have strong inclination one way or the other as its only adding 8 variables.

Note that we empty the history tapes in `Fates` to exclude CLM history, I'm assuming so as to reduce the runtime cost.  I think the same decision applied to your suggestions in 2 and 3 above.

I like your idea in suggestion 1.  Having FATES set the active flag for history based on the run mode sounds like a good improvement.  That said, I think it's out of scope for this PR.

ekluzek: Made this an issue #3811

ekluzek:
I sent the fates testlist to check on this. But unfortunately the tests have just been waiting in the queue.

PR #3761, opened 2026-02-14
Use pytest instead of unittest in run_ctsm_py_tests

### LINKED ISSUE #3616: Convert Python unit/system tests from unittest to pytest
Will enable us to get rid of our test helper script, along with a lot of other quality-of-life improvements.

samsrabin:
### Description of changes

This PR changes `run_ctsm_py_tests` to use `pytest` instead of `unittest`, enabling many modern features. One test file, `test_unit_longitude.py`, has been modified to use `pytest`'s "parametrize" functionality as an example. Most of the changed lines in that file are just whitespace differences.

Existing `run_ctsm_py_tests` arguments have been ported to replicate the same functionality in `pytest`. Unknown arguments will be passed to `pytest` directly.

Also: Made it so that the tests can be called from anywhere in the checkout, instead of only `python/`.

### Specific notes

**CTSM Issues Fixed:**
- Resolves #3616

**Testing performed, if any:** Python unit and system tests pass. New version finds the same number of tests as before.

### DIFF
--- a/python/ctsm/run_ctsm_py_tests.py
+++ b/python/ctsm/run_ctsm_py_tests.py
@@ -10,6 +10,7 @@ import argparse
 import logging
 import io
 import contextlib
+import re
 
 import pytest
 
@@ -158,7 +159,7 @@ def get_pytest_help_item(pytest_help_text: str, option: str) -> str:
     Extract a single pytest CLI option (e.g. "--debug") from the full ``pytest --help`` output and
     collapse its multi-line description into a single line.
 
-    The function locates the first line containing ``option`` and then collects all immediately
+    The function locates the first line starting with ``option`` and then collects all immediately
     following lines that are more indented than that header line. These indented lines are treated
     as the option's description. The first line break is replaced with a colon and the description
     lines are joined with single spaces.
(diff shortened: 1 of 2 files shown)


python/ctsm/run_ctsm_py_tests.py
```diff
@@ -4,83 +4,215 @@
...
 
-    output_level.add_argument(
-        "-d", "--debug", action="store_true", help="Run tests with even more verbosity"
+    parser.add_argument(
+        "-d",
+        "--debug",
+        nargs="?",
+        type=str,
+        const="",
+        metavar="DEBUG_FILE_NAME",  # Sam as pytest
```
samsrabin: Fix "Sam" typo.

python/ctsm/run_ctsm_py_tests.py
```diff
@@ -4,83 +4,215 @@
...
+    return debug_help
+
+
+def get_pytest_help_item(pytest_help_text: str, option: str) -> str:
+    """
+    Extract a single pytest CLI option (e.g. "--debug") from the full ``pytest --help`` output and
+    collapse its multi-line description into a single line.
+
+    The function locates the first line containing ``option`` and then collects all immediately
+    following lines that are more indented than that header line. These indented lines are treated
```
samsrabin, line 163: - [x] Check: I think we actually want the first line STARTING with `option` (ignoring leading whitespace).
- [x] Add testing.

slevis-lmwg (approved):
@samsrabin and I did a walkthrough. Thanks for this work, Sam, looks good!

Sam was going to improve comments on a certain method. Sam posted one or two other reminders for himself. 

One thing I forgot to mention in our meeting:
You pointed out a print statement at some point, which made me think that elsewhere we try to use logger tools. I'm not pushing for that, just mentioning.

samsrabin:
- [x] Use logging instead of `print()`

python/ctsm/run_ctsm_py_tests.py
samsrabin: Done with b105037d7.

python/ctsm/run_ctsm_py_tests.py
samsrabin: Done with 948cbd604.

PR #3776, opened 2026-02-19
b4b-dev: Update SpinupStability_BGC_v12_SE.ncl to work for FATES and ne16pg3

slevis-lmwg:
### Description of changes
See title.

### Specific notes

@adrifoster @olyson 

I think none.

No.

Does it? I didn't.

Testing performed, if any:
Tried the script on a couple of Adrianna's fates simulations.


slevis-lmwg:
@adrifoster this review should take <10 minutes. It contains my changes to the SpinupStability script for ne16 and for FATES. I'm not sure if we're done modifying it, but we could also consider it done for now and open another PR if more changes become necessary.

Issue #3775, opened 2026-02-19
NaN's in _FillValue of some input datasets


ekluzek:
### Brief summary of bug

Some of our input datasets have NaN's as the _FillValue, which causes problems with the ifx compiler. But, as a general practice it isn't good to have a NaN as a _FillValue since by nature you need to compare the values of a field to _FillValue and comparison to NaN's always return false.

### General bug information

**CTSM version you are using:** ctsm5.4.016-45-g157a07e51
**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** With ifx compiler
### Details of bug

Files in the cropdata directory and the Carbon isotope files have issues:

The list from @billsacks:

- various files in the cropdata directory: some of these have caused problems in testing
- lnd/clm2/isotopes/[ctsmforc.Graven.atm_delta_C14_CMIP7_360x720_1700-2023_yearly_v3.0_tweaked_latlons_c260108.n](http://ctsmforc.graven.atm_delta_c14_cmip7_360x720_1700-2023_yearly_v3.0_tweaked_latlons_c260108.nc/)

ekluzek:
The other type I found in #3769 was this test:

ERP_D_P64x2_Ld10.f10_f10_mg37.I2000Clm60Bgc.derecho_intel.clm-Hillslope

And it looks like it might be the file as it has NaN for _FillValue:

$DIN_LOC_ROOT/lnd/clm2/testdata/surfdata_10x15_hist_1850_78pfts_c240216.synth_hillslopes_241001.nc

slevis-lmwg:
I get the sense that this work is complete, so I will close the issue. Let me know if I'm mistaken.

Issue #3764, opened 2026-02-17
NaN values in some input files cause problems with ifx compiler


billsacks:
### Brief summary of bug

As found by @jedwards4b - `SMS_D.ne30pg3_ne30pg3_mt232.FHISTC_LTso.derecho_intel` is failing when updating to ccs_config_cesm1.0.75, which switches to the ifx compiler, due to some input files having NaN values. Based on consultation with @ekluzek we should change these files to have no NaN values.

### General bug information

**CTSM version you are using:** ctsm5.4.003

**Does this bug cause significantly incorrect results in the model's science?** Unknown, but currently believed to be No.

**Configurations affected:** Unknown

### Details of bug

From @jedwards4b - 

> I was looking at 
> 
> and it's crashing in dshr_strdata_mod.F90 on this line:
> if (stream%mapalgo == 'none' .and. any(data_dbl1d == fillvalue_r8)) then 
> 
> So I looked at the data in the file that it's trying to read and the file has NaN as a fill value:
> 
>   double p_ac_TBD(time, lat, lon) ;
>   p_ac_TBD:_FillValue = NaN ;
>   p_ac_TBD:long_name = "Air-conditioning penetration rate for Tall Building District (TBD) class" ;
>   p_ac_TBD:prediction_method = "Assume all 1; filled" ;
>   p_ac_TBD:units = "-" ;
> 
> I then tried a run  where I convert the NaN _FillValues to 1.e38 it got past that file and failed in a later 
> stream. 

and then:

> The next one is even trickier because the _FillValue is 
> double Z0a(time, lat, lon) ;
> Z0a:units = "cm" ;
> Z0a:_FillValue = 1.e+32 ;
> Z0a:long_name = "Prigent et al. (2005) 0.25x0.25 aeolian roughness length in cm" ;
> 
> But if you ncdump the file you see that the field is full of NaN values anyway - probably over ocean points?

### Important details of your setup / configuration so we can reproduce the bug

Testing was done in the context of CESM at `68fe606`, which uses ccs_config_cesm1.0.75.

Also pinging @briandobbins on this.

ekluzek:
@billsacks sent an email to CSEG about this, that I'm capturing the relevant part here for discussion

> [The above changes] ... allowed the prealpha I compset tests to pass. However, when I ran the more extensive aux_clm test suite, I ran into some additional issues. The first two I found were:
> - Some crop-related files also have NaNs as _FillValue and/or in data values
[...]

PR #3786, opened 2026-02-27
ctsm5.3.022: Bring in changes needed for the ifx compiler

### LINKED ISSUE #3794: Update to ccs_config1.0.75 with move to ifx compiler
Update to ccs_config1.0.75 which includes the move to the ifx compiler.

ekluzek:
### Description of changes

Updates ccs_config to 1.0.75 which has the new ifx compiler in use for derecho_intel, as well as updating to using ESMF8.9.0. This exposed problems in the code on two fronts when running with DEBUG mode:
- Reliance on short-circuiting in if statements
- Problems with NaN's, in datasets especially _FillValue

For the former, we changed code to break up if statements so this would work. And with the latter we point to new datasets with the NaN's handled. mosart and cdeps code also had short-circuiting problems.

Includes updating submodules: ccs_config, cdeps, mosart

### Specific notes

Contributors other than yourself, if any: @samsrabin @billsacks 

   work on #3764 
   work on #3775
   Fixes #3794

   We expect answers to be roundoff level different. But, there could be compiler bugs that create optimization issues beyond that.
  So we do need to evaluate to make sure the climate is the same

   input datasets changed to remove NaN's especially in _FillValue

Testing performed, if any: I ran almost all of the test lists on Derecho in the starting PR
  There are fails that need to be resolved before this comes in
  The CDEPS tag also needs to point to a tag rather than a personal branch

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 ===============================================================
 Tag name: ctsm5.4.022
 Originator(s): erik (Erik Kluzek)
-Date: Tue Mar  3 22:41:16 MST 2026
+Date: Wed Mar  4 10:52:05 AM MST 2026
 One-line Summary: Change to using the ifx compiler and some more fixes related to it
 
 Purpose and description of changes
--- a/doc/ChangeSum
+++ b/doc/ChangeSum
@@ -1,6 +1,6 @@
 Tag                   Who      Date  Summary
 ============================================================================================================================
-       ctsm5.4.022     erik 03/03/2026 Change to using the ifx compiler and some more fixes related to it
+       ctsm5.4.022     erik 03/04/2026 Change to using the ifx compiler and some more fixes related to it
        ctsm5.4.021     erik 02/20/2026 Bring b4b-dev to master
        ctsm5.4.020   slevis 02/19/2026 Address carbon cycle and fire biases in CESM3 development runs by updating namelist defaults
        ctsm5.4.019 samrabin 02/12/2026 Fix years used by generate_gdds


ekluzek:
We will most likely bring this in as it is at this point. There are more changes that will need to come in later, but are needed for specific CTSM configurations and not the coupled model.

ekluzek:
Whoops wrong PR for the last message,

billsacks:
@ekluzek and @samsrabin - thank you for your work on this! Was there separate testing done to check whether these changes were bit-for-bit with baselines that used the same compiler? i.e., my understanding is that there were two sets of changes here:

1. Fixes needed for aux_clm to pass with ifx
2. Updating the compiler environment

I know that (2) changes answers. Was there testing to check if (1) also changes answers or is bit-for-bit?

ekluzek:
> @ekluzek and @samsrabin - thank you for your work on this! Was there separate testing done to check whether these changes were bit-for-bit with baselines that used the same compiler? i.e., my understanding is that there were two sets of changes here:
> 
> 1. Fixes needed for aux_clm to pass with ifx
> 2. Updating the compiler environment
> 
> I know that (2) changes answers. Was there testing to check if (1) also changes answers or is bit-for-bit?

Yes, in @samsrabin PR #3787 he showed that the filename changes didn't change answers with the baseline with just the ifx compiler update. I showed that "2" changed answers and saved that one as a baseline that Sam used to make sure "1" was b4b.

billsacks:
Sounds good - thanks for clarifying.

PR #3787, opened 2026-03-01
ctsm5.4.022: Replace some netCDF inputs that have NaN fill values

samsrabin:
### Description of changes

For all netCDF input files that have NaN fill values, make new versions without NaN fill values and use those instead.

~~This PR will not include work on tower site input netCDF files.~~ None of those netCDF files have NaN fill values anyway.

### Specific notes

**CTSM Issues Fixed:**
- Contributes to #3764
- Contributes to #3775

**Are answers expected to change (and if so in what way)?** No, not relative to the latest on the `ifx` branch.

**Any User Interface Changes (namelist or namelist defaults changes)?** Yes, many default netCDF files have been replaced.

**Does this create a need to change or add documentation? Did you do so?** None

Derecho:
- [x] `aux_clm`: See #3788.
- [x] `fates`: Tests not included above: See #3789
- [x] `ctsm_sci`: Tests not included above: See #3790 and #3792
- [x] `crop_calendars`: Tests not included above: All okay.

Izumi:
- [x] `aux_clm`
- [x] `fates`

### DIFF
--- a/cime_config/testdefs/ExpectedTestFails.xml
+++ b/cime_config/testdefs/ExpectedTestFails.xml
@@ -60,6 +60,72 @@
       <issue>#2784</issue>
     </phase>
   </test>
+  <test name="SMS_Ld5.f09_g17.ISSP245Clm50BgcCrop.derecho_intel.clm-ciso_dec2050Start">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3790</issue>
+    </phase>
+  </test>
+  <test name="SMS_Ld5.f09_g17.ISSP370Clm50BgcCrop.derecho_intel.clm-ciso_dec2050Start">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3790</issue>
+    </phase>
+  </test>
+  <test name="SMS_Ld5.f09_g17.ISSP585Clm50BgcCrop.derecho_intel.clm-ciso_dec2050Start">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3790</issue>
+    </phase>
+  </test>
+  <test name="SMS_Ld5.f09_t232.ISSP245Clm60BgcCropCrujra.derecho_intel.clm-ciso_dec2050Start">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3790</issue>
+    </phase>
+  </test>
+  <test name="SMS_Ld5.f09_t232.ISSP370Clm60BgcCropCrujra.derecho_intel.clm-ciso_dec2050Start">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3790</issue>
+    </phase>
+  </test>
+  <test name="SMS_Ld5.f09_t232.ISSP585Clm60BgcCropCrujra.derecho_intel.clm-ciso_dec2050Start">
+    <phase name="RUN">
+      <status>FAIL</status>
+      <issue>#3790</issue>
+    </phase>
+  </test>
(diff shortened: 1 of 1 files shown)


bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -148,7 +148,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <atm_c14_filename use_c14=".true." use_c14_bombspike =".true." ssp_rcp="SSP5-8.5" cmip_era="cmip6">lnd/clm2/isotopes/atm_delta_C14_CMIP6_SSP5B_3x1_global_1850-2100_yearly_c181209.nc</atm_c14_filename>
 
 <!-- Settings for Carbon Isotope streams -->
-<stream_fldfilename_atm_c14 use_c14=".true." use_c14_bombspike =".true." ssp_rcp="hist" cmip_era="cmip7">lnd/clm2/isotopes/ctsmforc.Graven.atm_delta_C14_CMIP7_360x720_1700-2023_yearly_v3.0_tweaked_latlons_c260108.nc</stream_fldfilename_atm_c14>
+<stream_fldfilename_atm_c14 use_c14=".true." use_c14_bombspike =".true." ssp_rcp="hist" cmip_era="cmip7">lnd/clm2/isotopes/ctsmforc.Graven.atm_delta_C14_CMIP7_360x720_1700-2023_yearly_v3.0_tweaked_latlons_c260108.no_nan_fill.nc</stream_fldfilename_atm_c14>
```
ekluzek, line 151: I think I'm OK (ish) with the naming convention here <previous_name_and_date>.no_nan_fill.nc. It does make it obvious what the change was. But, I'm also not completely convinced, so I'll talk it through...

**Prefer creation dates at the end:**
Personally I prefer the creation date to be at the end just before the extension type (.nc). Overall we've been really consistent with that for CTSM. The exceptions are some ESMF mesh files where the date started appearing in the middle, but I'm not sure why that started happening. When I know I can rely on the form of the creation date I can search for file with wildcards, which does help in finding things.

**Can't tell modification date:**
With this I can't tell the date on when the files were modified -- except by doing an ncdump by seeing the history attribute. After they are rimported, we'll be able to see the GDEX date for it. But, when it's copied to another machine or re-downloaded, you won't see the date. I think it would also be more confusing if the file had two dates -- one for the original data and one for the small modification made. I've figured just that last date to be important.

**Future update names?:**
It also doesn't make it clear on what a future update should look like. Do we keep adding descriptors to the filenames? Sometimes changes even tweaks are hard to describe in the filename. And I don't think we want to keep adding descriptors to the filenames. We sometimes did that for parameter files, but gave up at some point. The advantage of having the creation date be the only thing that changes at the end is that the filenames don't have to continue to grow. If the changes are described in the history attribute of the file you can see what happened at each step with ncdump. And the progression of creation dates of files in the directory with the same name shows you when changes happened.

So in summary the PRO/CONs in the naming convention that I see:
- CON: The filename doesn't tell me when the file was most recently changed
- PRO: It does make it clear what the original data was and the one thing changed
- CON: It's not clear what the filename should change to for a future small change
- CON: I don't think we always want to keep adding descriptors to filenames, I think it's less helpful beyond a few
- CON/PRO: I prefer the creation date at the end per our convention / But, here it does make the lineage of the file clear

samsrabin: This is only a one-time thing. Going to keep it because it'd be an enormous pain to change at this point.

ekluzek (approved):

Issue #3130, opened 2025-05-14
Some FATES-SP configurations are calling patch%itype


rgknox:
The test: 
SMS_D.1x1_brazil.I2000Clm60FatesSpCruRsGs.derecho_gnu.clm-FatesColdDryDepSatPhen 

 generates a run fail with the following error message:

```
 At line 879 of file /glade/u/home/rgknox/ctsm/src/main/subgridWeightsMod.F90
dec2438.hsn.de.hpc.ucar.edu 0: Fortran runtime error: Index '-9998' of dimension 2 of array 'subgrid_weights_diagnostics%pct_nat_pft' below lower bound of 1
dec2438.hsn.de.hpc.ucar.edu 0:
```

Here is line [879](https://github.com/ESCOMP/CTSM/blob/ctsm5.3.043/src/main/subgridWeightsMod.F90#L879)

This issue is newly generated with issue #2933 and PR: #2935 , that PR sets itype to a bogus value that will trigger this type of error, so that it is not inappropriately used.

When fates is active, the itype should never be used, even in SP. Yes, with FATES-SP we do associate patches with a specific PFT, but FATES does not follow the same PFT indexing convention that is found in the surface file, so associating a patch with a pft on the clm-side is meaningless and impossible.

rgknox:
I was able to bypass the previous error by removing the call to have PCT_NAT_PFT in the diagnostics.  The ERP_P128x2_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen test still generates the following error further in the call sequence though:

```
dec0068.hsn.de.hpc.ucar.edu 9: forrtl: error (65): floating invalid
dec0068.hsn.de.hpc.ucar.edu 9: Image              PC                Routine            Line        Source             
dec0068.hsn.de.hpc.ucar.edu 9: libpthread-2.31.s  000014733E3D08C0  Unknown               Unknown  Unknown
dec0068.hsn.de.hpc.ucar.edu 9: cesm.exe           0000000002B39517  satellitephenolog         521  SatellitePhenologyMod.F90
dec0068.hsn.de.hpc.ucar.edu 9: cesm.exe           0000000002B304D2  satellitephenolog         303  SatellitePhenologyMod.F90
dec0068.hsn.de.hpc.ucar.edu 9: cesm.exe           0000000000AAEEBD  clm_initializemod         711  clm_initializeMod.F90

```

[...]

PR #3812, opened 2026-03-12
Fix running fates with HIST/SSP compsets.

mvdebolskiy:
### Description of changes

Enforce `irrigate=.false.` when running with fates.

Putting this on bfb, since the fix will not affect the current test-suites.

### Specific notes

Since `ctsm5.3.046`: `patch%itype` is set to -999, which produces out-of-bounds error, when running HIST with fates, since the defaults point to `irrigate=.true.`.

Though there is `irrigate=.true.` in use_cases, all the statements with `.true.` value have `use_crop=.true.` which can not be true when fates is on.

  Fixes NorESMhub/CTSM#204
  Works on #3130 

No, since there not tests with HIST and FATES in the test-suite.

`irrigate=.false.` when `use_fates=.true.`.

No/No

Testing performed, if any:

Testsuits don't cover this bug, so I ran: 
```
SMS_D_Ld1.f45_f45_mg37.HIST_DATM%GSWP3v1_CLM60%FATES_SICE_SOCN_SROF_SGLC_SWAV.derecho_intel.clm-FatesCold
SMS_D_Ld1.f45_f45_mg37.2000_DATM%GSWP3v1_CLM60%FATES_SICE_SOCN_SROF_SGLC_SWAV.derecho_intel.clm-FatesCold
SMS_D_Ld1.f45_f45_mg37.IHistClm60BgcCrop.derecho_gnu.clm-default
```
All passed. Irrigate was `.false.`,`.false.`,`.true` respectively.


src/main/clm_driver.F90
```diff
@@ -822,6 +823,10 @@ subroutine clm_drv(doalb, nextsw_cday, declinp1, declin, rstwr, nlend, rdate, ro
 
        if (irrigate) then
 
+          if (use_fates) then
+            call endrun(msg=' ERROR: somehow use_fates and irrigate are both set to TRUE ' // &
```
slevis-lmwg: @mvdebolskiy you probably wrote this error message with the idea that CLMBuildNamelist should have caught the problem before runtime. I would reword to a more user-friendly message, such as, "CLMBuildNamelist should not have allowed use_fates = .true. and irrigate = .true. in your namelist at the same time. Get around this error by setting one of the two to .false. in your user_nl_clm."

slevis-lmwg (changes requested):
Thanks @mvdebolskiy. I'm requesting a change to an error message.

ekluzek:
I was skeptical that we didn't have irrigate==TRUE. But, @mvdebolskiy did point out why that is guaranteed since the code aborts with an itype error. However, in any case I made sure FATES tests in aux_clm and fates testlists on Derecho didn't set irrigate==TRUE and they don't. So this is something that can come into b4b-dev.

Thanks for working on this @mvdebolskiy this will be good to bring in.

We should also reference the issue #3130 here. I'll add that to the top.

src/main/clm_driver.F90
mvdebolskiy: @slevis-lmwg I've added your suggestion, though got rid of "CLMBuildNamelist should not have allowed" since it makes it take too many lines in the log and look messy.

slevis-lmwg (approved):

ekluzek (approved):

slevis-lmwg:
@mvdebolskiy I forget whether you have permission to merge this yourself following instructions in https://github.com/ESCOMP/CTSM/wiki/b4b%E2%80%90dev-branch-procedures
or you need one of us (e.g. me) to do it.

wwieder:
This just needs testing now & merge to b4b-dev, @slevis-lmwg will do it

slevis-lmwg:
OK `./build-namelist_test.pl` on derecho
OK `./run_sys_tests -s aux_clm -c ctsm5.4.031 --skip-generate` on derecho and izumi
except derecho had two tests give the `ERROR in SparseMatrixMultiplyMod.F90 at line 973` so I'm rerunning them.

Issue #2169, opened 2023-09-22
Build-namelist logic to find the finidat file is fragile and needs to be redone


ekluzek:
In working on #2008 I'm finding that the logic to fine the finidat file is: complex, hard to understand, fragile, and will require a bunch of duplication. This is another reason for doing #585 which is the long term solution. Although there we should make sure the redesign in python has a better method of getting the finidat files.

This is similar to #549, but slightly different and focussed on just the namelist logic and the namelist default settings for: use_init_interp and init_interp_attributes. This is also related to #2149, as we need to have duplicated lnd_tuning_mode options for all the different clm physics options, as well as the cam options. Right now this is going to require a bunch of duplication.

ekluzek:
The other error that can happen is:

`use_init_interp is set BUT finidat is NOT, need to set both`

which can happen because the user explicitly set use_init_interp and didn't set finidat on their own. Or it can happen because the default settings aren't setup correctly. This should be improved as well.

ekluzek:
Working on this now as part of #2501 I see some improvements that could be made:

- Remove glc_nec and ic_tod (they are all 10 and 0)
- Possibly use one of ic_ymd and sim_year
- Replace bgc (sp, bgc, fates) for use_cn, use_cndv, use_fates) use_cndv can be removed as part of #2141
- use one of use_crop and maxpft

This would reduce the number of attributes needed to specify a finidat file from 15 to 9.

ekluzek:
Adding to this: https://github.com/ESCOMP/CTSM/pull/2821/files#r1797496942

so that 1979 and 2000 finidat files will work with do_transient_pfts=.true.

ekluzek:
One thing that's problematic with the current setup is that the following variables need to be handled together and synchronized appropriately:

- init_interp_sim_years
- use_init_interp
- init_interp_attributes
- finidat

A better way to handle this might be to do some checking of the namelist_default file to make sure it's constructed correctly keeping them all in sync.

Also in many cases the above items are copied and don't need to be. Also it can be constructed so that finidat files are NOT copied, and only init_interp_attributes and use_init_interp need to be copied for some cases. So that's another thing that could be checked for.

A checking script for this purpose could be implemented in python, which would make it more likely to be useful when we do move away from perl for build-namelist. It could also be run as a github action, which would catch these issues as they come in with each PR.

PR #3842, opened 2026-03-25
b4b: Work on build-namelist Initial Condition file (finidat) picking

### LINKED ISSUE #3832: Put use_init_interp in $nl_flags as either .true. or .false. so can be correctly handled in namelist_defaults_ctsm.xml
Logical attributes in namelist_defaults files need to have been converted to the $nl_flags dictionary so that they will be either .true. or .false. so that the comparison works correctly. Logical values are allowed to be different case, and shortened with or without the "." in front and after. As such they should be stored in $nl_flags in the two options, based on the logical value.

use_init_interp isn't handled this way, and as such if the user enters "T" or "TRUE" the namelist_defaults matches might not work correctly.

ekluzek:
### Description of changes

Some work on the initial condition picking in build-namelist. This makes it more robust by making sure use_init_interp is set in nl_flags as strictly either .true. or .false. so it can be used correctly in getting values from namelist-defaults.

Part of what is happening here is that a new subroutine to set the current iteration of use_init_interp is added

setup_preliminary_use_init_interp

which is called a few different places and called more than once. So it makes the handling of use_init_interp as it plays out to be more robust.

### Specific notes

  Fixes #3832   
  Work on #2169

Are answers expected to change (and if so in what way)? No, unless there is a bug in finidat picking

Testing performed, if any: will do regular


ekluzek:
This is blocked by needing ctsm5.4.027 to come to b4b-dev first.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -768,11 +768,11 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 
 <!--
 
-    Find Initial condition files: The settings use_init_interp, init_interp_attributes and finidat MUST all
-    be coordinated together!
-
-    NOTE: And they need to be coordinated with clm_start_type that's in namelist_defaults_overall.xml
-
+    Find Initial condition files: The settings use_init_interp, init_interp_attributes and finidat MUST allundef;
```
ekluzek, line 768: Looks like these changes should be in here.

ekluzek:
Because problems with #2169 are causing problems in code development this should be prioritized to come in soonish. And probably more work should be done ahead of having all of the finidat files in place so it can be done in a fairly reasonable way.

Issue #3360, opened 2025-07-25
Race condition with a threaded test with matrix on?


ekluzek:
### Brief summary of bug

In ctsm5.3.065 I had the test

ERP_P64x2_D_Ld5.f10_f10_mg37.I1850Clm50Bgc.derecho_intel.clm-ciso--clm-matrixcnOn_ignore_warnings

fail at runtime. But, after resubmitting it worked fine.

This is just to bring awareness that this is a potential problem and a simple resubmit solves it.

### General bug information

**CTSM version you are using:** ctsm5.3.064-72-g4037e5a8d

**Does this bug cause significantly incorrect results in the model's science?** no

**Configurations affected:** threading with soil matrix on

### Details of bug

It dies in a SHR_ASSERT statement that's ensuring the matrix size of two sparse matrices multiplied together are the same size.

### Important output or errors that show the problem

cesm.log:
```
dec0933.hsn.de.hpc.ucar.edu 63:  mosart decomp info proc =        63 begr =    255151 endr =    259200 numr =      4050
dec0933.hsn.de.hpc.ucar.edu 21:  ERROR in SparseMatrixMultiplyMod.F90 at line 973
dec0933.hsn.de.hpc.ucar.edu 21: Image              PC                Routine            Line        Source
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000004842AF1  shr_abort_mod_mp_         110  shr_abort_mod.F90
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000004842A83  shr_abort_mod_mp_          65  shr_abort_mod.F90
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000004843A06  shr_assert_mod_mp          95  shr_assert_mod.F90.in
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000004843E0A  shr_assert_mod_mp         112  shr_assert_mod.F90.in
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000003544EA7  sparsematrixmulti         973  SparseMatrixMultiplyMod.F90
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000001629F29  cnsoilmatrixmod_m         614  CNSoilMatrixMod.F90
dec0933.hsn.de.hpc.ucar.edu 21: cesm.exe           0000000003F2B6D6  cndrivermod_mp_cn        1101  CNDriverMod.F90
[...]

PR #3893, opened 2026-03-26
branch_tags/wetlands.n01.ctsm5.4.029: Wetland update

### LINKED ISSUE #3843: CH4 tuning for CLM6 & WIEMIP
@swensosc has a wetland CH4 branch that has both methane changes and hillslope hydrology modifications that we want to use for WIE-MIP.  

I've done a test showing that running this branch produces [nearly identical results](https://webext.cgd.ucar.edu/I20TR/ctsm5.4.024_f19_wetlandCH4_151_HIST/ctsm5.4.024_f19_wetlandCH4_151_HIST_1850_2023_vs_ctsm5.4.024_f19_150_HIST_1850_2023/website/index.html) to a 5.4.024 baseline control (but halves global CH4 production).

I'd love to consider merging Sean's branch to master to avoid having to maintain a long-term branch.

swensosc:
### Description of changes
biogeophysics: 1) a couple of modifications to how surface water is treated.  One deals with reconciling snow / h2osfc fractional areas when they coexist.  Previously fh2osfc was reduced, now reduce each proportionately to ensure a maximum sum of 1.  2) change drainage from surface water to be based on matric potential gradient instead of saturated hydraulic conductivity.
biogeochemistry: add a surface layer to the methane transport calculations.  This ensures that oxidation occurs in the unsaturated column calculations when the water table is at the surface.

### Specific notes

Testing performed, if any:


src/biogeochem/ch4Mod.F90
```diff
@@ -1934,18 +2021,39 @@ subroutine ch4 (bounds, num_soilc, filter_soilc, num_lakec, filter_lakec, &
...
                   conc_ch4_sat(c,j) = (fsat_bef(c)*conc_ch4_sat(c,j) + dfsat*conc_ch4_unsat(c,j)) / finundated(c)
-               else if (finundated(c) < fsat_bef(c)) then
-                  ch4_dfsat_flux(c) = ch4_dfsat_flux(c) + &
-                       (fsat_bef(c) - finundated(c))*(conc_ch4_sat(c,j) - conc_ch4_unsat(c,j)) * &
-                       dz(c,j) / dtime * catomw / 1000._r8 ! mol --> kg
+               else if (dfsat < 0._r8) then ! adjust conc_ch4_unsat
+                  if (finundated(c) < 1._r8) then
+                     conc_ch4_unsat(c,j) = ((1._r8-fsat_bef(c))*conc_ch4_unsat(c,j) - dfsat*conc_ch4_sat(c,j)) / (1._r8 - finundated(c))
+                  endif
                end if
```
wwieder, line 2047: Not sure I understand the motivation for these changes?

src/biogeochem/ch4Mod.F90
```diff
@@ -4194,128 +4312,1371 @@ subroutine ch4_tran (bounds, &
...
+    vmax_ch4_oxid   = params_inst%vmax_ch4_oxid
+    k_m             = params_inst%k_m
+    q10_ch4oxid     = params_inst%q10_ch4oxid
+    smp_crit        = params_inst%smp_crit
+    k_m_o2          = params_inst%k_m_o2
+    k_m_unsat       = params_inst%k_m_unsat
+    vmax_oxid_unsat = params_inst%vmax_oxid_unsat
 
-      f_sat = params_inst%f_sat
+    t0 = tfrz + 12._r8 ! Walter, for Michigan site where the 45 M/h comes from
```
wwieder, line 3572: What is this for?

wwieder (commented):
Sorry I didn't get very far on this.  We can talk more in a bit

src/biogeophys/SaturatedExcessRunoffMod.F90
```diff
@@ -273,8 +273,10 @@ subroutine SaturatedExcessRunoff (this, bounds, num_hydrologyc, filter_hydrology
        do fc = 1, num_hydrologyc
           c = filter_hydrologyc(fc)
           if(col%is_hillslope_column(c) .and. col%active(c)) then
-             ! Set fsat to zero for upland columns
-             if (col%cold(c) /= ispval) fsat(c) = 0._r8
+             ! Set fsat to zero for hillslope columns
+             fsat(c) = 0._r8
+             ! Set fsat to zero for upland columns only
```
wwieder: should this commented code just be removed?

src/biogeophys/SoilHydrologyInitTimeConstMod.F90
```diff
@@ -215,12 +217,19 @@ subroutine SoilHydrologyInitTimeConst(bounds, soilhydrology_inst, soilstate_inst
...
+            ! use standard value
+            pc = params_inst%pc
+         endif
+
          soilhydrology_inst%h2osfc_thresh_col(c) = 0._r8
          if (micro_sigma(c) > 1.e-6_r8 .and. (soilhydrology_inst%h2osfcflag /= 0)) then
             d = 0.0_r8
             do p = 1,4
-               fd   = 0.5_r8*(1.0_r8+shr_spfn_erf(d/(micro_sigma(c)*sqrt(2.0_r8)))) - params_inst%pc
+               fd   = 0.5_r8*(1.0_r8+shr_spfn_erf(d/(micro_sigma(c)*sqrt(2.0_r8)))) - pc
```
wwieder, line 232: should `params_inst%pc` still be used, not the hard coded value for pc?

wwieder: or is this now coming from the function in SurfaceWaterMod?

wwieder: Oh, now I think I understand, OK to resolve.

src/biogeophys/SurfaceWaterMod.F90
```diff
@@ -485,8 +520,7 @@ subroutine QflxH2osfcSurf(bounds, num_hydrologyc, filter_hydrologyc, &
           ! spatially variable k_wet
           k_wet=1.0e-4_r8 * sin((rpi/180._r8) * topo_slope(c))
           if (col%is_hillslope_column(c)) then
-             ! require a minimum value to ensure non-zero outflow
-             k_wet = 1e-4_r8 * max(col%hill_slope(c),min_hill_slope)
+             k_wet = max(col%hill_slope(c),0._r8)/dtime
```
wwieder, line 523: Assuming this change is intended for the hillslope changes, but wanted to flag it.

wwieder (commented):
few more comments

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -2603,6 +2603,13 @@ lnd/clm2/surfdata_esmf/NEON/ctsm5.4.0/surfdata_1x1_NEON_TOOL_hist_2000_78pfts_c2
 <stream_meshfile_ch4finundated finundation_method="ZWT_inversion"
 >lnd/clm2/paramdata/finundated_inversiondata_0.9x1_ESMFmesh_cdf5_130621.nc</stream_meshfile_ch4finundated>
 
+<!-- need a default for h2osfc even though not used  -->                               
+<stream_fldfilename_ch4finundated finundation_method="h2osfc" hgrid="0.9x1.25"         
+>lnd/clm2/paramdata/finundated_inversiondata_0.9x1.25_c170706.nc</stream_fldfilename_ch4finundated>
```
slevis-lmwg: Stream is not used with prognostic inundation but is still required by the error checking. 
@slevis-lmwg will look into clearing that up before the PR enters its final testing stage.

src/biogeochem/ch4Mod.F90
slevis-lmwg: @swensosc explained this during the paired review:
We do not exchange ch4 with the atm, so better to make the two branches of the if statemt consistent and calculate concentrations, not fluxes.

src/biogeochem/ch4Mod.F90
```diff
@@ -2091,7 +2199,8 @@ subroutine ch4 (bounds, num_soilc, filter_soilc, num_lakec, filter_lakec, &
 
          ! Solve CH4 reaction/diffusion equation 
          ! Competition for oxygen will occur here.
-         call ch4_tran (bounds, &
+         !         call ch4_tran (bounds, &
```
wwieder: since we're just replacing `ch4_tran` with this new function.  Let's rename `ch4_tran_surface_layers` as `ch4_tran` to avoid confusion (and get rid of the old `ch4_tran` subroutine)

src/biogeochem/ch4Mod.F90
```diff
@@ -2091,7 +2199,8 @@ subroutine ch4 (bounds, num_soilc, filter_soilc, num_lakec, filter_lakec, &
...
+         call ch4_tran_surface_layers (bounds, &
```
slevis-lmwg: I think Will commented here, so I may be repeating:
We agreed to keep the orig. subr. name and just modify the orig. subroutine. Easier to see the diffs that way in git/github.

src/biogeochem/ch4Mod.F90
```diff
@@ -4194,128 +4312,1371 @@ subroutine ch4_tran (bounds, &
...
+      ! Accumulate ebullition to place in first layer above water table, or directly to atmosphere
+      do j = 1,nlevsoi
+         do fc = 1, num_methc
+            c = filter_methc (fc)
+            if (j == 1) ch4_ebul_total(c) = 0._r8
+            ch4_ebul_total(c) = ch4_ebul_total(c) + ch4_ebul_depth(c,j) * dz(c,j)
+         enddo
+      enddo
+
+      if (1==1) then
```
slevis-lmwg: This can be removed.

src/biogeochem/ch4Mod.F90
```diff
@@ -4194,128 +4312,1371 @@ subroutine ch4_tran (bounds, &
...
+      ! surface flux is adjusted to conserve. This results in some inaccuracy as compared to a shorter timestep
+      ! or iterative solution.
+
+      ! surface (virtual) layer
+      j = 0
+      do fc = 1, num_methc
+         c = filter_methc (fc)
+         source(c,j,1) = -surface_layer_ch4_oxid(c)  ! [mol/m3-total/s]
+         source(c,j,2) = -surface_layer_o2_oxid(c)   ! O2 [mol/m3/s]
+         ! not adding error checks ...
```
slevis-lmwg: @swensosc will double-check this comment.

src/biogeochem/ch4Mod.F90
```diff
@@ -4194,128 +4312,1371 @@ subroutine ch4_tran (bounds, &
...
+
+         ! For history make sure that grnd_ch4_cond includes snow, for methane diffusivity
+         grnd_ch4_cond(c) = spec_grnd_cond(c,1)
+         
+         errch4(c) = errch4(c) + (ch4_surf_aere(c) + ch4_surf_ebul(c) + ch4_surf_diff(c))*dtime
+
+         if (abs(errch4(c)) < 1.e-8_r8) then 
+            ch4_surf_diff(c) = ch4_surf_diff(c) - errch4(c)/dtime
+
+            ! add change in surface layer concentration and oxidation to diffusion for to conserve ch4 (i.e. convert virtual changes to actual fluxes)
```
slevis-lmwg: This can go now.

src/biogeochem/ch4Mod.F90
```diff
@@ -4415,6 +5780,23 @@ subroutine ch4_totcolch4(bounds, num_nolakec, filter_nolakec, num_lakec, filter_
...
 
+    ! Calculate surface layer concentration
+    do fc = 1, num_nolakec
+       c = filter_nolakec(fc)
+!!$       l = col%landunit(c)
+!!$       if (lun%itype(l) .eq. isturb_tbd .or. lun%itype(l) .eq. isturb_hd .or. lun%itype(l) .eq. isturb_md) then
+!!$          totcolch4(c) = totcolch4(c) + &
+!!$               (finundated(c)*conc_ch4_sat(c,j)*surface_layer_thickness_sat(c)  &
+!!$               + (1._r8-finundated(c))*conc_ch4_unsat(c,j)*surface_layer_thickness_unsat(c)) * catomw
+!!$       else
```
slevis-lmwg: Reminder to see if these lines can be removed now.

src/biogeophys/SurfaceWaterMod.F90
```diff
@@ -276,7 +293,11 @@ subroutine BulkDiag_FracH2oSfc(bounds, num_soilc, filter_soilc, &
        if (frac_sno(c) > (1._r8 - frac_h2osfc(c)) .and. h2osno_total(c) > 0) then
 
           if (frac_h2osfc(c) > 0.01_r8) then             
-             frac_h2osfc(c) = max(1.0_r8 - frac_sno(c),0.01_r8)
+              f_overlap = frac_sno(c) + frac_h2osfc(c) - 1._r8
```
slevis-lmwg: nit: there's an extra space in the indent...

slevis-lmwg (changes requested):

wwieder (changes requested):
one more comment based on our paired review.

resolving this one will make it easier to see diffs in the PR.

src/biogeochem/ch4Mod.F90
swensosc: not an actual code change

src/biogeophys/SaturatedExcessRunoffMod.F90
swensosc: yes

src/biogeochem/ch4Mod.F90
swensosc: fixed

src/biogeochem/ch4Mod.F90
swensosc: fixed

src/biogeochem/ch4Mod.F90
swensosc: updated comment

src/biogeochem/ch4Mod.F90
swensosc: fixed

src/biogeophys/SurfaceWaterMod.F90
swensosc: fixed

src/biogeochem/ch4Mod.F90
swensosc: fixed

src/biogeophys/SurfaceWaterMod.F90
swensosc: correct, hillslope only

src/biogeochem/ch4Mod.F90
```diff
@@ -5780,21 +4958,11 @@ subroutine ch4_totcolch4(bounds, num_nolakec, filter_nolakec, num_lakec, filter_
        totcolch4(c) = 0._r8
     end do
 
-    ! Calculate surface layer concentration
+    ! Add surface layer ch4 and convert mol CH4 to g C
     do fc = 1, num_nolakec
        c = filter_nolakec(fc)
-!!$       l = col%landunit(c)
```
ekluzek: Thanks for removing the commented out code!

bld/namelist_files/namelist_defaults_ctsm.xml
swensosc: there is a check somewhere that a default file is needed.  In the code, if the file is '', then useStreams is false, which presumably is the behavior we want.  Is there a way to set the default to ''?  currently, it appends the xml value to din_loc_root.

src/biogeophys/SaturatedExcessRunoffMod.F90
```diff
@@ -273,10 +273,7 @@ subroutine SaturatedExcessRunoff (this, bounds, num_hydrologyc, filter_hydrology
        do fc = 1, num_hydrologyc
           c = filter_hydrologyc(fc)
           if(col%is_hillslope_column(c) .and. col%active(c)) then
-             ! Set fsat to zero for hillslope columns
```
ekluzek: I think removing this comment as it is, makes sense since it's just repeating in words -- what the code is doing.

However, I think it might be useful to say "why" we set fsat to zero for hillslope columns. That would be useful to have in a comment.

src/biogeochem/ch4Mod.F90
```diff
@@ -3495,822 +3493,6 @@ subroutine ch4_ebul (bounds, &
 
   end subroutine ch4_ebul
 
-  !#
```
ekluzek: I expected this would be moving this section of code somewhere else. But, it looks like it's just removing a ton of things.

I take it that ch4_tran and ch4_tran_surface_layers must have been almost identical, so you were able to just remove the differences in the one? That's a great improvement if so.

This does make it hard to assess what those differences were though. Maybe we could do the difference by hand and add it to the PR. I might do that, if I get to it...

ekluzek (commented):
I just took a quick look to see what we have here. I ask a couple questions. But, I think it's cool that you were able to improve the science, by also reducing code duplication.

src/biogeophys/SaturatedExcessRunoffMod.F90
swensosc: comment added

src/biogeochem/ch4Mod.F90
swensosc: Actually, there were many differences.  I had kept the original routine during development, but we decided it was not needed, and thus removed it.

bld/namelist_files/namelist_defaults_ctsm.xml
swensosc: or maybe this (from CLMBuildNamelist.pm) just needs to be in a "finundation_method not equal 'h2osfc'" if-block:

 add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_fldfilename_ch4finundated',
             'finundation_method'=>$finundation_method);
    if ($opts->{'driver'} eq "nuopc" ) {
        add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'stream_meshfile_ch4finundated',
                    'finundation_method'=>$finundation_method);
    }

swensosc: ok, I added a conditional and removed the placeholder defaults.

swensosc:
the buildnamelist change worked (i.e. no default is required for 'h2osfc' method), but I noticed that it means control%stream_fldFileName_ch4finundated (in src/cpl/share_esmf/ch4FInundatedStreamType.F90) will never be set because the namelist is not read if the 'h2osfc' method is active.  That means that later, in ch4Mod, this will always be true: ch4_inst%ch4findstream%useStreams(), because UseStreams checks whether stream_fldFileName_ch4finundated) == ''.  The former will be false because it is not set, nor initialized.

update: adding this 2nd conditional to useStreams would ensure correct behavior for 'h2osfc':
if ( trim(control%stream_fldFileName_ch4finundated) == '' .or. finundation_mtd == finundation_mtd_h2osfc)then

src/biogeochem/ch4Mod.F90
```diff
@@ -47,6 +47,7 @@ module ch4Mod
   ! Non-tunable constants
   real(r8) :: rgasm  ! J/mol.K; rgas / 1000; will be set below
   real(r8), parameter :: rgasLatm = 0.0821_r8 ! L.atm/mol.K
+  real(r8), parameter :: psi_fc = 3.4e3       ! matric potential at field capcity (mm)
```
slevis-lmwg:
```suggestion
  real(r8), parameter :: psi_fc = 3.4e3_r8    ! matric potential at field capcity (mm)
```

src/biogeochem/ch4Mod.F90
```diff
@@ -1404,7 +1486,7 @@ subroutine readParams ( ncid )
      call ncd_io(trim(tString),tempr, 'read', ncid, readvar=readv)
      if ( .not. readv ) call endrun(msg=trim(errCode)//trim(tString)//errMsg(sourcefile, __LINE__))
      params_inst%f_ch4=tempr
-
+
```
slevis-lmwg:
```suggestion

```
I just noticed these. Best to take out the blank spaces here and elsewhere to keep the code neat :-)

src/biogeochem/ch4Mod.F90
```diff
@@ -3484,9 +3701,12 @@ subroutine ch4_tran (bounds, &
     real(r8), pointer :: co2_decomp_depth (:,:)
     real(r8), pointer :: conc_o2          (:,:)         
     real(r8), pointer :: conc_ch4         (:,:)        
+    real(r8), pointer :: surface_layer_conc_ch4(:)
+    real(r8), pointer :: surface_layer_conc_o2(:)
+    real(r8), pointer :: surface_layer_thickness(:)
 
     integer  :: nstep                       ! time step number
-    character(len=32) :: subname='ch4_tran' ! subroutine name
+    character(len=32) :: subname='ch4_tran_surface_layers' ! subroutine name
```
slevis-lmwg:
```suggestion
    character(len=32) :: subname='ch4_tran'  ! subroutine name
```

src/biogeochem/ch4Mod.F90
```diff
@@ -3832,123 +4191,195 @@ subroutine ch4_tran (bounds, &
...
+                     eps =  watsat(c,1)-h2osoi_vol_min(c,j) ! Air-filled fraction of total soil volume
+                     if (organic_max > 0._r8) then
+                        om_frac = min(params_inst%om_frac_sf*cellorg(c,1)/organic_max, 1._r8)
+                        ! Use first power, not square as in iniTimeConst
+                     else
+                        om_frac = 1._r8
+                     end if
+                     diffus(c,j) = (d_con_g(s,1) + d_con_g(s,2)*t_soisno_c) * 1.e-4_r8 * &
+                          (om_frac * f_a**(10._r8/3._r8) / watsat(c,1)**2._r8 + &
+                          (1._r8-om_frac) * eps**2._r8 * f_a**(3._r8 / bsw(c,1)) ) &
```
slevis-lmwg: I'm told multiplication is much more efficient than raising exponents, so:

```suggestion
                          (1._r8-om_frac) * eps*eps * f_a**(3._r8 / bsw(c,1)) ) &
```

src/biogeochem/ch4Mod.F90
```diff
@@ -3832,123 +4191,195 @@ subroutine ch4_tran (bounds, &
...
+                     eps =  watsat(c,j)-h2osoi_vol_min(c,j) ! Air-filled fraction of total soil volume
+                     if (organic_max > 0._r8) then
+                        om_frac = min(params_inst%om_frac_sf*cellorg(c,j)/organic_max, 1._r8)
+                        ! Use first power, not square as in iniTimeConst
+                     else
+                        om_frac = 1._r8
+                     end if
+                     diffus (c,j) = (d_con_g(s,1) + d_con_g(s,2)*t_soisno_c) * 1.e-4_r8 * &
+                          (om_frac * f_a**(10._r8/3._r8) / watsat(c,j)**2._r8 + &
+                          (1._r8-om_frac) * eps**2._r8 * f_a**(3._r8 / bsw(c,j)) ) &
```
slevis-lmwg:
```suggestion
                          (1._r8-om_frac) * eps*eps * f_a**(3._r8 / bsw(c,j)) ) &
```

slevis-lmwg (changes requested):
Thank you @swensosc 
seems cleaner this way, and I was able (in this quick look) to notice a few minor things that I suggested fixing this time around.

Also, would you be willing to add/commit/push the conditional that you proposed in https://github.com/ESCOMP/CTSM/pull/3893#issuecomment-4144998197 (or in our separate emails), to provide me with a concrete starting point? If the question is whether to move the check from run-time to build, then I will feel more confident making the change after seeing your initial implementation.

slevis-lmwg:
- [x] Before anything else, starting aux_clm on derecho
`./run_sys_tests -s aux_clm -c ctsm5.4.028 --skip-generate`
though I probably expect failures due to needing new paramfiles...

slevis-lmwg:
Results from aux_clm in `/glade/derecho/scratch/slevis/tests_0330-162149de`:
- Seems I may have misunderstood before, thinking that we needed a new paramfile with this PR. My current understanding is that we do not.
- All clm versions show DIFFs from the .028 baseline. That's expected according to my understanding.
- Unexpected failures, mostly COMPARE_base_hybrid, but also three RUN at the end of the list:
```
ERI_D_Ld9.f10_f10_mg37.I1850Clm45Bgc.derecho_gnu.clm-default COMPARE_base_hybrid
ERI_D_Ld9.f10_f10_mg37.I1850Clm60BgcCrujra.derecho_gnu.clm-default COMPARE_base_hybrid
ERI_D_Ld9.f10_f10_mg37.I1850Clm60Bgc.derecho_gnu.clm-default COMPARE_base_hybrid
ERI_D_Ld9.f10_f10_mg37.I1850Clm60Bgc.derecho_gnu.clm-default--clm-matrixcnOn COMPARE_base_hybrid
ERI_D_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_gnu.clm-default COMPARE_base_hybrid
DAE_C2_D_Lh12.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-DA_multidrv COMPARE_base_da
ERI_D_Ld20.f10_f10_mg37.I2000Clm50Fates.derecho_intel.clm-FatesCold COMPARE_base_hybrid
ERI_D_Ld20.f10_f10_mg37.IHistClm60Bgc.derecho_intel.clm-default--clm-midDecStart--clm-allowTransientYearJump COMPARE_base_hybrid
ERI_D_Ld20.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdTwoStream COMPARE_base_hybrid
ERI_D_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default COMPARE_base_hybrid
ERI_D_Ld9.f45_f45_mg37.I2000Clm60Fates.derecho_intel.clm-FatesColdCamLndTuningMode COMPARE_base_hybrid
ERI_D_Ld9.ne30_g17.I2000Clm50BgcCru.derecho_intel.clm-vrtlay COMPARE_base_hybrid
ERI_D_Ld9.ne30_g17.I2000Clm50BgcCru.derecho_intel.clm-vrtlay--clm-matrixcnOn COMPARE_base_hybrid
ERI_D.ne30pg3_t232.I1850Clm60BgcCropG.derecho_intel.clm-clm60cam7LndTuningModeLDust COMPARE_base_hybrid
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default COMPARE_base_hybrid
ERI_Ld9.f45_g37.I2000Clm50BgcCru.derecho_intel.clm-nofire COMPARE_base_hybrid
ERI_Ld9.f45_g37.I2000Clm50BgcCru.derecho_intel.clm-nofire--clm-matrixcnOn COMPARE_base_hybrid
ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-flexCN_FUN--clm-matrixcnOn_ignore_warnings RUN
ERP_P256x2_D_Ld5.f19_g17_gris4.I1850Clm50BgcCropG.derecho_intel.clm-glcMEC_increase RUN
ERS_D_Ld12.f10_f10_mg37.I1850Clm50BgcCropG.derecho_intel.clm-glcMEC_spunup_inc_dec_bgc RUN
```
I will post info about the COMPARE_base_hybrid cases below, and I will rerun the RUN failures because I have seen flakiness in my b4b-dev PR.

slevis-lmwg:
Sean, could you start looking into these? I'm posting about the RUN failures first, because 
- the last two both stopped with a "Gridcell-level CH4 Conservation Error in CH4Mod" and `ERROR: Methane conservation errorERROR in ch4Mod.F90 at line 2459`; both mention glcMEC... is it coincidence or relevant to the error?
- I resubmitted the other RUN failure (third from the bottom) and produced the same outcome `ERROR in SparseMatrixMultiplyMod.F90 at line 973`, which (for now) I'm hoping may get resolved with the resolution of the other RUN failures, though see known issue #3360

slevis-lmwg:
For the COMPARE_base_hybrid tests, I grepped for `non-zero` in the cprnc.out files of this test and got:
```
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.clm2.h0a.2004-01-09-00000.nc.base.cprnc.out:          of which    431 had non-zero differences
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.clm2.h0a.2004-01-09-00000.nc.cprnc.out:          of which    468 had non-zero differences
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.clm2.h0i.2004-01-09-00000.nc.cprnc.out:          of which      1 had non-zero differences
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.clm2.h1a.2004-01-09-00000.nc.base.cprnc.out:          of which     32 had non-zero differences
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.clm2.h1a.2004-01-09-00000.nc.cprnc.out:          of which     32 had non-zero differences
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.cpl.hi.2004-01-09-00000.nc.base.cprnc.out:          of which     48 had non-zero differences
ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default.C.0330-162149de_int.cpl.hi.2004-01-09-00000.nc.cprnc.out:          of which     51 had non-zero differences
```
I got similar numbers of non-zero differences from another test, but the following tests had fewer non-zero diffs, which may make them easier to troubleshoot:
```
ERI_D_Ld9.f45_f45_mg37.I2000Clm60Fates.derecho_intel.clm-FatesColdCamLndTuningMode.C.0330-162149de_int.clm2.h0a.2004-01-09-00000.nc.base.cprnc.out:          of which      2 had non-zero differences
ERI_D_Ld9.f45_f45_mg37.I2000Clm60Fates.derecho_intel.clm-FatesColdCamLndTuningMode.C.0330-162149de_int.clm2.h0a.2004-01-09-00000.nc.cprnc.out:          of which     46 had non-zero differences
ERI_D_Ld9.f45_f45_mg37.I2000Clm60Fates.derecho_intel.clm-FatesColdCamLndTuningMode.C.0330-162149de_int.cpl.hi.2004-01-09-00000.nc.cprnc.out:          of which     28 had non-zero differences
```
```
DAE_C2_D_Lh12.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-DA_multidrv.C.0330-162149de_int.clm2_0001.h1a.2000-01-01-01800.nc.base.cprnc.out:          of which     11 had non-zero differences
DAE_C2_D_Lh12.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-DA_multidrv.C.0330-162149de_int.clm2_0002.h1a.2000-01-01-01800.nc.base.cprnc.out:          of which     11 had non-zero differences
```

slevis-lmwg:
The two RUN failures with conservation errors now PASS!

This ERI test also passed
`ERI_D_Ld9.f10_f10_mg37.I1850Clm45Bgc.derecho_gnu.clm-default.C.0330-162149de_gnu`
so I will rerun aux_clm.

wwieder:
I'm wondering how best to stage / plan things here since we don't yet have the green light to bring this into CESM and we know it will change answers.  Let's discuss at the SE meeting tomorrow.

slevis-lmwg:
`./run_sys_tests -s aux_clm -c ctsm5.4.029 --skip-generate` on derecho passed!

slevis-lmwg:
CTSM Software meeting 2026/4/2:
We will merge and tag this branch/PR in an escomp branch for use in simulations, including B-cases that will run parallel to Cecile's steps 3 and 5:
3) Cecile's extended run 400-500 years (May 12) + wetland parallel run for 100 yrs to show unchanged climate
5) Cecile's BGC spinup (July 16) starting from bare ground + wetland parallel case

slevis-lmwg (approved):
I'm approving the PR now that aux_clm has worked as expected on derecho.

@ekluzek do you need to make the escomp branch where we will merge this PR or do I have permission to do that and, if so, would you like me to start that branch?

ekluzek:
> I'm approving the PR now that aux_clm has worked as expected on derecho.
> 
> @ekluzek do you need to make the escomp branch where we will merge this PR or do I have permission to do that and, if so, would you like me to start that branch?

I'm planning on making the branch and setting up the rules for it. You do have permission to do it, so you could. It doesn't take a lot of time, but it would be easier to show what needs to be done.

If you'd like to learn about it, we could do it together.

slevis-lmwg:
> I'm planning on making the branch and setting up the rules for it. You do have permission to do it, so you could. It doesn't take a lot of time, but it would be easier to show what needs to be done.
> 
> If you'd like to learn about it, we could do it together.

Yes, that sounds good for me to learn the process. I have a vague memory of watching you go through the steps once before, but more than a year ago probably...

wwieder:
let me know when the branch is ready and I'll do another CH4 calibration run.

slevis-lmwg:
Now I need to merge this and make a tag and push the tag to escomp, same as for master tags except no ChangeLog here.

src/biogeochem/ch4Mod.F90
```diff
@@ -1446,7 +1545,7 @@ subroutine readParams ( ncid )
      params_inst%vmax_ch4_oxid=45.e-6_r8 * 1000._r8 / 3600._r8
      ! FIX(FIX(SPM,032414),032414) can't be read off of param file.  not bfb since it is a divide
      !params_inst%vmax_ch4_oxid=tempr
-
+
```
slevis-lmwg, line 1558:
```suggestion

```

src/biogeochem/ch4Mod.F90
```diff
@@ -2029,7 +2152,7 @@ subroutine ch4 (bounds, num_soilc, filter_soilc, num_lakec, filter_lakec, &
 
             call get_jwt (bounds, num_soilc, filter_soilc, jwt(begc:endc), &
                  soilstate_inst, waterstatebulk_inst, temperature_inst)
-
+
```
slevis-lmwg, line 2156:
```suggestion

```

src/biogeochem/ch4Mod.F90
```diff
@@ -2828,7 +2953,7 @@ subroutine ch4_oxid (bounds, &
                k_m_eff = k_m_unsat
                vmax_eff = vmax_oxid_unsat
             end if
-
+
```
slevis-lmwg, line 2958:
```suggestion

```

src/biogeochem/ch4Mod.F90
```diff
@@ -2861,7 +2985,7 @@ subroutine ch4_oxid (bounds, &
 
             ch4_oxid_depth(c,j) = oxid_a
             o2_oxid_depth(c,j) = ch4_oxid_depth(c,j) * 2._r8
-
+
```
slevis-lmwg, line 2991:
```suggestion

```

src/biogeochem/ch4Mod.F90
```diff
@@ -3210,7 +3334,7 @@ subroutine SiteOxAere(is_vegetated, &
           aerecond = area_tiller * rootfr(j) * diffus_aere / (z(j)*params_inst%rob)
           ! Add in boundary layer resistance
           aerecond = 1._r8 / (1._r8/(aerecond+smallnumber) + 1._r8/(grnd_ch4_cond+smallnumber))
-
+
```
slevis-lmwg, line 3341:
```suggestion

```

src/biogeochem/ch4Mod.F90
slevis-lmwg: @swensosc I'm making sure whether you already checked this before I resolve the post.

src/biogeochem/ch4Mod.F90
```diff
@@ -3832,123 +4203,195 @@ subroutine ch4_tran (bounds, &
...
+                     dp1_zp1(c,j) = 2._r8/(dz(c,j)/diffus(c,j)+dz(c,j+1)/diffus(c,j+1))
+                  else if (j == nlevsoi) then ! j=nlevsoi
+                     dm1_zm1(c,j) = 2._r8/(dz(c,j)/diffus(c,j)+dzm1/diffus(c,j-1))
+                  end if
+
+               endif ! sat; 
             enddo ! fp; patch
          end do ! j; nlevsoi
-
+
```
slevis-lmwg, line 4384:
```suggestion

```

src/biogeochem/ch4Mod.F90
```diff
@@ -3970,13 +4412,13 @@ subroutine ch4_tran (bounds, &
 
          do fc = 1, num_methc
             c = filter_methc (fc)
-            jtop(c) = 0
+            jtop(c) = -1
          end do
 
          if (s == 1) then  ! CH4
-
+
```
slevis-lmwg, line 4462:
```suggestion

```

src/biogeochem/ch4Mod.F90
```diff
@@ -4173,14 +4716,19 @@ subroutine ch4_tran (bounds, &
 
          ! For history make sure that grnd_ch4_cond includes snow, for methane diffusivity
          grnd_ch4_cond(c) = spec_grnd_cond(c,1)
-
+
```
slevis-lmwg, line 4763:
```suggestion

```

src/biogeochem/ch4Mod.F90
slevis-lmwg: In prep. for the merge, I will resolve this conversation. Pls unresolve if you feel that I was mistaken.

slevis-lmwg:
> let me know when the branch is ready and I'll do another CH4 calibration run.

@wwieder we need your approval of the PR, as well :-)

slevis-lmwg:
- [x] For sanity, I will submit aux_clm on izumi, as well, before merging and making the tag.

billsacks:
I was just thinking about the issue that @swensosc and I fixed in 981d2622e3d7b653902516c09a268091751852ae - which I believe meant that (before the fix) the mere act of writing a restart file would change answers. I realized that most of our restart tests wouldn't pick up an issue like this. That made me remember this discussion from a few months ago:

https://github.com/ESMCI/cime/issues/4859

where we discussed the value of an alternative ERS test that more closely mimics the production workflow and *would* catch problems where writing a restart file causes a change in answers. (My understanding is that that's the main additional value of that alternative "ERS2" test.)

Seeing this issue - which seems like a very easy kind of issue to accidentally introduce in CTSM - made me wonder if we should revisit the discussion in #4859 (we had decided to remove the ERS2 test because its limited additional value didn't seem to warrant the extra complexity). However, looking back through the above test results, it looks to me like this problem *was* picked up in an ERI test. @slevis-lmwg am I understanding that right? I can see why an ERI test would detect this: one of the comparisons done in an ERI test is the comparison of the hybrid run with a branch run done off of that hybrid run, and one (somewhat incidental) difference between the two is that the branch run writes an extra set of restart files mid-way through. So based on that, I feel more comfortable with the current state of testing, as long as we have ERI tests covering most / all major configurations, at least in terms of configurations of restart options.

I'd be interested to hear if anyone has any thoughts on this - and will probably add it as an agenda item for an upcoming cseg meeting for further discussion.

billsacks:
> Seeing this issue - which seems like a very easy kind of issue to accidentally introduce in CTSM - made me wonder if we should revisit the discussion in #4859 (we had decided to remove the ERS2 test because its limited additional value didn't seem to warrant the extra complexity). However, looking back through the above test results, it looks to me like this problem _was_ picked up in an ERI test. @slevis-lmwg am I understanding that right? I can see why an ERI test would detect this: one of the comparisons done in an ERI test is the comparison of the hybrid run with a branch run done off of that hybrid run, and one (somewhat incidental) difference between the two is that the branch run writes an extra set of restart files mid-way through. So based on that, I feel more comfortable with the current state of testing, as long as we have ERI tests covering most / all major configurations, at least in terms of configurations of restart options.

Answering this for myself: Yes, an ERI test detects this issue: `ERI_Ld9.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-default` passes now, but fails if I revert 981d2622e.

swensosc:
sam, please wait before merging.  bill and I are looking into a possible issue in an edge case related to dynamic glacier changes.

src/biogeochem/ch4Mod.F90
```diff
@@ -4415,6 +4967,13 @@ subroutine ch4_totcolch4(bounds, num_nolakec, filter_nolakec, num_lakec, filter_
        totcolch4(c) = 0._r8
     end do
 
+    ! Add surface layer ch4 and convert mol CH4 to g C
+    do fc = 1, num_nolakec
+       c = filter_nolakec(fc)
+       totcolch4(c) = totcolch4(c) + &
+            (finundated(c)*surface_layer_conc_ch4_sat(c)*surface_layer_thickness_sat(c) + (1._r8-finundated(c))*surface_layer_conc_ch4_unsat(c)*surface_layer_thickness_unsat(c)) *catomw
```
billsacks: This long line should be split - I'm surprised NAG didn't complain about it. Or are longer lines (> 132 characters) now permitted? (I think Fortran 2023 increases the max line length, but I don't think we should rely on that.)

While doing that, can you check for other long lines that should similarly be split? It looks like there are some long lines here but I haven't checked the line length.

billsacks:
To keep others in the loop:

As @swensosc noted, he and I have been working through a conservation error that still appears sometimes in some dynamic landunit cases - thanks to Sean for the careful additional testing that revealed this! Pending some additional testing by Sean, I think I have tracked it down to something that I neglected to put in the interface and/or documentation of dynColumnStateUpdaterMod: That code doesn't explicitly handle the layer thickness associated with each variable, but technically it should. Up until this point, I don't think this has caused problems because layer thicknesses are constant in time and the same between different columns (so we'd end up multiplying and dividing by the same number). But the novel thing in this new code is that layer thicknesses associated with the new methane variables change in time and space. This layer thickness therefore needs to be passed as an additional multiplying/dividing factor into the dynamic landunit column state updater.

Sean is running some tests to see if this solves the issue.

src/biogeochem/ch4Mod.F90
```diff
@@ -1299,8 +1363,49 @@ subroutine DynamicColumnAdjustments(this, bounds, clump_index, column_state_upda
...
+    call column_state_updater%update_column_state_no_special_handling( &
+         bounds = bounds, &
+         clump_index = clump_index, &
+         var    = this%conc_ch4_surface_layer_sat_col(begc:endc), &
+         fractional_area_old = surface_layer_volume_sat_col(begc:endc), &
+         fractional_area_new = surface_layer_volume_sat_new_col(begc:endc), &
+         adjustment = adjustment_one_level(begc:endc))
+
+    do c = bounds%begc, bounds%endc
+       this%dyn_ch4bal_adjustments_col(c) = &
```
billsacks, line 1395: I think it would be worth putting a comment here - and on the similar update of dyn_ch4bal_adjustments_col in the next block - pointing out that thickness is *not* included here because it is already folded into the adjustment_one_level variable due to the inclusion of thickness in the fractional_area_old and fractional_area_new variables.

src/biogeochem/ch4Mod.F90
```diff
@@ -1299,8 +1363,49 @@ subroutine DynamicColumnAdjustments(this, bounds, clump_index, column_state_upda
          clump_index = clump_index, &
          var = this%finundated_lag_col(begc:endc))
 
+    ! Because surface layer thicknesses can differ between landunit types, the volume is passed rather than the fractional area.  Passing solely the fractional area is acceptable when layer thicknesses are identical for different landunits, i.e. for the soil layers.
```
billsacks, line 1370: This comment is great. I think it would also be worth putting a comment above the 4 new variable declarations (`surface_layer_volume_*`) describing those variables - that they are fractional area times thickness. I'm particularly thinking: these aren't *really* volumes since the column area (m^2) isn't included here. In fact, since these aren't volumes, it may be worth renaming these to avoid confusion. Maybe something more generic like `surface_layer_multiplier_*`, with a comment clearly explaining what they are - since I can't think of a good name myself.

billsacks (changes requested):
These new changes look great - thank you for working through this with me, @swensosc ! I do have a couple of suggestions for some additional comments and possibly some variable renames for clarity.

billsacks:
@slevis-lmwg - to keep you in the loop here: After @swensosc addresses my latest comments, I'd like to make an additional change that renames a couple of variables in dynColumnStateUpdaterMod.F90. This is a straightforward change but will change a lot of lines and overlap with Sean's changes, so I want to wait until he's done. At that point, I'm hopeful that this will be ready for re-testing.

I'm open to feedback from anyone on these upcoming renames: As can be seen in Sean's recent changes, there are now things like this in calls to `update_column_state`:

```
fractional_area_old = surface_layer_volume_sat_col(begc:endc), &
```

where the passed-in variable really isn't just fractional_area anymore, but instead is a more complete multiplier that includes thickness. I'm thinking of renaming it to something generic like extra_multiplier, but welcome any input. Then I'd add some comments explaining what that does and giving some examples.

src/biogeochem/ch4Mod.F90
swensosc: comment added

src/biogeochem/ch4Mod.F90
swensosc: comment added

swensosc:
bill, I see what you are saying, but I feel that 'extra_multiplier' would be too vague.  To move mass between columns, concentration is scaled by a volume.  In the original code, both the gridcell area and the thickness were implicit, since their values were the same for both columns.  Now, just the gridcell area is implicit, but I think it's important to keep the idea clear that a volume is what scales the concentration.  I thought of changing the variable names to *_fractional_volume_*, but that maybe is too long.  If you want extra_multiplier though, that is fine with me.

billsacks (approved):

billsacks:
@swensosc - Thanks for your thoughts. Yes, I see your point. So, yes, maybe sticking with "volume" does make sense here.

I'm still thinking about what to name the variable in the dyn column state updater - i.e., the new name for fractional_area. Based on your good point, I'm thinking `volume_multiplier`. My thinking is that just using `volume` is misleading, because this will never actually be the true volume, but it includes all of the multiplier terms needed to fully convert to volume. (And I'd try to add a clear comment explaining this.) How does that sound to you?
- Update: maybe `extra_volume_multiplier` to better suggest its optionality?

I have also run through some hand calculations of different possibilities for getting conservation. With the scenario I used (which was simple, but I think is generic enough to apply to other scenarios), we get conservation as long as thickness is included in the updater (as it now is), regardless of whether thickness itself is updated. Looking at the equations, this makes sense: the new thickness appears in the denominator in the column state updater, and then in a multiplier in calculating total mass for conservation. So I think that we'd conserve with *any* non-zero final thickness. (I haven't tested this in an actual run of the code.)

So this implies to me that whether or not we update thicknesses in these dynamic column adjustments can depend on which one feels better scientifically. If this is a fairly quickly-varying term (say, order 1 year or less), then my gut feeling is that it probably isn't hugely important and we can stick with the simpler implementation of *not* updating it - i.e., what is now in place.

swensosc:
I think 'volume_multiplier' would be fine.  And a comment noting that thickness may be omitted when it is implicitly accounted for in the case of identical thickness for each landunit would help clarify the 'fractional_area' cases.

billsacks:
Great, thanks @swensosc ! I'll make those changes later today.

src/biogeochem/ch4Mod.F90
swensosc: fixed

src/biogeochem/ch4Mod.F90
swensosc: I have added error checks for the surface water layers.  I ran a couple of 5 year simulations successfully.

wwieder:
Testing the latest now with a [modified parameter file that's described in this issue](https://github.com/NCAR/LMWG_dev/issues/156)

wwieder:
Maybe my scientific test should I wait until @billsacks pushes changes too?

billsacks:
> Maybe my scientific test should I wait until @billsacks pushes changes too?

No need to wait: my changes are just variable renames and comments so won't change answers.

slevis-lmwg:
Generating baselines using the latest code in this PR:
`./run_sys_tests -s aux_clm -c ctsm5.4.029 -g wetland_update.1-ctsm5.4.029`
- [x] derecho
- [x] izumi

billsacks:
I have pushed the change I wanted to make (renaming a set of related arguments... the changes from this were fairly extensive because of the number of places this and similar names / comments appear).

I tested this change by running `ERS_D_Ld12.f10_f10_mg37.I1850Clm50BgcCropG.derecho_intel.clm-glcMEC_spunup_inc_dec_bgc` with comparison against the previous commit. This passed and is bit-for-bit. I also tested the Fortran unit tests on my Mac.

I don't have any more changes I wanted to make here.

swensosc:
I noticed that there were a number of places in ch4Mod where parameters were being overwritten after being read in from the parameter file.  I have confirmed that the values on the parameter file (I checked ctsm60_params.c260303.nc) were the same as the overwrites, so this should not be answer changing.

slevis-lmwg:
As a sanity test, I will run aux_clm against the baselines that I generated yesterday:
`./run_sys_tests -s aux_clm -c wetland_update.1-ctsm5.4.029 --skip-generate`
- [x] izumi
- [x] derecho

slevis-lmwg:
With 5 tests left to go on derecho, I see answer changes in the following tests located in `/glade/derecho/scratch/slevis/tests_0409-124516de`:
```
    FAIL ERI_D_Ld9.f10_f10_mg37.I1850Clm45Bgc.derecho_gnu.clm-default BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm45BgcCrop.derecho_gnu.clm-no_subgrid_fluxes BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERS_D_Ld6.f10_f10_mg37.I1850Clm45BgcCrop.derecho_gnu.clm-clm50CMIP6frc BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL LGRAIN2_Ly2_P128x1.f10_f10_mg37.I1850Clm45BgcCrop.derecho_gnu.clm-ciso--clm-cropMonthOutput BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL LREPRSTRUCT_Ly2_P128x1.f10_f10_mg37.I1850Clm45BgcCrop.derecho_gnu.clm-ciso--clm-cropMonthOutput BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL SMS_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.derecho_gnu.clm-till--clm-remove_residues BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERP_P64x2_D_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.derecho_intel.clm-crop BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERP_P64x2_D_Ld5.f10_f10_mg37.I1850Clm45BgcCru.derecho_intel.clm-ciso BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERP_P64x2_D_Ld5.f10_f10_mg37.I1850Clm45BgcCru.derecho_intel.clm-default BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERP_P64x2_D_Ld5.f10_f10_mg37.IHistClm45BgcCru.derecho_intel.clm-decStart BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL ERS_D_Ld6.f10_f10_mg37.I1850Clm45BgcCrop.derecho_intel.clm-clm50CMIP6frc BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL SMS_D_Ld65.f10_f10_mg37.I2000Clm45BgcCropQianRs.derecho_intel.clm-FireLi2014Qian BASELINE wetland_update.1-ctsm5.4.029: DIFF
    FAIL SMS_Ld5.f10_f10_mg37.I1850Clm45BgcCrop.derecho_intel.clm-crop BASELINE wetland_update.1-ctsm5.4.029: DIFF
```
I will rerun aux_clm on derecho with the code before Sean's last changes, i.e. with Bill's changes only, here: `/glade/derecho/scratch/slevis/tests_0409-155303de`. Izumi looked fine with Sean's changes, so I will not rerun there.

On derecho I started new baseline wetlands.n01.ctsm5.4.029, due to the diffs from wetland_update.1-ctsm5.4.029.
On izumi I linked wetlands.n01.ctsm5.4.029 to wetland_update.1-ctsm5.4.029 since I did not see diffs.

slevis-lmwg:
Testing with [penultimate commit in this PR](https://github.com/ESCOMP/CTSM/pull/3893/commits/8fc82033918ef1c25599d166ed67646e0040211e) in `/glade/derecho/scratch/slevis/tests_0409-155303de` is b4b, while testing with the final commit changed answers as shown in my previous post.

@swensosc before I proceed, how do you feel about this result?

swensosc:
sam, are you saying bill's commit is b4b, but this commit (remove parameter
over-writes) changes answers?

slevis-lmwg:
Correct.

swensosc:
I think that is expected then.  The last commit was correcting the fact
that the parameters were being overwritten after being read in from the
parameter file.  So now all the parameters are getting values from the
parameter file, so they could be different than the overwrite values.

wwieder (approved):

Issue #209, opened 2017-12-29
Add logic to reset flags for a crop column that just became inactive


billsacks:
_From @billsacks on November 3, 2017 20:37_

Currently, `croplive` is only updated for active crop columns (and this is probably true for other similar flags as well). I can envision how this might cause problems for crop columns that start out active, become inactive (due to their gridcell area going to 0), and then later become active again (due to their gridcell area becoming greater than 0). In this situation, `croplive` will start out as true when the column re-activates. I imagine that all sorts of oddities may result in this first year after re-activation.

There probably aren't many places where this happens in practice, so this may not be a big issue, but it's probably still worth fixing.

Some ways I can think of to fix this:

(1) Have some special logic in subgridWeightsMod: is_active_c (and/or is_active_p) that keeps a crop column/patch active for one time step after the time at which it first becomes inactive (i.e., if its weight was > 0 in the _last_ timestep). Crop columns / patches that are about to become inactive (i.e., would be inactive if it weren't for this special logic) would get a flag set (e.g., `crop_newly_inactive`). Then, in the crop phenology routine, any crop patch with `crop_newly_inactive` True (and `croplive` currently True) would go through the harvest logic, in order to set `croplive` to False, etc.

Notes about this:

- We need to keep the column active for this extra time step so that some other cleanup steps can happen. One example that I'm aware of is that we need the logic to reset GDDPLANT to accumResetValue to be triggered; with some changes I'm planning to make very soon, that would only happen properly if the patch were active.

- I'd like to have this new flag (`crop_newly_inactive`) rather than, e.g., checking (active and weight = 0). This is because I want to keep the capability to maintain crop columns as always-active, in which case they shouldn't go through this instant-harvest logic. Keeping the `crop_newly_inactive` logic in subgridWeightsMod makes it easier to change this to keep crop columns always active, if desired.

(2) Alternatively, we could do this via resetting the appropriate flags when a column/patch newly becomes active. For columns, we could potentially do this in dynInitColumnsMod, although (a) This isn't currently set up to reset flags for a newly-active column - though in principle it could be, and (b) We currently don't have analogous infrastructure for new patches.

[...]

Issue #3789, opened 2026-03-02
FatesColdLUH2 tests fail with input files that have non-NaN fills


samsrabin:
### Brief summary of bug

These tests are failing in RUN on the #3787 branch at hash 60a9c5d3d:
- `ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2`
- `ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2HarvestArea`
- `ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2HarvestMass`

### General bug information

**CTSM version you are using:** the #3787 branch at hash 60a9c5d3d

**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:** Unclear

### Important output or errors that show the problem

See testing in `/glade/derecho/scratch/samrabin/tests_0301-171531de/`. 

Here are the namelist diffs from the `ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2` test:
```
  BASE: flandusepftdat = fates_landuse_pft_map_4x5_240206.nc'
  COMP: flandusepftdat = fates_landuse_pft_map_4x5_240206.no_nan_fill.nc'
  BASE: fluh_timeseries = LUH2_states_transitions_management.timeseries_4x5_hist_simyr0850-2015_c240216.nc'
  COMP: fluh_timeseries = LUH2_states_transitions_management.timeseries_4x5_hist_simyr0850-2015_c240216.no_nan_fill.nc'
  BASE: paramfile = clm50_params.c250311.nc'
  COMP: paramfile = clm50_params.c250311.no_nan_fill.nc'
  BASE: stream_meshfile_cropcal = 360x720_120830_ESMFmesh_c20210507_cdf5.tweaked_latlons.nc'
  COMP: stream_meshfile_cropcal = 360x720_120830_ESMFmesh_c20210507_cdf5.tweaked_latlons.no_nan_fill.nc'
```

And here's the error for one task in that run:
```
dec1447.hsn.de.hpc.ucar.edu 433:  negative area           1           1           1  -999.000000000000
dec1447.hsn.de.hpc.ucar.edu 433:  ENDRUN:
[...]

PR #3902, opened 2026-04-01
ctsm5.4.034: bug fix to the FATES land use driver input code

adrifoster:
### Description of changes

This PR includes a cherry-picked fix from [NorESMhub/CTSM #209](https://github.com/NorESMhub/CTSM/pull/209) to address swap in the order of rangeland and pasture in the reading of the landuse drivers.

### Specific notes

Contributors other than yourself, if any: @rosiealice 

Are answers expected to change (and if so in what way)? Yes - for FATES land use runs

Testing performed, if any:

testing starting


samsrabin (approved):
Nice

adrifoster:
okay well it's hard to tell if this is working as intended because we don't have any LUH tests that don't die in the RUN phase?

glemieux:
> okay well it's hard to tell if this is working as intended because we don't have any LUH tests that don't die in the RUN phase?

@adrifoster I talked about this with @ckoven yesterday.  I could have a "quick" fix for this that would require an update to the LUH datasets and the FATES code.  I'll start working on that; what do you think about holding on this until that's updated?

samsrabin:
That seems like a reasonable approach to me, especially considering the b4b-dev merge is supposed to happen today.

rosiealice:
Just to confirm that this code change is what we implemented in NorESM and that it's all working fine there. But agree that this might be a good reason to create the relevant working tests!

glemieux:
> Just to confirm that this code change is what we implemented in NorESM and that it's all working fine there. But agree that this might be a good reason to create the relevant working tests!

@rosiealice @mvdebolskiy  Do y'all run the tests are we see failing here: https://github.com/ESCOMP/CTSM/issues/3789#issue-4012265453?

mvdebolskiy:
@glemieux no, we don't and the latest NorESM has not yet been updated to 5.4.030.

mvdebolskiy:
@rosiealice it is fine only  because we did not yet updated to sam's files.

samsrabin:
This PR will come in after @glemieux fixes the FATES land use tests, which will happen in a separate master tag.

glemieux:
The FATES-side fix appears to be working.  I'll make the PR on the fates side for review and link it here.

adrifoster:
All tests have now passed or fail as expected.

PR #3914, opened 2026-04-08
Remove unnecessary spmdMod from unit test build

billsacks:
### Description of changes

This inclusion of spmdMod in the unit test build appears to be unnecessary. I saw two issues with its inclusion:

(1) On my Mac, I was getting an error building spmdMod.F90 in unit tests because it couldn't find mpif.h. This makes sense because we don't include an MPI library in the build. I guess this was working on derecho - at least if you have the right modules loaded - but the unit test build is not designed to work with an mpi library.

(2) The stub version of spmdMod (spmdMod_stub.F90) is also still being included in the unit test build. It's unclear which of the two would actually be used at runtime.

### Specific notes

Testing performed, if any:

Ran Fortran unit tests on my Mac and on derecho

### DIFF
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -22,7 +22,6 @@ list(APPEND clm_sources
   SparseMatrixMultiplyMod.F90
   IssueFixedMetadataHandler.F90
   NumericsMod.F90
-  spmdMod.F90
   )
 
 sourcelist_to_parent(clm_sources)


ekluzek:
This was my fault. I started bringing in changes to allow use of a MPI library in 6b171e8af18b30c2b3dfb8ef7b87333c6ca8edd1. And then I backed a lot of that out in later commits, but missed this thing.

Sorry about that. And thanks for figuring it out.

ekluzek (approved):

ekluzek:
Works for me on Derecho, so merging.

billsacks:
No problem - this was one of those fortunate cases where it didn't take up much of my time because my first guess about what to do turned out to work (thanks to our hallway conversation where you made it sound like this would be the right solution).

PR #3927, opened 2026-04-13
[WIP] Add NVP (Non-Vascular Plant) Layer in CTSM

huitang-earth:
### Description:
This PR introduces a physically explicit representation of non-vascular plants (NVP; e.g., moss and lichen) in CTSM–FATES and its biophysical effect by assigning them to a dedicated vertical layer in CTSM, and a NVP PFT in FATES. The core architectural decision is to use layer 0 of the soil-snow layer structure to accommodate NVP and its growth (i.e., changes in height) when it is present. This layer integrates consistently into CLM’s radiation (under snow), thermal, and hydrology schemes, while coupling to FATES (NGEET/fates#1556) for canopy radiation (without snow) and photosynthesis (see the following call graph). 
Note: The implementation is not fully ready yet and is still open for discussion. 

#### Key implementation details in CLM:

- **Adds NVP as an explicit surface layer at index 0**, controlled by col%nvp_layer_active and col%jbot_sno, with geometry (dz, z, zi) updated each daily timestep via UpdateNVPLayer.
- **Extends snow radiation to include NVP absorption at layer 0** with NVP specific optical parameters (when under snow).
- **Integrates NVP into the thermal solver with its own thermal properties**. 
- **Adds NVP specific hydraulic properties, water retention curve, and water balance subroutine** `NVPWaterBalance_Column`.
- Adds interface with FATES for exchanging NVP structure properties, radiation flux, temperature and water status. 

diagram

### Collaborators:
@rosiealice

### Expectation of Answer Changes:
This PR is expected to produce answer changes when NVP is active and NVP layer is present:

- Surface energy balance modified due to new top boundary condition at layer 0 (CLM, FATES)
- Soil temperature profiles altered due to NVP thermal properties (CLM) 
- Hydrology affected by additional water storage and evaporation pathway (CLM)
- Carbon fluxes (GPP/NPP) changed due to NVP photosynthesis and modified PAR environment (FATES)

No answer changes are expected when NVP is disabled.

### Description of generative AI usage
Claude code has been used for porting changes into the new CTSM version and designing the overall structure. The call graph is also generated by Claude code.

### Does this create a need to change or add documentation? Yes


samsrabin:
@huitang-earth Your timing on this could not be better! Let's discuss on the FATES-side PR, NGEET/FATES#1556.
(Locking conversation on this for now, so that we don't get confused.)

src/biogeophys/NVPLayerDynamicsMod.F90
```diff
@@ -0,0 +1,763 @@
...
+    class(waterstate_type),  intent(inout) :: waterstate_inst
+    !
+    ! !LOCAL VARIABLES:
+    integer  :: c
+    real(r8) :: max_ice_nvp   ! NVP pore-space ice capacity [kg m-2]
+    !-----------------------------------------------------------------------
+    do c = bounds%begc, bounds%endc
+       if (col%nvp_layer_active(c) .and. col%dz(c,0) > 0._r8) then
+          max_ice_nvp = watsat_nvp * denice * col%dz(c,0)
+          waterstate_inst%h2osoi_ice_col(c,0) = max_ice_nvp
```
samsrabin, line 756: I will change this to be frozen only if temperature is ≤ freezing. That would be consistent with [`WaterStateType%InitCold()`](https://github.com/ESCOMP/CTSM/blob/334ba96fa4679a1bb29eca01c9702fb9ed1830f7/src/biogeophys/WaterStateType.F90#L444-L450).

Some text in the subroutine description suggests this is handled elsewhere, but it's cryptic, and I don't understand it. It would be easier to understand if this subroutine chose liquid or frozen itself. (Make sure to update the subroutine description!)

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2949,8 +3135,68 @@ subroutine wrap_canopy_radiation(this, bounds_clump, nc, fcansno, surfalb_inst)
           ftdd(p,:) = this%fates(nc)%bc_out(s)%ftdd_parb(ifp,:)
           ftid(p,:) = this%fates(nc)%bc_out(s)%ftid_parb(ifp,:)
           ftii(p,:) = this%fates(nc)%bc_out(s)%ftii_parb(ifp,:)
-          
+
        end do
+
+       ! [PORTED by Hui Tang: transfer NVP Beer's law absorptance from bc_out to surfalb_inst]
+       ! fabd_nvp_col/fabi_nvp_col are col-level; average equally over all patches in the site.
```
samsrabin, line 3151: > average equally over all patches in the site

Why? I would have thought these should be weighted by the area of the column on each patch.

src/utils/clmfates_interfaceMod.F90
```diff
@@ -2949,8 +3135,68 @@ subroutine wrap_canopy_radiation(this, bounds_clump, nc, fcansno, surfalb_inst)
...
+          surfalb_inst%alb_nvp_gnd_col(c) = 0._r8
+          nvp_frac_sum = 0._r8
+          do ifp = 1, npatches_site
+             surfalb_inst%alb_nvp_gnd_col(c) = surfalb_inst%alb_nvp_gnd_col(c) + &
+                  this%fates(nc)%bc_out(s)%alb_nvp_gnd_pa(ifp) * &
+                  this%fates(nc)%bc_out(s)%nvp_frac_pa(ifp)
+             nvp_frac_sum = nvp_frac_sum + this%fates(nc)%bc_out(s)%nvp_frac_pa(ifp)
+          end do
+          if (nvp_frac_sum > 1.e-5_r8) then
+             surfalb_inst%alb_nvp_gnd_col(c) = surfalb_inst%alb_nvp_gnd_col(c) / nvp_frac_sum
```
samsrabin, line 3205: Shouldn't the case of small `nvp_frac_sum` be handled?

samsrabin:
It looks like you're never passing the nvp-specific `stomatal_model` to FATES:

```fortran
if (trim(fates_stomatal_model) == 'ballberry1987') then
   pass_stomatal_model = 1
else if (trim(fates_stomatal_model) == 'medlyn2011') then
   pass_stomatal_model = 2
end if
call set_fates_ctrlparms('stomatal_model',ival=pass_stomatal_model)
```

Is that intentional? There are, after all, parts of the FATES PR ([e.g.](https://github.com/NGEET/fates/pull/1556/changes#r3509416327)) that look for the `nvp_model` value (3).

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -948,21 +939,40 @@ which processes the raw land use data from the THEMIS tool data sets
...
+</entry>
+
+<entry id="nvp_rad_model_ground" type="logical" category="physics"
+        group="clm_inparm" valid_values="" value=".false.">
+Use FATES Beer's law radiation model for the NVP ground layer instead of
+the CLM bare-ground scheme.
+Only relevant when use_nvp=".true.".
+</entry>
+
+<entry id="use_nvp_temp_for_patch_gas_params" type="logical" category="physics"
```
samsrabin, line 970: Added `hlm_use_nvp_temp_for_patch_gas_params` option to avoid having the entire patch take NVP's temperature, which would be bad when NVP coexists with other PFTs.

However, before merging, this should probably be deleted, with the FATES code being updated only allow the "false" behavior. Even in an NVP-only run, why should NVP be allowed to ignore snow and surface water effects on temperature when other PFTs can't? Especially since temperature of NVP—the shortest PFT—would be _most_ affected by those?

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -738,8 +836,11 @@ subroutine SoilThermProp (bounds, num_urbanc, filter_urbanc, num_nolakec, filter
 
             ! Thermal conductivity of snow
             ! Only examine levels from snl(c)+1 -> 0 where snl(c) < 1
-            if (snl(c)+1 < 1 .AND. (j >= snl(c)+1) .AND. (j <= 0)) then  
+            ! [PORTED by Hui Tang: skip j=0 when NVP occupies that layer; handled below]
+            if (snl(c)+1 < 1 .AND. (j >= snl(c)+1) .AND. (j <= 0) .AND. &
+                .NOT. (use_nvp .AND. jbot_sno(c) == -1 .AND. j == 0)) then
                bw(c,j) = (h2osoi_ice(c,j)+h2osoi_liq(c,j))/(frac_sno(c)*dz(c,j))
```
samsrabin, line 842: I'm running into a divide-by-zero error when `use_nvp` is true but there is no NVP layer present. `jbot_sno(c) == -1` evaluates to false, so the `NOT` is satisfied, and this conditional block gets entered. But then `dz` is zero, causing divide-by-zero in the calculation of `bw`.

PR #3955, opened 2026-04-23
Workflow to add PR comment with reason for doc build failure

samsrabin:
### Description of changes

When a PR has the "build docs" tests fail, it's hard to figure out why. This PR will make it so that the doc build log is posted in a PR comment after the test failure. Example [here](https://github.com/samsrabin/CTSM/pull/17#issuecomment-4309044220).

This is b4b but I think it should come straight to master, and we should not bother making a tag. It's purely a devops thing with the only changes being made to files in `.github/`. Coming straight to master means we won't need to wait 2+ weeks to start benefiting from this, and time is a factor since people will be submitting many docs update PRs soon.

### Specific notes

**Contributors other than yourself, if any:** Claude Sonnet 4.6 and Opus 4.7 1M

**Testing performed, if any:** See https://github.com/samsrabin/CTSM/pull/17.


slevis-lmwg (approved):
Nice feature. For the record, I'm not good at spotting problems in the workflow language, but I'm approving.

ekluzek:
Just looking this over so I understand what it's doing. The docs publish step was changed so that if it fails it keeps a failure artifact. Which then will be used by a new action that will publish a summary of what happened to the PR where the failure happened.

The other thing was a nice change to add a verbose option to a couple scripts so that some of the logging output only happens when verbose is on. This kind of thing is always helpful.

@samsrabin the one thing I'm wondering about is if these artifacts then should be removed by the new failure check action? Or does that happen automatically or maybe with some kind of automatic deletion of old artifcats?

samsrabin:
@ekluzek:
> the one thing I'm wondering about is if these artifacts then should be removed by the new failure check action? Or does that happen automatically or maybe with some kind of automatic deletion of old artifcats?

The artifacts are [retained for 90 days](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization) by default. Unfortunately it looks like this can only be changed at the org level—I would have preferred to set a 1-day retention for this workflow.

Issue #3739, opened 2026-02-05
User's Guide: Add instructions for FATES-nocomp spinup & calibration


samsrabin:
### What sort(s) of documentation issue is this?
- [x] Something is missing.
- [ ] Something is (or might be) incorrect or outdated.
- [ ] Something is confusing.
- [ ] Something is broken.

### What part(s) of the documentation does this concern?
- [ ] [Technical Note](https://escomp.github.io/CTSM/tech_note/index.html) (science and design of the model)
- [x] [User's Guide](https://escomp.github.io/CTSM/users_guide/index.html) (using the model and related tools)
- [ ] Somewhere else (e.g., README file, tool help text, or code comment): _Please specify_
- [ ] I don't know

### Additional context
There is apparently a notes doc that can be converted. If in Google Docs, consider downloading as Markdown as a starting point.

ekluzek:
@adrifoster @ckoven @rgknox @mvdebolskiy and others -- do we have the FATES spinup procedure documented somewhere? If not can someone write a few bullet points on what is done? Thanks

Issue #2648, opened 2024-07-12
Expand FATES land use spin up testing to exercise AD spinup with `use_fates_potentialveg`


glemieux:
> While AD will likely be used for spin-up the main goal of the test is to exercise this `use_fates_potentialveg` to transient fates `use_fates_luh` mode.  Per discussion with @ekluzek it would be good to include a test that exercises going from AD to `use_fates_potentialveg` spin up.

ckoven:
Just wanted to note that I think the standard workflow to spin up the model will be:
(1) AD_SPINUP = true, use_fates_potentialveg = true
(2) AD_SPINUP = false, use_fates_potentialveg = true
(3) AD_SPINUP = false, use_fates_potentialveg = false

not sure if that is how you meant it here, so wanted to just clarify

ekluzek:
An example of how we do a spinup test for CN-Matrix is here:

https://github.com/ESCOMP/CTSM/pull/640/files#diff-75ee12316b38f1c71563ffbebf623f245619a7459c3f148a0adb573d229103f7

The importance here is going through the standard spinup sequence as would be done to do a "real" spinup. This ensures that the entire process works from an infrastructure level (note the system tests for this only run a shortened length for each step). If the standard sequence for FATES spinup is: AD, Potential-Veg, FATES-transient then exercising the whole sequence helps ensure a "real" spinup can always be done by users.

Note, also that when Matrix comes in FATES could use the soil-matrix part of spinup so I think the sequence would be:

(1) AD_SPINUP = true, use_fates_potentialveg = true
(2) AD_SPINUP = true, use_fates_potentialveg = true, use_soil_matrixcn = true
(2) AD_SPINUP = false, use_fates_potentialveg = true
(3) AD_SPINUP = false, use_fates_potentialveg = false

@ckoven does that sound right to you? And I'm I right that FATES spinup should use soil CN-Matrix when it's available? Obviously we will need to test it that it works as expected and saves spinup time, but I at least hope both are true.

ckoven:
@ekluzek we'll have to see whether and how well FATES interacts with matrixcn (I suspect that spinning up FATES will become the rate-limiting step rather than spinning up soil carbon), but yes, that makes sense to me.

Issue #1879, opened 2022-10-25
FATES Accelerated Decomposition Issues


rgknox:
### Brief summary of bug

At the very least, some spin-up flags are not being broadcast when FATES is on, and it is not clear if flags related to AD are being set and applied correctly when FATES is active.  At a first order, I think we need to look closely at a FATES AD spinup and make sure that the correct rate-constants are being amplified during this phase, and returned to normal when starting the follow up phase.

### General bug information

No specific version has been identified, but I don't think the code has changed since we first identified this issue and [ctsm5.1.dev112](https://github.com/ESCOMP/CTSM/releases/tag/ctsm5.1.dev112) .  This issue is the result of investigation by @olyson and I earlier this year.  It is possible that there is no issue, as FATES does bypass some of the soil biogeochemical code, but not all.  At the least, we need to verify that even if AD flags are not being set, they don't need to be set.

### Details of bug

Appoligies in advance for how cloudy this report is, I'm reading throug old emails to reconstruct my understanding.

@olyson and I tried several ways of building the model and noticed an issue related to spinup_state and nitrogen supplementation, although things are still murkey.  It appears that if we didn't actively set suplnitro=ALL, the spinup_state was not being set to 2 in AD model.  It was odd, and it should be irrelevant anyway, as FATES is carbon only at the moment and shouldn't even work outside carbon only mode, which implies suplnitro=ALL.

@olyson also felt that even though we were able to get AD FATES to generate spinup_state = 2, he was suspicious that it was not actually setting amplified rate constatns during spinup.

@olyson did identify that when fates is active, we are not broadcasting flags, for instance:Although, for instance, suplnitro is not used when FATES is on anyway..

```
if (use_cn) then
       call mpi_bcast (suplnitro, len(suplnitro), MPI_CHARACTER, 0, mpicom, ier)
       call mpi_bcast (nfix_timeconst, 1, MPI_REAL8, 0, mpicom, ier)
       call mpi_bcast (spinup_state, 1, MPI_INTEGER, 0, mpicom, ier)
       call mpi_bcast (override_bgc_restart_mismatch_dump, 1, MPI_LOGICAL, 0, mpicom, ier)
    end if
```

Here is a branch we had been working with to resolve some of these issues:

https://github.com/rgknox/ctsm/tree/update_ADfates_histlist

olyson:
Thanks for the nice summary @rgknox .  This prompted me to go back over my own notes and re-setup the simulation we had been working with.
[...]

PR #3962, opened 2026-04-24
Update CLMBuildNamelist to allow AD spinup with FATES (not SP)

### LINKED ISSUE #3771: AD mode currently blocked from being turned on when FATES is on
### Brief summary of bug

When `use_fates=.true.` the xmlchange `CLM_ACCELERATED_SPINUP=on` has no effect, and the `spinup_state` is automatically set to `0`. 

If `spinup_state` is set to something other than `0` in `user_nl_clm` the build script fails with the error:

```
       $log->fatal_error("spinup_state is accelerated (=1 or 2) which is for a BGC mode of CN or BGC," .
                         " but the BGC mode is Satellite Phenology, change one or the other");
```
see [here](https://github.com/ESCOMP/CTSM/blob/90bbcdb8eedb5cf78ccd27a53f142f096e871c2a/bld/CLMBuildNamelist.pm#L1274C1-L1275C95)

We need to rewrite this section of the `CLMBuildNamelist` so that we can use AD mode with FATES.

### General bug information

**CTSM version you are using:** latest

**Configurations affected:** any AD mode

adrifoster:
### Description of changes

Updated CLMBuildNamelist to allow AD spinup with FATES (anything other than FATES-SP) turned on.

### Specific notes

* added some helper variables to check viability of spinup
* added check to only turn on reseed_dead_plants if spinup AND CN

CTSM Issues Fixed (include github issue #): resolves #3771

Are answers expected to change (and if so in what way)? No because we weren't able to run AD mode before?

Will need to test the buildnamelist but everything else should be b4b

@ekluzek I did my best but please feel free to suggest updates given your pearl expertise


bld/CLMBuildNamelist.pm
```diff
@@ -1263,7 +1270,7 @@ sub setup_cmdl_spinup {
        $nl_flags->{'bgc_spinup'} = "off";
     }
     # For AD spinup mode by default reseed dead plants
```
ekluzek: Change the comment to say that reseed dead plants is only for BGC and not FATES

ekluzek: Oh, and that makes me wonder. Does FATES have any kind of equivalent to reseed dead plants?

bld/CLMBuildNamelist.pm
```diff
@@ -1245,7 +1245,14 @@ sub setup_cmdl_spinup {
      }
   }
   $log->verbose_message("CLM accelerated spinup mode is $val");
-  if ( &value_is_true($nl_flags->{'use_cn'}) ) {
+  
+  # is this a BGC-capable mode that supports spinup?
+  # spinup is valid if using CN, or using FATES in non-SP mode.
+  my $use_fates_spinup_ok = &value_is_true($nl_flags->{'use_fates'}) &&
+                            ! &value_is_true($nl_flags->{'use_fates_sp'});
```
ekluzek, line 1252: So this will die if FATES and FatesSP mode which is straightforward enough.

I think the spinup sequence also starts off with particular FATES LUH2 settings. Is that part of the AD step or something that happens later?

ekluzek (changes requested):
The logic here looks right to me. And I think it's clear and easy to see what it's doing and it takes care of both fates and BGC type cases. So thumbs up there.

We do need to add some tests around this though to the namelist tester. There are a bunch of spinup tests there, but we should add some for FATES. And we need to make sure the current ones still work.

As something to work on after this PR, for usability we will want the FATES spinup sequence to be encapsulated into the CLM_ACCELERATED_SPINIP option. Or possibly in a different option. Could you lead me through how that works and we can think about how this should work?

bld/CLMBuildNamelist.pm
adrifoster: No, FATES does not have any kind of "reseed dead plants" mode

bld/CLMBuildNamelist.pm
adrifoster: No I don't think we want AD mode to set anything related to land use

adrifoster:
> The logic here looks right to me. And I think it's clear and easy to see what it's doing and it takes care of both fates and BGC type cases. So thumbs up there.
> 
> We do need to add some tests around this though to the namelist tester. There are a bunch of spinup tests there, but we should add some for FATES. And we need to make sure the current ones still work.
> 
> As something to work on after this PR, for usability we will want the FATES spinup sequence to be encapsulated into the CLM_ACCELERATED_SPINIP option. Or possibly in a different option. Could you lead me through how that works and we can think about how this should work?

Thanks for this review Erik. Just to clarify, are you asking for some bldnamelist tests for this PR?

samsrabin:
SE meeting: @adrifoster will run the existing build-namelist tests and make sure they pass, but she'll file an issue saying that such tests need to be added.

mvdebolskiy:
@adrifoster there is more to this.
https://github.com/ESCOMP/CTSM/blob/27841bdb21e9e444819b05a256a37cf7a7caa2f4/bld/namelist_files/namelist_defaults_ctsm.xml#L76-L77

https://github.com/ESCOMP/CTSM/blob/27841bdb21e9e444819b05a256a37cf7a7caa2f4/bld/namelist_files/namelist_defaults_ctsm.xml#L84-L85

If I understand correctly, the run will fail at runtime, because with fates GPP and NPP (and maybe more) does not exist. See NorESMhub/CTSM#233

adrifoster:
> @adrifoster there is more to this.
> 
> https://github.com/ESCOMP/CTSM/blob/27841bdb21e9e444819b05a256a37cf7a7caa2f4/bld/namelist_files/namelist_defaults_ctsm.xml#L76-L77
> 
> https://github.com/ESCOMP/CTSM/blob/27841bdb21e9e444819b05a256a37cf7a7caa2f4/bld/namelist_files/namelist_defaults_ctsm.xml#L84-L85
> 
> If I understand correctly, the run will fail at runtime, because with fates GPP and NPP (and maybe more) does not exist. See [NorESMhub#233](https://github.com/NorESMhub/CTSM/issues/233)

I usually set my own hist files for every FATES run because everything is so hard-coded. We could try to fix some of this here, but this is a much larger issue

adrifoster:
https://github.com/NorESMhub/CTSM/pull/235/changes look at this to update other things

wwieder:
First, @adrifoster will update this PR (making the easy fixes above).
Then, @slevis-lmwg will work on testing (and maybe creating a new test, which may require coordination with @ekluzek or @glemieux).

UPDATE: Sam L. and Erik agreed to check in about this early next week (July 20 or so).

slevis-lmwg:
@slevis-lmwg note to self from 2026/7/23 software meeting:
- [x] Run the namelist test

slevis-lmwg:
@adrifoster I submitted `./build-namelist_test.pl > & ! test.out &`
which returned (as expected) `# Looks like you failed 2 tests of 3407.`
and test.out shows the usual two expected fails:
```
1089/3407 < FAIL> <Test Id: 1089> <Desc: inventoryfileDNE>
1202/3407 < FAIL> <Test Id: 1202> <Desc: useFATESLUH2fileDNE>
```

adrifoster:
> @adrifoster I submitted `./build-namelist_test.pl > & ! test.out &` which returned (as expected) `# Looks like you failed 2 tests of 3407.` and test.out shows the usual two expected fails:
> 
> ```
> 1089/3407 < FAIL> <Test Id: 1089> <Desc: inventoryfileDNE>
> 1202/3407 < FAIL> <Test Id: 1202> <Desc: useFATESLUH2fileDNE>
> ```

ah okay great. @ekluzek does this mean it is ready to come in? I know we want to add tests to increase our coverage but perhaps that can be done later? Or what do you think?

slevis-lmwg:
Pinging @ekluzek reminder to meet about this soon.

slevis-lmwg:
From meeting with @ekluzek 
TODOs (not necessarily all in this PR):
- Add build-namelist tests (similar coverage as we have for BGC, `grep clm_accel build-namelist_test.pl`, including SASU although we're not encouraging or discouraging SASU with FATES), e.g. with
  - [ ] FATES AD that works
  - [ ] FATES SP AD that fails
- Add system tests to the fates test-list
  - [ ] SMS to test FATES spinup
  - SSP FATES specialized test: do in #2648 later because it's more difficult, need to know the FATES spinup sequence (see also #1879, see also #3739 about documentation)

Issue #627, opened 2019-01-29
Assess potential speedups from vectorization


billsacks:
I strongly suspect that little, if any, CTSM code is currently being vectorized, due to a combination of (1) our use of pointers (#235), (2) our indirect indexing via filters (#236), (3) the complex logic in many of our loops, and (4) the fact that many of our loops call other subroutines (including endrun). We should assess whether we could get a significant performance benefit if our code was more vector-friendly. Here are some steps we could take:

- [ ] For one or more simple loops, without any conditionals or function / subroutine calls: assess whether they're being vectorized.
  * Can check whether the loop is vectorized via intel's vectorization report that you can produce during compilation.
  * If our loops are NOT being vectorized, assess whether this is due to:
    * Our use of filters (#236)
    * Our use of pointers: does switching to allocatables help? (#235) (first just try this for the small set of variables used in the trial loop(s))

- [ ] For one or more loops that are getting vectorized: Assess whether we're getting much performance benefit from the vectorization. It would be good to check this both for single loops and for nested loops (with an outer loop over level, for example): Vectorization may have little performance benefit for non-nested loops, because data elements aren't reused, so CPU time is swamped by memory access time.
  * For a loop that is vectorized: can check performance improvement by comparing the vectorized version with a non-vectorized version (which can be forced with an in-code directive to the intel compiler: `!dir$ novector`.
  * If we're not getting much benefit from vectorization based on some sample loops (e.g., because we're much more memory-bound than cpu-bound), then we could give up on trying to make our code more vectorizable

- [ ] If we ARE getting significant speedups from vectorization for fairly simple loops, then look into:
  * Are conditionals inside loops preventing vectorization?
  * Are function / subroutine calls inside loops preventing vectorization? I expect this to be the case for calls to other modules, but what about calls to routines in the same module, which could potentially be inlined?

ekluzek:
We are looking into this with the OpenACC/NVIDIA Hackathon 2026 and it looks like the answer is a resounding yes. Vectorization makes a huge difference in performance. And since we don't have our loops vectorizing, working on this will provide significant gains in performance.

billsacks:
See https://github.com/ESCOMP/CTSM/pull/4017 for some investigation of this in one sample case. And see also my comment here: https://github.com/ESCOMP/CTSM/issues/235#issuecomment-4426274200

PR #4017, opened 2026-05-12
Documenting some experimentation to allow vectorization as part of the 2026 NVIDIA hackathon

billsacks:
I am opening this PR just to document some of the work I did for the recent NVIDIA hackathon with @samsrabin and @ekluzek , towards #627 and related issues. I don't think anything here is worth bringing in at this time, but wanted to save this for future reference. (See also https://github.com/billsacks/prototypes-fortran_vectorization.)

I did my testing in the context of the nvidia-hackathon-26 branch, but here rebased my changes onto master for ease of opening a PR.

I first looked at the intel vectorization reports (by adding the compilation flags `-qopt-report=3 -vec-threshold0`; the latter tells intel to ignore its heuristics on profitability of vectorization and vectorize whenever possible, giving the estimated speedup/slowdown in the optimization report). Conclusions from this were that the important changes for good vectorization speedup are:
- Declaring array arguments to be contiguous
- Changing indirect indexing to direct indexing (getting rid of filters)
- Adding the `-fimf-use-svml` compilation flag for the sake of the math routines in this target subroutine

Then I did timings. Because this target routine is such a small fraction of the overall runtime, I added a loop with 100 repetitions to get more precise results. I first tried putting the timer outside the subroutine call, but then wondered how much of the timing change was due to the subroutine call itself (e.g., due to a different amount of time being needed to do the subroutine call with local arrays vs. members of a derived type), so then put the timer inside the subroutine as it is in this PR... but from comparing timings from both, it seems like the overhead of the subroutine call is a relatively small amount of time, so I think the results / conclusions are about the same with the timer either inside or outside the subroutine.

In my initial timings, with our standard compilation flags, I saw unexpected results: Using filtered/compressed arrays and adding the `contiguous` attribute on the array arguments actually led to a *slowdown* relative to the baseline with ifx. My hypothesis (partly based on looking at intel's optimization report) is that this code change leads the compiler to predict a benefit from vectorization, so it vectorizes the loop, but in reality, performance is worse with vectorization, probably due to using non-vector forms of the math libraries.

Then I redid some timings after adding `-fimf-use-svml`. Here I focus on ifx timings, and am just giving the mean timing from the timing files.

Baseline timing (only code changes were adding a timer and adding 100 repetitions; no `-fimf-use-svml` flag): 0.1664 

Adding `-fimf-use-svml` flag: 0.1348

Additionally adding `contiguous` attribute: 0.1074

Additionally changing to filtered/compressed arrays (note that this does *not* include the time needed to filter / unfilter the arrays): 0.0816

Note that using filtered versions but without the `contiguous` attribute gave a time of 0.2518  – so it seems that contiguous is important when using vectorized code?

Back to using contiguous plus filtered, and additionally removing the `use_clm5_fpi` conditional (out of curiosity about how much the conditional is slowing things down): 0.0709.

### DIFF
--- a/src/biogeophys/CanopyHydrologyMod.F90
+++ b/src/biogeophys/CanopyHydrologyMod.F90
@@ -197,6 +197,7 @@ contains
      !
      ! !LOCAL VARIABLES:
      integer  :: i     ! index of water tracer or bulk
+     integer :: rep
      real(r8) :: dtime ! land model time step (sec)
 
      real(r8) :: qflx_liq_above_canopy_patch(bounds%begp:bounds%endp)        ! liquid water input above canopy (rain plus irrigation) [mm/s]
@@ -246,6 +247,7 @@ contains
           qflx_liq_above_canopy = qflx_liq_above_canopy_patch(begp:endp), &
           forc_snow_patch       = forc_snow_patch(begp:endp))
 
+     do rep = 1, 100
      call BulkFlux_CanopyInterceptionAndThroughfall(bounds, num_nolakep, filter_nolakep, &
           ! Inputs
           patch                 = patch, &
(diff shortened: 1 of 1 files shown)


billsacks:
Since I only opened this for future reference, I'm going ahead and closing it.

PR #4018, opened 2026-05-13
Add module for designating patch/col/etc. as "debug"

samsrabin:
### Description of changes

Often we want to add `write` statements as part of debugging model behavior, but this can quickly become hard to parse if we don't limit the scope of the messaging. If we're debugging just one patch, column, gridcell, or landunit, we can make the messaging conditional on the object having the correct index. This PR adds a new module, `debugMod`, that facilitates this.

`debugMod` allows the user to hard-code a global index for patch, column, landunit, and/or gridcell to debug. It then provides functions where the user can give either a local or global index and get back a boolean that's true if it's the debugged one and false otherwise.

### Specific notes

- None

- None

**Any user interface changes (namelist or namelist defaults changes)?** No

- [x] Tested during some column-level troubleshooting
- [ ] Add unit tests.

- [x] The code in this PR branch runs with no errors. **Column-level troubleshooting**
- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Which?:** A

- [x] I have reviewed relevant parts of the CLM documentation [Tech Note](https://escomp.github.io/CTSM/tech_note/index.html) or [User's Guide](https://escomp.github.io/CTSM/users_guide/index.html) to determine if anything needs to be changed or added. **If it does, describe:** Should add a bit in the [Troubleshooting](https://escomp.github.io/CTSM/users_guide/trouble-shooting/index.html) section of the User's Guide
- [ ] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates (see [guidelines for contributing documentation](https://escomp.github.io/CTSM/users_guide/working-with-documentation/docs-intro.html#contribution-guidelines)). **Which?:** Will include updates, but I haven't made them yet.


src/main/debugMod.F90
```diff
@@ -0,0 +1,202 @@
...
+  use decompMod        , only : bounds_level_proc
+  use decompMod        , only : subgrid_level_gridcell, subgrid_level_landunit
+  use decompMod        , only : subgrid_level_column, subgrid_level_patch
+  use decompMod        , only : get_global_index
+  use spmdMod          , only : masterproc, mpicom
+  use abortutils       , only : endrun
+  use shr_log_mod      , only : errMsg => shr_log_errMsg
+  use clm_time_manager , only : timemgr_restart_io, get_nstep, get_curr_date
+  use clm_varctl       , only : iulog
+  use clm_varcon       , only : nameg, namel, namec, namep
```
samsrabin: Need to remove a bunch of unused things.

src/main/debugMod.F90
```diff
@@ -0,0 +1,202 @@
...
+  !
+  ! !PUBLIC TYPES:
+  implicit none
+  private
+  !
+  ! !PUBLIC MEMBER FUNCTIONS:
+!   public :: do_debug_patch
+  public :: do_debug_column
+!   public :: do_debug_landunit
+!   public :: do_debug_gridcell
```
samsrabin: Need to uncomment all these.

src/main/debugMod.F90
```diff
@@ -0,0 +1,202 @@
...
+  !
+  ! !PRIVATE TYPES:
+
+  ! Global indices for objects to debug.
+  ! When merging to a CTSM branch, these must always be < 1!!
+  integer, parameter :: debug_p = -999
+  integer, parameter :: debug_c = 2787
```
samsrabin: Set this to -999.

samsrabin:
Good idea from @mvdebolskiy:
- ~~Add an option to the functions that would have them return true if any _parent_ is debug. So for instance, you could set a patch, column, or landunit to debug, then do `do_debug_patch(..., include_parent=.true.)`.~~ This is too messy. I would have to add optional `i_local_` and `i_global_` arguments for every possible parent.

samsrabin:
SE meeting:
- [ ]  Bill's [POI module](https://escomp.github.io/CTSM/users_guide/trouble-shooting/trouble-shooting.html#use-the-point-of-interest-module) does this with lat/lon. Bring that in here.
- [x] (Done with 97d16bcf8.) Make indices namelist params and just require that no merged code leave in `do_debug_*` calls.

src/main/debugMod.F90
samsrabin: Done with db92e6e38.

src/main/debugMod.F90
samsrabin: Done with 43e989df2.

src/main/debugMod.F90
samsrabin: Done with 6e22851e8.

PR #3810, opened 2026-03-12
ctsm5.4.027: Important coupled model  final answer changes: param file and n_melt_glcmec

ekluzek:
### Description of changes

Some important final answer changes needed for the coupled model to get the right parameter file and to set n_melt_glcmec=1 for clm6_0.

Also use the 2015 finidat file for SSP future scenarios. And fix some of the namelist picking for finidat files, especially for: 2003, 2010, and 2015.

### Specific notes

Contributors other than yourself, if any: @slevis-lmwg 

CTSM Issues Fixed (include github issue #):
  Fixes #3799
  Fixes #3824
  Fixes #3826
  Fixes #3835 
  Fixes #3836

Are answers expected to change (and if so in what way)? Yes!
   coupled model and clm60cam7LndTuning mode tests change answers due to parameter file
   clm6_0 cases change answers because of change in n_melt_glcmec
   SSP cases change answers for cases where the finidat file changes

Any User Interface Changes (namelist or namelist defaults changes)? Yes
   n_melt_glcmec is 1 for clm6_0
   paramfile for clm6_0_cam7.0 changes
   2015 finidat file for SSP cases starting in 2015
   2010 finidat files will be used for clm6_0
   2003 finidat file will be used for clm4_5 with cam6.0/cam7.0

Does this create a need to change or add documentation? Did you do so? No No 

Testing performed, if any: Will do regular

PR #4029, opened 2026-05-18
ctsm5.4.040: Update clm6 paramfiles and add a new one for hillslope hydrology

### LINKED ISSUE #3659: Check generic crop parameter values
For CLM BGC no crop (16 pft) compsets the parameter values for C3 irrigated and C3 non-irrigated crops should be identical to the generic C3 grass. 

@linniahawkins Needs to check and fix this in the parameter file. Assign this to me when it is appropriate to make the change.

slevis-lmwg:
### Description of changes

Updating the clm6 paramfiles.

### Specific notes

@wwieder @linniahawkins 

Resolves #3659 

Answers change from the baseline due to changing param values in the CLM6 paramfiles.

Changing two paramfiles and adding a new one for hillslope hydrology in namelist_defaults.

- [x] ./build-namelist_test.pl OK
- [x] derecho `./run_sys_tests -s aux_clm -c ctsm5.4.039 -g ctsm5.4.040`
- [x] izumi `./run_sys_tests -s aux_clm -c ctsm5.4.039 -g ctsm5.4.040`

- [x] The code in this PR branch runs with no errors. **Briefly describe tested configuration(s):** aux_clm
- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Answer: (c) Simulations https://github.com/NCAR/LMWG_dev/issues/164 and https://github.com/NCAR/LMWG_dev/issues/167**
- [x] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates (see [guidelines for contributing documentation](https://escomp.github.io/CTSM/users_guide/working-with-documentation/docs-intro.html#contribution-guidelines)). **Answer: (a)**
- [x] Update ChangeLog: Mention issues lmwg_dev 164 and 167 as the test runs, we are not pointing to new IC in the defaults because of various non-standard things in these runs

### DIFF
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -2264,9 +2264,10 @@ sub setup_logic_params_file {
   # get param data
   my ($opts, $nl_flags, $definition, $defaults, $nl) = @_;
 
+  $nl_flags->{'use_hillslope'} = $nl->get_value('use_hillslope');
   add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'paramfile',
               'phys'=>$nl_flags->{'phys'},
-              'use_hillslope'=>$nl->get_value('use_hillslope'),
+              'use_hillslope'=>$nl_flags->{'use_hillslope'},
               'lnd_tuning_mode'=>$nl_flags->{'lnd_tuning_mode'},
               'use_flexibleCN'=>$nl_flags->{'use_flexibleCN'} );
 }
--- a/bld/CLMBuildNamelist.pm
+++ b/bld/CLMBuildNamelist.pm
@@ -2266,6 +2266,7 @@ sub setup_logic_params_file {
 
   add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'paramfile',
               'phys'=>$nl_flags->{'phys'},
+              'use_hillslope'=>$nl->get_value('use_hillslope'),
               'lnd_tuning_mode'=>$nl_flags->{'lnd_tuning_mode'},
               'use_flexibleCN'=>$nl_flags->{'use_flexibleCN'} );
 }
--- a/bld/namelist_files/namelist_defaults_ctsm.xml
+++ b/bld/namelist_files/namelist_defaults_ctsm.xml
@@ -600,9 +600,10 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- ================================================================== -->
 <!-- The default filenames are given relative to the root directory
      for the CLM2 data in the CESM distribution -->
-<!-- Plant function types (relative to {csmdata}) -->
-<paramfile phys="clm6_0" >lnd/clm2/paramdata/ctsm60_params.c260303.nc</paramfile>
-<paramfile phys="clm6_0" lnd_tuning_mode="clm6_0_cam7.0">lnd/clm2/paramdata/ctsm60-cam70_params.c260305.nc</paramfile>
+<!-- CLM parameter files (relative to {csmdata}) -->
+<paramfile phys="clm6_0" >lnd/clm2/paramdata/ctsm60_params.c260518.nc</paramfile>
+<paramfile phys="clm6_0" lnd_tuning_mode="clm6_0_cam7.0">lnd/clm2/paramdata/ctsm60-cam70_params.c260518.nc</paramfile>
+<paramfile phys="clm6_0" use_hillslope=".true.">lnd/clm2/paramdata/ctsm60-HH_params.c260518.nc</paramfile>
 <paramfile phys="clm5_0" >lnd/clm2/paramdata/clm50_params.c260305.nc</paramfile>
 <paramfile phys="clm4_5" >lnd/clm2/paramdata/clm45_params.c260305.nc</paramfile>
 
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -66,6 +66,7 @@ Changes answers relative to baseline: Yes
        - LMWG_dev issue number(s):
  https://github.com/NCAR/LMWG_dev/issues/164
  https://github.com/NCAR/LMWG_dev/issues/167
+ Not changing finidat files in the defaults at this time due to various non-standard things in these test simulations.
 
 Other details
 -------------
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,3 +1,78 @@
+===============================================================
+Tag name: ctsm5.4.040
+Originator(s): slevis (Samuel Levis,UCAR/TSS,303-665-1310)
+Date: Mon May 18 03:29:57 PM MDT 2026
+One-line Summary: Paramfile updates
+
+Purpose and description of changes
+----------------------------------
+
+ Updating the clm6 paramfiles.
+
+Significant changes to scientifically-supported configurations
+--------------------------------------------------------------
+Does this tag change answers significantly for any of the following physics configurations?
+(Details of any changes will be given in the "Answer changes" section below.)
+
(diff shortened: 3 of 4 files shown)


bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -600,9 +600,10 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- ================================================================== -->
 <!-- The default filenames are given relative to the root directory
      for the CLM2 data in the CESM distribution -->
-<!-- Plant function types (relative to {csmdata}) -->
-<paramfile phys="clm6_0" >lnd/clm2/paramdata/ctsm60_params.c260303.nc</paramfile>
-<paramfile phys="clm6_0" lnd_tuning_mode="clm6_0_cam7.0">lnd/clm2/paramdata/ctsm60-cam70_params.c260305.nc</paramfile>
+<!-- CLM parameter files (relative to {csmdata}) -->
+<paramfile phys="clm6_0" >lnd/clm2/paramdata/ctsm60_params.c260518.nc</paramfile>
+<paramfile phys="clm6_0" lnd_tuning_mode="clm6_0_cam7.0">lnd/clm2/paramdata/ctsm60-cam70_params.c260518.nc</paramfile>
+<paramfile phys="clm6_0" use_hillslope=".true.">lnd/clm2/paramdata/ctsm60-HH_params.c260518.nc</paramfile>
```
slevis-lmwg, line 606: @wwieder 
I didn't end up moving this section to after the use_hillslope section because `use_hillslope` here needs to be .true. to pick up the HH file, while below `use_hillslope` always gets set to .false. and would have no effect on this line.

slevis-lmwg: In the two Hillslope tests, despite ` use_hillslope = .true.`, I got 
paramfile = '/glade/campaign/cesm/cesmdata/inputdata/lnd/clm2/paramdata/ctsm60_params.c260518.nc'
This reminded me to update to CLMBuildNamelist.pm but no luck. Now I remember that @ekluzek had to make more changes to get lnd_tuning_mode to work with the corresponding paramfile.
- I will look for Erik's PR #3810 in case I can reproduce what he had to do
- If not, then I can put the HH paramfile in the Hillslope testmods, and then we still need to figure out how to use the HH paramfile out of the box.

slevis-lmwg: Three Hillslope tests in aux_clm (the third on izumi):
```
ERP_D_P64x2_Ld10.f10_f10_mg37.I2000Clm60Bgc.derecho_intel.clm-Hillslope.GC.0518-142638de_int
SMS_D_Mmpi-serial_Ld5.5x5_amazon.I2000Clm60Bgc.derecho_gnu.clm-HillslopeC.GC.0518-142638de_gnu
SMS_D_Ld3.f10_f10_mg37.I2000Clm60Bgc.izumi_nag.clm-HillslopeD.GC.0518-150814iz_nag
```

slevis-lmwg: Ok, so it's not just the paramfile not seeing the `use_hillslope = .true.`. This other namelist parameter that has been supposed to be .true. with use_hillslope, is also not getting set correctly:
```
add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'hillslope_fsat_equals_zero', 'use_hillslope'=>$nl_flags->{'use_hillslope'} );
```
@wwieder this perl stuff may take me a bit to resolve, or I could open an issue. Preference?

slevis-lmwg: I may have found a solution to pick up the HH paramfile (testing now). Regardless, I will open an issue for hillslope_fsat_equals_zero, as out of scope in this PR.

slevis-lmwg:
Ok, now ready for review, and we can first discuss if you prefer @wwieder.

bld/CLMBuildNamelist.pm
```diff
@@ -2266,6 +2266,7 @@ sub setup_logic_params_file {
 
   add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'paramfile',
               'phys'=>$nl_flags->{'phys'},
+              'use_hillslope'=>$nl->get_value('use_hillslope'),
```
ekluzek: To make this robust, we should put use_hillslope into $nl_flags and handle it like other nl_flags options. This makes sure that it will work regardless of how use_hillslope is turned on. For example, it will work if you set " use_hillslope = .true.", but would fail if you set it to "use_hillslope=T", or one of the other forms that evaluate to TRUE in the Fortran code.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek, line 606: It's painful to start thinking about this, but are there any needed permutations of lnd_tuning_mode and use_hillslope needed here? I'm assuming that we only expect use_hillslope to be used with CRUJRA24b forcing?

ekluzek: The other part here is that this might come into play in how lnd_tuning_mode is handled in buildnml. At least in how it's reported. I should probably think about that and file an issue if it looks like we should make changes there. That would be an update that comes in later.

But, it basically shows that lnd_tuning_mode now is going to be different for the hillslope model than for the default flat model. And we probably aren't going to tune the hillslope model for as many configurations as the flat model. So that is likely something we'll want documented and the user to be aware of.

ekluzek: I forget if namelist items are default last or first. But, if you were to run coupled with LND_TUNING_MODE==clm6_0_cam7.0 and hillslope on -- I'm not sure what you will get. Nor am I sure what you'd want to get for that case either.

That could be something to decide on. To make it clear you could also add:

lnd_tuning_mode="clm6_0_crujra" to the attributes for the hillslope case to make it clear. But, it also depends on which file we think we want to have set for a hillslope case using different LND_TUNING_MODE.

ekluzek (approved):
The basic functionality is what needs to be done.

I'm hitting approve so you don't have to track me down. But, the one suggested change I gjive is something that I think should be done. And I think it's straightforward enough to do. Otherwise, we can do it later.

And there's a more subtle thing that came to my mind that will need more thought and some discussion to figure out, so should be done later.

bld/namelist_files/namelist_defaults_ctsm.xml
wwieder: Thanks for all these thoughts, @ekluzek.  I agree there is a lot of potential complication here.  We should try to keep it simple as much as possible by narrowly defining what we'll support (e.g. CLM6_CRUJRA_HH compset)?

bld/CLMBuildNamelist.pm
slevis-lmwg: Rerunning aux_clm with this change on derecho, comparing to the baseline that I already created, since I expect identical answers:
`./run_sys_tests -s aux_clm -c ctsm5.4.040 -g ctsm5.4.040b`

bld/namelist_files/namelist_defaults_ctsm.xml
slevis-lmwg: Thank you for starting this conversation @ekluzek. I will let you open the issue.

slevis-lmwg: Or rather I will open the issue right now with the above info, so that we can resolve this conversation.

slevis-lmwg: Opened #4034

Issue #2352, opened 2024-02-08
Move to the mpi_f08 library in CTSM


ekluzek:
We talked about this in the last CSEG meeting. CAM-SIMA is likely to be moving to the mpi_f08 library, and it would be good for CESM to as well. This is a full 2008 library which adds type checking and use modules for MPI. So it brings in a lot of error checking that would be beneficial.

It looks like CTSM widely uses shr_mpi_mod and most MPI calls are hidden under spmdMod. So the following would need to change:

utils/clmfates_interfaceMod.F90
utils/spmdMod.F90
main/decompInitMod.F90
main/controlMod.F90

ekluzek:
Note, that PR #4102 is bringing in a module where "use mpi_f08" is added. Previous to this I had seen builds fail when this was added (probably just on Izumi_nag), but with this working we should be freer to do this more generally. The error checking that you get with mpi_f08 is more robust, can fix errors, and will reduce compiler warnings so long term this will be a good project to take on. In CSEG we reiteratted the need to do this post-CESM30.

Issue #3817, opened 2026-03-17
Screwy ERP CN threaded tests start failing in ctsm5.4.025


ekluzek:
### Brief summary of bug

I didn't notice this when I made the ctsm5.4.025 tag since these tests are expected to fail for COMPARE_base_rest. But, they started failing at RUN at ctsm5.4.025. But, since ctsm5.4.025 was b4b -- maybe they would work on resubmit?

```
ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-flexCN_FUN--clm-matrixcnOn_ignore_warnings      ( RUN)         EXPECTED (COMPARE_base_rest)
ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-noFUN_flexCN--clm-matrixcnOn_ignore_warnings    ( RUN)         EXPECTED (COMPARE_base_rest)
```

### General bug information

**CTSM version you are using:** ctsm5.4.025
**Does this bug cause significantly incorrect results in the model's science?** No

**Configurations affected:**
   Just tests with CN matrix where threading is turned on. Since this is a screwy problematic test, it's probably fine for it to fail. And maybe that means we should remove these tests?

ekluzek:
This relates to #2619

ekluzek:
I got them to work in ctsm5.4.025, by submitting them a few times. So the baselines are good now.

samsrabin:
Are you sure that this only started happening in ctsm5.4.025? I looked back at the testing I did for what became ctsm5.4.023, and I got `ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-flexCN_FUN--clm-matrixcnOn_ignore_warnings` failing in RUN. This was before [reverting the `finidat` file changes](https://github.com/ESCOMP/CTSM/pull/3809/commits/a7098fa0168719f281de34f151d3a7ec056aacf7), though.

Test directory at `/glade/derecho/scratch/samrabin/tests_0309-141321de/ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-flexCN_FUN--clm-matrixcnOn_ignore_warnings.GC.0309-141321de_int/`.

samsrabin:
Per today's SE meeting: reduce to just one test, ideally one that only ever fails BASELINE. Probably the REP test.

[...]

PR #4102, opened 2026-06-25
Reduce the memory usage that is important for ne1024 simulation

### LINKED ISSUE #4103: Out of memory issue in CTSM at ne1024 resolution
### Brief summary of bug

CTSM initialization consumes excessive memory (and time) at very high resolution (e.g. ne1024), to the point of risking out-of-memory during startup. The cost is dominated by global-sized arrays replicated identically on every MPI rank, NetCDF/PIO file handles held open long past their last use, and duplicate ESMF mesh construction for CDEPS streams whose data already lives on the model grid.

### General bug information

**CTSM version you are using:** ctsm5.4.044-3-gd3078f4e0

**Configurations affected:** High-resolution land configurations, in particular `ne1024`

### Details of bug

1. Per-rank replication of the global land mask (src/cpl/share_esmf/lnd_set_decomp_and_domain.F90). In both lnd_set_lndmask_from_maskmesh and lnd_set_lndmask_from_lndmesh, lndmask_glob(gsize) is allocated on every MPI rank and built with an all-rank ESMF_VMAllReduce into a second global-sized integer temporary (itemp_glob). At ne1024 this is two global-sized integer arrays held per rank with identical contents across every rank on a node — pure duplication that scales with rank count.

2. NetCDF/PIO file handles held open past their last use. Several init routines open a file, finish reading, and only close the handle much later — or, on at least one path, never:
- src/main/clm_instMod.F90 — params_ncid closed at the end of init_accflds rather than right after its last use (bgc_vegetation_inst%Init).
- src/main/initVerticalMod.F90 — ncid closed at the end of initVertical rather than after the last read (STD_ELEV).
- src/biogeophys/UrbanParamsType.F90 — ncid leaked on the nlevurb == 0 early-return path in UrbanInput.
- src/main/organicFileMod.F90 — no close after reading ORGANIC in organicrd.
- src/main/surfrdMod.F90 — handles left open after the dimension reads in surfrd_get_num_patches and surfrd_get_nlevurb.

3. Duplicate ESMF mesh construction for redist streams (src/cpl/share_esmf/PrigentRoughnessStreamType.F90 and src/cpl/share_esmf/UrbanTimeVarType.F90). For streams already on the model grid (mapalgo == 'redist') — as for the ne1024 Prigent-roughness and urban-time-varying inputs — shr_strdata_init_from_inline still reads the stream mesh file and builds a second full ESMF mesh identical to the already-built CLM model mesh. This duplicate is a large init memory/time cost at ne1024. Avoiding it requires CDEPS to accept a caller-supplied mesh (stream_mesh_in), added in ESCOMP/CDEPS#414.

FHISTC_LTso compset at ne1024 resolution, Derecho, Intel compiler

sjsprecious:
### Description of changes

This PR aims at reducing memory usage of a simulation at ne1024 resolution. All the changes are done by Claude under my supervisory.

---

The goal is to cut CTSM initialization memory (and some init time) at high resolution (like `ne1024`), where per-rank data replication and duplicate ESMF mesh construction dominate startup cost.

**The detailed edits:**

1. New per-node shared-memory helper: clm_shmem_mod.F90

A MPI-3 shared-memory module and specialized for CTSM's decomposition setup. The idea is that arrays that are otherwise allocated identically on every MPI rank instead get one physical copy per shared-memory node, mapped into every rank on that node — freeing **ranks_per_node − 1** copies per node.

- clm_shmem_alloc_i4_1d(ptr, win, n) — allocate a node-shared default-integer rank-1 array (only the node leader requests storage via `MPI_Win_allocate_shared`; peers map the leader's segment via `MPI_Win_shared_query`).

- clm_shmem_leader_allreduce_sum_i4(ptr, win, n) — fence → node leaders sum partials across nodes over a leader-only communicator → fence to publish. Builds a globally-summed array in the shared buffer without every rank holding a global-sized copy.

- clm_shmem_free / clm_shmem_fence / clm_shmem_is_leader / clm_shmem_leader_comm / clm_shmem_npes_per_node — lifecycle and query helpers; lazily build node-local and node-leader communicators via `mpi_comm_split_type(MPI_COMM_TYPE_SHARED)`.

2. lnd_set_decomp_and_domain.F90 — apply the shmem helper to the global land mask

The global land mask `lndmask_glob(gsize)` was previously allocated on every rank and built with an all-rank `ESMF_VMAllReduce` into a second global-sized temporary (`itemp_glob`). Now, in both code paths (`lnd_set_lndmask_from_maskmesh` and `lnd_set_lndmask_from_lndmesh`):

- `lndmask_glob` is allocated once per node via `clm_shmem_alloc_i4_1d`, with a new `lndmask_win` window handle threaded through both subroutine signatures.

- Leader zeroes it, fence, each rank fills its disjoint local indices, then `clm_shmem_leader_allreduce_sum_i4` replaces the `ESMF_VMAllReduce + itemp_glob` temporary (the temporary is deleted entirely).

- Cleanup is now branch-aware: the cmeps driver paths free via `clm_shmem_free(lndmask_glob, lndmask_win)`; the `lilac` path still uses plain `deallocate` (it uses a plain allocate).

This removes two global-sized integer arrays per rank (the mask copy + the all-reduce temp), replaced by one node-shared copy.

3. NetCDF file-handle close fixes

Closing pio file handles that were opened but closed late or never — frees buffers earlier in init:

- clm_instMod.F90: moves `ncd_pio_closefile(params_ncid)` earlier — to right after its last use (`bgc_vegetation_inst%Init`) instead of at the end of `init_accflds`.
- initVerticalMod.F90: moves `ncd_pio_closefile(ncid)` to right after the last read (`STD_ELEV`) instead of the end of `initVertical`.
- UrbanParamsType.F90: adds a missing `ncd_pio_closefile(ncid)` on the early-return path (`nlevurb == 0`) that previously leaked the handle.
- organicFileMod.F90: adds `ncd_pio_closefile(ncid)` after reading ORGANIC.
surfrdMod.F90: adds two `ncd_pio_closefile(ncid)` calls after dimension reads complete (after the pft/cft dims, and after nlevurb).

4. reuse already-built model mesh for `redist` streams

- PrigentRoughnessStreamType.F90 / UrbanTimeVarType.F90: the changes are now done in CDEPS whenever these streams use `stream_mapalgo='redist'`

_____

### Specific notes

- (Replace this text and add more list items as needed)

- Resolves #4103
- Some work on #2352
- We did add a pointer array which works against #235, but we felt it should be addressed later
- A test will be added later see #4137
- The mpi_shmem_mod.F90 module will be moved to cesm_share later as well see #4138

No.

aux_clm

- [x] The code in this PR branch runs with no errors. **Briefly describe tested configuration(s): aux_clm**
- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **(a) does not change answers**

- [x] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates (see [guidelines for contributing documentation] (https://escomp.github.io/CTSM/users_guide/working-with-documentation/docs-intro.html#contribution-guidelines)). **Which?:** a


samsrabin:
Thanks for this, @sjsprecious! A couple of questions:
1. Do you have a date you need this in by?
2. Do you expect this to give bit-for-bit identical results to the previous version?

@ekluzek, I'm assigning you for now given your recent work on our task decomposition, but I'm also adding Next so we can discuss in our SE meeting.

sjsprecious:
Thanks @samsrabin for your quick reply. To answer your questions:

> 1. Do you have a date you need this in by?

We are waiting for a new tag for these CTSM changes so that our collaborators can start their scientific runs soon. Thus  I would say no hard date, but the sooner, the better.

> 2. Do you expect this to give bit-for-bit identical results to the previous version?

Yes, these changes should not change the answers for CTSM. I am happy to do some tests on Derecho if you can share the detailed instructions.

Let me know if you or Erik has any comments/suggestions about these code changes.

samsrabin:
If you could run the `aux_clm` test suite on Derecho to make sure it all works, that'd be awesome. From the top level of your CTSM checkout:

```shell
conda run -n ctsm_pylib ./run_sys_tests -s aux_clm --compare ctsm5.4.044 --skip-generate
```

If you don't have the `ctsm_pylib` conda environment installed, follow the instructions [here](https://escomp.github.io/CTSM/users_guide/using-clm-tools/using-ctsm-pylib.html).

samsrabin:
Also, please fill out the PR template. I've added it to the bottom of your PR description. Thanks!

sjsprecious:
Thanks @samsrabin for your detailed instructions. While I was still running the `aux_clm` test suite on Derecho, most of my tests finished successfully but hit the following error:
```
ERROR BFAIL baseline directory '/glade/campaign/cgd/tss/ctsm_baselines/ctsm5.4.044/SMS.f45_f45_mg37.I2000Clm60FatesSpRsGs.derecho_nvhpc.clm-FatesColdSatPhen' does not exist
```
I think this is because I did not have access to the directory `/glade/campaign/cgd/tss/ctsm_baselines/ctsm5.4.044`. Can you help grant me the access to that directory or can you help check whether my output is BFB?

samsrabin:
Ah, sorry, I probably should have expected that. If you send me the path of your test directory, I can check myself. Thanks so much for doing this!

sjsprecious:
Thanks @samsrabin . My test output is at `/glade/derecho/scratch/sunjian/tests_0630-094247de` and let me know what you find out later.

I saw some tests failed at runtime. However, I am not familiar with those setups and not sure how my changes affect them. Can you provide more details about those failed tests, too?

samsrabin:
It looks like the only test failing RUN was `ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-noFUN_flexCN--clm-matrixcnOn_ignore_warnings`. There's a known issue about that (#3817); it might take a few resubmissions (using `./case.submit` in the test directory), but it should eventually work. If it doesn't work after two resubmissions, just note that here and don't worry about it.

Looks like all the other tests pass as expected!

sjsprecious:
> It looks like the only test failing RUN was `ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-noFUN_flexCN--clm-matrixcnOn_ignore_warnings`. There's a known issue about that (#3817); it might take a few resubmissions (using `./case.submit` in the test directory), but it should eventually work. If it doesn't work after two resubmissions, just note that here and don't worry about it.
> 
> Looks like all the other tests pass as expected!

Thanks @samsrabin for checking it. Indeed after resubmitting the same test a few times, it finished successfully. The output is available at `/glade/derecho/scratch/sunjian/tests_0630-094247de/ERP_D_P64x2_Ld3.f10_f10_mg37.I2000Clm50BgcCru.derecho_intel.clm-noFUN_flexCN--clm-matrixcnOn_ignore_warnings.C.0630-094247de_int/run`. Can you please help check whether it is BFB against the baseline?

samsrabin:
Yup, looks good!

samsrabin:
SE meeting:
- This is blocked by the required CDEPS update.
- Since this has submodule updates, we will merge this to master just to be safe—even though it should be bit-for-bit.
- @ekluzek will handle this.

sjsprecious:
> SE meeting:
> 
> * This is blocked by the required CDEPS update.
> * Since this has submodule updates, we will merge this to master just to be safe—even though it should be bit-for-bit.
> * @ekluzek will handle this.

Thanks @samsrabin . My CDEPS PR was merged and I just updated the CDEPS tag here. Let me know if you or @ekluzek have any questions or comments before merging it.

samsrabin:
@briandobbins I might have misremembered. Is there also a CMEPS PR this depends on?

sjsprecious:
My CMEPS PR is independent of my CTSM PR here.

briandobbins:
How critical is your CMEPS PR to the memory reduction you need to run, Jian?  (On a conversation about it right now - there's some fundamental complexities in CMEPS that this is highlighting, so I would just like your assessment.)

sjsprecious:
> How critical is your CMEPS PR to the memory reduction you need to run, Jian? (On a conversation about it right now - there's some fundamental complexities in CMEPS that this is highlighting, so I would just like your assessment.)

Thanks @briandobbins . We have a very tight memory headroom on Derecho so I need all the changes in my CDEPS, CMEPS, CTSM, and future CAM PRs to get the ne1024 simulation to work with more ranks per node. Bill mentioned some concerns about CMEPS but I think that is unrelated to my changes directly.

My CMEPS PR seems to be BFB so if Bill or you can merge it, I can update the tag here as well.

briandobbins:
Unfortunately it's not that simple -- Bill, Bob and Mariana are speaking this afternoon about the CMEPS issues, and I'll know more after that, but I can't commit to the CMEPS PR yet.

sjsprecious:
> Unfortunately it's not that simple -- Bill, Bob and Mariana are speaking this afternoon about the CMEPS issues, and I'll know more after that, but I can't commit to the CMEPS PR yet.

I see. Just want to clarify again: my CTSM PR is independent of my CMEPS PR, so it should not affect the review or merge process here.

briandobbins:
Right, we're not worried about the CTSM one, just the CMEPS one -- and I'll
know more about that later.

- Brian

sjsprecious:
Hi @samsrabin and @ekluzek , just wants to give you a heads up: Both my CDEPS and CMEPS PRs are now merged. The CDEPS PR is **BFB** but the CMEPS PR is answer-changing at roundoff level.

Since my CTSM PR does not depend on the CMEPS PR, I prefer not to include the CMEPS changes in my PR here so that my CTSM changes remain **BFB** as shown in the [previous test](https://github.com/ESCOMP/CTSM/pull/4102#issuecomment-4860234861). Let me know if you would like to proceed with a different option.

ekluzek:
> Hi @samsrabin and @ekluzek , just wants to give you a heads up: Both my CDEPS and CMEPS PRs are now merged. The CDEPS PR is **BFB** but the CMEPS PR is answer-changing at roundoff level.
> 
> Since my CTSM PR does not depend on the CMEPS PR, I prefer not to include the CMEPS changes in my PR here so that my CTSM changes remain **BFB** as shown in the [previous test](https://github.com/ESCOMP/CTSM/pull/4102#issuecomment-4860234861). Let me know if you would like to proceed with a different option.

Thanks so much @sjsprecious this is great. Our preference is to seperate out any answer changes from things that are expected to be bit-for-bit. And actually it might be good to remove the CDEPS submodule update here to a seperate thing.

Before we bring this in we need to see how it merges with the mpi_scan branch that was merged onto b4b-dev. So it might be best to remove the CDEPS update here, and rebase this to the b4b-dev branch. See the instructions under "1" here: https://github.com/ESCOMP/CTSM/wiki/b4b%E2%80%90dev-branch-procedures#merging-to-b4b-dev

Since you've already done work here, and I'm asking for more -- I'll create a PR on master to bring the bit-for-bit CDEPS update in, as well as a different one to bring in the answer changing CMEPS one. How does that sound?

sjsprecious:
> Thanks so much @sjsprecious this is great. Our preference is to seperate out any answer changes from things that are expected to be bit-for-bit. And actually it might be good to remove the CDEPS submodule update here to a seperate thing.

Thanks @ekluzek . It sounds good to me.
 
> Before we bring this in we need to see how it merges with the mpi_scan branch that was merged onto b4b-dev. So it might be best to remove the CDEPS update here, and rebase this to the b4b-dev branch. See the instructions under "1" here: https://github.com/ESCOMP/CTSM/wiki/b4b%E2%80%90dev-branch-procedures#merging-to-b4b-dev

I will do it as suggested and let you know if I run into any problem later.

> Since you've already done work here, and I'm asking for more -- I'll create a PR on master to bring the bit-for-bit CDEPS update in, as well as a different one to bring in the answer changing CMEPS one. How does that sound?

That sounds good to me. Thanks for your help.

sjsprecious:
Hi @ekluzek , I have removed the CDEPS update in my PR and rebased my branch to the latest `b4b-dev` branch. Let me know if you have any questions about it.

src/utils/shr_mpishmem_mod.F90
ekluzek, line 1: From looking at this file, it looks like it's really a utility that builds on top of MPI rather than ESMF. As such it should be moved from src/cpl/share_esmf to src/utils.

Because it really extends MPI capability, I also wonder if it shouldn't have MPI in the name? So something like mpi_shmem.F90? Note, we've also been going away from the convention of having the "_mod" (for module) in the ending of files.

I also didn't see where there was a strict need for "clm" in the names? This seems fairly generic. Couldn't there be a version of this in CESM share code that is used by more than one model? Or did I miss something in here that's specific to CLM and CLM data structures?

src/utils/shr_mpishmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+  ! implement the MPI-2/MPI-3 one-sided / shared-memory interfaces, so for
+  ! mpi-serial builds (CPP macro NO_MPI2, set by CIME for MPILIB=mpi-serial) a
+  ! single-task fallback is compiled instead: each "node-shared" array is a plain
+  ! local allocation (one task is its own node and its own leader, so there is no
+  ! cross-rank sharing and the leader sum-reduce is a no-op).  The F90 'mpi' module
+  ! (not mpif.h) is used because the TYPE(C_PTR) overloads of MPI_WIN_ALLOCATE_SHARED
+  ! / MPI_WIN_SHARED_QUERY are only guaranteed there (MPI-3.0).
+  !-----------------------------------------------------------------------------
+
+  use mpi
```
ekluzek, line 48: I recall that having "use mpi" caused problems with at least one of the builds we support for CLM. So we might need to have this enclosed in a CPP ifdef. I have to figure out what build that was though and get back to you.

src/cpl/share_esmf/clm_shmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+  ! single-task fallback is compiled instead: each "node-shared" array is a plain
+  ! local allocation (one task is its own node and its own leader, so there is no
+  ! cross-rank sharing and the leader sum-reduce is a no-op).  The F90 'mpi' module
+  ! (not mpif.h) is used because the TYPE(C_PTR) overloads of MPI_WIN_ALLOCATE_SHARED
+  ! / MPI_WIN_SHARED_QUERY are only guaranteed there (MPI-3.0).
+  !-----------------------------------------------------------------------------
+
+  use mpi
+  use, intrinsic :: iso_c_binding, only : c_ptr, c_f_pointer
+  use spmdMod   , only : mpicom
```
ekluzek: Use of spmdMod and abortutils are two CLM specific things. They could be replaced by passing mpicom down into subroutines, and using shr_sys_abort rather than endrun.

src/utils/shr_mpishmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+  use, intrinsic :: iso_c_binding, only : c_ptr, c_f_pointer
+  use spmdMod   , only : mpicom
+  use abortutils, only : endrun
+
+  implicit none
+  private
+
```
ekluzek, line 54: This implements specifc versions for 4-byte integer 1D data. I think maybe that should be disguised within a generic interface like how it handles clm_shmem_free? This also seems like something that could be done using a *.in file to handle different data and vector dimensions and be appropriate for CESM-share.

src/utils/shr_mpishmem_mod.F90
ekluzek: It looks like pio.F90.in in share/unit_test_stubs/pio/pio.F90.in gets around this by having this pragma around it..

``` fortran
#ifndef NO_MPIMOD
   use mpi, only : MPI_OFFSET_KIND ! _EXTERNAL
#endif
```

so that could be used here.

src/utils/shr_mpishmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+  !
+  ! Usage (collective over the land communicator mpicom):
+  !   call clm_shmem_alloc_i4_1d(ptr, win, n)        ! all ranks
+  !   if (clm_shmem_is_leader()) ptr(:) = 0          ! leader owns the storage
+  !   call clm_shmem_fence(win)                       ! publish the zeros
+  !   <each rank stores into its disjoint indices of ptr>
+  !   call clm_shmem_leader_allreduce_sum_i4(ptr,win,n) ! fence; sum across nodes; fence
+  !   <all ranks may now read ptr for the rest of its lifetime>
+  !   call clm_shmem_free(ptr, win)                   ! collective over the node comm
+  !
```
ekluzek, line 24: Above and in the implementation there's use of some terms that could use definition:
- fence
- leader
- window
- color

src/utils/shr_mpishmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+  integer, save :: node_size   = 1
+  logical, save :: is_leader   = .true.
+
+contains
+
+  !=============================================================================
+  subroutine init_comms()
+    ! Lazily build the node-local and node-leader communicators.  Collective over
+    ! mpicom; safe to call from every shared-memory request.
+#ifndef NO_MPI2
```
ekluzek, line 97: This is true in other parts of the code as well, but the use of NO_MPI2 is confusing. So it would be good to have a comment to say what this is really trying to do. Which in this case is really a way to distinquish between mpi-serial and full MPI (which is this part of the branch).

src/utils/shr_mpishmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+    integer :: ierr, color
+#endif
+
+    if (initialized) return
+
+#ifndef NO_MPI2
+    call mpi_comm_split_type(mpicom, MPI_COMM_TYPE_SHARED, 0, MPI_INFO_NULL, &
+                             node_comm, ierr)
+    call mpi_comm_rank(node_comm, node_rank, ierr)
```
ekluzek, line 106: OK, so node_rank is the rank within a node. And node_size will be the number of ranks on a single node.

So for a case on Derecho with 2 processors, and without theading: node_size == 128, rank 0, and rank 128 will be a leader and hence color 0. So leader_comm will be a new communicator for rank 0 and rank 128.

ekluzek: I wonder about what will happen when the model is running hybrid OpenMP/MPI though? So if I'm still running on two nodes, but both have two threads. Then there will be 128 ranks total, and I want rank 0 and 64 to be leaders. So node_size would be 64 in this case.

Was this run with threading so we know this works in that case?

src/cpl/share_esmf/clm_shmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+    ! leader requests storage; peers map the leader's contiguous segment.
+    integer, pointer, intent(out) :: ptr(:)
+    integer,          intent(out) :: win
+    integer,          intent(in)  :: n
+
+#ifndef NO_MPI2
+    integer(kind=MPI_ADDRESS_KIND) :: winsize, qsize
+    integer :: ierr, disp_unit, qdisp
+    integer :: itmp
+    type(c_ptr) :: baseptr
```
ekluzek: Is this needed to be a "c_ptr" because of how mpi_win* routines handles it? It looks like that's true...

src/cpl/share_esmf/clm_shmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+#endif
+
+    initialized = .true.
+  end subroutine init_comms
+
+  !=============================================================================
+  subroutine clm_shmem_alloc_i4_1d(ptr, win, n)
+    ! Allocate a node-shared default-integer array of length n.  Only the node
+    ! leader requests storage; peers map the leader's contiguous segment.
+    integer, pointer, intent(out) :: ptr(:)
```
ekluzek: Add some comments for what each of these are.

src/cpl/share_esmf/clm_shmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+#ifndef NO_MPI2
+    integer, allocatable :: tmp(:)
+    integer :: ierr
+#endif
+
+    call clm_shmem_fence(win)             ! all node stores complete and visible to leader
+#ifndef NO_MPI2
+    if (is_leader) then
+       allocate(tmp(n))
+       call mpi_allreduce(ptr, tmp, n, MPI_INTEGER, MPI_SUM, leader_comm, ierr)
```
ekluzek: It looks like ptr could be checked to make sure it's size is n. This would probably be good to do here and maybe a few other places to make sure things are as expected.

src/cpl/share_esmf/clm_shmem_mod.F90
ekluzek: I ask, because avoiding pointers is a good goal. Using allocatables is a better practice and something we want to move toward. We use pointers too much in CLM, because of a reason that's no longer true, so moving them toward allocatable would be a good thing to do. See #235.

It looks like here lndmask_glob is a pointer, but should be able to become an allocatable.

src/cpl/share_esmf/clm_shmem_mod.F90
```diff
@@ -0,0 +1,228 @@
...
+                                 baseptr, win, ierr)
+    if (ierr /= MPI_SUCCESS) call endrun('clm_shmem_mod: MPI_Win_allocate_shared failed')
+
+    ! Non-leaders learn the address of the leader's (rank 0) contiguous segment.
+    if (.not. is_leader) then
+       call mpi_win_shared_query(win, 0, qsize, qdisp, baseptr, ierr)
+       if (ierr /= MPI_SUCCESS) call endrun('clm_shmem_mod: MPI_Win_shared_query failed')
+    end if
+
+    call c_f_pointer(baseptr, ptr, [n])
```
ekluzek: Here's where ptr as a Fortran pointer gets converted to the C pointer. Can this be done for an allocatable array as well? Or if it's allocatable can the whole pointer thing be avoided?

src/utils/shr_mpishmem_mod.F90
ekluzek: It also looks like this infrastructure is really a general use utility, although here it's just used for the calculation of the decomposition. But, this could be used for other global arrays to save memory. I'm not sure where the best savings would be for that, but it's something that occurs to me. That's another reason it would be good to make this a general use share utility code that could be called by all CESM models and multiple contexts.

src/utils/shr_mpishmem_mod.F90
ekluzek: I think this also means it might only work for only mpi-serial or MPI3. So we need to assess if there is an important build that's full MPI, but only MPI2. It looks like it could be possible to do it for MPI2, by just letting the arrays be non-shared. But, it would take work to figure out.

ekluzek (changes requested):
@sjsprecious thanks so much for your work here! This is a cool feature for me to look at. And this is something that will help for ultra high resolution cases.

I have a ton of questions and asks that I give. I think it's enough that it would be good to review together and talk about it. Let's setup a time to do that. I'll propose something in gcal, and let me know if it works for you.

Another part of this is that I'm thinking about some things that may be more long term rather than something you need to get in now. So there might be multiple steps for this. That would be another thing to figure out, "what would be the minimum possible in order to bring it in now?" and what could be done later.

src/utils/shr_mpishmem_mod.F90
sjsprecious: @ekluzek , thanks for your comments. You are correct here that this module is quite general and not specific to CTSM. I actually need to use the same module for CAM to reduce some memory usage there as well.

Maybe I can move this module to `cesm_share`? But my concern is that this may involve more PRs/reviews, and I do not want to block my CTSM PR here for this reason. If you have a suggestion or you can quickly make a new CESM_share tag for this module, I am happy to do that.

src/utils/shr_mpishmem_mod.F90
sjsprecious: Thanks for your comments. I ran the `aux_clm` test suite and everything seemed to work except some known fails. I incline not to include extra `#ifdef` CPP unless we have a test that fails without it.

You are correct that this module only works for mpi-serial or MPI-3, which is supported by the Cray-MPICH library on Derecho. It will be good to figure out the compatibility with MPI-2 but I am not sure whether it is worth the efforts to do it. MPI-3 is supported by OpenMPI v2.0 or newer, MPICH v3.0 or newer, so we really need a very old MPI library to test MPI-2 compliance here.

src/cpl/share_esmf/clm_shmem_mod.F90
sjsprecious: In CAM I also used `spmdMod` and `abortutils` in a similar module. Maybe let's decide where this module should be placed first before I make your suggested changes?

src/utils/shr_mpishmem_mod.F90
sjsprecious: Agreed. But the `NO_MPI2` CPP variable is enabled by `cime` at https://github.com/ESMCI/cime/blob/master/CIME/Tools/Makefile#L101. Changing it to be something more meaningful like `no_mpi` would require a change on the cime's side, as well as a whole change through the CESM for consistency. I think this is something good to be done in the future but not in this PR.

src/utils/shr_mpishmem_mod.F90
sjsprecious: Thanks for your comment. This is a good point.

Based on my tests, CAM7 physics has a runtime error with threading on. Thus all my ne1024 tests are done with MPI-exclusive configuration, though I did not see obviously why this implementation did not work with threading.

Does `aux_clm` test suite have any threading test? If so, then my results suggest that the current implementation works with threading. If not, is there a test in CTSM that I can run with threading?

src/cpl/share_esmf/clm_shmem_mod.F90
sjsprecious: I see your point here and thanks for sharing Bill's investigation.

For `mpi_win*` subroutines, `c_ptr` is needed based on the examples from OpenMPI manual: https://www.open-mpi.org/doc/v3.0/man3/MPI_Win_allocate_shared.3.php. Thus we should keep `lndmask_glob` as a pointer to be compatible with the MPI interface.

src/utils/shr_mpishmem_mod.F90
briandobbins: @ekluzek Given the need to get the StormSPEED runs going, can we add this as-is to CTSM, and _also_ open an issue in CTSM and CAM to compare implementations and arguably make this move to shared code happy at a later time?  All our code is an evolution; I don't mind doing changes like this in two steps versus one, especially when doing so enables leading-edge experiments.  

I'd also like to learn from the CAM folks how much this capability is used there.

src/utils/shr_mpishmem_mod.F90
briandobbins: We certainly don't need to worry about MPI2 at this point -- that's ten years old, and I don't know of any MPI libraries that aren't MPI3.0 compliant.  If this works fine for MPI-serial, too, I'm good with this.

(Separately, we discovered during a review of the init time fix that the '-DNO_MPI2' flag is an artifact of much older development, and likely meant _no MPI_.  We'll fix that at a future point, since it's confusing now.)

src/utils/shr_mpishmem_mod.F90
briandobbins: Ah, I just mentioned this above -- should've skimmed through the whole thing first.  Yes, this'll be done in the future, but I'm good with keeping the NO_MPI2 for now, for consistency, as that makes a future, CESM-wide change easier via search/replace/check.

ekluzek: To be clear, I'm not objecting to "NO_MPI2" in and of itself. I agree that's a longer term thing that we should do CESM-wide.

What I would like here is to add a comment when it's used to help make it clear.

src/utils/shr_mpishmem_mod.F90
ekluzek: Ahhh, good. aux_clm has several threading tests so if you ran it through there that would validate that threading is working correctly.

ekluzek:
@sjsprecious do you have access to our CGD machine Izumi? I tried running some tests there and many are failing at run time. We test with the NAG compiler on Izumi and often find that it will catch legitimate bugs not caught by other compilers.

Here's an example test that fails there:

SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.izumi_nag.clm-crop


Here's a part of the cesm.log that I get from that case.

```
 (t_initf)       profile_global_stats=     T
 (t_initf)       profile_ovhd_measurement= F
 (t_initf)       profile_add_detail=       F
 (t_initf)       profile_papi_enable=      F
Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassocRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
Program terminated by fatal error
/fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
Program terminated by fatal error
/fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: ErrRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
Program terminated by fatal error
/fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Error occRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
Program terminated by fatal error
/fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Error occurred in LND_SET_DECOMP_AND_DOMAIN:LND_SET_LNDMASK_FROM_MASKMESH
Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
```

ekluzek:
@sjsprecious I'd also like to add a ne1024 test to the uhr_decomp_init testlist, so it can demostrate this working. I also suggest you run that test list as well, because that will run with mpasa3p75 which would be a good stress test for this.

The problem I saw is that ccs_config doesn't have the ne1024 grid in it, and there isn't an issue or PR for doing it there as well. I'm assuming that's something that's planned to do? What's the story on that?

src/utils/shr_mpishmem_mod.F90
sjsprecious: Thanks for your clarification. I will add a comment as you suggested later.

sjsprecious:
> @sjsprecious do you have access to our CGD machine Izumi? I tried running some tests there and many are failing at run time. We test with the NAG compiler on Izumi and often find that it will catch legitimate bugs not caught by other compilers.
> 
> Here's an example test that fails there:
> 
> SMS_D.f10_f10_mg37.I2000Clm60BgcCrop.izumi_nag.clm-crop
> 
> Here's a part of the cesm.log that I get from that case.
> 
> ```
>  (t_initf)       profile_global_stats=     T
>  (t_initf)       profile_ovhd_measurement= F
>  (t_initf)       profile_add_detail=       F
>  (t_initf)       profile_papi_enable=      F
> Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassocRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
> Program terminated by fatal error
> /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
> Program terminated by fatal error
> /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: ErrRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
> Program terminated by fatal error
> /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Error occRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
> Program terminated by fatal error
> /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Error occurred in LND_SET_DECOMP_AND_DOMAIN:LND_SET_LNDMASK_FROM_MASKMESH
> Runtime Error: /fs/cgd/data0/erik/ctsm_worktree/b4b-dev/src/cpl/share_esmf/lnd_set_decomp_and_domain.F90, line 625: Reference to disassociated POINTER LNDMASK_GLOB
> ```

Thanks @ekluzek for reporting this error. I did not have access to Izumi but I will take a look at the error and see if I can fix it on my end.

sjsprecious:
> @sjsprecious I'd also like to add a ne1024 test to the uhr_decomp_init testlist, so it can demostrate this working. I also suggest you run that test list as well, because that will run with mpasa3p75 which would be a good stress test for this.
> 
> The problem I saw is that ccs_config doesn't have the ne1024 grid in it, and there isn't an issue or PR for doing it there as well. I'm assuming that's something that's planned to do? What's the story on that?

Correct, @jtruesdal is still working on some cleanup work and he should be able to commit a PR in `css_config` to add the grid definition and file for ne1024 later. But I think it is better to be done in a separate PR.

Shall I just replace `aux_clm` by `uhr_decomp_init` in the command below to run this testlist on Derecho?
```
conda run -n ctsm_pylib ./run_sys_tests -s aux_clm --compare ctsm5.4.044 --skip-generate
```

src/utils/shr_mpishmem_mod.F90
sjsprecious: Thanks @ekluzek for pointing it out. I have added some comments about these terms at https://github.com/ESCOMP/CTSM/pull/4102/changes/6e46c4f98a776ce4a1d72839287908bfea716898.

src/utils/shr_mpishmem_mod.F90
sjsprecious: @ekluzek , I just added the comment about the actual usage of "NO_MPI2" at https://github.com/ESCOMP/CTSM/pull/4102/changes/6e46c4f98a776ce4a1d72839287908bfea716898.

src/cpl/share_esmf/clm_shmem_mod.F90
sjsprecious: Thanks @ekluzek . I have added some comments as required at https://github.com/ESCOMP/CTSM/pull/4102/changes/6e46c4f98a776ce4a1d72839287908bfea716898.

src/cpl/share_esmf/clm_shmem_mod.F90
sjsprecious: Thanks @ekluzek for your comments. Good point and I have added the size check at https://github.com/ESCOMP/CTSM/pull/4102/changes/7572cf3721c5fdfd505565c355ab3c55b5183f1e.

sjsprecious:
@ekluzek Hmm, I guess the error you got on Izumi may be due to its MPI version. I pushed some changes at https://github.com/ESCOMP/CTSM/pull/4102/changes/f0dc92f664380843becf4115d9fd95ecaf79e3af and could you please let me know if it works on Izumi now?

ekluzek:
> @ekluzek Hmm, I guess the error you got on Izumi may be due to its MPI version. I pushed some changes at [f0dc92f](https://github.com/ESCOMP/CTSM/commit/f0dc92f664380843becf4115d9fd95ecaf79e3af) and could you please let me know if it works on Izumi now?

Hi @sjsprecious -- yes your changes get that case to work! Good job, I'm impressed that you could suggest a fix for a machine you didn't have access to. But, I'm rerunning the other tests that failed and we'll see if anything else comes up there.

ekluzek:
> > @sjsprecious I'd also like to add a ne1024 test to the uhr_decomp_init testlist, so it can demostrate this working. I also suggest you run that test list as well, because that will run with mpasa3p75 which would be a good stress test for this.
> > The problem I saw is that ccs_config doesn't have the ne1024 grid in it, and there isn't an issue or PR for doing it there as well. I'm assuming that's something that's planned to do? What's the story on that?
> 
> Correct, @jtruesdal is still working on some cleanup work and he should be able to commit a PR in `css_config` to add the grid definition and file for ne1024 later. But I think it is better to be done in a separate PR.

Yes, updates to submodules will be done on a tag on master.

> 
> Shall I just replace `aux_clm` by `uhr_decomp_init` in the command below to run this testlist on Derecho?
> 
> ```
> conda run -n ctsm_pylib ./run_sys_tests -s aux_clm --compare ctsm5.4.044 --skip-generate
> ```

Yep, exactly.

And you might as well run the decomp_init test suite as well. The two of those were designed to test the processor decomposition at initialization. If we run a baseline for ctsm5.4.047 for those test suites we should be able to see the difference in memory it makes for a selection of configurations. That could be a helpful confirmation that the code is working as we expect.

src/utils/shr_mpishmem_mod.F90
sjsprecious: I talked with @ekluzek today and we decided to keep this code in CTSM for now and move it to `cesm_share` in the future for long-term maintenance.

It is better to rename this file to `shr_mpishmem_mod.f90.in` and use `TYPE` or `DIMS` to template some functions later.

src/utils/shr_mpishmem_mod.F90
sjsprecious: Go through the `aux_clm` with @ekluzek and it seems this CPP variable is no longer needed.

src/cpl/share_esmf/clm_shmem_mod.F90
sjsprecious: @ekluzek suggested:
- using `shr_abort_abort` in the `share`, `shr_abort_mod.F90` file
- passing `mpicom` as an argument to the new interface in this file

src/utils/shr_mpishmem_mod.F90
sjsprecious: Talked to @ekluzek and we will do this when we merging the files here and CAM's one in `cesm_share`.

sjsprecious:
@ekluzek the output of the `uhr_decomp_init` test suite on Derecho is available at `/glade/derecho/scratch/sunjian/tests_0722-104618de`. Let me know what you find out later.

src/utils/shr_mpishmem_mod.F90
ekluzek: Yes, @sjsprecious will open an issue in cesm_share about adding this as a generic utility there. And that will be done as a future refinement step.

src/utils/shr_mpishmem_mod.F90
ekluzek: Yes, @sjsprecious seems to have the Izumi cases working, so it looks like we don't need to do anything different here, which is good. I'm going to rerun the rest of the Izumi tests and we'll see if anything else comes up, but looking fine so far. We also made sure that the Derecho nvhpc tests were working and they did. So looking all good so far.

src/cpl/share_esmf/clm_shmem_mod.F90
ekluzek: Yes, and we should:
- move to the src/utils directory
- Remove clm_ from the name (maybe call it shr_mpishmem_mod.F90 as it will be called when moved to share?)
- Remove clm_ from the internal module data and procedures

src/utils/shr_mpishmem_mod.F90
ekluzek: There's 54 threading tests that @sjsprecious ran with aux_clm and they all pass. 49 of them are ERP tests where the pe-layout is halved on restart and those worked and gave identical results. So threading is clearly fine here.

src/cpl/share_esmf/clm_shmem_mod.F90
sjsprecious: Thanks @ekluzek . The changes are made at https://github.com/ESCOMP/CTSM/pull/4102/commits/436e2b7a4afb4ede2b2a80354d74c7e2e5ed80a0 and https://github.com/ESCOMP/CTSM/pull/4102/commits/8716b48f3b18b934aa6a76b64146270210f43d6f. Let me know if they look good to you.

sjsprecious:
@ekluzek the output of the `decomp_init` test suite on Derecho is available at `/glade/derecho/scratch/sunjian/tests_0722-160125de`. I used the https://github.com/ESCOMP/CTSM/pull/4102/changes/8716b48f3b18b934aa6a76b64146270210f43d6f commit and make sure your suggested changes still build and run successfully.

src/cpl/share_esmf/clm_shmem_mod.F90
ekluzek: Looks good.

ekluzek (approved):
All my suggestions have been addressed. Or they have been put into an issue to do later.

sjsprecious:
@ekluzek could you please help merge this PR if everything looks good to you?

ekluzek:
The one thing I wanted to do here was to compare the decomp_init simulations for memory to the baselines I ran for them. It should show the memory improvement.

The other thing to do is to rerun aux_clm testing to make sure everything is good there. I didn't complete the izumi testing, so might as well redo that.

We want this in place and in a tag by next week when there is a meeting that will discuss it.

ekluzek:
@sjsprecious could you give me collaborator access to your CTSM fork?  I may not need it, but that way I could push something to the branch if needed.

sjsprecious:
> @sjsprecious could you give me collaborator access to your CTSM fork? I may not need it, but that way I could push something to the branch if needed.

Sure @ekluzek , I just sent you an invite.

ekluzek:
I reran aux_clm on Derecho and everything was as expected. MEMCOMP shows a consistent drop in memory by a small amount for the tests from 0  to 4% (with the caveat that one test showed a small increase in memory by 2%). I tried seeing the memory difference for the decomp_init testlist, but because the tests are short it doesn't show a memory comparison.

I'm rerunning aux_clm on Izumi now.

ekluzek:
@sjsprecious did you have a way to show the memory drop in the change? I'm wondeirng if we should add some additional memory checking to show what's going on? I thought the standard tests would help with this, but it doesn't look like they really do.

sjsprecious:
> @sjsprecious did you have a way to show the memory drop in the change? I'm wondeirng if we should add some additional memory checking to show what's going on? I thought the standard tests would help with this, but it doesn't look like they really do.

Hi @ekluzek , I did a 480-node run on Derecho for the FHISTC_LTso compset at ne1024pg2 resolution, and used the `qhist` command to extract the memory usage for a job run on Derecho. Below is the plot and you can see given the same number of MPI tasks, the BASE version (without any optimization) uses about 5% to 10% more memory than the OPT version (including all the memory optimizations in CAM/CTSM/CMEPS/CDEPS). Can you run the `qhist` command on Derecho for your runs and see if they show a memory difference?

image

ekluzek:
OK, I'm going to do a little more work looking at memory, I sent some tests to do that.

But, the Izumi and Derecho aux_clm tests passed as expected so I'm merging.

sjsprecious:
> OK, I'm going to do a little more work looking at memory, I sent some tests to do that.
> 
> But, the Izumi and Derecho aux_clm tests passed as expected so I'm merging.

Thanks @ekluzek for merging it. Since it is merged to `b4b-dev` branch, will I get a new CTSM tag or I need to wait until you merge `b4b-dev` to the `main` branch?

ekluzek:
> > OK, I'm going to do a little more work looking at memory, I sent some tests to do that.
> > But, the Izumi and Derecho aux_clm tests passed as expected so I'm merging.
> 
> Thanks @ekluzek for merging it. Since it is merged to `b4b-dev` branch, will I get a new CTSM tag or I need to wait until you merge `b4b-dev` to the `main` branch?

It won't have a tag until merged to the master branch. I plan to do that early next week.

PR #4113, opened 2026-07-09
Improve restart history-option error messages (#3593)

### LINKED ISSUE #3593: Improve "number of ntapes differs from restart file" error msg
If you change the number of history files in the middle of a run (`CONTINUE_RUN=TRUE`), you get this error:
```
 ENDRUN: ERROR in histFileMod.F90 at line 4886
  ERROR: number of ntapes differs from restart file. You can NOT change history
 options on restart.
```

This would be greatly improved by telling the user that they can avoid this by doing a branch run.

There may be similar changed-history-files errors that should also get this improvement.

enu-emu:
Fixes #3593.

The error raised when history options change across a restart told the user what was forbidden ("You can NOT change history options on restart") but not what to do about it. This adds guidance to use a branch run instead, and clarifies that the restriction applies to a continuation run.

While tracing the code I noticed the same phrasing appears in **three** sibling `endrun` calls in the restart-read block of `histFileMod.F90` — the `ntapes`, `history_tape_in_use`, and number-of-fields checks — so I updated all three for consistency. Happy to narrow it to just the `ntapes` message if you'd prefer.

This is a message-text-only change (no logic touched), so it should be bit-for-bit identical. One note: I wasn't able to build/run CTSM locally (M1 Mac, limited RAM/disk), so I verified this by code inspection and would appreciate it being checked via your regression testing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ekluzek (approved):
I think this is an improvement in the documentation. So we should bring it in.

I'm rebasing to b4b-dev, and we need to run standard testing.

wwieder:
@slevis-lmwg will run aux_clm with no-run and bring this to b4b-dev

- [x] derecho: `./run_sys_tests -s aux_clm -c ctsm5.4.052 --skip-generate --extra-create-test-args " --no-run"`
`FAIL MKSURFDATAESMF_P128x1.f10_f10_mg37.I1850Clm50BgcCrop.derecho_intel SHAREDLIB_BUILD`
```
fatal: No names found, cannot describe anything.
fatal: not a git repository (or any parent up to mount point /glade/derecho)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
I will ignore the failure, merge to b4b-dev, and run the test again from a b4b-dev checkout for my sanity.
- [x] izumi: `./run_sys_tests -s aux_clm -c ctsm5.4.052 --skip-generate` (skipping the --no-run here as sanity check)

slevis-lmwg:
> I will ignore the failure, merge to b4b-dev, and run the test again from a b4b-dev checkout for my sanity

PASS `/glade/derecho/scratch/slevis/MKSURFDATAESMF_P128x1.f10_f10_mg37.I1850Clm50BgcCrop.derecho_intel.C.20260828_165047_mxysoj`

Issue #4115, opened 2026-07-09
Deprecate and remove clm4_5 physics


samsrabin:
From the [SE meeting today](https://docs.google.com/document/d/1euq06v1pQOoN6NcGwdUqJsfgDNPqXzWB2cdMOJcydI8/edit?tab=t.0#heading=h.nq7hvqdxmrw): To ease our maintenance burden, before the release of CTSM6.0, we would like to remove `clm4_5` physics.

Definition of done:
- [x] Add as deprecated to IMPORTANT_NOTE, as well as the features that it uses
- [ ] Make notes elsewhere as needed about it being deprecated
- [x] Remove from tests
- [ ] Remove from compsets
- [ ] Remove from RTM testlist (https://github.com/ESCOMP/RTM/issues/68)
- [ ] Remove features from build-namelist and namelist-defaults
- [ ] Remove the code that uses those options
- [ ] Remove it from documentation
- [ ] Make sure it isn't mentioned anywhere

billsacks:
I was thinking about how this might help the upcoming water isotope/tracer work - if there are options we can avoid supporting when we add water tracers.

Can we remove support for soilwater_movement_method = 0 (the clm4_5 default - I think this is zengdecker2009)?

And if we do that, can we remove support for lower_boundary_condition = 4 (bc_aquifer)? If we further remove support for vichydro (which may or may not be on the table), could we then remove support for lower_boundary_condition = 3 (bc_waterTable)? I believe that removing support for those two lower boundary conditions could let us remove support for use_aquifer_layer = .true. (see the `use_aquifer_layer` function in SoilWaterMovement.F90). That, in turn, would resolve #596.

On the flip side, I was hoping to rely on the existence of the clm4.5 finundation_method (ZWT_inversion) as a workaround for #658 in doing incremental testing for the water tracer work.

To be clear: I'm not pushing hard myself for the removal of these various options: if they are still useful to people, then we should keep them... and if needed, we can just disallow water tracers with certain hydrology options. I'm still holding out some hope that CTSM can replace NoahMP for various applications eventually, and to do that, we might need to maintain more hydrology options??

@swensosc - keeping you in the loop here.

wwieder:
I'm happy to deprecate 4.5 physics options especially as we near the 6.0 release.

ekluzek:
[...]

PR #4146, opened 2026-07-31
Fix restart failures for 35 diagnostic fields (#3661, #4142)

### LINKED ISSUE #3661: FAIL ERP_P64x2_D_Ld3...clm-default--clm-all_outputs COMPARE_base_rest

### LINKED ISSUE #4142: Get all_outputs tests working cleanly

johnpaulalex:
### Description of changes
This is intended as a followup to https://github.com/ESCOMP/CTSM/pull/4145.

This PR fixes exact restart failures for 35 diagnostic fields originally identified in Issue #3661 (including `LEAFCN_STORAGE`, numerous `GAMMA` fields, `RRESIS`, `USTAR`, etc.). The root cause was that these fields were placed on the history tape as "inactive" by default, but were missing the appropriate read/write logic for the restart files. This omission led to mismatches between `h0a` and `h0i` history output upon an exact restart.

To resolve this, explicit `restartvar()` calls and `Restart` subroutine hooks have been added for these diagnostic variables across several modules:
- `NutrientCompetitionFlexibleCNMod` and `NutrientCompetitionMethodMod`
- `VOCEmissionMod`
- `EnergyFluxType`
- `FrictionVelocityMod`
- `SoilStateType`
- `WaterDiagnosticBulkType`
- `clm_instMod`

Additionally, because these fields now restart perfectly, the `all_outputs` integration test override for Issue `#3661` in `ExpectedTestFails.xml` has been removed, and the `hist_fexcl1` exclusion list is no longer needed in `testmods_dirs/clm/all_outputs/user_nl_clm`.

### Specific notes

- Gemini LLM authored this fix and this description.

- Resolves #3661
- Resolves #4142

None.

- [ ] TODO: need to verify exact restart test (`ERP_P64x2_D_Ld3.f10_f10_mt232.IHistClm60BgcCropCrujra.derecho_gnu.clm-default--clm-all_outputs`) which previously failed under #3661.

- [ ] The code in this PR branch builds with no errors.
- [ ] The code in this PR branch runs with no errors. 
- [ ] 
- [ ] **Briefly describe tested configuration(s):**  Exact restart tests (`all_outputs`).
- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Which?:** (a) does not change answers for active prognostic fields (only fixes restart states for inactive diagnostic fields).
- [x] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates. **Which?:** (a) does not create a need to update the documentation.

### DIFF
--- a/src/biogeochem/NutrientCompetitionCLM45defaultMod.F90
+++ b/src/biogeochem/NutrientCompetitionCLM45defaultMod.F90
@@ -35,6 +35,7 @@ module NutrientCompetitionCLM45defaultMod
    contains
      ! public methocs
      procedure, public :: init                                ! Initialize the class
+     procedure, public :: restart                             ! restart
      procedure, public :: calc_plant_nutrient_competition     ! calculate nutrient yield rate from competition
      procedure, public :: calc_plant_nutrient_demand          ! calculate plant nutrient demand
      !
@@ -74,6 +75,20 @@ contains
     type(bounds_type), intent(in) :: bounds
 
   end subroutine Init
+
+  !------------------------------------------------------------------------
+  subroutine Restart(this, bounds, ncid, flag)
+    !
+    ! !DESCRIPTION:
+    ! Restart for the class (currently empty for this version)
+    !
+    use ncdio_pio              , only : file_desc_t
+    class(nutrient_competition_clm45default_type) :: this
+    type(bounds_type), intent(in)    :: bounds
+    type(file_desc_t), intent(inout) :: ncid
+    character(len=*) , intent(in)    :: flag
+
+  end subroutine Restart
 
   !-----------------------------------------------------------------------
   subroutine calc_plant_nutrient_competition (this,                   &
--- a/src/biogeochem/VOCEmissionMod.F90
+++ b/src/biogeochem/VOCEmissionMod.F90
@@ -62,6 +62,7 @@ module VOCEmissionMod
      real(r8) , pointer, private :: efisop_grc        (:,:) ! gridcell isoprene emission factors
    contains
      procedure, public  :: Init
+     procedure, public  :: Restart
      procedure, private :: InitAllocate
      procedure, private :: InitHistory
      procedure, private :: InitCold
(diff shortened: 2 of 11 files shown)


johnpaulalex:
SG, I need to add more filters to my issue-chooser.  :)

That said, https://github.com/ESCOMP/CTSM/pull/4146 seems like it still might be valid, deeper fix? Maybe?

ekluzek:
@johnpaulalex cool this is a nice thing to get working.

I think this is something that we'll want to look a bit closer at. In many cases adding the field to the restart file is the right thing to do. But, in some cases it shouldn't HAVE to be there, so it would be good to see why it's needed and figure it out if it could be changed. Disk space is becoming more of a premium so avoiding writing to the restart file is good if it can be avoided. But, the analysis of figuring out why it has to be added can be tricky as it depends on the order that things are done both at initialization and runtime.

I'm adding next though so we can discuss as a group. It might be best to just take this in, after you show it's working. I also think that this should probably be moved to the b4b-dev branch. We'll discuss that in the group as well.

src/biogeochem/NutrientCompetitionCLM45defaultMod.F90
```diff
@@ -75,6 +76,20 @@ subroutine Init(this, bounds)
 
   end subroutine Init
 
+  !------------------------------------------------------------------------
+  subroutine Restart(this, bounds, ncid, flag)
+    !
```
ekluzek, line 81: I'm not sure we should add empty restart methods. This is a standard part of the pattern for these science modules, so I see that perspective. Adding it now, makes it easier for when it is needed later...

ekluzek: Oh, and actually this is the CLM45 one which is up for removal, so we shouldn't bring this in. See #4115

ekluzek: Oh, although I see what is going on, this was added because the other NutrientCompetition interface now requires it. So it's actually needed here.

src/biogeochem/VOCEmissionMod.F90
```diff
@@ -362,8 +363,92 @@ subroutine InitHistory(this, bounds)
...
+
+    call restartvar(ncid=ncid, flag=flag, varname='GAMMAC', xtype=ncd_double,  & 
+         dim1name='pft', long_name='gamma C for VOC calc', units='non', &
+         interpinic_flag='interp', readvar=readvar, data=this%gammaC_out_patch)
+
+  end subroutine Restart
+
   !-----------------------------------------------------------------------
-  subroutine InitCold(this, bounds)
+  subroutine InitCold(this, bounds, soilstate_inst)
```
ekluzek, line 451: This change must be a mistake. soilstate_inst is being passed in, but there it isn't declared as an argument, and not added to the call for InitCold, and it doesn't seem to have a reason for being added into the subroutine.

ekluzek (changes requested):
I added some comments at the top about a change that I didn't think was needed at first. But, it turns out it is, as it's part of a class that needs it in one of the other instances so needed in all versions even though it's empty.

There was also a bit that seemed to be a mistake, that should be removed. It would be found in testing as well. It also might have been the start of more work that wasn't completed. But, just putting it there as I saw it.

johnpaulalex:
Hey Erik, well that sounds like a challenge so I fed your reply into AI and it analyzed all 35 variables, walking up the call stack by itself, and concluded none of them need to be stored (!), they all just fail to initialize a subset of their array space for certain cases. Let me whip up a PR for that and we can discuss.

slevis-lmwg:
> [...] concluded none of them need to be stored (!), they all just fail to initialize a subset of their array space for certain cases. Let me whip up a PR for that and we can discuss.

Confirming my understanding:
@johnpaulalex does your statement mean that #4172 supersedes this PR? If so, feel free to close this one as "will not fix."

johnpaulalex:
yes and will do. Thanks!

PR #4155, opened 2026-08-06
ctsm5.4.050: Remove FatesColdSatPhen testmod

### LINKED ISSUE #4150: FatesColdSatPhen testmod doesn't actually set SP mode
### Brief summary of bug

The `FatesColdSatPhen` testmod, misleadingly, doesn't actually set anything having to do with SP mode. Why does it exist? Or at least, why is it named that?

### General bug information

**CTSM version you are using:** ctsm5.4.049

### Details of bug

All it does is include `FatesColdBasic` (which also doesn't set SP mode) and then add these lines to `user_nl_clm`:
https://github.com/ESCOMP/CTSM/blob/a96fb3f836afe2284a618c740a9fa3977a558313/cime_config/testdefs/testmods_dirs/clm/FatesColdSatPhen/user_nl_clm#L2-L3

All `fates` suite tests with `FatesColdSatPhen*` testmods use a `*FatesSp*` compset, which is what actually enables SP mode.

samsrabin:
### Description of changes

Removes the FatesColdSatPhen testmod, which didn't actually set anything having to do with SP mode (see #4150). Tests that included FatesColdSatPhen now include FatesColdDailyHist instead. The latter has replaced FatesColdBasic, which was previously included by FatesColdSatPhen and has now been updated to include the history file setting changes from FatesColdSatPhen's `user_nl_clm`.

All remaining `Fates*SatPhen*` testmods will now fail if a non-FatesSp compset is chosen. Why not just set use_fates_sp in user_nl_clm? Because there are (or may be) *other* things set in a FatesSp compset, so if this test doesn't use one, it's not representative.

- [x] **Remove temporary remove-fatescoldsatphen-testmod test suite**

### Specific notes

- Resolves #4150

**If answers are expected to change, describe (delete this line otherwise):** Answers are not expected to change in any meaningful sense. However, there will be diffs in the former
```
SMS_Lm1.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdBasic
```
test, which is now
```
SMS_Lm1.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdDailyHist
```
and is for the first time getting the history file setting changes that were in FatesColdSatPhen's `user_nl_clm`.

- [x] Generate new baselines for the tests in the temporary remove-fatescoldsatphen-testmod test suite
- [x] `aux_clm` suite: Check namelist match against ctsm5.4.049. If OK, softlink ctsm5.4.049 baselines into ctsm5.4.050.
- [x] `fates` suite: Check namelist match against ctsm5.4.049. If OK, softlink ctsm5.4.049 baselines into ctsm5.4.050.

- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Which?:** (a)

- [x] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates (see [guidelines for contributing documentation](https://escomp.github.io/CTSM/users_guide/working-with-documentation/docs-intro.html#contribution-guidelines)). **Which?:** (b)


cime_config/testdefs/testmods_dirs/clm/FatesColdDryDepSatPhen/user_nl_clm
```diff
@@ -1,4 +1 @@
-
-hist_mfilt        = 1
-hist_nhtfrq       = -24
-
+use_fates_sp = .true.
```
ekluzek, line 1: I really think that we shouldn't set this in the testmod, that we should use the FatesSp compset. I think there are other things besides use_fates_sp that are set there now, and even if there aren't they may well be more in the future.

cime_config/testdefs/testmods_dirs/clm/FatesColdMeganSatPhen/user_nl_clm
```diff
@@ -1,4 +1 @@
-
-hist_mfilt        = 1
-hist_nhtfrq       = -24
-
+use_fates_sp = .true.
```
ekluzek, line 1: Same as above add a comment that talks about this and that in general this should be used with a FatesSP compset.

cime_config/testdefs/testmods_dirs/clm/FatesColdSatPhen_prescribed/user_nl_clm
```diff
@@ -1,3 +1,4 @@
+use_fates_sp = .true.
```
ekluzek, line 1: Same as above add a comment about this.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -2539,10 +2539,11 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
-  <test name="SMS" grid="f45_f45_mg37" compset="I2000Clm60FatesSpRsGs" testmods="clm/FatesColdSatPhen">
+  <test name="SMS" grid="f45_f45_mg37" compset="I2000Clm60FatesSpRsGs" testmods="clm/FatesColdDailyHist">
     <machines>
       <machine name="derecho" compiler="nvhpc" category="aux_clm"/>
       <machine name="derecho" compiler="nvhpc" category="fates"/>
+      <machine name="derecho" compiler="nvhpc" category="remove-fatescoldsatphen-testmod"/>
```
ekluzek: From your notes you'll remove this new temporary testlist when you are done, which is good.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -2539,10 +2539,11 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
-  <test name="SMS" grid="f45_f45_mg37" compset="I2000Clm60FatesSpRsGs" testmods="clm/FatesColdSatPhen">
```
ekluzek, line 2542: I do really like this more descriptive name.

cime_config/testdefs/testmods_dirs/clm/FatesColdDryDepSatPhen/user_nl_clm
ekluzek: Now, your notes give this:

> "The justification: If you're only testing that standard FATES SP mode works, use a FatesSp compset so that you also test the compset. That's why we deleted the FatesColdSatPhen testmod—to ensure the FatesSp compsets are tested. However, these other testmods exercise special SP-mode settings. They should still exist, and they should all set FATES SP mode as a backup in case a non-FatesSp compset is chosen."

OK, I can see that.

But, add a comment that says that in the user_nl_clm. I think the tests for these modes should use the FatesSp compset, but as they are generally testing something slightly outside of that, it's probably OK to have that here.

cime_config/testdefs/testlist_clm.xml
```diff
@@ -4164,35 +4169,39 @@
   <test name="SMS_D_Ld3" grid="f09_g17" compset="I2000Clm60FatesSpCruRsGs" testmods="clm/FatesColdSatPhen_prescribed">
     <machines>
       <machine name="derecho" compiler="gnu" category="fates"/>
+      <machine name="derecho" compiler="gnu" category="remove-fatescoldsatphen-testmod"/>
     </machines>
     <options>
       <option name="wallclock">00:30:00</option>
       <option name="comment">Test with prescribed LAI and soil moisture with FatesSP mode</option>
```
ekluzek, line 4170: You could add something here that points out that this should use a FatesSp compset.

ekluzek:
```suggestion
      <option name="comment">Test with prescribed LAI and soil moisture with FatesSP mode (be sure to use a FatesSp compset along with the FatesColdSatPhen_prescribed testmod)</option>
```

ekluzek: This might also go other places as well...

ekluzek (approved):
OK this is good. I don't really like the testmods that explicitly set use_fates_sp==TRUE. But, I can relent and accept that. However, do put comments in the user_nl_clm files that give the justification and that in general you should use them in tests with FatesSp compsets. And then I can accept that. This could also be pointed out in the testlist, and I have one suggestion at a change to the comment in the testlist to say that. I think that's helpful, but not as important as a comment in the testmod.

cime_config/testdefs/testlist_clm.xml
samsrabin: Why, though? Is the thinking that these testmods won't work without a `FatesSp` compset?

ekluzek: OK, good question. This is really a testing philosphy type thing for me. Which might need a longer conversation. 

But, as one point, I think our testing should be small deviations from standard ways that the model is run. And we should test and retest those standards ways as often as possible. So if you need FatesSp for a test you do it with the FatesSp compset, so that you are testing the standard compset alongside the test that's a deviation from the standard compset.

The other part is to ensure that as things change things don't work in the test -- but would've broken if the FatesSp compset was used. With this the two can evolve seperately, and that could cause problems as those change.

Now, an alturnative way to do this would be to add this to the include_user_mods:

../../../usermods_dirs/clm/fates_sp

as that will ensure they are aligned as the FatesSp compsets are done using user-mods right now. That may change in the future, but adding it explicitly would make it more visible.

Is that sufficient? Or would you like to do a quick chat on this? A longer term discussion on testing philosphy would also be good to do with the group, but that can be later.

samsrabin: Thanks, Erik—your explanation for wanting to always run these tests with a FatesSp compset makes a lot of sense, so I will make a change. (No need for a larger group discussion, as far as I'm concerned.) But since the general motivation behind this PR is "make it hard to do the wrong thing," just adding a note like "make sure to..." doesn't feel sufficient.

The way I see it, there are two options:
1. Your idea of including `../../../usermods_dirs/clm/fates_sp`.
2. Adding a `shell_commands` line that fails the run if it doesn't use a FatesSp compset.

Which would you prefer?

ekluzek: Ahh. You are right it's better to make have a way to ensure it's followed over just a reminder to do it.

I would be good with either. But I'd say I prefer the later with it aborting if you do a mismatch.

samsrabin: Done with 5a5c2b327.

ekluzek: Perfect

PR #4171, opened 2026-08-12
test(ci): Re-enable FatesColdAllVars exact restart tests

johnpaulalex:
### Description of changes
This change removes the `FatesColdAllVars` exact restart (ERP) test entries from `cime_config/testdefs/ExpectedTestFails.xml`, re-enabling automated testing for FATES exact restarts.

**Dependencies:**
- Depends on FATES PR [NGEET/fates#1596](https://github.com/NGEET/fates/pull/1596).
- Depends on CTSM PR [ESCOMP/CTSM#4172](https://github.com/ESCOMP/CTSM/pull/4172).

**Linked issues addressed, if any:**
- Resolves [ESCOMP/CTSM#3660](https://github.com/ESCOMP/CTSM/issues/3660) (Re-enables `FatesColdAllVars` exact restart tests in CI)

**Description of generative AI usage:**
- Google Antigravity was used to write the code and tests, followed by human-guided verification.

### Answer Changes & Scientific Impact
- [x] **Bit-for-Bit (B4B)** with baseline master
- [ ] **Roundoff-level differences only**
- [ ] **Expected Answer Changes (ECA)**

### User Interface & Namelist Changes
- **Namelist / Defaults modified?** No
- **XML / Build script changes?** No

### Testing planned or performed, if any:
- [x] Executed pFUnit / CIME regression tests

**CTSM / CESM baseline hash-tag:** 334ba96fa4679a1bb29eca01c9702fb9ed1830f7
**PR branch hash-tag:** 173f76bf70f08ce8d1a59e1c42535b492ff84f35

### Requirements before merge:
- [x] I have followed the [CTSM contribution guidelines](https://github.com/ESCOMP/CTSM/blob/master/CONTRIBUTING.md).
- [x] The code in this PR branch builds with no errors.
- [x] The code in this PR branch runs with no errors. **Briefly describe tested configuration(s):** macOS gfortran pFUnit / CIME ERP regression tests
- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Which?:** (a) Bit-for-bit (B4B) with baseline master
- [x] I have reviewed relevant parts of the CLM documentation [Tech Note](https://escomp.github.io/CTSM/tech_note/index.html) or [User's Guide](https://escomp.github.io/CTSM/users_guide/index.html) to determine if anything needs to be changed or added.
[...]

Issue #4124, opened 2026-07-16
mizuRoute tests fail namelist comparison


samsrabin:
### Brief summary of bug

The following `aux_clm` tests fail `mizuRoute.control` namelist comparison for no apparent reason:
```
SMS_D_Ld5.5x5_amazon_rHDMA.I2000Clm60SpMizGs.derecho_gnu.mizuroute-default
SMS_D_Ld5.5x5_amazon_rHDMA.I2000Clm60SpMizGs.derecho_intel.mizuroute-default
```

### General bug information

**CTSM version you are using:** ctsm5.4.047

**Does this bug cause significantly incorrect results in the model's science?** No

### Important output or errors that show the problem

See `/glade/derecho/scratch/samrabin/tests_0709-162611de/SMS_D_Ld5.5x5_amazon_rHDMA.I2000Clm60SpMizGs.derecho_gnu.mizuroute-default.GC.0709-162611de_gnu/`.

olyson:
See my comments in https://github.com/ESCOMP/CTSM/pull/4111.  I had the same problem with mizuroute tests on Izumi.  It seemed to be due to the fact that the case name is embedded in mizuRoute.control.
I'm not sure why I didn't get similar failures in my Derecho testing as you did.

samsrabin:
@ekluzek will link the relevant mizuRoute and CIME issues here.

slevis-lmwg:
I was looking at ExpectedTestFails.xml and came across the cime issue, so I'm linking it:
ESMCI/cime#4966

ekluzek:
Issues in mizuRoute are:

https://github.com/ESCOMP/mizuRoute/issues/395

Issue #3660, opened 2025-12-12
FAIL ERP_Ld9...clm-FatesColdAllVars COMPARE_base_rest


slevis-lmwg:
`FAIL ERP_Ld9.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdAllVars COMPARE_base_rest`
FATES_C13DISC_SZPF field appears different in h0a and h0i output suggesting a diagnostic field that was possibly not tested for perfect restarts before.

_Originally posted by @slevis-lmwg in https://github.com/ESCOMP/CTSM/issues/2061#issuecomment-3619052227_
            
This is a FATES test. The change that revealed the problem came in with the commit e53049f8655f62e8152b295418e908c653da2ad2 in PR #2061, which put default inactive fields in history.

ekluzek:
The FATES test only has one field that fails:

```
tail -n20 /glade/derecho/scratch/erik/tests_ctsm546acl/ERP_Ld9.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdAllVars.GC.ctsm546acl_int/run/ERP_Ld9.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdAllVars.GC.ctsm546acl_int.clm2.h0i.2000-01-10-00000.nc.base.cprnc.out

                2932   0.100000000000000E+01          0.358681290328578E+00
                2932   0.100000000000000E+01          0.358681290328578E+00
                6624  (     4,    41,     1,     1) (    60,    10,     1,     1)
          avg abs field values:    7.777429871382585E-01
                                   7.777429871382585E-01
************************************************************************************************************************************
  
SUMMARY of cprnc:
 A total number of   1188 fields were compared
          of which      1 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
               and      0 had different data types
 A total number of      2 fields could not be analyzed
 A total number of      0 time-varying fields on file 1 were not found on file 2.
 A total number of      0 time-constant fields on file 1 were not found on file 2.
 A total number of      0 time-varying fields on file 2 were not found on file 1.
 A total number of      0 time-constant fields on file 2 were not found on file 1.
  diff_test: the two files seem to be DIFFERENT 

[...]

PR #4172, opened 2026-08-12
fix(biogeophys): Correct diagnostic array initialization for exact restarts

### LINKED ISSUE #3661: FAIL ERP_P64x2_D_Ld3...clm-default--clm-all_outputs COMPARE_base_rest

johnpaulalex:
### Description of changes
This change fixes exact restart (ERP) failures when `hist_all_fields = .true.` by ensuring diagnostic variables across `BiogeophysPreFluxCalcsMod`, `CanopyFluxesMod`, `FrictionVelocityMod`, `SoilHydrologyMod`, `SoilMoistStressMod`, `SoilStateType`, and `SoilTemperatureMod` are properly initialized to `spval` each timestep on inactive layers or patches. Without explicit per-timestep initialization, uninitialized array space on inactive points retained memory across timestep boundaries, causing state discrepancies when comparing base vs. restart history output.

**Linked issues addressed, if any:**
- Resolves [ESCOMP/CTSM#4124](https://github.com/ESCOMP/CTSM/issues/4124) (Fixes CTSM diagnostic array cold-start initialization when `hist_all_fields = .true.`)
- Resolves [ESCOMP/CTSM#3661](https://github.com/ESCOMP/CTSM/issues/3661) (Fixes CTSM diagnostic array resetting on deactivated patches during transient landuse)
- Pre-requisite for [ESCOMP/CTSM#3660](https://github.com/ESCOMP/CTSM/issues/3660) (Fixes CTSM diagnostic array portion of `FatesColdAllVars` restart failure)

**Dependencies / Blockers:**
- Blocked by [NGEET/fates#1596](https://github.com/NGEET/fates/pull/1596)
- Blocks [ESCOMP/CTSM#4171](https://github.com/ESCOMP/CTSM/pull/4171)

**Description of generative AI usage:**
- Google Antigravity was used to write the code and tests, followed by human-guided verification.

### Answer Changes & Scientific Impact
- [x] **Bit-for-Bit (B4B)** with baseline master
- [ ] **Roundoff-level differences only**
- [ ] **Expected Answer Changes (ECA)**

### User Interface & Namelist Changes
- **Namelist / Defaults modified?** No
- **XML / Build script changes?** No

### Testing planned or performed, if any:
- [x] Executed pFUnit / CIME regression tests

**CTSM / CESM baseline hash-tag:** 7409d3218513e1d95dbc2828de9a11a8972b1386
**PR branch hash-tag:** 3a0dc8a1455978e66838fa5cef20c3ac60305190

- [x] The code in this PR branch runs with no errors. **Briefly describe tested configuration(s):** macOS gfortran pFUnit / CIME ERP regression tests
- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Which?:** (a) Bit-for-bit (B4B) with baseline master
- [x] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates. **Which?:** (a) Internal fix/test; no documentation updates required.


slevis-lmwg:
@johnpaulalex I think this PR blocks #4171, so I added the corresponding label. Pls correct me if I'm wrong.

johnpaulalex:
Thanks @slevis-lmwg ! True, but also note that this issue itself depends on https://github.com/NGEET/fates/pull/1596. Should I mark 1596 as a blocker? And then does 1596 get submitted, then cesm is pointed to the updated FATES version, then this PR can be submitted, and finally 4171?

slevis-lmwg:
That sounds correct.

src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -737,6 +737,11 @@ subroutine SoilThermProp (bounds, num_urbanc, filter_urbanc, num_nolakec, filter
             endif
 
             ! Thermal conductivity of snow
+            ! Initialize bw and thk to spval for inactive snow layers
```
ekluzek: This is good to do each timestep, since snow can change each timestep.

ekluzek: Should this be a tiny subroutine though?

src/biogeophys/SoilStateType.F90
```diff
@@ -393,4 +394,28 @@ subroutine Restart(this, bounds, ncid, flag)
     
   end subroutine Restart
 
+  !------------------------------------------------------------------------
+  subroutine TimestepInit(this, bounds)
```
ekluzek: I wonder about the name of the subroutine here. To match some of the biogeochem code, a name like SetValues could be done, with spval passed in, which would also allow it to be more general. And give a clearer idea of what's going on from the call.

src/biogeophys/SoilMoistStressMod.F90
```diff
@@ -383,6 +383,7 @@ subroutine calc_root_moist_stress_clm45default(bounds, &
             ! rootr effectively defines the active root fraction in each layer      
             if (h2osoi_liqvol(c,j) .le. 0._r8 .or. t_soisno(c,j) .le. tfrz-2._r8) then
                rootr(p,j) = 0._r8
+               rresis(p,j) = 0._r8
```
ekluzek, line 386: This looks like the right thing to do.

src/biogeophys/SoilHydrologyMod.F90
```diff
@@ -251,6 +251,16 @@ subroutine SetSoilWaterFractions(bounds, num_hydrologyc, filter_hydrologyc, &
        end do
     end do
 
+    ! Initialize deeper bedrock layers (which do not have active hydrology) to spval
```
ekluzek, line 254: This is something that should only have to happen once. The level of bedrock doesn't change through the simulation even for transient simulations.

ekluzek: I think it should also be done in a tiny subroutine. And some way of only doing it once.

src/biogeophys/FrictionVelocityMod.F90
```diff
@@ -404,6 +405,36 @@ subroutine InitCold(this, bounds)
    
   end subroutine InitCold
 
+  !------------------------------------------------------------------------------
+  subroutine TimestepInit(this, bounds)
```
ekluzek: With the similar subroutine below, I thought the name might be changed to SetValue, and spval passed in (to match a name used in biogeochem). I think the same thing here.

As below I also think that likely these variables only need to be reset this way once.

src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -640,6 +640,9 @@ subroutine CanopyFluxes(bounds,  num_exposedvegp, filter_exposedvegp,
         bsha                    => energyflux_inst%bsha_patch                       ! Output: [real(r8) (:)   ]  sunlit canopy transpiration wetness factor (0 to 1)
       end if
 
+      ! Initialize diagnostic variables to spval so inactive patches don't retain stale memory
```
ekluzek, line 643: I think this is another that problably only needs to be done once. But, for a single array like this I don't know that it needs a lot of logic around it or moved into a tiny subroutine. But, since our code is so un-modular, spinning off tiny subroutines isn't a bad thing to do either.

src/biogeophys/BiogeophysPreFluxCalcsMod.F90
```diff
@@ -91,6 +91,9 @@ subroutine BiogeophysPreFluxCalcs(bounds, &
     character(len=*), parameter :: subname = 'BiogeophysPreFluxCalcs'
     !-----------------------------------------------------------------------
 
+    call frictionvel_inst%TimestepInit(bounds)
```
ekluzek: Talked about below. Maybe the name should be SetValue, and spval passed into it?

src/biogeochem/VOCEmissionMod.F90
```diff
@@ -587,6 +587,24 @@ subroutine VOCEmission (bounds, num_soilp, filter_soilp, &
     vocflx(bounds%begp:bounds%endp,:)   = 0._r8
     vocflx_tot(bounds%begp:bounds%endp) = 0._r8
 
+    gamma_out(bounds%begp:bounds%endp) = spval
```
ekluzek: Since, there's a bunch of these spinning off into a subroutine seems worthwhile. Also this might not have to be done every timestep? This is talked about more below as well...

src/biogeochem/NutrientCompetitionFlexibleCNMod.F90
```diff
@@ -673,6 +673,8 @@ subroutine calc_plant_cn_alloc(this, bounds, num_soilp, filter_soilp,   &
                this%actual_storage_leafcn(p) = cnveg_carbonstate_inst%leafc_storage_patch(p)  &
                     / cnveg_nitrogenstate_inst%leafn_storage_patch(p)
             end if
+         else
```
ekluzek, line 677: Yes, this is something that should be done like this every timestep.

ekluzek (changes requested):
Interesting, this allows us to not have to add the diagnostic variables to restarts which is good. I'm also impressed that the AI figured that out.

It has the cost of resetting diagnostic variables to spval each time step. I doubt it has to be redone every time step though. The pattern of where the spval need to be shouldn't change for most of these variables. Only if it's a transient simulation in some cases, or if they are bound to something like whether snow is on the ground or not. And I don't think either applies here. At least in general there are a couple cases that I spell out in comments where it does need to be each timestep. Others I think can probably be done only once.

I suggest maybe changing the name of TimeStepInit to SetValue with spval passed in (matches biogeochem code). And I think spinning off some of these things into tiny subroutines might be appropriate.

So marking this with requesting changes, but also happy to discuss if you disagree.

But, there must be something going on where the pattern can be wrong on restart. From the earlier PR #4142 we can surmise that it's correct when written to the restart file and when read in from restart. But, if not read on restart, the pattern must not be correct and it causes problems. So maybe this just has to be done at initialization?

@johnpaulalex could you try (or did you try) having this done just at initialization and not each timestep? Although that should be taken care of by the normal initial setting of things to spval because they are output to history files. But, maybe it needs to be reset after that point and before used? What if TimeStepInit was only done once? So you added a flag to see if it's already been set and it only allows it to be set the first time?

ekluzek:
@johnpaulalex I reset the PR to go to b4b-dev as we talked about. It looks like you don't need to rebase the branch, but you should at least pull in the latest b4b-dev. And b4b-dev and master is going to be reconciled early next week, so you could wait for that.

johnpaulalex:
yeah good point @ekluzek . ~3 vars truly needed to be reset every time step. The rest are now done at init time.  And for the transisent case, that could apparently happen for all those other vars? See the second new commit for that, ResetDeactivatedDiagnostics in dynPatchStateUpdaterMod.F90.  I also switched the name to SetValue and added helper subroutines.

src/biogeophys/BiogeophysPreFluxCalcsMod.F90
johnpaulalex: done, and moved into InitCold

src/biogeophys/CanopyFluxesMod.F90
johnpaulalex: done, in InitCanopyFluxDiagnostics

src/biogeochem/VOCEmissionMod.F90
johnpaulalex: now in InitVOCDiagnostics, called once

src/biogeochem/NutrientCompetitionFlexibleCNMod.F90
johnpaulalex: ack

src/biogeophys/SoilTemperatureMod.F90
johnpaulalex: done, in InitSnowThermPropDiagnostics

src/biogeophys/SoilStateType.F90
johnpaulalex: done, renamed to SetValues and moved into InitCold

src/biogeophys/SoilMoistStressMod.F90
johnpaulalex: ack

src/biogeophys/SoilHydrologyMod.F90
johnpaulalex: done, removed per-timestep reset

src/biogeophys/FrictionVelocityMod.F90
johnpaulalex: done, renamed to SetValues and moved into InitCold

johnpaulalex:
> That sounds correct.

Ah I can't set labels on the FATES repo.  @ekluzek fyi, not sure if I should have permissions or ask you to do it.

johnpaulalex:
> @johnpaulalex I reset the PR to go to b4b-dev as we talked about. It looks like you don't need to rebase the branch, but you should at least pull in the latest b4b-dev. And b4b-dev and master is going to be reconciled early next week, so you could wait for that.

ack, I just rebased to b4b-dev @ekluzek

johnpaulalex:
@slevis-lmwg @ekluzek @glemieux I have updated description of CTSM PRs 4171, 4172, and FATES 1596 to establish their relationship to fixing CTSM issues 3660, 3661, and 4124.

TL;DR the PRs need to be submitted in 1596->4172->4171 order. PR 4172 resolves Issues 3661 and 4124, and PR 4171 resolves issue 3660.

src/biogeophys/SoilStateType.F90
```diff
@@ -337,6 +338,8 @@ subroutine InitCold(this, bounds)
     this%smp_l_col(bounds%begc:bounds%endc,1:nlevgrnd) = -1000._r8
     this%hk_l_col(bounds%begc:bounds%endc,1:nlevgrnd) = 0._r8
 
+    call this%SetValues(bounds, spval)
```
mvdebolskiy, line 341: Why? everything above is already assigned. in the allocate statements? There is no need for an extra routine for 2 variables that is only called ones in the same module.

johnpaulalex: Thanks @mvdebolskiy , I had neglected to call the new method in two places, see the new commit .

src/biogeophys/SoilStateType.F90
```diff
@@ -401,24 +400,29 @@ subroutine Restart(this, bounds, ncid, flag)
   end subroutine Restart
 
   !------------------------------------------------------------------------
-  subroutine SetValues(this, bounds, setval)
+  subroutine SetValues(this, bounds, setval, p)
```
mvdebolskiy: I still would recommend to remove this routine.
It's not needed. and with optional index argument I am not sure if will work with any mpi decomposition routines.  You can directly set the values of these two variables in `src/dyn_subgrid/dynPatchStateUpdaterMod.F90` and `src/biogeophys/SoilStateType.F90`

slevis-lmwg: @mvdebolskiy I find myself leaning in favor of keeping the three calls to this routine at this point.

slevis-lmwg (approved):
I will test by uncommenting the fields that I previously commented out in #4136 for the all_outputs test to pass:
`./create_test ERP_P64x2_D_Ld3.f10_f10_mt232.IHistClm60BgcCropCrujra.derecho_gnu.clm-default--clm-all_outputs -c /glade/campaign/cgd/tss/ctsm_baselines/ctsm5.4.058` FAIL
Putting back the commented out code (git stash): FAIL
Update branch: 
Vanilla ctsm5.4.058: 
Ultimately (once the above cases work) I should run with fates code that includes https://github.com/NGEET/fates/pull/1596

slevis-lmwg:
@johnpaulalex did you run tests and did they pass at some point? My tests fail so far, so I will bounce the PR back to you. 

See my last post for what I tested. To replicate my work, start with my tests 3 (your branch) and 4 (baseline). I included the `create_test` command in my post above, and I submit that from the /cime/scripts directory. I could pick this up again when you get test 3 to pass.

Issue #4188, opened 2026-08-28
Consider one or more changes to reduce DYNBAL fluxes (dynamic landunit conservation correction fluxes)


billsacks:
### Introduction

When subgrid areas change due to dynamic landunits / columns, [fluxes are generated to compensate for changes in gridcell-average water and energy](https://escomp.github.io/CTSM/tech_note/Transient_Landcover/CLM50_Tech_Note_Transient_Landcover.html#water-and-energy-conservation) (QFLX_LIQ_DYNBAL, QFLX_ICE_DYNBAL, EFLX_DYNBAL). These fluxes can (and should) be reduced via [careful](https://github.com/escomp/ctsm/issues/4186) setting of [reset_dynbal_baselines](https://escomp.github.io/cism-docs/cism-in-cesm/versions/master/html/clm-cism-coupling.html#resetting-clm-s-baseline-states-for-dynamic-landunits), but that still doesn't reduce these fluxes to zero.

The current implementation seems good enough in situations where there is a gradual change in a given grid cell's subgrid areas over many years. Unfortunately, though, that is not always the situation when coupling to CISM, [as has been seen in recent CESM3 runs](https://github.com/NCAR/cesm_dev/discussions/393): In grid cells around the ice sheet margin, large fractions of a CTSM grid cell can blink between mostly-vegetated/bare and mostly-glacier on a 2-year cycle, and this oscillatory behavior can go on for many years, possibly indefinitely. This can lead to large dynbal fluxes in each year, even when dribbled throughout the year (as is currently done).

[It is problematic for the ocean to get negative dynbal fluxes](https://github.com/NCAR/cesm_dev/discussions/393#discussioncomment-18178231), but it is also problematic to remove these negative dynbal fluxes by spreading them globally, as is currently done, because then regional fluxes that should average to 0 spatially and temporally (e.g., into the Arctic ocean) end up being biased positive (as explained [here](https://github.com/NCAR/cesm_dev/discussions/393#discussion-10692756)). Therefore, I have been thinking about possible strategies to further reduce these DYNBAL fluxes. I'm laying out a few possibilities here, and we may want to implement one or more of these.

### (1) Add pools that slowly release the dynbal fluxes, with multi-year residence times

[...]

Issue #1452, opened 2021-08-04
Make endrun calls consistently use errMsg which gives line number, remove subname, and "ERROR::" prepending it


ekluzek:
@billsacks did some work on this in #1447. In there I suggested a few other changes that would be good to do.

Make sure errMsg is always used which gives the line number of the file, so allows you to easily find the problem. As such we should remove the use of "subname" which needs to be maintained if the subroutine name changes whereas the errMsg automatically updates the line number. We can also remove prepending the error message to endrun with a "ERROR::" as it's already in the shr_sys_abort call in endrun.

ekluzek:
We recommend using the endrun form where __sourcefile__ and __LINE__ is sent into endrun. We might need to test this?

ekluzek:
There was some work in share adding a shr_abort_mod module with an interface to do a better job with this that we should adopt in endrun for CTSM. This is available at least in ctsm6.5.065 and probably before that.

As such we should update the interface for endrun to use it and also add the options for file and line to endrun.

Issue #1423, opened 2021-07-08
FORTRAN 2008 newunit feature allows us to remove fileutils getavu/relavu code


ekluzek:
The FORTRAN 2008 feature "newunit" for open statements can allows us to remove the getavu/relavu code in fileutils.F90. Most everything in fileutils.F90 is now obsolete. So we likely could refactor to remove all of it.

ekluzek:
@mvertens started utilizing this in #1420

DavidBergeron-eng:
If unit is obsolete upgrade please

PR #4189, opened 2026-08-31
ctsm5.4.055: Add temporal smoothing of DYNBAL conservation correction fluxes

billsacks:
### Description of changes

Spread the DYNBAL water and energy fluxes over many years: Rather than releasing all of the DYNBAL fluxes in the year in which they're generated, instead put them in storage pools that are released with a 20-year residence time (controllable via a new namelist parameter). This reduces these fluxes, especially in the oscillatory case that is somewhat common with a dynamic ice sheet.

The new storage pools are reset when reset_dynbal_baselines is set to .true., since this feels conceptually similar to the other resetting done by that flag, so it feels appropriate to do this resetting at the same time.

This PR also changes the sign convention for the DYNBAL fluxes: the previous sign convention was unintuitively the opposite of the sign convention of other fluxes to river/atm; while I'm changing these fluxes anyway, I'm also flipping the sign convention to be more intuitive and hopefully less error-prone.

I have evaluated the changes here by examining QFLX_ICE_DYNBAL, QFLX_LIQ_DYNBAL and EFLX_DYNBAL, FSH_TO_COUPLER, QRUNOFF_ICE_TO_COUPLER and QRUNOFF_TO_COUPLER in 3 cases with compset I1850Clm60BgcCropG at resolution f09_g17:
- A baseline case
- The new code, out-of-the-box - residence time 20 years
- The new code, with residence time set at 1 year

### Specific notes

- Much of the code was written by Claude code. However, I gave detailed guidance and carefully reviewed all code changes.

- Partially addresses #4188 (specifically, option (1) in that issue)

- Changes answers for all cases with landcover change - Hist/SSP cases, CISM cases, and FATES cases
- Changes in Satellite Phenology cases are limited to the DYNBAL fluxes and a few related variables, including liquid/ice runoff and sensible heat flux (which then impacts ROF state and fluxes)
- Changes in BGC cases are more widespread, I think because of the feedback of VOLR on CH4. (I tested this by running a version of `ERS_Ly3_P64x2.f10_f10_mg37.IHistClm50BgcCropG.derecho_intel.clm-cropMonthOutput` with SROF and verified that diffs were then limited to just the expected variables.)
- Also changes answers for some additional cases that start mid-year and use a restart file that has non-zero state in the dynbal dribblers, despite not having transient landcover themselves (since this PR removes these dribblers, so no longer has fluxes out of these dribblers in these cases). On derecho aux_clm, this impacts the tests `ERS_D_Ld5_Mmpi-serial.1x1_mexicocityMEX.I1PtClm60SpRsGs.derecho_gnu.clm-CLM1PTStartDate`, `SMS_Ld12_Mmpi-serial.1x1_vancouverCAN.I1PtClm60SpRsGs.derecho_gnu.clm-output_sp_highfreq`, and `ERP_D_Ld5.f10_f10_mg37.I2000Clm60Sp.derecho_intel.clm-decStart--clm-nofireemis`.

- New namelist flag, `dynbal_storage_residence_time`

- Ran aux_clm on derecho. Most tests passed; I'll look into a couple of failures. Answers changed as noted above
- Will run aux_clm on izumi

- [x] This either (a) does not change answers, (b) it only changes answers at roundoff level, or (c) I have performed a scientific evaluation of the answer changes. **Which?: (c)**

- [x] I have reviewed relevant parts of the CLM documentation [Tech Note](https://escomp.github.io/CTSM/tech_note/index.html) or [User's Guide](https://escomp.github.io/CTSM/users_guide/index.html) to determine if anything needs to be changed or added. **If it does, describe:** Made a small necessary change to the tech note.
- [x] This PR either (a) does not create a need to update the documentation or (b) includes required documentation updates (see [guidelines for contributing documentation](https://escomp.github.io/CTSM/users_guide/working-with-documentation/docs-intro.html#contribution-guidelines)). **Which?: (b)**


wwieder:
@billsacks when this is ready for review maybe this can be a paired review with @slevis-lmwg?  I think we may want this in the CESM3_alpha10a tag?

billsacks:
@slevis-lmwg - I feel this is ready for review. I'm still reviewing some test results, but from my initial look things looked basically good.

doc/source/tech_note/Transient_Landcover/CLM50_Tech_Note_Transient_Landcover.rst
```diff
@@ -47,7 +47,7 @@ When subgrid areas change, the water and energy states remain unchanged on a per
...
 
-Water and energy are conserved by summing up the total water and energy content of each gridcell before and after a change in area. Differences in liquid and ice water content are balanced by liquid and ice runoff terms, which can be either positive or negative. (Negative runoff is effectively a withdrawal of water from the ocean.) Differences in energy content are balanced by a sensible heat flux term, which again can be either positive or negative. These balancing fluxes are spread evenly throughout the following year.
+Water and energy are conserved by summing up the total water and energy content of each gridcell before and after a change in area. Differences in liquid and ice water content are balanced by liquid and ice runoff terms, which can be either positive or negative. (Negative runoff is effectively a withdrawal of water from the ocean.) Differences in energy content are balanced by a sensible heat flux term, which again can be either positive or negative. These balancing fluxes are released gradually from temporary storage pools with 20-year residence times.
```
slevis-lmwg: ...so am I saying this correctly, that this PR changes the release of balancing fluxes from an exact term of 1 year to a residence time of 20 years? If so, I would suggest a comment in `bld/namelist_files/namelist_definition_ctsm.xml` (could be after the last paragraph that you added) stating that, prior to this new namelist option, the balancing fluxes were spread evenly throughout the following year.

src/dyn_subgrid/dynSubgridControlMod.F90
```diff
@@ -190,9 +202,23 @@ subroutine read_namelist( NLFilename )
...
+       write(iulog,*) 'ERROR: dynbal_storage_residence_time must be greater than 0'
+       write(iulog,*) 'Value given: ', dynbal_storage_residence_time
+       call endrun(msg='ERROR: dynbal_storage_residence_time must be greater than 0 '// &
+            errMsg(sourcefile, __LINE__))
+    end if
+    ! It isn't important to account for leap years in this conversion: we just want a
+    ! value that roughly equates to the desired residence time in years; it doesn't need
+    ! to be exact.)
+    dynbal_storage_turnover_rate = &
+         1._r8 / (dynbal_storage_residence_time * (365._r8 * SHR_CONST_CDAY))
```
slevis-lmwg: How about
```suggestion
         1._r8 / (dynbal_storage_residence_time * (get_average_days_per_year() * SHR_CONST_CDAY))
```
or along the lines of how I see this done in CNFUNMod.F90.

src/main/lnd2atmMod.F90
```diff
@@ -463,7 +463,7 @@ subroutine lnd2atm(bounds, &
          c2l_scale_type= 'urbanf', l2g_scale_type='unity' )
     do g = bounds%begg, bounds%endg
        water_inst%waterlnd2atmbulk_inst%qflx_rofice_grc(g) = &
-         water_inst%waterlnd2atmbulk_inst%qflx_rofice_grc(g) - &
+         water_inst%waterlnd2atmbulk_inst%qflx_rofice_grc(g) + &
```
slevis-lmwg, line 466: You mentioned sign conventions in some variable declarations and possibly other code comments. I don't remember if you stated explicitly that switching from dribbler fluxes to dynbal storage pools leads to the sign switch here. I don't know if you need to. I'm just typing my thoughts :-)

ekluzek:
> @billsacks when this is ready for review maybe this can be a paired review with @slevis-lmwg? I think we may want this in the CESM3_alpha10a tag?

I think you mean cesm3_0_alpha11a. We don't need this in beta10 as it's only changing transient cases, and beta10 is about 1850 and CPLHIST spinup. It would also further delay the beta10 tag.

The pic of the whiteboard from yesterday also agrees with this being in beta11 rather than beta10.

slevis-lmwg (approved):
@billsacks thank you for writing clear code and comments. I made minor suggestions and asked for confirmation of my understanding once or twice. I may get a bit more out of this PR from a paired review, but it also does not seem necessary. If you would rather meet, I would be open.

billsacks:
> I think you mean cesm3_0_alpha11a. We don't need this in beta10 as it's only changing transient cases, and beta10 is about 1850 and CPLHIST spinup. It would also further delay the beta10 tag.
> 
> The pic of the whiteboard from yesterday also agrees with this being in beta11 rather than beta10.

I'm not clear on the vision for the different beta tags, but as I just noted to @ekluzek - this is desired for the coupled spinup immediately following the land spinup - so probably around the end of next week.

wwieder:
> > I think you mean cesm3_0_alpha11a. We don't need this in beta10 as it's only changing transient cases, and beta10 is about 1850 and CPLHIST spinup. It would also further delay the beta10 tag.
> > The pic of the whiteboard from yesterday also agrees with this being in beta11 rather than beta10.
> 
> I'm not clear on the vision for the different beta tags, but as I just noted to @ekluzek - this is desired for the coupled spinup immediately following the land spinup - so probably around the end of next week.

I agree with @billsacks here, I think this is something that's used in the CESM3 PI control runs as soon as is realistically possible.  I'm also not really clear what that means for the staging of CESM alpha tags, but I thought it could be alpha10e based on one version of the plan I saw today?

doc/source/tech_note/Transient_Landcover/CLM50_Tech_Note_Transient_Landcover.rst
slevis-lmwg: Bill and I discussed an alternative comment that conveys the lack of backwards compatibility.

src/dyn_subgrid/dynSubgridControlMod.F90
slevis-lmwg: Bill correctly points out that he should use get_average_days_per_year() instead.

src/main/lnd2atmMod.F90
slevis-lmwg: Bill explained to me that the sign convention is changing for consistency with other fluxes out of the land model, which are mostly (if not always) positive exiting the land model.

doc/source/tech_note/Transient_Landcover/CLM50_Tech_Note_Transient_Landcover.rst
billsacks: Done in 1eee35d

ekluzek:
> > > I think you mean cesm3_0_alpha11a. We don't need this in beta10 as it's only changing transient cases, and beta10 is about 1850 and CPLHIST spinup. It would also further delay the beta10 tag.
> > > The pic of the whiteboard from yesterday also agrees with this being in beta11 rather than beta10.
> > 
> > 
> > I'm not clear on the vision for the different beta tags, but as I just noted to @ekluzek - this is desired for the coupled spinup immediately following the land spinup - so probably around the end of next week.
> 
> I agree with @billsacks here, I think this is something that's used in the CESM3 PI control runs as soon as is realistically possible. I'm also not really clear what that means for the staging of CESM alpha tags, but I thought it could be alpha10e based on one version of the plan I saw today?

I don't have too strong of a sense if it should be in beta10 or beta11. I mainly didn't want to add more complexity to beta10. From the planning we talked about today, I don't think the MOM, CICE, CAM, and CTSM tags will be ready and in a cesm alpha tag for about 2 weeks. So unless they want to start another coupled segment without all of those in place, but just the DYNBAL tag in place -- I think it's either the tail end of beta10 (cesm3_0_alpha10d) or it's the beginning of beta11 (cesm3_0_alpha11a). I don't think there will be much of a difference between either option in terms of timing. It's just if it's in beta10, it WILL have had prebeta testing done on it, and if it's in beta11, it would have to WAIT for the prebeta testing of beta10 to happen.

I know Dave wanted to start by the 9th, based on what we talked about today, I think the 18th is a realistic timeline, with maybe a few days sooner if things go well. But, if things don't go great it probably slips into the next week after the 18th.

bld/namelist_files/namelist_definition_ctsm.xml
```diff
@@ -2893,6 +2898,38 @@ continue run; it is an error for this to be set in a branch
...
+Must be greater than 0. This can be set to less than 1 year, but should be much longer
+than the model time step: if the residence time approaches the length of a time step, the
+pools will overshoot and the fluxes will oscillate.
+
+(Prior to the introduction of this namelist option, the balancing fluxes were spread
+evenly throughout the following year. There is no way to exactly recover the old CLM5
+behavior, but it can be approximated by setting this option to something between 0.5 and
+1.0, depending on the behaviors you want to recover - peak fluxes, effective residence
+time, etc.)
+</entry>
```
ekluzek: @billsacks this is a really clear explaination that I find really helpful. I'm just wondering if this is the right place to put it all? The CTSM UG has a link to the definition file, but it's not presented very well. I imagine we'll have something for presenting namelist values somewhere like we did for CESM2., but that's not there yet. So this isn't that accessabile to users.

But, if it doesn't go here, it's not clear where the best place for it would be. But, the long version of this should be in the Tech Note.

bld/namelist_files/namelist_defaults_ctsm.xml
```diff
@@ -2750,6 +2750,7 @@ lnd/clm2/surfdata_esmf/NEON/ctsm5.4.0/surfdata_1x1_NEON_TOOL_hist_2000_78pfts_c2
 <do_transient_lakes>.false.</do_transient_lakes>
 <do_transient_urban>.false.</do_transient_urban>
 <reset_dynbal_baselines>.false.</reset_dynbal_baselines>
+<dynbal_storage_residence_time>20.d00</dynbal_storage_residence_time>
```
ekluzek: It does seem like this should be 20 for clm6_0 physics, but maybe 0.75 for clm5_0 physics? From the comment below you can get similar behavior for a value between 0.5 and 1. Having clm5_0 act similar to what it used to seems like a good idea to me.

src/biogeophys/BalanceCheckMod.F90
```diff
@@ -283,38 +279,14 @@ subroutine WaterGridcellBalanceSingle(bounds, &
        enddo
     endif
     
-    ! Call the beginning or ending version of the subroutine according
```
ekluzek, line 286: So this is ripping out the previous method, which I think I see the argument for, since it didn't work that well? You did say that changing dynbal_storage_residence_time to something between 0.5-1.0 would approximate the old behavior. So that might be a reason to just rip it all out.

I just want to make sure we are convinced we really don't want the old behavior before complete removal. Especially since it affects what clm5_0 cases will be.

ekluzek: And if we ARE ripping it out. We also nee to remov AnnualFluxDribbler and it's unit tester. Although that could also be done in a follow on PR as a cleanup step just to get this one done quicker. If that's the case we should make an issue for that removal step.

bld/CLMBuildNamelist.pm
```diff
@@ -2827,6 +2827,12 @@ sub setup_logic_dynamic_subgrid {
         &remove_leading_and_trailing_quotes($nl_flags->{'clm_start_type'}) eq "branch") {
       $log->fatal_error("reset_dynbal_baselines has no effect in a branch run");
    }
+
+   add_default($opts, $nl_flags->{'inputdata_rootdir'}, $definition, $defaults, $nl, 'dynbal_storage_residence_time');
+   my $dynbal_storage_residence_time = $nl->get_value('dynbal_storage_residence_time');
+   if ( $dynbal_storage_residence_time <= 0.0 ) {
+      $log->fatal_error("dynbal_storage_residence_time must be greater than 0");
```
ekluzek, line 2848: We could add a test for this to the namelist tester.

I think maybe that could be something I do and push to this PR?

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -559,10 +584,16 @@ subroutine dyn_hwcontent_final(bounds, &
          liquid_water_temp2 = temperature_inst%liquid_water_temp2_grc(begg:endg), &
          delta_heat = delta_heat(begg:endg))
 
-    call energyflux_inst%eflx_dynbal_dribbler%set_curr_delta(bounds, &
-         delta_heat(begg:endg))
-    call energyflux_inst%eflx_dynbal_dribbler%get_curr_flux(bounds, &
-         energyflux_inst%eflx_dynbal_grc(begg:endg))
+    ! Add this time step's spurious heat gain to the storage pool (with a negative to use
```
ekluzek, line 671: It seems like this new behavior should have unit testing around it? You might be reluctant to do that now to save time and if so I get it.

ekluzek: I do also wonder if the dribblers couldn't be extended to do this new behavior rather than the dribbling over a year? They nicely encapsulate the data and methods for the dribbling and allow it to be reused for both water and energy. I'm just wondering if you explored that idea? Or maybe did and gave up on it for some reason?

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -625,15 +663,20 @@ subroutine dyn_water_content_final(bounds, &
...
-    call waterflux_inst%qflx_liq_dynbal_dribbler%set_curr_delta(bounds, &
-         delta_liq(begg:endg))
-    call waterflux_inst%qflx_liq_dynbal_dribbler%get_curr_flux(bounds, &
-         waterflux_inst%qflx_liq_dynbal_grc(begg:endg))
-
-    call waterflux_inst%qflx_ice_dynbal_dribbler%set_curr_delta(bounds, &
-         delta_ice(begg:endg))
-    call waterflux_inst%qflx_ice_dynbal_dribbler%get_curr_flux(bounds, &
-         waterflux_inst%qflx_ice_dynbal_grc(begg:endg))
+    ! Add this time step's spurious water gain to the storage pools (with a negative to
```
ekluzek, line 748: This is an example where the old dribblers seem to encapsulate what's going on in a better way. There's two calls for each flux. Whereas in the new code the fluxes and the change to storage is directly visible, and you have to have the time-step and turnover_rate out in the open, and the loop is right there as well.

ekluzek (changes requested):
This is all great stuff, that will be good to have in place to counter the oscilitory behavior that's been seen. It's more or less transforming the dribblers that were fixed on a year return time to something with a flexible time (currently 20 years). But, it's also easy to change that time now.

I have a bunch of questions and comments. The most important ones are:

- I think we should set clm5_0 physics to use a value of 0.75 to get similar behavior to how it worked previously
- The long description in namelist_definition is very clear and very useful. But, I think for usability/accessability it should go somewhere else. I'm thinking the Tech Note.
- I ask if we think we really want to remove the old dribblers?
- IF we do want to remove the old dribblers the AnnualFluxDribbler module should be removed and the unit tests for it.
- If we are removing unit tests -- shouldn't we add some for the new behavior?

I think some decisions might have been made to get something robust in place quickly. And that is probably what we need here. But, I think we might want to make sure we are good with all of them for long term? And if not we might want to plan some follow on refactoring of the most important things.

We should probably talk about some of this together as well. But, we probably don't want/need to do that until next week....

src/biogeophys/BalanceCheckMod.F90
billsacks: The AnnualFluxDribbler is still used in other parts of the code. I considered trying to keep things completely backwards compatible here, but felt it would be a big mess, since the conditional logic would appear in numerous parts of the code, and it felt relatively unimportant in terms of the science.

But very good point about making the default residence time differ for clm5. I think that's a good compromise approach and I'll do that.

bld/CLMBuildNamelist.pm
billsacks: > We could add a test for this to the namelist tester.
> 
> I think maybe that could be something I do and push to this PR?

I personally feel like there are higher priorities than doing that, but you have my permission to push to the branch if you want.

src/dyn_subgrid/dynConsBiogeophysMod.F90
billsacks: I think it would be difficult to put this new behavior in the dribbler.

I gave serious consideration to introducing a similar new class to encapsulate this behavior, which would be unit testable, but felt that this would add extra infrastructure and indirection for very little benefit in this case. I was particularly thinking about some possible future uses of similar pools (for the sake of water tracers): there, we might need to do some special-purpose adjustments to the fluxes before updating the storage pools, so the design would need to accommodate that, and it felt at that point that, for these relatively simple residence time equations, it would be considerably simpler and less error-prone to just do the equations inline like this.

And then, given the way it's currently done, I felt that unit testing would be infeasible without a considerable refactor that didn't fit into the time pressure for this change.

src/dyn_subgrid/dynConsBiogeophysMod.F90
billsacks: See my above comment: It's true that the current code has three lines instead of two, but in some future uses I'm envisioning, it wouldn't work well to encapsulate things in the way that's done in the dribbler, because we'd need to do some adjustment to the fluxes before updating the state variable, so you'd need to separate it into these three separate steps anyway.

bld/namelist_files/namelist_defaults_ctsm.xml
ekluzek:
```suggestion
<dynbal_storage_residence_time >20.d00</dynbal_storage_residence_time>
<dynbal_storage_residence_time phys="clm5_0">0.75d00</dynbal_storage_residence_time>
```

src/biogeophys/BalanceCheckMod.F90
ekluzek: Just talked with @billsacks and I'm good with this. I wanted to make sure he was thinking about design and he did. It seems likely that changes in water isotopes might drive a new design to help facilitate them. But, for now this is simple enough.

src/dyn_subgrid/dynConsBiogeophysMod.F90
ekluzek: As above we talked it over some as well, and it didn't seem the right thing to do for several reasons. Maybe later with future water isotope work.

src/dyn_subgrid/dynSubgridControlMod.F90
ekluzek: I updated the suggestion to use the right option.

bld/namelist_files/namelist_defaults_ctsm.xml
billsacks: > It does seem like this should be 20 for clm6_0 physics, but maybe 0.75 for clm5_0 physics? From the comment below you can get similar behavior for a value between 0.5 and 1. Having clm5_0 act similar to what it used to seems like a good idea to me.

Very good point. Done in 1475d298d5b906c532e51a5b49fef4a59d18234c

src/dyn_subgrid/dynSubgridControlMod.F90
billsacks: Fixed in e5768b9701

doc/source/tech_note/Transient_Landcover/CLM50_Tech_Note_Transient_Landcover.rst
billsacks: @slevis-lmwg - in 54ff1feab4 I undid the text that we had added to namelist_definition in 1eee35dda2ecb7f2dcaa8c59de38d14bc05460b7. My rationale was: The main motivation I saw for adding this was to explain to a user how to best reproduce the clm5.0 behavior; but now this is done via namelist_defaults (per @ekluzek 's suggestion), so this text that explained the change felt unnecessarily verbose. Please let me know if you're okay with this.

slevis-lmwg: I'm OK with this AND I suggest a one-line comment in `namelist_defaults_ctsm.xml` pointing out the lack of actual backwards compatibility despite providing a clm5 value.

billsacks: > I'm OK with this AND I suggest a one-line comment in `namelist_defaults_ctsm.xml` pointing out the lack of actual backwards compatibility despite providing a clm5 value.

Excellent point. I did this in 223b6e7344 and added some more information describing the choice of default. And actually, I'll add another comment describing the choice of a 20-year default while I'm at it.

billsacks: > I'll add another comment describing the choice of a 20-year default while I'm at it

See 3f516e8d89 for that extra documentation, and let me know if you feel this is too much or have other comments on this

slevis-lmwg: I'm comfortable with these explanations. Feel free to resolve this conversation. Thanks, Bill.

bld/CLMBuildNamelist.pm
ekluzek: Yeah I won't be able to get to this. So just resolving.

ekluzek (approved):
@billsacks it looks like all review points were taken care of. So I resolved most of the conversations. The only one I left open was the one about the long (but clear) instructions in the namelist_definition file, in case you were going to do more with it. But, I would also be fine with your going ahead with it as is.

So approving. Are you planning on running the baselines on Derecho and Izumi and making the ChangeLog/ChangeSum files to close this one out?

Since, you are ready now, we could switch tag order and make this "55" and mine "56", which won't be ready on Friday anyway.

ekluzek:
Oh, actually @billsacks it looks like @slevis-lmwg is assigned to this, so maybe he's going to finish the baselines for it? It would just be good if you start the ChangeLog file @billsacks.

@slevis-lmwg we are thinking of changing the tag order so this becomes "55" and goes next. I won't have a tag on master on Friday, so this could come in now.

ekluzek:
@billsacks is going to take this all the way home and it's now next up in the tag queue.

bld/namelist_files/namelist_definition_ctsm.xml
billsacks: I have moved pieces of this description out of the namelist definition into the tech note as you suggest - see 219555257f.

ekluzek: Perfect

billsacks:
I have confirmed that the recent change to move the namelist variable is bit-for-bit with the previous version of this branch, by running `ERS_Ly3_P64x2.f10_f10_mg37.IHistClm50BgcCropG.derecho_intel.clm-cropMonthOutput`.

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -57,12 +59,102 @@ module dynConsBiogeophysMod
...
+    ! Read the dyn_cons_biogeophys_inparm namelist.
+    !
+    ! Note that this must be called AFTER the time manager has been initialized, because
+    ! converting dynbal_storage_residence_time to a turnover rate depends on the model
+    ! calendar.
+    !
+    ! !USES:
+    use shr_const_mod  , only : SHR_CONST_CDAY
+    use abortutils     , only : endrun
+    use fileutils      , only : getavu, relavu
```
ekluzek: Remove these. See below...

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -57,12 +59,102 @@ module dynConsBiogeophysMod
...
+    character(len=*), parameter :: subname = 'dynConsBiogeophys_readnl'
+    !-----------------------------------------------------------------------
+
+    namelist /dyn_cons_biogeophys_inparm/ &
+         dynbal_storage_residence_time
+
+    dynbal_storage_residence_time = 0._r8
+
+    if (masterproc) then
+       nu_nml = getavu()
```
ekluzek: You can remove the old unit processing things we had. Fortran now allows you to use a new unit keyword in the open statement to give you a free unit.

I'd like to move to that pattern so we can remove the unit processing stuff.

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -57,12 +59,102 @@ module dynConsBiogeophysMod
...
+       open( nu_nml, file=trim(NLFilename), status='old', iostat=nml_error )
+       call find_nlgroup_name(nu_nml, 'dyn_cons_biogeophys_inparm', status=nml_error)
+       if (nml_error == 0) then
+          read(nu_nml, nml=dyn_cons_biogeophys_inparm, iostat=nml_error)
+          if (nml_error /= 0) then
+             call endrun(msg='ERROR reading dyn_cons_biogeophys_inparm namelist'//errMsg(sourcefile, __LINE__))
```
ekluzek: You can now remove the bit about errMsg and pass file and line options directly in the endrun call. It's a little less cumbersome.

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -57,12 +59,102 @@ module dynConsBiogeophysMod
...
+          end if
+       else
+          call endrun(msg='ERROR finding dyn_cons_biogeophys_inparm namelist'//errMsg(sourcefile, __LINE__))
+       end if
+       close(nu_nml)
+       call relavu( nu_nml )
```
ekluzek: This can be removed with the newunit option to open.

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -57,12 +59,102 @@ module dynConsBiogeophysMod
...
+       nu_nml = getavu()
+       open( nu_nml, file=trim(NLFilename), status='old', iostat=nml_error )
+       call find_nlgroup_name(nu_nml, 'dyn_cons_biogeophys_inparm', status=nml_error)
+       if (nml_error == 0) then
+          read(nu_nml, nml=dyn_cons_biogeophys_inparm, iostat=nml_error)
+          if (nml_error /= 0) then
+             call endrun(msg='ERROR reading dyn_cons_biogeophys_inparm namelist'//errMsg(sourcefile, __LINE__))
+          end if
+       else
+          call endrun(msg='ERROR finding dyn_cons_biogeophys_inparm namelist'//errMsg(sourcefile, __LINE__))
```
ekluzek: Same thing here. Line and file can be used as arguments directly in endrun.

src/dyn_subgrid/dynConsBiogeophysMod.F90
```diff
@@ -57,12 +59,102 @@ module dynConsBiogeophysMod
...
+    character(len=*), intent(in) :: NLFilename ! Namelist filename
+    !
+    ! !LOCAL VARIABLES:
+    ! temporary variable corresponding to the namelist variable:
+    real(r8) :: dynbal_storage_residence_time ! residence time of the dynbal storage pools [years]
+    ! other local variables:
+    integer :: nu_nml    ! unit for namelist file
+    integer :: nml_error ! namelist i/o error flag
+
+    character(len=*), parameter :: subname = 'dynConsBiogeophys_readnl'
```
ekluzek: I suggest removing this. With the file and line options to endrun I've found this not useful. And it has the possibility of changing the subroutine name and forgetting to change this. So it has a maintenance burden that file and line to endrun don't have.

src/dyn_subgrid/dynConsBiogeophysMod.F90
ekluzek: #1423 talks about this.

src/dyn_subgrid/dynConsBiogeophysMod.F90
ekluzek: #1452 talks about this

ekluzek (approved):
Some comments about handling file units with newunit, and about endrun.

billsacks:
@ekluzek - thank you for pointing out these newer patterns. I agree with all of them. This is fixed in e10c2e0eb1e0966222fe15a34ef3780b17f691f8

Update 2026-09-07: I have confirmed that the changes in that commit are bit-for-bit with the earlier code, via `ERS_Ly3_P64x2.f10_f10_mg37.IHistClm50BgcCropG.derecho_intel.clm-cropMonthOutput`.

wwieder:
I just wanted to confirm that this will be in the CESM_beta10 tag.  @dlawrenncar asked for it to be used in the PI control runs to start next week. @billsacks, @briandobbins, @ekluzek

